PR #889, opened 2022-03-08
Added Connection shutdown complete analogue 

### LINKED ISSUE #849: Add Connection.ShutdownComplete
We want to add a Connection.ShutdownComplete property that mirrors Server.ShutdownComplete.

ReeceHumphreys:
This is an attempt at adding similar functionality of Server.Shutdown complete to Connection. 
Not really sure if this is the best solution or if I am missing anything but would love some feedback.

The connection tests are currently quite hard to follow so I wasn't sure where I should be testing this new functionality but I added it to a test where it should work and it did. Additionally, I modified the Bidir example and it seems to be behaving as expected but again, I could be missing something as Connection is quite complex.

* I definitely need to add more tests, just not sure where.

Resolves #849


externl (commented):
You should also added a test to verify it's not called on a resumable connection. Do I understand it right that if the connection is resumable this completion source will never be set?

bentoi (commented):
I think we first need to decide on the semantics of resumable connection. See https://github.com/zeroc-ice/icerpc-csharp/issues/891

PR #1087, opened 2022-04-20
Removed 'FrameHeader'

InsertCreativityHere:
I'm not a `FrameHeader` expert, but this looks like dead code.
If there's a reason to keep this around, just let me know and I'll close this PR!


bernardnormier (approved):
The comment suggests it is or used to be used in Slic. Guess it no longer is?

bentoi (approved):
It's indeed no longer used.

PR #3654, opened 2023-09-13
Fix width of code samples

### LINKED ISSUE #3653: Update sample code in READMEs
The README are displayed on nuget.org, e.g.
https://www.nuget.org/packages/IceRpc

We want to update them to:
 - avoid horizontal scrolling
 - for Slice files, make it clear it's Slice and not C# (and maybe do the same for C#?)

bernardnormier:
This PR fix the width of the code samples in README.md, so that they dispatch properly (without scroll bar) on nuget.org.

I think the max width is 99 or 100 columns.

Fixes #3653.


externl (commented):
I think it's already obvious that the code is C#. The rest of the changes are good.

externl (approved):
Looks good!

Issue #3663, opened 2023-09-27
The generated code for a field of type enum? does not compile.


bernardnormier:
### Description

The C# code generated by slicec-cs does not compile a struct with an optional enum field.

For example:
```slice
struct MyStructWithOptionalFields {
    i: int32
    j: int32
    k: int32?
    l: int32?
    e: MyEnum?
}
```

results in:
```
C:\Users\bernard\source\repos\icerpc-csharp\tests\ZeroC.Slice.Tests\generated\StructTests.cs(113,68): error CS1503: Argument 2: cannot convert from 'ZeroC.Slice.Tests.MyEnum?' to 'ZeroC.Slice.Tests.MyEnum' [C:\Users\bernard\source\repos\
icerpc-csharp\tests\ZeroC.Slice.Tests\ZeroC.Slice.Tests.csproj]
```

It seems that slicec-cs mistakenly treats an enum like a reference type in this context, but I didn't actually check the Rust code.

### Reproduction steps

See draft PR that updates the Struct test.

### Expected behavior

The generated code should compile without error or warning.

### Actual behavior

The generated code is incorrect and does not compile.

### Configuration

icerpc-csharp main, i.e. 0.1.1

PR #3664, opened 2023-09-27
Update test to show that generated code for field of type enum? does not compile

bernardnormier:
This draft update the Struct test to reproduce #3663.


bernardnormier:
Incorporated in #3666.

PR #3670, opened 2023-10-03
Refactors how Slic keeps connections alive

### LINKED ISSUE #3665: Slic and client-to-server streaming
I am writing a new example where a client slowly streams data to a server (a new element every 5 seconds).

After a while, the connection dies and the client Debug log shows me:
```
dbug: IceRpc.ClientConnection[7]
      Client connection from '[::1]:54372' to '[::1]:10000' failed to shutdown
      IceRpc.IceRpcException: The connection did not receive any bytes for over 30 s.
```

The only activity on this connection is this slow streaming. As I understand it, the client should receive Slic StreamWindowUpdate frames from time to time from the server, but apparently doesn't? 

Maybe we don't send these StreamWindowUpdate frames often enough in this scenario?

bernardnormier:
This PR refactors how Slic keeps connections alive.

Fixes #3665.

With this new implementation, Slic keeps connection alive by sending Ping frames from client connections (like before).

The algorithm is now:
 - send a Ping frame idle timeout * 0.6 after the last write even if there is an outstanding pong (= like before, except it's now 0.6 instead of 0.5)
 -  send a Ping frame idle timeout * 0.5 after the last read unless there is already an outstanding pong (new)

This PR also refactors the idle timeout decorator and associated tests. They are now 2 separate decorators, one for Ice and one for Slic. This PR didn't change the Ice logic.

### DIFF
--- a/src/IceRpc/Internal/IceDuplexConnectionDecorator.cs
+++ b/src/IceRpc/Internal/IceDuplexConnectionDecorator.cs
@@ -12,7 +12,7 @@ namespace IceRpc.Internal;
 internal class IceDuplexConnectionDecorator : IDuplexConnection
 {
     private readonly IDuplexConnection _decoratee;
-    private readonly Timer _keepAliveTimer;
+    private readonly Timer _writerTimer;
     private readonly CancellationTokenSource _readCts = new();
     private readonly TimeSpan _readIdleTimeout;
     private readonly TimeSpan _writeIdleTimeout;
@@ -22,8 +22,8 @@ internal class IceDuplexConnectionDecorator : IDuplexConnection
         TransportConnectionInformation connectionInformation = await _decoratee.ConnectAsync(cancellationToken)
             .ConfigureAwait(false);
 
-        // Schedules or reschedules a keep alive after a successful connection establishment.
-        ScheduleKeepAliveAfterWrite();
+        // Schedule or reschedule a keep alive after a successful connection establishment.
+        ResetWriteTimer();
         return connectionInformation;
     }
 
@@ -33,7 +33,7 @@ internal class IceDuplexConnectionDecorator : IDuplexConnection
         _readCts.Dispose();
 
         // Using Dispose is fine, there's no need to wait for the keep alive action to terminate if it's running.
-        _keepAliveTimer.Dispose();
+        _writerTimer.Dispose();
     }
 
     public ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken)
@@ -83,13 +83,12 @@ internal class IceDuplexConnectionDecorator : IDuplexConnection
             // After each successful write, we (re)schedule one ping (heartbeat) at _writeIdleTimeout / 2 in the future.
             // Since each ping is itself a write, if there is no application activity at all, we'll send successive
             // pings at _writeIdleTimeout / 2 intervals.
-            ScheduleKeepAliveAfterWrite();
+            ResetWriteTimer();
         }
     }
 
     /// <summary>Constructs a decorator that ensures a call to <see cref="ReadAsync" /> will fail after readIdleTimeout.
-    /// This decorator also schedules a keepAliveAction after each write (see
-    /// <see cref="ScheduleKeepAliveAfterWrite" />).</summary>
+    /// This decorator also schedules a keepAliveAction after each write (see <see cref="ResetWriteTimer" />).</summary>
     internal IceDuplexConnectionDecorator(
         IDuplexConnection decoratee,
         TimeSpan readIdleTimeout,
@@ -100,12 +99,11 @@ internal class IceDuplexConnectionDecorator : IDuplexConnection
         _decoratee = decoratee;
         _readIdleTimeout = readIdleTimeout; // can be infinite i.e. disabled
         _writeIdleTimeout = writeIdleTimeout;
-        _keepAliveTimer = new Timer(_ => keepAliveAction());
+        _writerTimer = new Timer(_ => keepAliveAction());
 
         // We can't schedule a keep alive right away because the connection is not connected yet.
     }
 
-    /// <summary>Schedules one keep alive in writeIdleTimeout / 2.</summary>
-    private void ScheduleKeepAliveAfterWrite() =>
-        _keepAliveTimer.Change(_writeIdleTimeout / 2, Timeout.InfiniteTimeSpan);
+    /// <summary>Resets the write timer. We send a keep alive when this timer expires.</summary>
+    private void ResetWriteTimer() => _writerTimer.Change(_writeIdleTimeout / 2, Timeout.InfiniteTimeSpan);
 }
--- a/src/IceRpc/Internal/IceDuplexConnectionDecorator.cs
+++ b/src/IceRpc/Internal/IceDuplexConnectionDecorator.cs
@@ -23,7 +23,7 @@ internal class IceDuplexConnectionDecorator : IDuplexConnection
             .ConfigureAwait(false);
 
         // Schedules or reschedules a keep alive after a successful connection establishment.
-        ScheduleKeepAlive();
+        ScheduleKeepAliveAfterWrite();
         return connectionInformation;
     }
 
@@ -83,13 +83,13 @@ internal class IceDuplexConnectionDecorator : IDuplexConnection
             // After each successful write, we (re)schedule one ping (heartbeat) at _writeIdleTimeout / 2 in the future.
             // Since each ping is itself a write, if there is no application activity at all, we'll send successive
             // pings at _writeIdleTimeout / 2 intervals.
-            ScheduleKeepAlive();
+            ScheduleKeepAliveAfterWrite();
         }
     }
 
     /// <summary>Constructs a decorator that ensures a call to <see cref="ReadAsync" /> will fail after readIdleTimeout.
-    /// This decorator also schedules a keepAliveAction after each write (see <see cref="ScheduleKeepAlive" />).
-    /// </summary>
+    /// This decorator also schedules a keepAliveAction after each write (see
+    /// <see cref="ScheduleKeepAliveAfterWrite" />).</summary>
     internal IceDuplexConnectionDecorator(
         IDuplexConnection decoratee,
         TimeSpan readIdleTimeout,
@@ -106,5 +106,6 @@ internal class IceDuplexConnectionDecorator : IDuplexConnection
     }
 
     /// <summary>Schedules one keep alive in writeIdleTimeout / 2.</summary>
-    private void ScheduleKeepAlive() => _keepAliveTimer.Change(_writeIdleTimeout / 2, Timeout.InfiniteTimeSpan);
+    private void ScheduleKeepAliveAfterWrite() =>
+        _keepAliveTimer.Change(_writeIdleTimeout / 2, Timeout.InfiniteTimeSpan);
 }
--- a/src/IceRpc/Internal/IceDuplexConnectionDecorator.cs
+++ b/src/IceRpc/Internal/IceDuplexConnectionDecorator.cs
@@ -1,0 +1,110 @@
+// Copyright (c) ZeroC, Inc.
+
+using IceRpc.Transports;
+using System.Buffers;
+using System.Diagnostics;
+
+namespace IceRpc.Internal;
+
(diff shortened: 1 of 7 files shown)


src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -598,6 +571,48 @@ async Task PerformDisposeAsync()
...
+                Debug.Fail($"The sending of a Ping frame failed with an unexpected exception: {exception}");
+                throw;
+            }
+        }
+
+        void SendReadPing()
+        {
+            // ReadKeepAlive is no-op unless _pendingPongCount == 0 before the Increment below: we send a Ping only if
+            // there is no outstanding Pong.
+
```
externl, line 596: I'd remove this line for consistency with SendWritePing
```suggestion
```

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1190,8 +1205,8 @@ async Task ReadPongFrameAsync(int size, CancellationToken cancellationToken)
                     (ref SliceDecoder decoder) => new PongBody(ref decoder),
                     cancellationToken).ConfigureAwait(false);
 
-                // For now, we only send a 0 payload value.
-                if (pongBody.Payload != 0L)
+                // For now, we only send a 0 and 1 payload value.
```
externl, line 1208:
```suggestion
                // For now, we only send a 0 or 1 payload value.
```

src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionDecorator.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            try
+            {
+                using CancellationTokenRegistration _ = cancellationToken.UnsafeRegister(
+                    cts => ((CancellationTokenSource)cts!).Cancel(),
+                    _readCts);
+                _readCts.CancelAfter(_idleTimeout); // enable idle timeout before reading
+
+                int bytesRead = await _decoratee.ReadAsync(buffer, _readCts.Token).ConfigureAwait(false);
+                // Debug.Assert(bytesRead > 0); // TODO: uncomment when bug is fixed
```
externl, line 49: What bug? It would be nicer to link to a GitHub issue.
```suggestion
                // Debug.Assert(bytesRead > 0); // TODO: uncomment when bug is fixed
```

src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionDecorator.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+        Debug.Assert(idleTimeout != Timeout.InfiniteTimeSpan);
+        _idleTimeout = idleTimeout;
+
+        SchedulePingAfterRead();
+        SchedulePingAfterWrite();
+    }
+
+    /// <summary>Schedules one ping in idleTimeout * 0.5.</summary>
+    private void SchedulePingAfterRead() => _readTimer?.Change(_idleTimeout * 0.5, Timeout.InfiniteTimeSpan);
```
externl, line 113: The name of these functions describes the behavior around where you use them, not what they actually do. 

This function doesn't schedule a print after reading.

How about `ResetReadTimer` or `UpdateReadTimer`

src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionDecorator.cs
```diff
@@ -0,0 +1,117 @@
...
+
+    /// <summary>Schedules one ping in idleTimeout * 0.6.</summary>
```
externl, line 115: I would maybe mention that 0.6 is specifically chosen to be bigger than the read time value. It's not just an arbitrary magic number.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
externl: We should document somewhere that 0L is for write timer pings, and 0 is for read timer. Is there a benefit to using a different value?

externl (approved):
Looks good. See comments.

src/IceRpc/Internal/IceDuplexConnectionDecorator.cs
```diff
@@ -1,31 +1,39 @@
...
+    private readonly TimeSpan _writeIdleTimeout;
 
-    public Task<TransportConnectionInformation> ConnectAsync(CancellationToken cancellationToken) =>
-        _decoratee.ConnectAsync(cancellationToken);
+    public async Task<TransportConnectionInformation> ConnectAsync(CancellationToken cancellationToken)
+    {
+        TransportConnectionInformation connectionInformation = await _decoratee.ConnectAsync(cancellationToken)
+            .ConfigureAwait(false);
+
+        // Schedules or reschedules a keep alive after a successful connection establishment.
```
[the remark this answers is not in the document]
bernardnormier: On the server-side, we reschedule here: the server sends a ValidateConnection to the client during the connection establishment. It's a write that schedules a "keep alive".

InsertCreativityHere (approved):
Looks sound to me!

src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionDecorator.cs
bernardnormier: Fixed

src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionDecorator.cs
bernardnormier: Fixed as suggested

src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionDecorator.cs
bernardnormier: Added comment.

externl (approved):

src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionDecorator.cs
```diff
@@ -0,0 +1,122 @@
...
+/// decorates both <see cref="ReadAsync"/> and <see cref="WriteAsync" /> to schedule pings that prevent both the local
+/// and remote idle timers from expiring.</summary>
+internal class SlicDuplexConnectionDecorator : IDuplexConnection
+{
+    private readonly IDuplexConnection _decoratee;
+    private TimeSpan _idleTimeout = Timeout.InfiniteTimeSpan;
+    private readonly CancellationTokenSource _readCts = new();
+
+    private Timer? _readTimer;
+    private Timer? _writeTimer;
```
pepone, line 18: The timers can be read only

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -265,9 +264,7 @@ async Task<TransportConnectionInformation> PerformConnectAsync()
 
             if (idleTimeout != Timeout.InfiniteTimeSpan)
             {
-                // Only client connections send ping frames when idle to keep the server connection alive. The server
-                // sends back a Pong frame in turn to keep alive the client connection.
-                _enableIdleTimeoutAndKeepAlive(idleTimeout, IsServer ? null : KeepAlive);
+                _duplexConnection.Enable(idleTimeout);
```
pepone, line 267: Would be better to keep `IDuplexConnection` and then do 

```
if (_duplexConnection is SlicDuplexConnectionDecorator duplexConnectionDecoator)
{
    duplexConnectionDecoator.Enable();
} 
```

Then we can make `SlicDuplexConnectionDecorator` timers non-nullable, and keep a single constructor.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -598,6 +571,47 @@ async Task PerformDisposeAsync()
...
+            catch (Exception exception)
+            {
+                Debug.Fail($"The sending of a Ping frame failed with an unexpected exception: {exception}");
+                throw;
+            }
+        }
+
+        void SendReadPing()
+        {
+            // ReadKeepAlive is no-op unless _pendingPongCount == 0 before the Increment below: we send a Ping only if
```
pepone, line 594: What is ReadKeepAlive there is no such operation?

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -598,6 +571,47 @@ async Task PerformDisposeAsync()
...
+            // there is no outstanding Pong.
```
pepone, line 595: is `outstanding Pong.` correct here?

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -598,6 +571,47 @@ async Task PerformDisposeAsync()
...
+            if (Interlocked.Increment(ref _pendingPongCount) == 1)
+            {
+                SendPing(1L);
+            }
+            else
+            {
+                Interlocked.Decrement(ref _pendingPongCount);
```
pepone, line 602: Can you explain this? It is not obvious to me why the call to SendReadPing decrements the pong count here.

src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionDecorator.cs
bernardnormier: Fixed

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
bernardnormier: I don't see how this would work.

The difficulty here is the idle timeout is negotiated during connection establishment. At construction time, we don't know if the negotiated idle timeout is good to be infinite or some other value.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
bernardnormier: Fixed. It's the current local function, SendReadPing.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
bernardnormier: Replaced by pending Pong, like the field name.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
pepone: We can pass the `idleTimeout` I forget to add it there, and that is not the point. The point is to avoid creating this decorator for the Server connections.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
bernardnormier:
```csharp
           if (Interlocked.Increment(ref _pendingPongCount) == 1)
            {
                SendPing(1L);
            }
            else
            {
                Interlocked.Decrement(ref _pendingPongCount);
            }
```

The code is quite simple. 

1 - we always increment _pendingPongCount
2 - if the incremented value is 1 -- meaning there was no pending Pong -- we send a Ping. The _pendingPongCount will be decremented when we receive the Pong
3 - if the incremented value has a different value, we decremented it to make the whole local function overall no-op

pepone: We can do the same with `CompareExchange` without increment/decrement
```
if (Interlocked.CompareExchange(ref _pendingPongCount, 1, 0) == 0)
{
    SendPing(1L);
}
```

pepone (approved):

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
bernardnormier: We need the decorator for server connections too. It implements the idle timer.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
bernardnormier: Updated as suggested

PR #3675, opened 2023-10-04
New Thermostat example

bernardnormier:
This PR adds a new thermostat example.


examples/Thermostat/Client/Client.csproj
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net7.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <!-- Copy the PDBs from the NuGet packages to get file names and line numbers in stack traces. -->
```
ReeceHumphreys, line 8: I know this was a generated file but should we include these comments?

examples/Thermostat/Client/Program.cs
```diff
@@ -0,0 +1,88 @@
...
+using Igloo;
+using Microsoft.Extensions.Logging;
+using System.CommandLine;
+
+using ILoggerFactory loggerFactory = LoggerFactory.Create(builder =>
+    builder
+        .AddSimpleConsole()
+        .AddFilter("IceRpc", LogLevel.Information));
+
+// Create a client connection to the cloud server.
```
ReeceHumphreys, line 13:
```suggestion
// Create a client connection to the "cloud" server.
```

To avoid confusion I would always put double quoutes around cloud.

examples/Thermostat/README.md
```diff
@@ -0,0 +1,87 @@
+# Thermostat
+
+This larger example presents an application with a client (think app running on your phone) that interacts with a device
```
ReeceHumphreys, line 3:
```suggestion
This larger example presents an application with a client (ex. an app running on your phone) that interacts with a device
```

I think saying think three times in a row feels a bit repetitive. Breaking this up with ex. might be good.

examples/Thermostat/README.md
```diff
@@ -0,0 +1,87 @@
...
+(think thermostat in your home) via an intermediary server (think server deployed on a public cloud).
```
ReeceHumphreys, line 4:
```suggestion
(ex. a server deployed on a public cloud) via an intermediary server (think server deployed on a public cloud).
```

examples/Thermostat/README.md
ReeceHumphreys, line 3:
```suggestion
This more extensive example presents an application with a client (think app running on your phone) that interacts with a device
```

I think large almost have a bit of a negative context, I think more extensive is a better word choice.

examples/Thermostat/Server/ThermoBridge.cs
```diff
@@ -0,0 +1,42 @@
...
+        IAsyncEnumerable<Reading> readings,
+        IFeatureCollection features,
+        CancellationToken cancellationToken)
+    {
+        IDispatchInformationFeature? dispatchInfo = features.Get<IDispatchInformationFeature>();
+
+        // We installed the dispatch middleware in Router to get this dispatch info:
+        Debug.Assert(dispatchInfo is not null);
+
+        _deviceConnection.SetInvoker(dispatchInfo.ConnectionContext.Invoker);
```
pepone, line 27: I would add a comment here something like:

A device has called home to start reporting readings, we update the device connection invoker, with the invoker from the connection context, this invoker allows to send requests to the just connected device

pepone (approved):
Looks great!

examples/Thermostat/slice/Reading.slice
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) ZeroC, Inc.
+
+module Igloo
+
+/// Represents a snapshot of the thermostat state at a point in time.
```
InsertCreativityHere, line 5:
```suggestion
/// Represents a snapshot of the thermostat's state at a point in time.
```

examples/Thermostat/slice/Reading.slice
```diff
@@ -0,0 +1,21 @@
...
+struct Reading {
+    /// The point in time where this Reading was taken.
```
InsertCreativityHere, line 7:
```suggestion
    /// The point in time when this Reading was taken.
```
Since we're talking about a point in time (when), not a point in space (where).

examples/Thermostat/slice/Thermostat.slice
```diff
@@ -0,0 +1,15 @@
...
+
+/// Monitors and controls a residential cool-only AC system.
+/// This interface corresponds to the public API used by client applications.
+interface Thermostat {
+
+    /// Returns a new {@link Reading} immediately and then every time the actual device reports a value.
+    monitor() -> stream Reading
+
+    /// Changes the target temperature. If it's below the current temperature, the AC will start cooling.
+    /// @param setPoint: The target temperature in degrees Fahrenheit
```
InsertCreativityHere, line 13: This is missing a period, and I'd write this like we do in `Reading`.
```suggestion
    /// @param setPoint: The target temperature (in °F).
```

examples/Thermostat/slice/ThermoControl.slice
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) ZeroC, Inc.
+
+module Igloo
+
+/// Controls the actual thermostat.
+interface ThermoControl {
+    /// Sets a new target temperature (known as set point).
+    /// @param setPoint: The new target temperature.
```
InsertCreativityHere, line 8: Should we specify this is `(in °F)`? We seem to mention that fact everywhere else we talk about the temperature.

examples/Thermostat/slice/Reading.slice
```diff
@@ -0,0 +1,21 @@
...
+
+    /// The target temperature (in °F).
+    setPoint: float32
+
+    /// The state of the AC unit.
+    cooling: Stage
+}
+
+/// Represents the possible states of the AC unit.
+enum Stage : uint8 { Off, Stage1, Stage2 }
```
InsertCreativityHere, line 21: It might be worth writing short doc comments for these enumerators.
Most people who don't obsess over A/C systems like us will be unfamiliar with "stages".

examples/Thermostat/slice/Reading.slice
```diff
@@ -0,0 +1,21 @@
...
+/// Represents a snapshot of the thermostat state at a point in time.
+struct Reading {
+    /// The point in time where this Reading was taken.
+    timeStamp: WellKnownTypes::TimeStamp
+
+    /// The current temperature (in °F).
+    temperature: float32
+
+    /// The target temperature (in °F).
+    setPoint: float32
```
InsertCreativityHere, line 14: After reading the whole example, I understand, but I was initially confused by these names. I think these would be clearer:
```suggestion
    /// The actual temperature (in °F).
    actual_temperature: float32

    /// The target temperature (in °F).
    target_temperature: float32
```

1) While skimming, I thought "current temperature" was the "current _target_ temperature". Actual temperature better reflects what it's measuring.
`temperature` stores the current actual temperature
`setPoint` stores the current target temperature.

2) Most readers will be unfamiliar with the term 'setPoint', even though we explain it one place. Plus, in our comments, we always say "target temperature", might as well embrace it.
Also, then `changeSetPoint` could be renamed to `setTargetTemperature`.
getters and setters I'm familiar with, but a function named `changeX` sounds... weird...

examples/README.md
```diff
@@ -22,6 +22,7 @@ a specific feature or programming technique.
 | [Secure](./Secure/)                   | Shows how to secure TCP connections with TLS.                                                                   |
 | [Stream](./Stream/)                   | Shows how to stream data from a client to a server.                                                             |
 | [Telemetry](./Telemetry/)             | Shows how to use the telemetry interceptor and middleware.                                                      |
+| [Thermostat](./Thermostat/)           | Shows how to send requests via an intermediary server; includes sending requests the "other way around", from a server to a client.|
```
InsertCreativityHere, line 25: I personally like the way this is written, but it would be nice to have it with fit better.
Both fit on my screen (it's too long) and fit within the markdown table (it pushes the column right) How about: 
```suggestion
| [Thermostat](./Thermostat/)           | Shows how to send requests via an intermediary server and how to send requests from a server to a client.       |
```

examples/Thermostat/Device/ThermoBot.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    private Stage _cooling;
+
+    // Protects all read-write fields.
+    private readonly object _mutex = new();
+
+    // Used to signal when the server has stopped reading.
+    private readonly TaskCompletionSource _readTcs = new(TaskCreationOptions.RunContinuationsAsynchronously);
+
+    private float _setPoint = 75.0F;
```
InsertCreativityHere, line 25: Total coincidence, but it's kind of nice how looks and feels like `75°F` == `75.0F`

examples/Thermostat/Device/ThermoBot.cs
```diff
@@ -0,0 +1,104 @@
...
+            }
+            _setPoint = setPoint;
+            if (_temperature > setPoint)
+            {
+                _cooling = _temperature - setPoint > 5.0F ? Stage.Stage2 : Stage.Stage1;
+            }
+            else
+            {
+                _cooling = Stage.Off;
+            }
```
InsertCreativityHere, line 50: I think this would look cleaner as a switch expression, instead of (what's effectively) nested ternary operators.
```suggestion
            _cooling = (_temperature - setPoint) switch {
			    > 5.0F => Stage.Stage2,
			    > 0.0F => Stage.Stage1,
			    _ => Stage.Off,
            };
```
Maybe this is too fancy though.

examples/Thermostat/README.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+Then in another window, start the Client program:
+
+```shell
+cd Client
+dotnet run
+```
+
+You can start multiple instances of the Client program. When you start Client without argument, it monitors the device
```
InsertCreativityHere, line 78: Maybe this? I'm not sure if you were trying to say something else though.
```suggestion
You can start multiple instances of the Client program. When you start Client without any arguments, it monitors the device
```

InsertCreativityHere (commented):
Looks good so far, but still reviewing!

examples/Thermostat/Client/Program.cs
```diff
@@ -0,0 +1,88 @@
...
+await using var connection = new ClientConnection(
+    new Uri("icerpc://localhost"),
+    logger: loggerFactory.CreateLogger<ClientConnection>());
+
+Pipeline pipeline = new Pipeline()
+    .UseDeadline(TimeSpan.FromSeconds(3))
+    .UseLogger(loggerFactory)
+    .Into(connection);
+
+var proxy = new ThermostatProxy(pipeline);
```
externl, line 23: I think it would be clearer to rename "proxy" to "thermostat". 

I think `thermostat.ChangeSetPointAsync()` vs `proxy.ChangeSetPointAsync()`.

examples/Thermostat/slice/Reading.slice
bernardnormier: "set point" is the correct term. Target temperature is a way to describe "set point" for people unfamiliar with this term.

examples/Thermostat/slice/ThermoControl.slice
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) ZeroC, Inc.
+
+module Igloo
+
+/// Controls the actual thermostat.
```
externl, line 5: This comment seems to vaguely reference other interfaces that maybe control a not actual thermostat. 
```suggestion
/// Controls a physical thermostat device.
```

examples/Thermostat/Client/Client.csproj
bernardnormier: It's not a generated file. We have this comment in all our example project files.

examples/Thermostat/slice/Thermostat.slice
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) ZeroC, Inc.
+
+module Igloo
+
+/// Monitors and controls a residential cool-only AC system.
+/// This interface corresponds to the public API used by client applications.
+interface Thermostat {
```
externl, line 7: I think we're somehow missing out here by having two interfaces with `changeSetPoint(setPoint: float32)`.  One option is to rename this interface to `ThermoMonitor` and remove the `chagneSetPoint` operations. The implement would then implement `ThermoMonitor` and `ThermoControl`.

examples/Thermostat/slice/Reading.slice
ReeceHumphreys: > Most readers will be unfamiliar with the term 'setPoint', even though we explain it one place

I agree with Austin on this while 'set point' is the more correct term I feel like the example is clearer for people who do not know much about AC and Thermostats with the terms "actual" and "target" temp.

examples/Thermostat/Device/ThermoBot.cs
ReeceHumphreys: > Maybe this is too fancy though.

I think it's too fancy. I think its more readable with the if statement.

examples/Thermostat/Client/Program.cs
ReeceHumphreys: I agree I think it's cleaner. If we wanted to keep the idea of it being a proxy everywhere we could also do `thermostatProxy`

examples/Thermostat/Device/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+using Igloo;
+using Microsoft.Extensions.Logging;
+using ThermostatDevice;
+
+using ILoggerFactory loggerFactory = LoggerFactory.Create(builder =>
+    builder
+        .AddSimpleConsole()
+        .AddFilter("IceRpc", LogLevel.Debug));
+
+var thermoBot = new ThermoBot();
```
ReeceHumphreys, line 14: Maybe `IglooThermostat` instead of `ThermoBot`. I find the ThermoBot name a bit confusing.

externl (approved):

examples/Thermostat/Server/ThermoBridge.cs
```diff
@@ -0,0 +1,42 @@
...
+
+using IceRpc.Features;
+using IceRpc.Slice;
+using Igloo;
+using System.Diagnostics;
+
+namespace ThermostatServer;
+
+/// <summary>Implements Slice interface `ThermoHome`.</summary>
+
```
externl, line 11:
```suggestion
```

examples/Thermostat/Client/Program.cs
bernardnormier: Removed cloud. There is no cloud server here.

examples/Thermostat/README.md
bernardnormier: Reworked text.

examples/Thermostat/README.md
bernardnormier: Removed larger.

examples/Thermostat/Server/ThermoFacade.cs
```diff
@@ -0,0 +1,176 @@
...
+
+    /// <summary>Changes the target temperature by forwarding the call to the device.</summary>
+    public async ValueTask ChangeSetPointAsync(
+        float setPoint,
+        IFeatureCollection features,
+        CancellationToken cancellationToken)
+    {
+        try
+        {
+            await _thermoControl.ChangeSetPointAsync(setPoint, cancellationToken: cancellationToken);
```
pepone, line 38: I think it is uncommon to use an argument label in this case:

```suggestion
            await _thermoControl.ChangeSetPointAsync(setPoint, cancellationToken);
```

examples/Thermostat/slice/ThermoControl.slice
bernardnormier: Fixed

examples/Thermostat/slice/Reading.slice
bernardnormier: Added enumerator doc-comments.

examples/Thermostat/slice/Reading.slice
bernardnormier: I much prefer set point. It's the correct term.

externl: I think set point is fine.

examples/README.md
bernardnormier: You're supposed to read the rendered version, not the GitHub review diff. 

Then, your "formatting" update changes the semantics. The example includes this server->client requests. The big picture is requests through an intermediary server. Both are not at the same level.

examples/Thermostat/README.md
bernardnormier: Reworked text

examples/Thermostat/Client/Program.cs
bernardnormier: I renamed it to thermostatProxy. `thermostat` would be more appropriate if we put it in an `IThermostat` variable.

examples/Thermostat/slice/Thermostat.slice
bernardnormier: I discussed this with Joe and I prefer keeping the Thermostat interface as-is.

examples/Thermostat/Server/ThermoFacade.cs
bernardnormier: How does this work? We have a features parameter in between.

examples/Thermostat/Device/Program.cs
bernardnormier: It should be ThermoSomething. 

I first wanted to use ThermoDevice, but the namespace is already ThermostatDevice so it's too repetitive.

I picked ThermoBot because it's the actual service (just like ChatBot in the Greeter examples) and it's a fake / (ro)bot. Can you think of a better name?

examples/Thermostat/Server/ThermoFacade.cs
pepone: You are right, we need the label

PR #3681, opened 2023-10-10
Upgrades to use .NET 8 RC1

### LINKED ISSUE #3678: Upgrade to .NET 8
We want to upgrade the code to build only with .NET 8, and remove a few .NET 7 work-arounds in the source code.

pepone:
Fix #3678


src/IceRpc/Internal/ProtocolLoggerExtensions.cs
```diff
@@ -62,7 +62,7 @@ internal static partial class ProtocolLoggerExtensions
             remoteNetworkAddress);
 
     // Multiple logging methods are using same event id.
-#pragma warning disable SYSLIB1006
+#pragma warning disable SYSLIB1025
```
externl, line 65: The number was changed?

pepone: Yes I get this new error with Visual Studio 2022 preview

externl (approved):

src/IceRpc/Internal/ProtocolLoggerExtensions.cs
bernardnormier: I think we should use 2 separate EventId here - for another PR!

bernardnormier (approved):
Looks good.

PR #3685, opened 2023-10-11
Connect_fails_if_listener_is_disposed test simplifications

pepone:
- For the QUIC Connect_fails_if_listener_is_disposed conformance test, successful connections must be disposed to ensure timely failures for new connections.
- Updated the CI job to use a larger 8-core runner.
- Limit test runners to 8 globally: some SSL tests were running slow with higher concurrency.

### DIFF
--- a/tests/IceRpc.Conformance.Tests/Transports/MultiplexedListenerConformanceTests.cs
+++ b/tests/IceRpc.Conformance.Tests/Transports/MultiplexedListenerConformanceTests.cs
@@ -110,16 +110,11 @@ public abstract class MultiplexedListenerConformanceTests
         IListener<IMultiplexedConnection> listener = provider.GetRequiredService<IListener<IMultiplexedConnection>>();
         var clientTransport = provider.GetRequiredService<IMultiplexedClientTransport>();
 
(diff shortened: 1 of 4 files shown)


tests/IceRpc.Conformance.Tests/Transports/MultiplexedListenerConformanceTests.cs
```diff
@@ -109,43 +109,20 @@ public async Task Connect_fails_if_listener_is_disposed()
         IListener<IMultiplexedConnection> listener = provider.GetRequiredService<IListener<IMultiplexedConnection>>();
         var clientTransport = provider.GetRequiredService<IMultiplexedClientTransport>();
 
-        Task connectTask = ConnectAsync(clientTransport);
+        await using IMultiplexedConnection connection = clientTransport.CreateConnection(
```
bernardnormier, line 112: Should we make a successful connection first to make sure the ConnectAsync doesn't fail for some other reason?

pepone: Yes, I reworked this test

bernardnormier (approved):

PR #3689, opened 2023-10-12
Remove Proxy suffix from example variables

### LINKED ISSUE #3679: Rename greeterProxy and other proxy variables
We want to drop the Proxy suffix of most proxy variables in the examples. 

For example, in the Greeter client, it should be simply:
```csharp
var greeter = new GreeterProxy(connection);
```

Rationale:

1) In all these examples, we use the proxy as an IGreeter (or the applicable client interface). The "proxy" API is irrelevant. And yet we want to keep "var" (which exposes this proxy API) for conciseness.

2) If we passed a proxy as a parameter through a Slice operation, the parameter name would be typically Greeter, without a proxy suffix. For example:
```slice
op(greeter: Greeter) // the greeter is a GreeterProxy in C#
```

We could keep the Proxy suffix for variables in examples where we use the proxy API. This would typically happen when we encode a proxy parameter - but none of the examples currently show that.

bernardnormier:
This PR removes the Proxy suffix from many example variables. It's unnecessary / confusing since we don't use the proxy API (= change the invoker or encode the service address) but only the interface implemented by the struct.

For example:
```
var greeter = new GreeterProxy(...);
// the only thing that matters below is that greeter is an IGreeter. It's like a shortcut for
// IGreeter greeter = new GreeterProxy(...);
```

Fixes #3679.


pepone (approved):

externl (approved):
Looks good!

PR #3692, opened 2023-10-13
Replace Service class with a Service source generator.

pepone:
This PR introduces the following changes:

- Replaces the existing service class with a source generator.
- The source generator generates `IDispatcher` implementations for classes annotated with `SliceServiceAttribute`.
- The implementation is inspired by the .NET runtime `LoggerMessage` source generator. For reference, see [LoggerMessage Generator](https://github.com/dotnet/runtime/tree/e6f99449193cfc3fe12d46f7e8c9a8096d06ce53/src/libraries/Microsoft.Extensions.Logging.Abstractions/gen).


src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -0,0 +1,97 @@
...
+                serviceClass.IsSealed ? "public" : "public virtual";
+
+            string dispatcherBlock;
+            if (serviceClass.ServiceMethods.Count > 0)
+            {
+                string dispatchBlocks = "";
+                foreach (ServiceMethod serviceMethod in serviceClass.ServiceMethods)
+                {
+                    dispatchBlocks += @$"
+
```
externl, line 30: Is the goal of this extra space to provide spacing between each case statement? Maybe we could use a list instead to hold the blocks, then trim them and join with a double. newline separator later when added to `dispatcherBlock`?

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -0,0 +1,97 @@
...
+                containerDefinition = parent;
+            }
+
+            generated += WriteContainer($"namespace {serviceClass.ContainingNamespace}", container);
+            generated += $"{_newLine}{_newLine}";
+        }
+        return generated.TrimTrailingWhiteSpaces();
+    }
+
+    private static string WriteContainer(string header, string body)
```
externl, line 89: This function doesn't actually write anything. Maybe we can name it something like `GenerateContainer`

src/IceRpc.Slice.Generators/Internal/StringExtensions.cs
externl, line 1: These methods use `\n` for newlines. `Emitter` uses this function as well as special `\r\n` or `\n` when it generates its own newlines. Are the `\r\n` required?

src/IceRpc.Slice.Generators/ServiceGenerator.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        IEnumerable<ClassDeclarationSyntax> distinctClasses = classes.Distinct();
+
+        var parser = new Parser(compilation, context.ReportDiagnostic, context.CancellationToken);
+        IReadOnlyList<ServiceClass> serviceClasses = parser.GetServiceDefinitions(distinctClasses);
+        if (serviceClasses.Count > 0)
+        {
+            var emitter = new Emitter();
+            string result = emitter.Emit(serviceClasses, context.CancellationToken);
+            context.AddSource("Service.g.cs", SourceText.From(result, Encoding.UTF8));
```
externl, line 53: So all services classes get generated into one file?

src/IceRpc.Slice/SliceServiceAttribute.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) ZeroC, Inc.
+
+namespace IceRpc.Slice;
+
+/// <summary>Represents an attribute used to mark classes implementing Slice services.</summary>
+/// <remarks>The Slice source generator implements <see cref="IDispatcher"/> for classes marked with this attribute.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Class, Inherited = false)]
```
externl, line 8: Does this mean we can no longer use structs to implement services?

tests/IceRpc.Slice.Tests/OperationDerivedTests.slice
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) ZeroC, Inc.
+
+module IceRpc::Slice::Derived::Tests
+
+interface MyDerivedOperations :  IceRpc::Slice::Tests::MyBaseOperations {
```
externl, line 5:
```suggestion
interface MyDerivedOperations : IceRpc::Slice::Tests::MyBaseOperations {
```

externl (approved):
Still reviewing but have a few questions:
1. What's the expected behavior if you put `SliceService` on a class that doesn't implement any Slice operations? Should we error?

src/IceRpc.Slice.Generators/Internal/DiagnosticDescriptors.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.CodeAnalysis;
+
+namespace IceRpc.Slice.Generators.Internal;
+
+internal static class DiagnosticDescriptors
+{
+    internal static DiagnosticDescriptor DuplicateOperationNames { get; } = new DiagnosticDescriptor(
+        id: "SLICE1001",
+        title: "Multiple RPC operations cannot use the same name within a service class",
+        messageFormat: "Multiple RPC operations are using name {0} in class {1}",
```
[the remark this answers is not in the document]
bernardnormier: Same here

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Runtime.InteropServices;
+
+namespace IceRpc.Slice.Generators.Internal;
+
+internal class Emitter
+{
+    private static readonly string _newLine = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "\r\n" : "\n";
```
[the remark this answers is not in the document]
pepone: It is banned by the analyzer used by generators

There is a discussion about this at https://github.com/dotnet/roslyn-analyzers/issues/6467

bernardnormier:
You also need to update src/IceRpc.Slice/README.md

src/IceRpc.Slice.Generators/ServiceGenerator.cs
pepone: Yes there is a single generated file

src/IceRpc.Slice/SliceServiceAttribute.cs
pepone: Yes, this attribute only applies to classes.

src/IceRpc.Slice.Generators/Internal/Emitter.cs
pepone: Yes, this provides the spacing between each case statement, I will check if using a list simplifies things.

src/IceRpc.Slice/SliceServiceAttribute.cs
bernardnormier: I think that's fine:
 - it's not a regression (previously you have to have a class inherit class Service)
 - we could allow struct now but that's the point since you'd immediately box it to use it as a dispatcher

src/IceRpc.Slice.Generators/Internal/Emitter.cs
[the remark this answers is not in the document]
pepone: I reworked the code to only use `\n`

src/IceRpc.Slice.Generators/Internal/Emitter.cs
pepone: Reworked to use a List

src/IceRpc.Slice.Generators/Internal/StringExtensions.cs
pepone: Right, reworked to only use `\n`

pepone:
> You also need to update src/IceRpc.Slice/README.md

Updated

src/IceRpc.Slice.Generators/Internal/Parser.cs
```diff
@@ -0,0 +1,218 @@
...
+    internal const string OperationAttribute = "IceRpc.Slice.SliceOperationAttribute";
+    internal const string ServiceAttribute = "IceRpc.Slice.SliceServiceAttribute";
+    internal const string SliceTypeIdAttribute = "ZeroC.Slice.SliceTypeIdAttribute";
+
+    private readonly CancellationToken _cancellationToken;
+    private readonly Compilation _compilation;
+    private readonly INamedTypeSymbol? _operationAttribute;
+    private readonly Action<Diagnostic> _reportDiagnostic;
+    private readonly INamedTypeSymbol? _serviceAttribute;
+    private readonly INamedTypeSymbol? _typeIdAttribute;
```
bernardnormier, line 23: We should not use or reference type ID attributes anywhere in this code.

src/IceRpc.Slice.Generators/Internal/ServiceClass.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace IceRpc.Slice.Generators.Internal;
+
+/// <summary>Represents a C# class that has the <c>IceRpc.Slice.SliceServiceAttribute</c> attribute and for which this
+/// generator implements the <c>IceRpc.IDispatcher</c> interface.</summary>
+internal class ServiceClass : ContainerDefinition
+{
+    /// <summary>Gets a value indicating whether the service has a base service class.</summary>
+    internal bool HasBaseServiceClass { get; }
+
+    /// <summary>Gets the C# namespace containing this definition.</summary>
```
bernardnormier, line 12: Is this the namespace or the scope?

This service class can presumably be nested in some other class or struct.

src/IceRpc.Slice.Generators/ServiceGenerator.cs
bernardnormier: Is there anyway to see the generated file?

src/IceRpc.Slice.Generators/Internal/Parser.cs
```diff
@@ -0,0 +1,218 @@
...
+                default:
+                    Debug.Assert(false, "Unexpected number of arguments in attribute constructor.");
+                    break;
+            }
+
+            Debug.Assert(operationName is not null);
+
+            serviceMethods.Add(new ServiceMethod
+            {
+                OperationName = operationName!,
```
bernardnormier, line 194: Don't need "!"

src/IceRpc.Slice.Generators/Internal/ServiceMethod.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) ZeroC, Inc.
+
+namespace IceRpc.Slice.Generators.Internal;
+
+/// <summary>Represents an RPC operation annotated with the <c>IceRpc.Slice.SliceOperationAttribute</c> attribute.
+/// </summary>
+internal record struct ServiceMethod
```
bernardnormier, line 7: Given the usage, I think it should be a readonly record struct.

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -0,0 +1,90 @@
...
+                serviceClass.IsSealed ? "public" : "public virtual";
+
+            string dispatcherBlock;
+            if (serviceClass.ServiceMethods.Count > 0)
+            {
+                var dispatchBlocks = new List<string>();
+                foreach (ServiceMethod serviceMethod in serviceClass.ServiceMethods)
+                {
+                    dispatchBlocks.Add(@$"
+case ""{serviceMethod.OperationName}"":
```
bernardnormier, line 26: Why don't you generate a switch expression here?

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -0,0 +1,90 @@
...
+        foreach (ServiceClass serviceClass in serviceClasses)
+        {
+            // stop if we're asked to.
+            cancellationToken.ThrowIfCancellationRequested();
+
+            string methodModifier =
+                serviceClass.HasBaseServiceClass ? "public override" :
+                serviceClass.IsSealed ? "public" : "public virtual";
+
+            string dispatcherBlock;
```
bernardnormier, line 19: We should not have a variable called "dispatcherBlock" and another called "dispatchBlocks". That's too confusing. Also, what is a "block"?

src/IceRpc.Slice.Generators/ServiceGenerator.cs
pepone: In Visual Studio you can see the generated code by expanding the analyzer

https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/media/source-generators/solution-explorer-program.png#lightbox

src/IceRpc.Slice.Generators/Internal/ServiceClass.cs
pepone: > Is this the namespace or the scope?

This contains both, mimics the nesting of the partial definition with the SliceService attribute

src/IceRpc.Slice.Generators/Internal/Parser.cs
```diff
@@ -0,0 +1,218 @@
...
+using System.Diagnostics;
+
+namespace IceRpc.Slice.Generators.Internal;
+
+internal sealed class Parser
+{
+    internal const string IceObjectService = "IceRpc.Slice.Ice.IIceObjectService";
+    internal const string OperationAttribute = "IceRpc.Slice.SliceOperationAttribute";
+    internal const string ServiceAttribute = "IceRpc.Slice.SliceServiceAttribute";
+    internal const string SliceTypeIdAttribute = "ZeroC.Slice.SliceTypeIdAttribute";
```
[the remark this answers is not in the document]
pepone: It is also a leftover

src/IceRpc.Slice.Generators/Internal/ServiceMethod.cs
pepone: The generator targets `netstandard2.0` we cannot use `init` properties, aren't these required to make it readonly?

src/IceRpc.Slice.Generators/Internal/Emitter.cs
pepone: it was to mean a code block

src/IceRpc.Slice.Generators/Internal/Emitter.cs
pepone: I will update the code to do so

src/IceRpc.Slice.Generators/Internal/ServiceMethod.cs
bernardnormier: Given the usage, you don't need init property either. You can just use the primary constructor. You could also simplify your record struct:
```
internal readonly record struct ServiceMethod(...);
```

src/IceRpc.Slice.Generators/Internal/ServiceClass.cs
bernardnormier: So it should be scope? For me, scope corresponds to both namespace(s) and enclosing structs/classes.

src/IceRpc.Slice.Generators/ServiceGenerator.cs
bernardnormier: It does indeed work with Visual Studio.

src/IceRpc.Slice.Generators/IceRpc.Slice.Generators.csproj
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+    <IsPackable>false</IsPackable>
+    <developmentDependency>true</developmentDependency>
```
bernardnormier, line 7: Is this correct?
The camel case developmentDependency looks like an attribute name, not a property name.

src/IceRpc.Slice/SliceServiceAttribute.cs
bernardnormier, line 8: Reading between the lines here: `Inherited = false` because we want users to always specify explicitly with `SliceService` which classes should be extended via generated code. 

For example, in this case:
```csharp
[SliceService] internal partial class Chatbot : IGreeter { ... }

internal class SuperChatbot : Chatbot { ... }
```

We don't want the Slice Service code generator to extend SuperChatbot. And it wouldn't work anyway since SuperChatbot is not marked partial. If I want generated code for SuperChatbot, I need both partial and SliceService, as in:
```csharp
[SliceService] internal partial class SuperChatbot : Chatbot, ISuperHero { ... }
```

It would be good to clarify this point in the summary or the remarks.

src/IceRpc.Slice.Generators/Internal/Parser.cs
```diff
@@ -0,0 +1,202 @@
...
+    {
+        foreach (AttributeData attribute in classSymbol.GetAttributes())
+        {
+            if (SymbolEqualityComparer.Default.Equals(attribute.AttributeClass, _serviceAttribute))
+            {
+                return true;
+            }
+        }
+
+        if (classSymbol.BaseType is INamedTypeSymbol baseType)
```
bernardnormier, line 195: I don't understand why we do that.

We go out of our way to mark the SliceService attribute Inherited = false, and yet here we implement the equivalent of Inherited = true behavior here.

src/IceRpc.Slice.Generators/Internal/Parser.cs
```diff
@@ -0,0 +1,202 @@
...
+            serviceMethods.Add(new ServiceMethod
+            {
+                OperationName = operationName,
+                DispatchMethodName = GetFullName(method)
+            });
+        }
+        return serviceMethods;
+    }
+
+    private bool HasServiceAttribute(INamedTypeSymbol classSymbol)
```
bernardnormier, line 185: I was confused by this code (see below). It would be good to rename classSymbol to baseClassSymbol and add a comment like:
```csharp
<summary>Checks whether this base class or any of its base class(es) has the SliceService attribute.
When true, we'll override the DispatchAsync method generated for this base class; otherwise, we'll implement the IDispatcher interface.</summary>
```

src/IceRpc.Slice.Generators/Internal/Parser.cs
```diff
@@ -0,0 +1,202 @@
...
+                SemanticModel semanticModel = _compilation.GetSemanticModel(classDeclaration.SyntaxTree);
+                INamedTypeSymbol? classSymbol = semanticModel.GetDeclaredSymbol(classDeclaration, _cancellationToken);
+                if (classSymbol is null)
+                {
+                    continue;
+                }
+
+                IReadOnlyList<ServiceMethod> baseServiceMethods = Array.Empty<ServiceMethod>();
+                bool hasBaseServiceDefinition = false;
+                if (classSymbol.BaseType is INamedTypeSymbol baseType)
```
bernardnormier, line 61: The logic does not seem correct to me. Take:
```csharp
[SliceService] internal partial class MyBase : IBase { ... }

internal class MyDerived : MyBase, IDerived { ... } // no SliceService attribute, so no code generation

[SliceService] internal partial class MyMoreDerived : MyDerived, IMoreDerived { ...}
```

I believe you'll generate code that skips the dispatching to operations on Derived.

The `baseServiceMethods` that you filter out must correspond to the methods on the nearest base class where you found the SliceService attribute.

Would be good to add a test!

src/IceRpc.Slice.Generators/Internal/Parser.cs
bernardnormier: Maybe you should replace this method by:
```csharp
/// <summary>Returns the nearest base class with the SliceService attribute.</summary>
/// <value>The nearest base class with the SliceService attribute. null when there is no such base class.</value>
private INamedTypeSymbol? GetBaseServiceClass(INamedTypeSymbol classSymbol)
```

See other comment about the base class logic.

src/IceRpc.Slice.Generators/IceRpc.Slice.Generators.csproj
pepone: Fixed the case

src/IceRpc.Slice/SliceServiceAttribute.cs
pepone: Right, the generator only needs to implement IDispatcher for the classes containing the attribute, we don't need to provide another implementation for derived classes.

I will add a comment.

src/IceRpc.Slice.Generators/Internal/ServiceMethod.cs
pepone: That doesn't work with .netstandard2.0, it results in:

```
Predefined type 'System.Runtime.CompilerServices.IsExternalInit' is not defined or imported[CS0518](https://msdn.microsoft.com/query/roslyn.query?appId%3Droslyn%26k%3Dk%28CS0518%29)
```

src/IceRpc.Slice/IceRpc.Slice.csproj
```diff
@@ -28,17 +28,24 @@
     <ItemGroup>
       <ProjectReference Include="..\IceRpc\IceRpc.csproj" ExactVersion="true" />
       <ProjectReference Include="..\ZeroC.Slice\ZeroC.Slice.csproj" ExactVersion="true" />
-        <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
-        <PackageReference Include="StyleCop.Analyzers" Version="$(StyleCopAnalyzersVersion)">
-            <PrivateAssets>all</PrivateAssets>
-            <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-        </PackageReference>
+      <!-- Includes the analyzer in the package without consuming it. -->
```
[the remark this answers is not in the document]
pepone: This come straight from the docs see https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/libraries-packaging.md#analyzers--source-generators

src/IceRpc.Slice/IceRpc.Slice.csproj
```diff
@@ -28,17 +28,24 @@
...
-        <Content Include="README.md" Pack="true" PackagePath="\" />
-        <None Include="..\..\build\icerpc-icon.png" Pack="true" PackagePath="\" />
-        <Content Include="IceRpc.Slice.props" Pack="true" PackagePath="buildTransitive" />
+      <Content Include="..\..\LICENSE" Pack="true" PackagePath="\" />
+      <Content Include="README.md" Pack="true" PackagePath="\" />
+      <None Include="..\..\build\icerpc-icon.png" Pack="true" PackagePath="\" />
+      <Content Include="IceRpc.Slice.props" Pack="true" PackagePath="buildTransitive" />
+      <Content Include="..\IceRpc.Slice.Generators\bin\$(Configuration)\netstandard2.0\IceRpc.Slice.Generators.dll"
+               Pack="true"
+               PackagePath="analyzers\dotnet\roslyn4.4\cs"/>
```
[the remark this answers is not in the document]
pepone: `analyzers\dotnetcs` is probably enough here as we don't build for different roslyn APIs

src/IceRpc.Slice.Generators/Internal/Emitter.cs
pepone: Reworked this code

src/IceRpc.Slice.Generators/Internal/ContainerDefinition.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) ZeroC, Inc.
+
+namespace IceRpc.Slice.Generators.Internal;
+
+/// <summary>Represents a C# definition that is the parent of a service definition.</summary>
```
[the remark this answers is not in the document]
pepone: Replaced with "enclosing"

src/IceRpc.Slice.Generators/Internal/ServiceClass.cs
pepone: Renamed as "Scope"

src/IceRpc.Slice.Generators/Internal/Parser.cs
pepone: Added a test

src/IceRpc.Slice.Generators/Internal/ServiceClass.cs
pepone: I was mistaken here, this is just the namespace and it can be null

See https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.isymbol.containingnamespace?view=roslyn-dotnet-4.6.0

tests/IntegrationTests/Pingable.slice
```diff
@@ -4,6 +4,6 @@ mode = Slice1
 
 module IceRpc::IntegrationTests
 
-interface Pingable {
+interface Pingable : Ice::Object {
```
bernardnormier, line 7: I don't understand this update, and I'd rather not do that.

tests/IceRpc.Slice.Tests/ServiceTests.slice
```diff
@@ -11,3 +11,15 @@ interface Service2 {
...
+
+interface Base {
+    op1()
+}
+
+interface Derived : Base {
+    op2()
+}
+
+interface MoreDerived : Base {
```
bernardnormier, line 23: I believe you meant "MoreDeviced : Derived"

tests/IceRpc.Slice.Tests/ServiceTests.cs
```diff
@@ -1,26 +1,41 @@
...
+        var proxy = new MoreDerivedProxy(invoker);
 
-        public ValueTask Op3Async(IFeatureCollection features, CancellationToken cancellationToken) =>
-            throw new NotImplementedException();
+        // Act/Assert
+        Assert.That(async () => await proxy.Op3Async(), Throws.Nothing);
     }
 }
+
+    [SliceService]
```
bernardnormier, line 24: Extra space

tests/IceRpc.Slice.Tests/ServiceTests.cs
```diff
@@ -1,26 +1,41 @@
...
+internal partial class MyBaseService : IBaseService
+{
+    public ValueTask Op1Async(IFeatureCollection features, CancellationToken cancellationToken) => default;
+}
+
+// This class doesn't use the SliceService attribute, and no dispatch method would be generated for it. It would
+// be possible to call `IDerivedService` using the code generated for the derived class `MyMoreDerivedService`.
+internal partial class MyDerivedService : MyBaseService, IDerivedService
```
bernardnormier, line 32: I would remove "partial".

bernardnormier: You also need a test that calls Op2Async. Otherwise, how do you test we generate the correct DispatchAsync for MyMoreDerivedService?

bernardnormier (approved):
Looks good.

tests/IntegrationTests/Pingable.slice
pepone: The previous version of `PingableService` was doing

```
private class PingableService : Service, IPingableService, IIceObjectService
```

My understanding is we need `Ice::Object` inheritance to test `Ice_operations`, Maybe we should use a separate service for the `Ice_operation` test?

PR #3698, opened 2023-10-17
Use FrozenDictionary in Router

bernardnormier:
According to https://devblogs.microsoft.com/dotnet/performance-improvements-in-aspnet-core-8/, Frozen dictionary is significantly faster for the "router" use-case.


InsertCreativityHere (approved):
Seems pretty cool.                                       (get it?)

I wonder how much a performance difference it will make for us.

pepone (approved):

PR #3706, opened 2023-10-18
Handle QuicError.AlpnInUse

### LINKED ISSUE #2155: Handle QuicError.AlpnInUse
.NET 8 adds a new quic error `QuicError.AlpnInUse`, once we update we should handle it, it should probably be mapped to TransportError.AddressInUse as the Alpn is fixed for "icerpc".

see https://github.com/zeroc-ice/icerpc-csharp/pull/2152

pepone:
Fix #2155


pepone:
Failures seem unrelated to the PR. We should probably rename `Listen_twice_on_the_same_address_fails_with_a_transport_exception` there are no transport exceptions anymore.

PR #3715, opened 2023-10-19
Simplify GreeterProtobuf example

bernardnormier:
This PR simplifies the GreeterProtobuf example to be more on-par with the Greeter example.


examples/GreeterProtobuf/Client/Program.cs
```diff
@@ -8,11 +8,10 @@
 await using var connection = new ClientConnection(new Uri("icerpc://localhost"));
 
 // GreeterClient is a struct generated by protoc-gen-icerpc-csharp.
```
pepone, line 10:
```suggestion
// GreeterClient is a record struct generated by protoc-gen-icerpc-csharp.
```

pepone (approved):
Looks good!

examples/GreeterProtobuf/Client/Program.cs
ReeceHumphreys: I think the fact that it is a record struct in particular is more of an implementation detail. A user reading the example does not need to know it is a record struct, just that it is a struct.

ReeceHumphreys (approved):

PR #3717, opened 2023-10-19
Add test to ensure disposing listener doesn't affect existing connections

### LINKED ISSUE #3686: Disposing the listener must not affect existing connections
We have to add a conformance test to verify that disposing the listener doesn't affect existing connections.

pepone:
Fixes #3686


bernardnormier (approved):
Looks good.

Issue #3699, opened 2023-10-17
Default service path


bernardnormier:
With Slice, we currently compute the "default service path" using the Slice type ID attributes included the generated code for mapped Slice interfaces.

And as a result, `Router.Map<IXxxService>` is a Slice extension:
https://github.com/icerpc/icerpc-csharp/blob/06ceee30fbeced504a32d2c0a068d326daf063d8/src/IceRpc.Slice/SliceRouterExtensions.cs#L20

With the addition of Protobuf support, I think it would be cleaner to generalize this notion of "default service path" and move it to the core--without a dependency on Slice type IDs.

Namely:
1.  we provide a new IceRpc- core attribute to specify the default service path on a "service" interface:

```csharp
// IceRPC + Slice generated code
[DefaultServicePath("/VisitorCenter.Greeter")
public partial interface IGreeterService { ... }

// IceRPC + Protobuf generated code
[DefaultServicePath("/visitor_center.Greeter")
public partial interface IGreeterService { ... }
```

2. We move the Router.Map extension method to IceRpc and use this DefaultServicePath attribute to lookup the default service path.

3. For consistency, in the Slice proxy generated code, we stop using type IDs to compute the default service address and simply generate the default service path as a string literal.

PR #3718, opened 2023-10-19
Add default service path attribute

pepone:
See #3699


src/IceRpc/IceRpcTypeExtensions.cs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) ZeroC, Inc.
+
+using ZeroC.Slice;
+
+namespace IceRpc;
+
+/// <summary>Provides extension methods for <see cref="Type" />.</summary>
```
bernardnormier, line 7: The file name should be simply TypeExtensions.cs

src/IceRpc/IceRpcTypeExtensions.cs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) ZeroC, Inc.
+
+using ZeroC.Slice;
```
bernardnormier, line 3: Not clear why you need this using.

src/IceRpc/IceRpcTypeExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            object[] attributes = type.GetCustomAttributes(typeof(DefaultServicePathAttribute), false);
+            if (attributes.Length == 1 && attributes[0] is DefaultServicePathAttribute defaultServicePath)
+            {
+                return defaultServicePath.Value;
+            }
+            else
+            {
+                throw new ArgumentException(
+                    $"The type '{type}' doesn't have a {nameof(DefaultServicePathAttribute)} attribute.",
```
bernardnormier, line 28: I would change the message to:
Interface '{type}' does not have ...

src/IceRpc/RouterExtensions.cs
```diff
@@ -8,6 +8,21 @@ namespace IceRpc;
 /// feature.</summary>
 public static class RouterExtensions
 {
+    /// <summary>Registers a route to a service that uses the service's default path as the route path. If there is an
+    /// existing route at the same path, it is replaced.</summary>
+    /// <typeparam name="TService">The service type used to get the default path.</typeparam>
```
bernardnormier, line 13: ```<typeparam name="TService">An interface with the DefaultServicePathAttribute attribute. The path of the mapped service corresponds to the value of this attribute.</typeparam>```

+ no remarks.

bernardnormier (approved):

src/IceRpc/IceRpcTypeExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        if (type.IsInterface)
+        {
+            object[] attributes = type.GetCustomAttributes(typeof(DefaultServicePathAttribute), false);
+            if (attributes.Length == 1 && attributes[0] is DefaultServicePathAttribute defaultServicePath)
+            {
+                return defaultServicePath.Value;
+            }
+            else
+            {
+                throw new ArgumentException(
```
ReeceHumphreys, line 27: If you accidentally have two `[IceRpc.DefaultServicePath(""/{service.Name}"")]` on the same interface by mistake then `attributes.Length` == 1 would be false and we would get into the error case here. The error message then would not be correct as it does have the `DefaultServicePathAttribute ` attribute, there are just multiple.

This would probably be super uncommon so it might not be worth handling with a seperate error.

ReeceHumphreys (approved):
Looks good

src/IceRpc/IceRpcTypeExtensions.cs
pepone: The compiler handles this case, the default is AllowMultiple = false

See https://learn.microsoft.com/en-us/dotnet/api/system.attributeusageattribute.allowmultiple?view=net-7.0

Issue #3707, opened 2023-10-18
Add support for protobuf streaming RPCs


pepone:
The Protobuf generator in https://github.com/icerpc/icerpc-csharp/tree/main/tools/protobuf/IceRpc.ProtocGen only handles Unary RPCs.

We should support streaming RPCs just like we do for Slice.

bernardnormier:
I propose to encode streams of Protobuf messages like gRPC, i.e. the payload of a Protobuf-encoded request or response is a
"repeated sequence of Length-Prefixed-Message items" where:

- Length-Prefixed-Message → Compressed-Flag Message-Length Message
- Compressed-Flag → 0 / 1 # encoded as 1 byte unsigned integer
- Message-Length → {length of Message} # encoded as 4 byte unsigned integer (big endian)
- Message → *{binary octet}

See https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

And unary RPCs should be the same, except there is only one Protobuf message.

I would also map a stream of Protobuf messages to an `IAsyncEnumerable<T>`, like we do for Slice.

pepone:
I understand the idea is to be able to forward gRPC messages over icerpc without re-encoding. How would an IceRPC server handle a compressed message like this?

bernardnormier:
> How would an IceRPC server handle a compressed message like this?

The cleanest would be to leave this decompression to the IceRPC + Protobuf integration.

At the core level (IceRPC), it's just bytes. Initially, we would simply not support this gRPC-style compression in IceRPC + Protobuf.

PR #3736, opened 2023-10-20
Add support for Protobuf length prefixed messages

### LINKED ISSUE #3719: Add IProtobufFeature
We need an IProtobufFeature in IceRpc.Protobuf, similar to ISliceFeature.

```csharp
public interface IProtobufFeature
{
    /// <summary>Gets the maximum length of an encoded Protobuf message, in bytes.</summary>
    /// <value>The maximum length of a Protobuf message, in bytes.</value>
    int MaxMessageLength { get; }

   ... other properties? ...
}
```

It's used during decoding.  For historical reasons, we use "size" with Slice. Protobuf / gRPC prefer "length".

### LINKED ISSUE #3734: Protobuf: encode into IBufferWriter, decode from ReadOnlySequence
We should use these APIs to encode / decode Protobuf messages into / from IceRPC payloads:

https://cloud.google.com/dotnet/docs/reference/Google.Protobuf/latest/Google.Protobuf.MessageExtensions#Google_Protobuf_MessageExtensions_WriteTo_Google_Protobuf_IMessage_System_Buffers_IBufferWriter_System_Byte__

https://cloud.google.com/dotnet/docs/reference/Google.Protobuf/latest/Google.Protobuf.MessageParser-1#Google_Protobuf_MessageParser_1_ParseFrom_System_Buffers_ReadOnlySequence_System_Byte__

### LINKED ISSUE #3735: Add ProtobufEncodeOptions
Once we start creating Pipe to encode payloads - including stream payload - we should add a ProtobufEncodOptions class similar to
https://github.com/icerpc/icerpc-csharp/blob/main/src/IceRpc.Slice/SliceEncodeOptions.cs

See also #3719 and #3734.

pepone:
This PR updates the encoding of Protobuf payloads to use the Length prefixes message encoding. It also add `IProtobufFeature` and `ProtobufEncodeOptions`

Fixes #3719
Fixes #3734
Fixes #3735

See also #3707


src/IceRpc.Protobuf/IProtobufFeature.cs
```diff
@@ -0,0 +1,17 @@
...
+/// <summary>Represents a feature used to customize the encoding and decoding of request and response payloads.
+/// </summary>
+public interface IProtobufFeature
+{
+    /// <summary>Gets the maximum length of an encoded Protobuf message, in bytes.</summary>
+    /// <value>The maximum length of a Protobuf message, in bytes.</value>
+    int MaxMessageLength { get; }
+
+    /// <summary>Gets the options to use when encoding the payload of an outgoing response.</summary>
+    /// <value>The Protobuf encode options. <see langword="null" /> is equivalent to <see cref="ProtobufEncodeOptions.Default"
```
bernardnormier, line 14: Line too long.

src/IceRpc.Protobuf/MessageExtensions.cs
```diff
@@ -2,28 +2,30 @@
...
-    {
-        using var stream = new MemoryStream();
-        await reader.CopyToAsync(stream).ConfigureAwait(false);
-        stream.Seek(0, SeekOrigin.Begin);
-        message.MergeFrom(stream);
-    }
-
-    /// <summary>Converts an <see cref="IMessage" /> into a <see cref="PipeReader" /> containing the protobuf-encoded
-    /// binary data.</summary>
+    /// <summary>Encodes an <see cref="IMessage" /> as a Protobuf length prefixed message into a
```
bernardnormier, line 13: I would simplify:
Encodes an IMessage as a length-prefixed message, using the Protobuf encoding.

"length-prefixed message" is actually a gRPC term, not a Protobuf term.

src/IceRpc.Protobuf/MessageExtensions.cs
```diff
@@ -2,28 +2,30 @@
...
+    /// <see cref="PipeReader" />.</summary>
     /// <param name="message">The <see cref="IMessage" /> to encode.</param>
+    /// <param name="pipeOptions">The options used to create the pipe.</param>
     /// <returns>The <see cref="PipeReader" /> containing the protobuf-encoded data.</returns>
```
bernardnormier, line 17: Returns: A PipeReader containing the length-prefixed message.

src/IceRpc.Protobuf/MessageExtensions.cs
```diff
@@ -2,28 +2,30 @@
...
-    public static PipeReader ToPipeReader(this IMessage message) =>
-        PipeReader.Create(new ReadOnlySequence<byte>(message.ToByteArray()));
+    public static PipeReader EncodeAsLengthPrefixedMessage(this IMessage message, PipeOptions pipeOptions)
+    {
+        var pipe = new Pipe(pipeOptions);
+        Span<byte> compressed = [0];
```
bernardnormier, line 21: Do you need this temporary variable?

src/IceRpc.Protobuf/MessageParserExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Buffers.Binary;
+using System.Diagnostics;
+using System.IO.Pipelines;
+
+namespace IceRpc.Protobuf;
+
+/// <summary>Provides an extension method for <see cref="MessageParser{T}" />.</summary>
+public static class MessageParserExtensions
+{
+    /// <summary>Decodes a Protobuf length prefixed message from a <see cref="PipeReader" />.</summary>
```
bernardnormier, line 14: That's not quite correct. Here, you're decoding a Protobuf message _from_ a stream of bytes (the pipe reader). And the format of these bytes is a length-prefixed message.

bernardnormier: Maybe it would be cleaner to make it a PipeReader extension method `DecodeProtobufMessageAsync(...)`?

src/IceRpc.Protobuf/MessageParserExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// <param name="maxMessageLength">The maximum allowed length.</param>
+    /// <param name="cancellationToken">A cancellation token that receives the cancellation requests.</param>
+    /// <returns>The decoded message object.</returns>
+    public static async ValueTask<T> DecodeFromLengthPrefixedMessageAsync<T>(
+        this MessageParser<T> parser,
+        PipeReader reader,
+        int maxMessageLength,
+        CancellationToken cancellationToken) where T : IMessage<T>
+    {
+        ReadResult readResult = await reader.ReadAtLeastAsync(5, cancellationToken).ConfigureAwait(false);
```
bernardnormier, line 26: Check readResult.IsCanceled.

This is similar to this code:
https://github.com/icerpc/icerpc-csharp/blob/ecbe38c0fd584046384b0589b9c94b278fb5ca6f/src/IceRpc.Slice/Internal/IncomingFrameExtensions.cs#L39

src/IceRpc.Protobuf/MessageParserExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+            throw new NotSupportedException("IceRPC does not support Protobuf compressed messages");
+        }
+        int messageLength = DecodeMessageLength(readResult.Buffer.Slice(1, 4));
+        reader.AdvanceTo(readResult.Buffer.GetPosition(5));
+        if (messageLength >= maxMessageLength)
+        {
+            throw new InvalidDataException("The message length exceeds the maximum value.");
+        }
+
+        readResult = await reader.ReadAtLeastAsync(messageLength, cancellationToken).ConfigureAwait(false);
```
bernardnormier, line 46: check IsCanceled (see above)

src/IceRpc.Protobuf/MessageParserExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+        if (readResult.Buffer.Length < messageLength)
+        {
+            reader.AdvanceTo(readResult.Buffer.End);
+            throw new InvalidDataException(
+                $"The payload has {readResult.Buffer.Length} bytes, but {messageLength} bytes were expected.");
+        }
+
+        T message = parser.ParseFrom(readResult.Buffer.Slice(messageLength));
```
bernardnormier, line 54: Does ParseFrom check it read all the bytes? 

Assume it does, please add a comment.

In the equivalent Slice code, we have 
https://github.com/icerpc/icerpc-csharp/blob/ecbe38c0fd584046384b0589b9c94b278fb5ca6f/src/IceRpc.Slice/Internal/IncomingFrameExtensions.cs#L53

src/IceRpc.Protobuf/MessageParserExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+    public static async ValueTask<T> DecodeFromLengthPrefixedMessageAsync<T>(
+        this MessageParser<T> parser,
+        PipeReader reader,
+        int maxMessageLength,
+        CancellationToken cancellationToken) where T : IMessage<T>
+    {
+        ReadResult readResult = await reader.ReadAtLeastAsync(5, cancellationToken).ConfigureAwait(false);
+        if (readResult.Buffer.Length < 5)
+        {
+            reader.AdvanceTo(readResult.Buffer.End);
```
bernardnormier, line 29: We don't need this AdvanceTo (the caller will complete the reader on exception). We also don't AdvanceTo consistently, e.g. if ParseFrom below throws.

See equivalent code in Slice:
https://github.com/icerpc/icerpc-csharp/blob/main/src/IceRpc.Slice/Internal/IncomingFrameExtensions.cs

src/IceRpc.Protobuf/MessageParserExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+        static int DecodeMessageLength(ReadOnlySequence<byte> buffer)
+        {
+            Debug.Assert(buffer.Length == 4);
+            if (buffer.IsSingleSegment)
+            {
+                return BinaryPrimitives.ReadInt32BigEndian(buffer.FirstSpan);
+            }
+            else
+            {
+                var span = new Span<byte>(buffer.ToArray());
```
bernardnormier, line 67: I would allocate a 4-bytes span and copy this buffer into the span:

https://learn.microsoft.com/en-us/dotnet/api/system.buffers.buffersextensions.copyto?view=net-7.0#system-buffers-buffersextensions-copyto-1(system-buffers-readonlysequence((-0))-system-span((-0)))

src/IceRpc.Protobuf/ProtobufEncodeOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace IceRpc.Protobuf;
+
+/// <summary>Represents a property bag used to configure the encoding of payloads.</summary>
+public sealed class ProtobufEncodeOptions
+{
+    /// <summary>Gets the default instance of <see cref="ProtobufEncodeOptions" />.</summary>
+    public static ProtobufEncodeOptions Default { get; } = new();
+
+    /// <summary>Gets the pipe options that the Slice engine uses when creating pipes. The Slice engine creates a pipe
```
bernardnormier, line 14: No "Slice engine".

It's really the IceRPC + Protobuf integration or IceRPC + Protobuf for short. (Same for Slice btw).

src/IceRpc.Protobuf/ProtobufFeature.cs
```diff
@@ -0,0 +1,38 @@
...
+public sealed class ProtobufFeature : IProtobufFeature
+{
+    /// <summary>Gets a <see cref="IProtobufFeature" /> with default values for all properties.</summary>
+    public static IProtobufFeature Default { get; } = new DefaultProtobufFeature();
+
+    /// <inheritdoc/>
+    public int MaxMessageLength { get; init; }
+
+    /// <inheritdoc/>
+    public ProtobufEncodeOptions? EncodeOptions {  get; init; }
```
bernardnormier, line 15: extra white space before get

bernardnormier (approved):
Looks good.

src/IceRpc.Protobuf/MessageExtensions.cs
pepone: No

tools/IceRpc.ProtocGen/ClientGenerator.cs
```diff
@@ -57,15 +57,17 @@ internal static string GenerateImplementation(ServiceDescriptor service)
             Invoker,
             ServiceAddress,
             ""{method.Name}"",
-            message.ToPipeReader(),
+            message.EncodeAsLengthPrefixedMessage(EncodeOptions?.PipeOptions ?? ProtobufEncodeOptions.Default.PipeOptions),
             payloadContinuation: null,
             async (response, request, cancellationToken) =>
```
bernardnormier, line 62: It seems this could be much simpler:
- call `Invoke.InvokeAsync<T>(...)`
- don't include a ResponseDecodeFunc param since it's always the same code

bernardnormier: With Slice, there is no special T that represents the result, hence we need this decode-response-func parameter.

src/IceRpc.Protobuf/MessageParserExtensions.cs
pepone: The parse ends with a call to `CheckReadEndOfStreamTag`

```
        /// <summary>
        /// Verifies that the last call to ReadTag() returned tag 0 - in other words,
        /// we've reached the end of the stream when we expected to.
        /// </summary>
        /// <exception cref="InvalidProtocolBufferException">The 
        /// tag read was not the one specified</exception>
        public static void CheckReadEndOfStreamTag(ref ParserInternalState state)
        {
            if (state.lastTag != 0)
            {
                throw InvalidProtocolBufferException.MoreDataAvailable();
            }
        }
```

This seems more like a tag end marked, than an end of buffer check.

pepone: When using the ReadOnlySequence overload, we don't know how many bytes the ParseFrom method read.

src/IceRpc.Protobuf/MessageParserExtensions.cs
pepone: Yes indeed

src/IceRpc.Protobuf/MessageParserExtensions.cs
bernardnormier: Maybe just add a TODO for now?

tools/IceRpc.ProtocGen/ClientGenerator.cs
bernardnormier: I would defer this to a follow-up PR.

bernardnormier (approved):

PR #3737, opened 2023-10-20
Remove ProtobufD methods

bernardnormier:
This PR removes the ProtobufD methods generated by IceRpc.ProtocGen.

They are replaced by better code generated by the Protobuf Service source generator.

### DIFF
--- a/src/IceRpc.Protobuf.Generators/Internal/Emitter.cs
+++ b/src/IceRpc.Protobuf.Generators/Internal/Emitter.cs
@@ -19,26 +19,40 @@ internal class Emitter
             string dispatchImplementation;
             if (serviceClass.ServiceMethods.Count > 0)
             {
+                methodModifier += " async";
+
                 dispatchImplementation = "";
                 foreach (ServiceMethod serviceMethod in serviceClass.ServiceMethods)
                 {
-                    string operationName = serviceMethod.OperationName;
-                    string dispatchMethodName = serviceMethod.DispatchMethodName;
-                    dispatchImplementation +=
-                        $"\"{operationName}\" => global::{dispatchMethodName}(this, request, cancellationToken),\n";
+                    dispatchImplementation += @$"
+case ""{serviceMethod.OperationName}"":
+{{
+    var inputParam = new {serviceMethod.InputTypeName}();
+    await inputParam.MergeFromAsync(request.Payload).ConfigureAwait(false);
+    var outputParam = await (this as {serviceMethod.InterfaceName}).{serviceMethod.MethodName}(
+        inputParam,
+        request.Features,
+        cancellationToken).ConfigureAwait(false);
+    return new IceRpc.OutgoingResponse(request)
+    {{
+        Payload = outputParam.ToPipeReader()
+    }};
+}}".Trim();
                 }
-
                 if (serviceClass.HasBaseServiceClass)
                 {
-                    dispatchImplementation += "_ => base.DispatchAsync(request, cancellationToken)";
+                    dispatchImplementation += @$"
+default:
+    return base.DispatchAsync(request, cancellationToken);";
                 }
                 else
                 {
-                    dispatchImplementation +=
-                        "_ => new(new IceRpc.OutgoingResponse(request, IceRpc.StatusCode.NotImplemented))";
+                    dispatchImplementation += @$"
+default:
+    return new IceRpc.OutgoingResponse(request, IceRpc.StatusCode.NotImplemented);";
                 }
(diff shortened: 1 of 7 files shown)


src/IceRpc.Protobuf.Generators/Internal/Emitter.cs
```diff
@@ -19,26 +19,40 @@ internal string Emit(IReadOnlyList<ServiceClass> serviceClasses, CancellationTok
...
+default:
+    return base.DispatchAsync(request, cancellationToken);";
                 }
                 else
                 {
-                    dispatchImplementation +=
-                        "_ => new(new IceRpc.OutgoingResponse(request, IceRpc.StatusCode.NotImplemented))";
+                    dispatchImplementation += @$"
+default:
+    return new IceRpc.OutgoingResponse(request, IceRpc.StatusCode.NotImplemented);";
```
bernardnormier, line 52: I think there is a small bug here: should be `new(new...)` when not async. Let's fix it when we add more tests.

src/IceRpc.Protobuf.Generators/Internal/Emitter.cs
```diff
@@ -19,26 +19,40 @@ internal string Emit(IReadOnlyList<ServiceClass> serviceClasses, CancellationTok
...
-                    string operationName = serviceMethod.OperationName;
-                    string dispatchMethodName = serviceMethod.DispatchMethodName;
-                    dispatchImplementation +=
-                        $"\"{operationName}\" => global::{dispatchMethodName}(this, request, cancellationToken),\n";
+                    dispatchImplementation += @$"
+case ""{serviceMethod.OperationName}"":
+{{
+    var inputParam = new {serviceMethod.InputTypeName}();
+    await inputParam.MergeFromAsync(request.Payload).ConfigureAwait(false);
+    var outputParam = await (this as {serviceMethod.InterfaceName}).{serviceMethod.MethodName}(
```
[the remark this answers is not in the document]
pepone: I think we should use a regular cast, `as` suggests this can fail but it should not.

src/IceRpc.Protobuf.Generators/Internal/Emitter.cs
```diff
@@ -19,26 +19,40 @@ internal string Emit(IReadOnlyList<ServiceClass> serviceClasses, CancellationTok
...
+    }};
+}}".Trim();
                 }
-
                 if (serviceClass.HasBaseServiceClass)
                 {
-                    dispatchImplementation += "_ => base.DispatchAsync(request, cancellationToken)";
+                    dispatchImplementation += @$"
+default:
+    return base.DispatchAsync(request, cancellationToken);";
```
pepone, line 46: We need to `await` this call if we are in an async method

src/IceRpc.Protobuf.Generators/Internal/Parser.cs
```diff
@@ -190,7 +190,12 @@ private IReadOnlyList<ServiceMethod> GetServiceMethods(INamedTypeSymbol interfac
                 items.Length == 1,
                 "Unexpected number of arguments in attribute constructor.");
             string operationName = (string)items[0].Value!;
-            serviceMethods.Add(new ServiceMethod(dispatchMethodName: GetFullName(method), operationName));
+            serviceMethods.Add(
+                new ServiceMethod(
+                    operationName,
+                    interfaceName: $"global::{GetFullName(interfaceSymbol)}",
+                    methodName: method.Name,
+                    inputTypeName: $"global::{GetFullName(method.Parameters[0].Type)}"));
```
pepone, line 198: This is going to result in some over-qualification if the generated service is in the same namespace. We can look into this in a separate PR

pepone (approved):
Looks good!

src/IceRpc.Protobuf.Generators/Internal/Emitter.cs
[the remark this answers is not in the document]
bernardnormier: replaced by plain cast expression

src/IceRpc.Protobuf.Generators/Internal/Emitter.cs
bernardnormier: Fixed

src/IceRpc.Protobuf.Generators/Internal/Parser.cs
bernardnormier: You're right, let's fix this in another PR. This overqualification isn't critical since it's pretty hard to see the file generated by the source generator.

Issue #3708, opened 2023-10-18
Add more Protobuf tests


pepone:
We need more tests for Protobuf support in https://github.com/icerpc/icerpc-csharp/tree/main/tests/IceRpc.Protobuf.Tests

PR #3743, opened 2023-10-24
Add support for protobuf streaming RPCs

pepone:
See #3707

### DIFF
--- a/src/IceRpc.Protobuf.Generators/Internal/Parser.cs
+++ b/src/IceRpc.Protobuf.Generators/Internal/Parser.cs
@@ -224,7 +224,7 @@ internal sealed class Parser
                     operationName,
                     interfaceName: $"global::{GetFullName(interfaceSymbol)}",
                     methodName: method.Name,
-                    inputTypeName,
+                    inputTypeName: $"global::{inputTypeName}",
                     isClientStreaming,
                     isServerStreaming));
         }
--- a/src/IceRpc.Protobuf.Generators/Internal/Parser.cs
+++ b/src/IceRpc.Protobuf.Generators/Internal/Parser.cs
@@ -10,9 +10,11 @@ namespace IceRpc.Protobuf.Generators.Internal;
 
 internal sealed class Parser
 {
-    internal const string OperationAttribute = "IceRpc.Protobuf.ProtobufOperationAttribute";
-    internal const string ServiceAttribute = "IceRpc.Protobuf.ProtobufServiceAttribute";
+    private const string IAsyncEnumerableName = "System.Collections.Generic.IAsyncEnumerable`1";
+    private const string OperationAttributeName = "IceRpc.Protobuf.ProtobufOperationAttribute";
+    private const string ServiceAttributeName = "IceRpc.Protobuf.ProtobufServiceAttribute";
 
+    private readonly INamedTypeSymbol? _iasyncEnumerableSymbol;
     private readonly CancellationToken _cancellationToken;
     private readonly Compilation _compilation;
     private readonly INamedTypeSymbol? _operationAttribute;
@@ -28,13 +30,14 @@ internal sealed class Parser
         _reportDiagnostic = reportDiagnostic;
         _cancellationToken = cancellationToken;
 
-        _operationAttribute = _compilation.GetTypeByMetadataName(OperationAttribute);
-        _serviceAttribute = _compilation.GetTypeByMetadataName(ServiceAttribute);
+        _iasyncEnumerableSymbol = _compilation.GetTypeByMetadataName(IAsyncEnumerableName);
+        _operationAttribute = _compilation.GetTypeByMetadataName(OperationAttributeName);
+        _serviceAttribute = _compilation.GetTypeByMetadataName(ServiceAttributeName);
     }
 
     internal IReadOnlyList<ServiceClass> GetServiceDefinitions(IEnumerable<ClassDeclarationSyntax> classes)
     {
-        if (_operationAttribute is null || _serviceAttribute is null)
+        if (_operationAttribute is null || _serviceAttribute is null || _iasyncEnumerableSymbol == null)
         {
             // nothing to do if these types aren't available
             return Array.Empty<ServiceClass>();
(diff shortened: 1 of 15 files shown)


tests/IceRpc.Protobuf.Tests/OperationTests.cs
```diff
@@ -21,12 +21,142 @@ public void Unary_rpc_with_empty_param_and_return()
...
+        var client = new MyOperationsClient(invoker);
+
+        // Act/Assert
+        Assert.That(async () => await client.ClientStreamingOpAsync(GetDataAsync()), Throws.Nothing);
+        Assert.That(service.Empties, Has.Count.EqualTo(3));
+
+        static async IAsyncEnumerable<Empty> GetDataAsync()
+        {
+            await Task.Yield();
+            yield return new Empty();
```
bernardnormier, line 39: While I have nothing against Empty messages, it would be nice to also test "real" messages,

src/IceRpc.Protobuf/InvokerExtensions.cs
```diff
@@ -45,21 +42,21 @@ public static class InvokerExtensions
         this IInvoker invoker,
         ServiceAddress serviceAddress,
         string operation,
-        PipeReader payload,
-        PipeReader? payloadContinuation,
-        MessageParser<T> responseMessageParser,
+        IMessage inputMessage,
+        ProtobufEncodeOptions? encodeOptions,
+        ResponseDecodeFunc<T> responseDecodeFunc,
```
bernardnormier, line 47: It's not clear to me why we need this ResponseDecodeFunc. Why not 4 InvokeAsync overloads, for all combinations of unary / stream request / response?

src/IceRpc.Protobuf.Generators/Internal/Parser.cs
```diff
@@ -190,12 +190,30 @@ private IReadOnlyList<ServiceMethod> GetServiceMethods(INamedTypeSymbol interfac
                 items.Length == 1,
                 "Unexpected number of arguments in attribute constructor.");
             string operationName = (string)items[0].Value!;
+            // An IAsyncEnumerable input parameter denotes a client streaming RPC.
+            string inputTypeName = GetFullName(method.Parameters[0].Type);
+            bool isClientStreaming = inputTypeName == "System.Collections.Generic.IAsyncEnumerable";
```
bernardnormier, line 195: You can't compare Type objects/properties instead of this string comparison?

tools/IceRpc.ProtocGen/ClientGenerator.cs
```diff
@@ -48,21 +52,59 @@ internal static string GenerateImplementation(ServiceDescriptor service)
...
+            {
+                if (method.IsClientStreaming)
+                {
+                    genericArg = $"<{returnType}, {method.InputType.GetType(scope, false)}>";
+                }
+                else
+                {
+                    genericArg = $"<{returnType}>";
+                }
+                responseDecodeFunc = $@"
```
bernardnormier, line 67: I'd rather not generate this. See InvokeAsync comment.

bernardnormier (approved):

src/IceRpc.Protobuf.Generators/Internal/Parser.cs
pepone: I think we can compare the symbols yes, we already do that for the attributes

tests/IceRpc.Protobuf.Tests/OperationTests.cs
pepone: I will fix this along https://github.com/icerpc/icerpc-csharp/issues/3708

src/IceRpc.Protobuf/InvokerExtensions.cs
pepone: The issue with using overloads for the `InvokeAsync` method is the ambiguity introduced when overloads differ only in their return types. Specifically:

- `InvokeAsync` for unary RPC:
  * Input: `IMessage<T>, MessageParser<T>`
  * Return: `IMessage<T>`
  
- `InvokeAsync` for client-streaming RPC:
  * Input: `IAsyncEnumerable<IMessage<T>>`
  * Return: `IMessage<T>`
  
- `InvokeAsync` for server-streaming RPC:
  * Input: `IMessage<T>`
  * Return: `IAsyncEnumerable<IMessage<T>>`
  
- `InvokeAsync` for bidirectional-streaming RPC:
  * Input: `IAsyncEnumerable<IMessage<T>>`
  * Return: `IAsyncEnumerable<IMessage<T>>`

There's ambiguity between the overloads for unary and server-streaming RPCs. Similarly, ambiguity exists between the overloads for client-streaming and bidirectional-streaming RPCs.

We can of course give these methods different names:

- `UnaryInvokeAsync`
- `ClientStreamingInvokeAsync`
- `ServerStreamingInvokeAsync`
- `BidirectionalStreamingInvokeAsync`

Which I think would be fine, the principal caller of these methods is the Protobuf client-generated code.

pepone: Fixed the code to use separate extension methods as mentioned above.

src/IceRpc.Protobuf/InvokerExtensions.cs
```diff
@@ -73,33 +60,216 @@ public static class InvokerExtensions
             throw;
         }
 
-        // ReadResponseAsync is responsible for disposing the request
-        return ReadResponseAsync(responseTask, request);
+        // ReceiveResponseAsync is responsible for disposing the request
+        return ReceiveResponseAsync(messageParser, responseTask, request, cancellationToken);
+    }
+
+    /// <summary>Sends a request to a service and decodes the response.</summary>
```
externl, line 68: It would be nice for the summaries to explain the different behavior between `ClientStreamingInvokeAsync`, `ServerStreamingInvokeAsync`, etc.

externl (approved):
Looks good!

src/IceRpc.Protobuf/AsyncEnumerableExtensions.cs
```diff
@@ -0,0 +1,209 @@
...
+                int written = 0;
+                do
+                {
+                    _pipe.Writer.Write(new Span<byte>([0])); // Not compressed
+                    Span<byte> lengthPlaceholder = _pipe.Writer.GetSpan(4)[..4];
+                    _pipe.Writer.Advance(4);
+                    _asyncEnumerator.Current.WriteTo(_pipe.Writer);
+                    int length = checked((int)_pipe.Writer.UnflushedBytes - written);
+                    written += length;
+                    BinaryPrimitives.WriteInt32BigEndian(lengthPlaceholder, length - 5);
```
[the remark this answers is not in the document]
pepone: This is to allow forwarding this payload to gRPC without having to reencode them, Bernard explained this in #3707

src/IceRpc.Protobuf/InvokerExtensions.cs
bernardnormier: I would prefer: InvokeXxxAsync
with Xxx = Unary, ClientStreaming, ServerStream, BidiStreaming

per the names in https://protobuf.com/docs/language-spec#methods

PR #3753, opened 2023-11-13
Protobuf idempotent test

pepone:
This PR adds a test for Protobuf idempotency_level usage with IceRPC

### DIFF
--- a/tests/IceRpc.Protobuf.Tests/OperationTests.proto
+++ b/tests/IceRpc.Protobuf.Tests/OperationTests.proto
@@ -15,6 +15,14 @@ service MyOperations {
   rpc serverStreamingOp(google.protobuf.Empty) returns (stream OutputMessage);
 
   rpc bidirectionalStreamingOp(stream InputMessage) returns (stream OutputMessage);
+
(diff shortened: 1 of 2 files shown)


externl (approved):

tests/IceRpc.Protobuf.Tests/OperationTests.proto
```diff
@@ -15,6 +15,14 @@ service MyOperations {
   rpc serverStreamingOp(google.protobuf.Empty) returns (stream OutputMessage);
 
   rpc bidirectionalStreamingOp(stream InputMessage) returns (stream OutputMessage);
+
+  rpc idempotentOp(google.protobuf.Empty) returns(google.protobuf.Empty) {
```
bernardnormier, line 19: All the Protobuf tests and examples should use the standard Protobuf naming conventions. For RPC methods, it's Pascal case.

bernardnormier (approved):

bernardnormier:
It would be good to create an issue for this PR and discuss that unlike Slice, we don't check for client/server mismatch with Protobuf, even though we could.

PR #3750, opened 2023-11-13
Restructure example layout

pepone:

PR #3761, opened 2023-11-16
Rework examples layout

pepone:
This is a rework of  https://github.com/icerpc/icerpc-csharp/pull/3750

Mostly renamed `core` to `icerpc` and kept `interop` for examples that show interoperability with ZeroC Ice.


bernardnormier:
I think you should:
 - move all examples that use Slice into the slice sub-directory
 - keep the renaming examples in "icerpc" but rename icerpc to "no-idl"
 - in follow-up PRs, add more protobuf examples (Logger, Thermostat, Quic and more) that duplicate the slice examples, except they use Protobuf.

pepone:
> I think you should:

Applied these suggestions, and let additional Protobuf examples for follow up PRs

bernardnormier (approved):
Looks good

PR #3762, opened 2023-11-16
Remove extra blank lines in Protobuf generated code

### LINKED ISSUE #3759: Extra blank line in ServiceTests.IceRpc.cs
This generated file is currently the only Protobuf service tests and shows some bugs:

```csharp
public partial interface IMyWidget
{
     <-- extra blank line   
}

...

 }

  <-- extra blank lime   
}

```

pepone:
Fix #3759

I added a helper method to remove that allows removing extra empty lines, because I think is nice to be able to use the string templates in the generators. I think the string templates make it easy to see what the generator produces.

### DIFF
--- a/tools/IceRpc.ProtocGen/StringExtensions.cs
+++ b/tools/IceRpc.ProtocGen/StringExtensions.cs
@@ -9,7 +9,7 @@ internal static class StringExtensions
     /// <param name="value">The string to be processed.</param>
     /// <returns>The processed string without consecutive empty lines, and without empty lines that are between an open
     /// and a close brace '{', '}' or between consecutive close '}'.</returns>
-    internal static string RemoveSuperflousEmptyLines(this string value)
+    internal static string RemoveSuperfluousEmptyLines(this string value)
     {
         string[] lines = value.Split('\n');
         var processedLines = new List<string>();
--- a/tools/IceRpc.ProtocGen/StringExtensions.cs
+++ b/tools/IceRpc.ProtocGen/StringExtensions.cs
@@ -1,48 +1,33 @@
 // Copyright (c) ZeroC, Inc.
 
+using System.Text;
+
 namespace IceRpc.ProtocGen;
 
 internal static class StringExtensions
 {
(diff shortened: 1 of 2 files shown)


externl:
Did you consider writing a builder like we do for the rust code generator? This will probably be useful in the for a future C# version of the Slice code generator.

pepone:
>  We can't just not generate these empty lines to start with?

> I don't see how this relates to our use of string templates.

```
 return @$"
/// <remarks>protoc-gen-icerpc-csharp generated this client-side interface from Protobuf service <c>{service.FullName}</c>.
/// It's implemented by <c>{service.Name.ToPascalCase()}Client</c></remarks>
public partial interface I{service.Name.ToPascalCase()}
{{
    {methods.Trim()}
}}";
```

Consider the above template: if `methods` is empty, it will result in an empty line. We could structure the code differently, using if/else statements to avoid adding empty lines in the first place. However, I find that the current approach simplifies understanding what is being generated

> Did you consider writing a builder like we do for the rust C# code generator?

This started as a simple protoc plug-in, and using plain strings was the simpler way to get things rolling.

externl (approved):
Looks good. I think it could be nice to replace this with a builder pattern in the future.

tools/IceRpc.ProtocGen/StringExtensions.cs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) ZeroC, Inc.
+
+namespace IceRpc.ProtocGen;
+
+internal static class StringExtensions
+{
+    /// <summary>Removes consecutive empty lines, and empty lines that are between an open and a close brace '{', '}'
```
bernardnormier, line 7: As far as I can tell, the only superfluous empty lines we currently generate are:
```
 -- extra empty line here
}
```

And an empty line before a closing } is always incorrect. So can we simplify this code?

pepone: You are right, I simplified the code.

PR #3764, opened 2023-11-17
Protobuf test for decoding response with dispatch exception

pepone:
This PR adds a test to ensure that `DispatchException` decoded by Protobuf invoker has `ConvertToInternal` set to true. It is equivalent to the `IncomingResponseTests` we have for Slice.


src/IceRpc.Protobuf/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Runtime.CompilerServices;
+
+// Make internals visible to the tests assembly, to allow writing unit tests for the internal classes
```
externl, line 5:
```suggestion
// Make internals visible to the tests assembly, to allow writing unit tests for the internal classes.
```

bernardnormier (commented):
Making these 2 private methods of IceRpc.Protobuf internal and then exposing all internals to the test assembly does not seem clean or necessary to me.

It shouldn't be difficult to write a test where the client receives a DispatchException "over the wire". And this test could then simply rely on the public API.

externl:
> Making these 2 private methods of IceRpc.Protobuf internal and then exposing all internals to the test assembly does not seem clean or necessary to me.
> 
> It shouldn't be difficult to write a test where the client receives a DispatchException "over the wire". And this test could then simply rely on the public API.

Probably, but we already do the same thing for `Zeroc.Slice`.

externl (approved):

bernardnormier:
> Probably, but we already do the same thing for `Zeroc.Slice`.

Presumably you meant IceRpc.Slice? If we have the same structure in IceRpc.Slice, we should fix it too.

externl:
> > Probably, but we already do the same thing for `Zeroc.Slice`.
> 
> Presumably you meant IceRpc.Slice? If we have the same structure in IceRpc.Slice, we should fix it too.

I was just looking to see which other projects a similar assembly info `ZeroC.Slice` was the first one I found. `IceRpc.Slice` does **not**  have this.

pepone:
Reworked the test to use the public API.

PR #3772, opened 2023-11-18
Simplify generated code for Protobuf service dispatch

### LINKED ISSUE #3770: Simplify Protobuf generated code
The Protobuf generator generates too much code for service dispatch. 

Just like we encapsulate all the client logic in 4 InvokeXxxAsync, we could / should do the same for the dispatch logic instead of generating this logic over and over. 

```csharp
case ""{serviceMethod.OperationName}"":
{{
    var protobufFeature = request.Features.Get<IProtobufFeature>() ?? ProtobufFeature.Default;
    {inputParamDecodeCode};
    var outputParam = await (({serviceMethod.InterfaceName})this).{serviceMethod.MethodName}(
        inputParam,
        request.Features,
    return new IceRpc.OutgoingResponse(request)
    {{
        {outputParamEncode.Trim()}
    }};
}}".Trim();
```

^ always the same, we should just write & use the appropriate helper.

Then, various extension methods like ToPipeReader and ToAsyncEnumerable could become internal.

bernardnormier:
and reduce IceRpc.Protobuf public API.

Fixes #3770.

This PR also makes the generated code for IceRpc.Protobuf.Tests persistent. On my mac, the generated code is in:
```
cat ./tests/IceRpc.Protobuf.Tests/obj/Debug/net8.0/generated/IceRpc.Protobuf.Generators/IceRpc.Protobuf.Generators.ServiceGenerator/Service.g.cs
```

### DIFF
--- a/src/IceRpc.Protobuf/Internal/AsyncEnumerableExtensions.cs
+++ b/src/IceRpc.Protobuf/Internal/AsyncEnumerableExtensions.cs
@@ -6,11 +6,11 @@ using System.Buffers.Binary;
 using System.Diagnostics;
 using System.IO.Pipelines;
 
-namespace IceRpc.Protobuf;
+namespace IceRpc.Protobuf.Internal;
 
 /// <summary>Provides an extension method for <see cref="IAsyncEnumerable{T}" /> to encode elements into a <see
 /// cref="PipeReader"/>.</summary>
-public static class AsyncEnumerableExtensions
+internal static class AsyncEnumerableExtensions
 {
     /// <summary>Encodes an async enumerable into a stream of bytes represented by a <see cref="PipeReader"/>.</summary>
     /// <typeparam name="T">The async enumerable element type.</typeparam>
@@ -18,7 +18,7 @@ public static class AsyncEnumerableExtensions
     /// <param name="encodeOptions">The Protobuf encode options.</param>
     /// <returns>A pipe reader that represents the encoded stream of bytes.</returns>
     /// <remarks>This extension method is used to encode streaming parameters and streaming return values.</remarks>
-    public static PipeReader ToPipeReader<T>(
+    internal static PipeReader ToPipeReader<T>(
         this IAsyncEnumerable<T> asyncEnumerable,
         ProtobufEncodeOptions? encodeOptions = null) where T : IMessage<T> =>
         new AsyncEnumerablePipeReader<T>(asyncEnumerable, encodeOptions);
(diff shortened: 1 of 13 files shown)


src/IceRpc.Protobuf/Internal/AsyncEnumerableExtensions.cs
```diff
@@ -6,19 +6,19 @@
 using System.Diagnostics;
 using System.IO.Pipelines;
 
-namespace IceRpc.Protobuf;
+namespace IceRpc.Protobuf.Internal;
 
 /// <summary>Provides an extension method for <see cref="IAsyncEnumerable{T}" /> to encode elements into a <see
 /// cref="PipeReader"/>.</summary>
-public static class AsyncEnumerableExtensions
+internal static class AsyncEnumerableExtensions
```
pepone, line 13: The equivalent Slice class is part of the public API, why are we making this internal?

pepone (approved):
Looks good

src/IceRpc.Protobuf/Internal/AsyncEnumerableExtensions.cs
bernardnormier: In Slice, it's part of the public API only because the generated code uses this API. If it didn't (like here for Protobuf), we would make it internal as well.

pepone (approved):

tests/IceRpc.Protobuf.Tests/ServiceTests.cs
```diff
@@ -55,6 +55,32 @@ public void Base_class_implements_multiple_services()
...
+        var myFirstWidgetClient = new MyFirstWidgetClient(invoker);
+        var mySecondWidgetClient = new MySecondWidgetClient(invoker);
+
+        // Act/Assert
+        Assert.That(async () => await myFirstWidgetClient.OpFirstAsync(new Empty()), Throws.Nothing);
+        Assert.That(async () => await mySecondWidgetClient.OpSecondAsync(new Empty()), Throws.Nothing);
+    }
+
+    [Test]
+    public async Task Proto_less_service_returns_not_implemented()
```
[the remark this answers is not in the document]
bernardnormier: I think the current name is reasonable clear. The alternative is ultra long and sounds odd, as in:
Protobuf_service_class_that_does_not_implement_any_protobuf_service_returns_not_implemented

externl (approved):
Looks good!

PR #3774, opened 2023-11-19
Remove references to Proxy from ZeroC.Slice

bernardnormier:
The goal is to make ZeroC.Slice even more proxy-independent. See also https://github.com/icerpc/slicec/issues/673.


externl (approved):
Looks good!

PR #3780, opened 2023-11-20
Add Protobuf to the doc index page

externl:
This PR updates the doc index page to include links for protobuf.

### DIFF
--- a/docfx/index.md
+++ b/docfx/index.md
@@ -5,7 +5,7 @@ layout: landing
 # C# API Reference
 
 | Namespace                               | Description                                                                                                                                        |
-|-----------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
+| --------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------- |
 | [IceRpc]                                | The core of the IceRPC framework and various extensions for this core.                                                                             |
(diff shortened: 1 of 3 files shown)


pepone:
We need to also fix https://github.com/icerpc/icerpc-csharp/tree/main/docfx/namespaces

pepone (approved):

docfx/index.md
```diff
@@ -14,6 +14,8 @@ layout: landing
 | [IceRpc.Locator]                        | Provides the locator interceptor. This interceptor enables interop with Ice servers registered with a locator such as IceGrid.                     |
 | [IceRpc.Logger]                         | Provides the logger interceptor and middleware, for logging requests and responses to an ILogger.                                                  |
 | [IceRpc.Metrics]                        | Provides the metrics interceptor and middleware.                                                                                                   |
+| [IceRpc.Protobuf]                       | Provides support for the IceRPC + Protobuf integration.                                                                                            |
+| [IceRpc.Protobuf.Generators]            | Provides the Protobuf code generators for the IceRPC + Protobuf integration.                                                                       |
```
bernardnormier, line 18: Should we really include IceRpc.Protobuf.Generators and IceRpc.Slice.Generators in the public API reference?

I think it would be more confusing than helpful.

externl: I've removed this entry and excluded the both Slice and Protobuf generators.

bernardnormier (approved):

PR #3781, opened 2023-11-20
Fix bool decoding to be strict

### LINKED ISSUE #3752: Should we Strictly Enforce the Decoding of `bool`s?
The [encoding referece](https://docs.icerpc.dev/slice2/encoding/primitive-types) states that bools _must_ be `0` for `false` or `1` for `true`, and that all other values are invalid.

However, the logic for `DecodeBool` doesn't enforce this part of the spec. It treats `0` as `false`, and any other value as `true`.
Should we be better enforcing the spec and rejecting values like `2` as invalid, instead of silently treating them as valid?

For reference:
https://github.com/icerpc/icerpc-csharp/blob/fbd611d007b8cadfc995e1474885262307decd1b/src/ZeroC.Slice/SliceDecoder.cs#L130-L131

Or alternatively, should we amend the spec to follow icerpc-csharp's behavior (`0` is `false`, otherwise `true`).
This would be a strict relaxing of the spec, and hence backwards compatible.
It would simplify the job of a decoder (since it's impossible to decode an invalid bool with this change), but might be 'less safe'.

pepone:
Fix #3752

### DIFF
--- a/tools/slicec-cs/src/decoding.rs
+++ b/tools/slicec-cs/src/decoding.rs
@@ -221,10 +221,25 @@ pub fn decode_sequence(sequence_ref: &TypeRef<Sequence>, namespace: &str, encodi
                 // We always read an array even when mapped to a collection, as it's expected to be
                 // faster than decoding the collection elements one by one.
                 Some(format!(
-                    "decoder.DecodeSequence<{}>()",
+                    "decoder.DecodeSequence<{}>({})",
                     element_type.cs_type_string(namespace, TypeContext::Decode, true),
+                    if matches!(primitive, Primitive::Bool) {
+                        r#"checkElement: element =>
+    {
+        if (System.Runtime.CompilerServices.Unsafe.As<bool, byte>(ref element) > 1)
+        {
+            throw new InvalidDataException("The value is out of the bool type accepted range.");
+        }
+    }"#
+                    } else {
+                        ""
+                    }
                 ))
             }
+            Types::Primitive(primitive) if matches!(primitive, Primitive::Bool) => Some(format!(
+                "decoder.DecodeSequence<{}>()",
+                element_type.cs_type_string(namespace, TypeContext::Decode, true),
+            )),
             Types::Enum(enum_def)
                 if enum_def.underlying.is_some()
                     && enum_def.fixed_wire_size().is_some()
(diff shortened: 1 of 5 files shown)


bernardnormier (commented):
We also need to fix & test the decoding of a sequence (array) of bools.

bernardnormier:
It's significant faster to decode by copying an array (like we do for fixed-size numeric types) and then checking the values versus decoding the elements one by one.

InsertCreativityHere:
My C# knowledge has fallen by the wayside, but could we do something less verbose like:
```csharp
sut.DecodeSequence(
    (bool element) =>
    {
        var inMemoryValue = Unsafe.As<bool, byte>(ref element);
        if (inMemoryValue > 1)
        {
            throw new InvalidDataException("The value is out of the bool type accepted range.");
        }
    });
```

I agree with @bernardnormier though, I'd put a higher value on performance than conciseness.

pepone:
Added tests for sequences, and fixed the sequence decoding to include this check.

tools/slicec-cs/src/decoding.rs
```diff
@@ -221,10 +221,25 @@ pub fn decode_sequence(sequence_ref: &TypeRef<Sequence>, namespace: &str, encodi
                 // We always read an array even when mapped to a collection, as it's expected to be
                 // faster than decoding the collection elements one by one.
                 Some(format!(
-                    "decoder.DecodeSequence<{}>()",
+                    "decoder.DecodeSequence<{}>({})",
                     element_type.cs_type_string(namespace, TypeContext::Decode, true),
+                    if matches!(primitive, Primitive::Bool) {
+                        r#"checkElement: element =>
+    {
+        if (System.Runtime.CompilerServices.Unsafe.As<bool, byte>(ref element) > 1)
```
bernardnormier, line 229: I would create a helper static method, e.g. `SliceDecoder.CheckBoolValue`, and simplify this generated code.

bernardnormier (approved):

tools/slicec-cs/src/decoding.rs
```diff
@@ -301,11 +316,22 @@ decoder.DecodeSequenceOfOptionals(
         )
     } else {
         match element_type.concrete_type() {
-            Types::Primitive(primitive) if primitive.fixed_wire_size().is_some() => {
+            Types::Primitive(primitive) if primitive.is_numeric_or_bool() && primitive.fixed_wire_size().is_some() => {
```
InsertCreativityHere, line 319: This extra `is_numeric_or_bool` check is redundant.
We could probably remove it from the other half of this massive if/else.

All primitives with fixed wire size are either numeric or a bool.
And for decoding, what we really care about is that's fixed size, not so much the actual type.

InsertCreativityHere (approved):
Looks good!

tools/slicec-cs/src/decoding.rs
```diff
@@ -221,10 +221,25 @@ pub fn decode_sequence(sequence_ref: &TypeRef<Sequence>, namespace: &str, encodi
...
+    }"#
+                    } else {
+                        ""
+                    }
                 ))
             }
+            Types::Primitive(primitive) if matches!(primitive, Primitive::Bool) => Some(format!(
+                "decoder.DecodeSequence<{}>()",
+                element_type.cs_type_string(namespace, TypeContext::Decode, true),
+            )),
```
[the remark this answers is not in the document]
pepone: Yes, that is a leftover, I will remove it.

PR #3797, opened 2023-11-23
Add tests for streaming messages on background

### LINKED ISSUE #3792: Protobuf client streaming stops after receiving the response
I think the issue is that we are passing the dispatch cancellation token in the call to `ToAsyncEnumerable`

https://github.com/icerpc/icerpc-csharp/blob/97fca04090ee56b65f0f2381200c3f3e0333cc8e/src/IceRpc.Protobuf/IncomingRequestExtensions.cs#L70-L73

I found this while porting the Thermostat example to use Protobuf. Unfortunately, our client stream test didn't catch this because it consumes the client stream before it returns a response."

pepone:
Fix #3792 

I first thought of testing this with IceRpcConnection, but I took a different approach and added unit tests for the DispatchXxx extension methods.

There is a fix for InvokerExtensions, which uses the invocation cancellation token for the response async enumerable.
I committed a similar fix to the Dispatch helper as part of the Protobuf examples in 64db29358f7933b4d5ac3490a6b502859541efed


externl (approved):

tests/IceRpc.Protobuf.Tests/StreamTests.cs
```diff
@@ -389,6 +390,424 @@ static async Task EncodeDataAsync(PipeWriter writer)
         }
     }
 
+    /// <summary>Ensure that the async enumerable stream provided to DispatchClientStreamingAsync can be consumed
```
bernardnormier, line 393: Ensures

tests/IceRpc.Protobuf.Tests/StreamTests.cs
```diff
@@ -389,6 +390,424 @@ static async Task EncodeDataAsync(PipeWriter writer)
...
+    /// after the dispatch returns and the cancellation token provided to dispatch has been canceled.</summary>
+    [Test]
+    public async Task Dispatch_client_streaming_rpc_continues_on_background()
```
bernardnormier, line 396: I think "in background" or "in the background" would be more correct.

tests/IceRpc.Protobuf.Tests/StreamTests.cs
```diff
@@ -389,6 +390,424 @@ static async Task EncodeDataAsync(PipeWriter writer)
...
+                yield return new InputMessage()
+                {
+                    P1 = $"P{i}",
+                    P2 = i,
+                };
+            }
+        }
+    }
+
+    /// <summary>Ensure that the async enumerable stream returned by DispatchServerStreamingAsync can be consumed
```
bernardnormier, line 452: Same here.

bernardnormier (approved):

Issue #3805, opened 2023-11-28
Allow cs::identifier on custom


bernardnormier:
It would be helpful to support `cs::identifier` on custom.

The mapped identifier is used to lookup the encode and decode methods, and occasionally you want to customize this identifier.

Use case:
```slice
[cs::identifier("IceObject")]
interface Object {
   ...
}

/// The ObjectProxy data type, encoded as a service address.
[cs::type("IceRpc.Slice.Ice.IceObjectProxy")]
custom ObjectProxy
```
In this case, it would be preferable to have C# encode/decode methods with the "IceObjectProxy" name, as opposed to the default (ObjectProxy).

InsertCreativityHere:
This was implemented by #3807.

PR #3806, opened 2023-11-28
Generate encode/decode extension methods for proxies

### LINKED ISSUE #3803: Generate custom helper methods for proxies
See https://github.com/icerpc/slicec/issues/673

bernardnormier:
This PR generates encode/decode methods for proxies, and adds a number of custom *Name*Proxy types.

Fixes #3803.

### DIFF
--- a/tools/slicec-cs/src/generators/proxy_generator.rs
+++ b/tools/slicec-cs/src/generators/proxy_generator.rs
@@ -128,12 +128,18 @@ public static implicit operator {base_impl}({proxy_impl} proxy) =>
     if interface_def.supported_encodings().supports(Encoding::Slice1) {
         proxy_encoder_builder.add_comment(
             "summary",
-            "Provides extension methods for <see cref=\"SliceEncoder\" />.",
+            format!(
+                r#"
+Provides extension methods for <see cref="SliceEncoder" /> to encode a <see cref="{proxy_impl}" />."#
+            ),
         );
     } else {
         proxy_encoder_builder.add_comment(
             "summary",
-            "Provides an extension method for <see cref=\"SliceEncoder\" />.",
+            format!(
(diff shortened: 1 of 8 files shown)


tools/slicec-cs/src/generators/proxy_generator.rs
```diff
@@ -120,6 +120,98 @@ public static implicit operator {base_impl}({proxy_impl} proxy) =>
 
     code.add_block(&proxy_impl_builder.build());
 
+    let mut proxy_encoder_builder = ContainerBuilder::new(
+        &format!("{access} static class"),
+        &format!("{proxy_impl}SliceEncoderExtensions"),
```
bernardnormier, line 125: I think it makes sense to use `{proxy_impl}` in the name of this extension class and the Encode/Decode methods, as opposed to some other name.

See also #3805.

pepone (approved):

tools/slicec-cs/src/generators/proxy_generator.rs
```diff
@@ -120,6 +120,98 @@ public static implicit operator {base_impl}({proxy_impl} proxy) =>
...
+    );
+
+    if interface_def.supported_encodings().supports(Encoding::Slice1) {
+        proxy_encoder_builder.add_comment(
+            "summary",
+            "Provides extension methods for <see cref=\"SliceEncoder\" />.",
```
InsertCreativityHere, line 131: I think we should be a little more descriptive for these summaries (on the encoder and decoder).
Even just something like:
```suggestion
            "Provides extension methods for <see cref=\"SliceEncoder\" to encode a <see cref=\"{proxy_impl}\">/>.",
```

InsertCreativityHere (approved):
Looks good!

PR #3808, opened 2023-11-28
Use cs::identifier for ObjectProxy, add test

bernardnormier:
This PR adds `cs::identifier` to ObjectProxy and adds a test to make sure it works.


InsertCreativityHere (approved):
Looks good to me!

pepone (approved):

Issue #3812, opened 2023-11-29
Collections of Optional Custom Types do not Compile


InsertCreativityHere:
### Description

For collection types, we need to generate EncodeAction functions, and the implementation of this function depends on whether the collection is of value or reference types.

For custom types, the generated code assumes it's always backed by a reference type.
If the custom type you map to is actually a value type, `slicec-cs` produces code that doesn't compile.

### Reproduction steps

For example:
```
interface Pingable {}

[cs::type("IceRpc.Slice.Tests.PingableProxy")]
custom PingableProxy

compact struct MyCompactStructWithSequenceOfNullableProxies {
    i: Sequence<PingableProxy?>
}
```
Produces the following code:
```
(ref SliceEncoder encoder, IceRpc.Slice.Tests.PingableProxy? value) =>
    PingableProxySliceEncoderExtensions.EncodePingableProxy(ref encoder, value));
```
Instead of:
```
(ref SliceEncoder encoder, IceRpc.Slice.Tests.PingableProxy? value) =>
    PingableProxySliceEncoderExtensions.EncodePingableProxy(ref encoder, value!.Value));
```

InsertCreativityHere:
Actually, I don't think this is specific to collections, even this produces uncompilable code:
```
compact struct MyCompactStructWithNullableProxy {
    i: PingableProxy?
}
```

InsertCreativityHere:
I ran into this when updating the tests in `icerpc-csharp` to use the new interface-proxy-custom logic.
Everything ran fine, except this one test:
https://github.com/icerpc/icerpc-csharp/blob/main/tests/IceRpc.Slice.Tests/StructTests.Slice1.slice
_everything_ in that test produces uncompilable code.

For now, I'm just going to uncomment it.

Issue #3809, opened 2023-11-28
Add Support for Enum with Associated Values


InsertCreativityHere:
Full parser support was added for this, but we still need to implement the C# side of this: encoding / decoding these correctly and generating code as described in https://github.com/icerpc/slicec/issues/30.

PR #3813, opened 2023-11-29
Updated slicec-cs to use the latest version of slicec

InsertCreativityHere:
This PR updates the dependencies of `slicec-cs` including bringing in the latest `slicec`.

## Big changes:
- Interfaces can no longer be used as types. To use them, you must create a custom type instead.
    - While implementing this I came across a bug and had to comment out a test for now. See #3812,
- `slicec` added better location tracking for doc comment components and `slicec-cs` now uses these.
- Visitor function provide a default no-op implementation, so we don't need to have them anymore.
- Added a check that rejects enums without underlying types in `Slice2` files (now that the parser supports them). See #3809.

### DIFF
--- a/tests/IceRpc.Slice.Tests/StructTests.cs
+++ b/tests/IceRpc.Slice.Tests/StructTests.cs
@@ -9,6 +9,9 @@ namespace IceRpc.Slice.Tests;
 [Parallelizable(ParallelScope.All)]
 public sealed class StructTests
 {
+    /*
+    TODO: re-enable these tests when 'https://github.com/icerpc/icerpc-csharp/issues/3812' is fixed.
+
     [Test]
     public void Decode_slice1_compact_struct_with_nullable_proxy(
         [Values("icerpc://localhost/service", null)] string? serviceAddress)
@@ -291,4 +294,5 @@ public sealed class StructTests
                 (ref SliceDecoder decoder) => decoder.DecodeProxy<PingableProxy>() as PingableProxy?),
             Is.EqualTo(expected.I));
     }
+    */
 }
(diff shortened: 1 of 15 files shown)


externl (approved):
Looks good to me!

tests/IceRpc.Slice.Tests/StructTests.cs
```diff
@@ -9,6 +9,9 @@ namespace IceRpc.Slice.Tests;
 [Parallelizable(ParallelScope.All)]
 public sealed class StructTests
 {
+    /*
+    TODO: re-enable these tests when 'https://github.com/icerpc/icerpc-csharp/issues/3812' is fixed.
```
bernardnormier, line 13: You can also use simply #3812. It's all in the same GitHub repo.

tools/slicec-cs/src/cs_compile.rs
```diff
@@ -62,6 +65,26 @@ fn ensure_custom_types_have_type_attribute(compilation_state: &mut CompilationSt
...
+    for node in compilation_state.ast.as_slice() {
+        if let Node::Enum(enum_ptr) = node {
+            let enum_def = enum_ptr.borrow();
+            if enum_def.underlying.is_none() && !enum_def.supported_encodings().supports(Encoding::Slice1) {
+                let identifier = enum_def.identifier();
+                Diagnostic::new(Error::Syntax {
+                    message: "enums with associated fields are not supported by slicec-cs".to_owned(),
+                })
+                .add_note(
+                    format!("Try specifying an underlying type on your enum: `enum {identifier}: varint32`"),
```
bernardnormier, line 78: There is a space before the `:` with our preferred Slice formatting.

bernardnormier (approved):

tools/slicec-cs/src/cs_compile.rs
InsertCreativityHere: I always forget that. Fixed.

PR #3814, opened 2023-11-29
Simplified how we cast optional types in decoding functions.

InsertCreativityHere:
When we decode a type, it's obviously non-null (we just decoded it).
If the type it's being assigned to _is_ null though, we need to cast it. Currently, the decoding code-gen does this in 2 separate ways.
- For optional _value_ types, we do `... as type?`
- For optional _reference_ types, we do `(type?)`

It's simpler and easier to only use one of these, instead of using a different cast for each.
Both would work, I ultimately chose the `(type?)` cast as it seems slightly safer (it throws an exception for bad casts).
I tried both. Neither allowed for any more simplification than the other.
It would take a matter of seconds to switch to using the `as` syntax though. Just let me know what everyone's preferences are.

### DIFF
--- a/tools/slicec-cs/src/decoding.rs
+++ b/tools/slicec-cs/src/decoding.rs
@@ -365,17 +365,14 @@ pub fn decode_func(type_ref: &TypeRef, namespace: &str, encoding: Encoding) -> C
 }
 
 fn decode_func_body(type_ref: &TypeRef, namespace: &str, encoding: Encoding) -> CodeBlock {
-    // For value types the type declaration includes ? at the end, but the type name does not.
-    let type_name = if type_ref.is_optional && type_ref.is_value_type() {
-        type_ref.cs_type_string(namespace, TypeContext::Decode, true)
-    } else {
-        type_ref.cs_type_string(namespace, TypeContext::Decode, false)
-    };
-
     let mut code = CodeBlock::default();
-    if type_ref.is_optional && !type_ref.is_value_type() {
-        write!(code, "({type_name})");
+
+    // When we decode the type, we decode it as a non-optional.
+    // If the type is supposed to be optional, we cast it after decoding.
+    if type_ref.is_optional {
+        write!(code, "({})", type_ref.cs_type_string(namespace, TypeContext::Decode, false));
     }
+
     match &type_ref.concrete_typeref() {
         _ if type_ref.is_class_type() => {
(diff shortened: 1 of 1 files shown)


pepone (approved):
Looks good, I think we can still simplify it a bit:

```
diff --git a/tools/slicec-cs/src/decoding.rs b/tools/slicec-cs/src/decoding.rs
index 8450d16c4..3db6d7842 100644
--- a/tools/slicec-cs/src/decoding.rs
+++ b/tools/slicec-cs/src/decoding.rs
@@ -367,10 +367,11 @@ pub fn decode_func(type_ref: &TypeRef, namespace: &str, encoding: Encoding) -> C
 fn decode_func_body(type_ref: &TypeRef, namespace: &str, encoding: Encoding) -> CodeBlock {

+    let type_name = type_ref.cs_type_string(namespace, TypeContext::Decode, true);
     // When we decode the type, we decode it as a non-optional.
     // If the type is supposed to be optional, we cast it after decoding.
     if type_ref.is_optional {
-        write!(code, "({})", type_ref.cs_type_string(namespace, TypeContext::Decode, false));
+        write!(code, "({}?)", type_name);
     }

     match &type_ref.concrete_typeref() {
@@ -381,10 +382,9 @@ fn decode_func_body(type_ref: &TypeRef, namespace: &str, encoding: Encoding) ->
                 write!(
                     code,
                     "decoder.DecodeNullableClass<{}>()",
-                    type_ref.cs_type_string(namespace, TypeContext::Decode, true),
+                    type_name,
                 )
             } else {
-                let type_name = type_ref.cs_type_string(namespace, TypeContext::Decode, false);
                 write!(code, "decoder.DecodeClass<{type_name}>()")
             }
         }
@@ -405,7 +405,6 @@ fn decode_func_body(type_ref: &TypeRef, namespace: &str, encoding: Encoding) ->
         }
         TypeRefs::Struct(_) => {
             // For value types the type declaration end with ?, but the type name does not, so we set `ignore_optional`.
-            let type_name = type_ref.cs_type_string(namespace, TypeContext::Decode, true);
             write!(code, "new {type_name}(ref decoder)")
         }
         TypeRefs::CustomType(custom_type_ref) => {
```

InsertCreativityHere:
> Looks good, I think we can still simplify it a bit:

You're right, I was being too careful!
I implemented your suggestion and it saved a good number of lines.

PR #3819, opened 2023-11-30
Fix encoding of optional custom types

### LINKED ISSUE #3812: Collections of Optional Custom Types do not Compile

bernardnormier:
Fixes #3812.

### DIFF
--- a/tools/slicec-cs/src/encoding.rs
+++ b/tools/slicec-cs/src/encoding.rs
@@ -399,10 +399,18 @@ fn encode_action_body(
             let encoder_extensions_class =
                 custom_type_ref.escape_scoped_identifier_with_suffix("SliceEncoderExtensions", namespace);
             let identifier = custom_type_ref.cs_identifier(Case::Pascal);
-            write!(
-                code,
-                "{encoder_extensions_class}.Encode{identifier}(ref encoder, {value})",
-            )
+
+            if type_ref.is_optional && encoding == Encoding::Slice1 {
+                write!(
+                    code,
+                    "{encoder_extensions_class}.EncodeNullable{identifier}(ref encoder, value)",
+                )
+            } else {
+                write!(
+                    code,
+                    "{encoder_extensions_class}.Encode{identifier}(ref encoder, {value})",
+                )
+            }
         }
     }
 
--- a/tools/slicec-cs/src/encoding.rs
+++ b/tools/slicec-cs/src/encoding.rs
@@ -78,6 +78,8 @@ fn encode_type(
         concrete_typeref => {
             let value = if type_ref.is_optional && type_ref.is_value_type() {
                 format!("{param}.Value")
+            } else if type_ref.is_optional && !type_ref.is_reference_type() {
+                format!("{param} ?? default!")
             } else {
                 param.to_owned()
             };
@@ -169,8 +171,10 @@ fn encode_tagged_type(
 
     let value = if data_type.is_value_type() {
         format!("{param}.Value")
-    } else {
+    } else if data_type.is_reference_type() {
         param.to_owned()
+    } else {
+        format!("{param} ?? default!")
     };
 
     // For types with a known size, we provide a size parameter with the size of the tagged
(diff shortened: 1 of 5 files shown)


tools/slicec-cs/src/encoding.rs
```diff
@@ -351,9 +355,10 @@ fn encode_action_body(
     let mut code = CodeBlock::default();
     let is_optional = type_ref.is_optional && !is_tagged;
 
-    let value = match (is_optional, type_ref.is_value_type()) {
-        (true, false) => "value!",
-        (true, true) => "value!.Value",
+    let value = match (is_optional, type_ref.is_value_type(), type_ref.is_reference_type()) {
```
bernardnormier, line 358: I didn't change the logic of this code in this PR, even though it's definitely inelegant.

We compute this value here and then use "value" or value in the code below, which is pretty confusing. Not having any comment doesn't help either.

tools/slicec-cs/src/encoding.rs
```diff
@@ -78,6 +78,8 @@ fn encode_type(
         concrete_typeref => {
             let value = if type_ref.is_optional && type_ref.is_value_type() {
                 format!("{param}.Value")
+            } else if type_ref.is_optional && !type_ref.is_reference_type() {
```
InsertCreativityHere, line 81: Maybe add a comment about how this handles custom types?
Otherwise this branch seems redundant on the face of it.

tools/slicec-cs/src/encoding.rs
```diff
@@ -351,9 +355,10 @@ fn encode_action_body(
     let mut code = CodeBlock::default();
     let is_optional = type_ref.is_optional && !is_tagged;
 
-    let value = match (is_optional, type_ref.is_value_type()) {
-        (true, false) => "value!",
-        (true, true) => "value!.Value",
+    let value = match (is_optional, type_ref.is_value_type(), type_ref.is_reference_type()) {
+        (true, false, true) => "value!",
+        (true, true, false) => "value!.Value",
+        (true, false, false) => "(value ?? default!)",
```
InsertCreativityHere, line 361: I would explicitly mark this case as unreachable, since it's impossible for something to be both a value and reference type.
```suggestion
        (true, false, false) => "(value ?? default!)",
        (true, true, true) => unreachable!(),
```

InsertCreativityHere (approved):
Looks good!

tools/slicec-cs/src/encoding.rs
pepone: It is possible for custom, isn't it? it just mean we don't know

tools/slicec-cs/src/encoding.rs
pepone: Can we just check for custom type? I think would make the intent clear

```suggestion
            } else if type_ref.is_optional && matches!(type_ref.concrete_type(), Types::CustomType(_)) {
```

pepone (approved):

tools/slicec-cs/src/encoding.rs
InsertCreativityHere: The functions return true if we know-for-sure this is mapped to a value/reference type.
So custom is handled by `(_, false, false)`
Since we don't know that it's value type and we also don't know that it's a reference type.

`(_, true, true)` would mean we know something is a value type, and we also know it's a reference type, which is impossible.

tools/slicec-cs/src/encoding.rs
bernardnormier: I removed is_reference_type as suggested by Jose.

pepone (approved):

PR #3827, opened 2023-11-30
Add more tests for custom types

### LINKED ISSUE #3820: Need more tests for custom
As shown by #3812, we don't have enough tests for custom, especially optional of custom.

Those tests could use well-known types such as:
 - Uri - maps to a C# reference type (class Uri)
 - TimeStamp - maps to a C# value type (struct DateTime)

pepone:
Fix #3820


bernardnormier (approved):
Looks good.

PR #3831, opened 2023-12-01
Make IProxy.Invoker not nullable

bernardnormier:
This PR makes IProxy.Invoker not nullable, just like IProtobufClient.Incoker.

Rather than null, the code now uses  InvalidInvoker.Instance when it needs an invalid invoker - typically when the proxy is used for encoding and not for invoking.

### DIFF
--- a/src/IceRpc.Slice/IProxy.cs
+++ b/src/IceRpc.Slice/IProxy.cs
@@ -1,7 +1,5 @@
 // Copyright (c) ZeroC, Inc.
 
-using System.ComponentModel.DataAnnotations;
-
 namespace IceRpc.Slice;
 
 /// <summary>Represents a local ambassador for a remote service.</summary>
--- a/src/IceRpc.Slice/IProxy.cs
+++ b/src/IceRpc.Slice/IProxy.cs
@@ -1,5 +1,7 @@
 // Copyright (c) ZeroC, Inc.
 
+using System.ComponentModel.DataAnnotations;
+
 namespace IceRpc.Slice;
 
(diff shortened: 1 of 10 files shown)


src/IceRpc.Slice/IProxy.cs
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) ZeroC, Inc.
 
+using System.ComponentModel.DataAnnotations;
```
pepone, line 3: Why is this using required?

pepone (approved):
Looks good

tools/slicec-cs/src/generators/proxy_generator.rs
```diff
@@ -71,10 +71,10 @@ This remote service must implement Slice interface {slice_interface}."#
 public const string DefaultServicePath = "{default_service_path}";
 
 /// <inheritdoc/>
-public SliceEncodeOptions? EncodeOptions {{ get; init; }} = null;
+public SliceEncodeOptions? EncodeOptions {{ get; init; }}
 
 /// <inheritdoc/>
-public IceRpc.IInvoker? Invoker {{ get; init; }} = null;
+public required IceRpc.IInvoker Invoker {{ get; init; }}
```
pepone, line 77: We should do the same for the Protobuf client

src/IceRpc.Slice/IProxy.cs
bernardnormier: Not sure how this using was added. Will remove.

PR #3848, opened 2023-12-06
Fix protobuf clean target

### LINKED ISSUE #3845: Clean target doesn't work for IceRpc.Protobuf.Tests or Protobuf examples
After a "dotnet clean" on the enclosing solution:
```
C:\Users\bernard\source\repos\icerpc-csharp\tests\IceRpc.Protobuf.Tests\generated>dir
 Volume in drive C is Windows
 Volume Serial Number is DCB2-68BB

 Directory of C:\Users\bernard\source\repos\icerpc-csharp\tests\IceRpc.Protobuf.Tests\generated

12/05/2023  04:32 PM    <DIR>          .
12/05/2023  04:32 PM    <DIR>          ..
12/05/2023  04:32 PM            18,117 OperationTests.cs
12/05/2023  04:32 PM            21,685 OperationTests.IceRpc.cs
12/05/2023  04:32 PM             1,756 ServiceTests.cs
12/05/2023  04:32 PM            18,629 ServiceTests.IceRpc.cs
```

```
C:\Users\bernard\source\repos\icerpc-csharp\examples\protobuf\Greeter\Client\generated>dir
 Volume in drive C is Windows
 Volume Serial Number is DCB2-68BB

 Directory of C:\Users\bernard\source\repos\icerpc-csharp\examples\protobuf\Greeter\Client\generated

12/05/2023  04:35 PM    <DIR>          .
12/05/2023  04:35 PM    <DIR>          ..
12/05/2023  04:35 PM            15,245 Greeter.cs
12/05/2023  04:35 PM             5,509 Greeter.IceRpc.cs
```

Same "dotnet clean" works fine for the Slice tests and examples.

pepone:
Fixes  #3845

### DIFF
--- a/tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
+++ b/tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
@@ -1,8 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Copyright (c) ZeroC, Inc. -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <!-- Import ProtocTask -->
+    <!-- Import Tasks -->
     <UsingTask TaskName="IceRpc.Protobuf.Tools.ProtocTask" AssemblyFile="$(IceRpcProtobufToolsTaskAssembliesPath)IceRpc.Protobuf.Tools.dll" />
+    <UsingTask TaskName="IceRpc.Protobuf.Tools.OutputFileNamesTask" AssemblyFile="$(IceRpcProtobufToolsTaskAssembliesPath)IceRpc.Protobuf.Tools.dll" />
     <ItemGroup>
         <PropertyPageSchema Include="$(MSBuildThisFileDirectory)ProtoFile.ItemDefinition.xaml" />
         <AvailableItemName Include="ProtoFile" />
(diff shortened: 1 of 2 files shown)


externl (approved):

tools/IceRpc.Protobuf.Tools/OutputFileNamesTask.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    public override bool Execute()
+    {
+        var computedSources = new List<ITaskItem>();
+        foreach (ITaskItem source in Sources)
+        {
+            string fullPath = source.GetMetadata("FullPath");
+            ITaskItem computedSource = new TaskItem(source.ItemSpec);
+            source.CopyMetadataTo(computedSource);
+            computedSource.SetMetadata("OutputFileName", Path.GetFileNameWithoutExtension(fullPath).ToPascalCase());
```
bernardnormier, line 32: So `computedSource` is a TaskItem with a metadata OutputFileName = proto file without extension converted to Pascal case?

It would be good to clarify the doc-comment on ComputedSources above.

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -70,6 +71,11 @@
         </ItemGroup>
     </Target>
     <Target Name="ProtoClean" BeforeTargets="Clean">
+        <OutputFileNamesTask Sources="@(ProtoFile)">
```
bernardnormier, line 74: I find it odd to use this new task only for clean. Why not use it for the Compile too?

Or maybe we don't need a task for this simple substitution?

pepone: The thing is we have to convert the proto file name to PascalCase, which is not that simple to do without a C# method.

pepone: > I find it odd to use this new task only for clean. Why not use it for the Compile too?

To avoid running two tasks, as we are already have to run the compile task

PR #3852, opened 2023-12-06
Compute if Protobuf files need to be rebuilt

### LINKED ISSUE #3841: Determine whether Proto files need to be rebuilt
The `ProtoCompile` task from the `IceRpc.Protobuf.Tools` package always runs the `protoc` compiler; as a result, the generated `.cs` and `.IceRpc.cs` files are updated regardless of any changes.  Consequently, the project is never considered up to date and is rebuilt each time the `dotnet build` command is executed.

Ideally, the system should determine which .proto files have changed since the last build and only rebuild those.

pepone:
Fixes #3841


bernardnormier (approved):
Looks good.

externl (approved):
Looks good!

PR #3853, opened 2023-12-07
Use forward slashes for paths

### LINKED ISSUE #3838: Consistently use forward slashes or back slashes for paths
I noticed in this `.csproj` file that we are not consistent with using forward and back slashes when specifying paths. We should pick a style and make sure we consistently use it.

https://github.com/icerpc/icerpc-csharp/blob/70d266bcb9c26a834061d6ff0b7b0b661ea63b47/tests/IntegrationTests/IntegrationTests.csproj#L22-L30

pepone:
Fix #3838

### DIFF
--- a/build/Sign.targets
+++ b/build/Sign.targets
@@ -10,7 +10,7 @@
     <!-- Authenticode sign task that retries on failures -->
     <UsingTask TaskName="SignTask"
         TaskFactory="RoslynCodeTaskFactory"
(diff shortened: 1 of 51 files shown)


ReeceHumphreys (approved):
Looks good. Im assuming the default when dotnet creates a project is for it to be forward slash?

pepone:
There aren't any slashes for default projects.

PR #3867, opened 2023-12-12
Map Slice fields to C# properties

### LINKED ISSUE #3858: Map fields to C# properties
We currently map fields (in structs, classes, exceptions) to C# fields, occasionally readonly C# fields.

We should map them to C# properties instead because:
 - that's the common C# practice
 - it makes struct fields (mapped to C# properties in record structs) "with" capable. "with" doesn't work with plain C# fields.

bernardnormier:
This PR changes the mapping for Slice fields. Prior to this PR, they were mapped to C# fields. This PR maps them to C# properties.

Rationale:
 - C# prefers auto-implemented public properties over public fields
 - for `cs::readonly` struct, you can now clone a readonly struct with `with` - this didn't work with readonly fields
 - it's consistent with the mapping for enum fields

Fixes #3858 

Note: this PR changes the mapping for all fields (struct, class, exception), not just struct fields.

### DIFF
--- a/tools/slicec-cs/src/member_util.rs
+++ b/tools/slicec-cs/src/member_util.rs
@@ -33,15 +33,13 @@ pub fn field_declaration(field: &Field, field_type: FieldType) -> String {
         writeln!(prelude, "[{obsolete}]");
     }
 
-    let setter = if field.is_cs_readonly() { "init" } else { "set" };
-
     format!(
         "\
 {prelude}
 {access} {type_string} {name} {{ get; {setter}; }}",
         access = field.parent().access_modifier(),
         name = field.field_name(field_type),
-        field = if field.is_cs_readonly() { "init" } else { "set" },
+        setter = if field.is_cs_readonly() { "init" } else { "set" },
     )
 }
 
--- a/tools/slicec-cs/src/member_util.rs
+++ b/tools/slicec-cs/src/member_util.rs
@@ -1,11 +1,10 @@
 // Copyright (c) ZeroC, Inc.
 
 use crate::comments::CommentTag;
-use crate::cs_attributes::CsReadonly;
 use crate::cs_util::*;
 use crate::slicec_ext::*;
 use slicec::code_block::CodeBlock;
-use slicec::grammar::{Attributable, Contained, Field, Member};
+use slicec::grammar::{Contained, Field, Member};
 use slicec::utils::code_gen_util::TypeContext;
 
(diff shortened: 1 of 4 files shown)


pepone (approved):

externl (approved):

tools/slicec-cs/src/member_util.rs
```diff
@@ -34,20 +33,13 @@ pub fn field_declaration(field: &Field, field_type: FieldType) -> String {
...
 
-    // Fields inherit the visibility of their parents.
-    let mut modifiers = vec![field.parent().access_modifier()];
-
-    // Check if this field, or it's parent struct, are marked with `cs::readonly`.
-    let mut attributes = field.all_attributes().concat().into_iter();
-    if attributes.any(|a| a.downcast::<CsReadonly>().is_some()) {
-        modifiers.push("readonly");
-    }
+    let setter = if field.is_cs_readonly() { "init" } else { "set" };
```
InsertCreativityHere, line 36: For simple one line computations like this, which are _only_ used by a `format`,
it's customary to just move it directly into the `format` call.

tools/slicec-cs/src/member_util.rs
```diff
@@ -34,20 +33,13 @@ pub fn field_declaration(field: &Field, field_type: FieldType) -> String {
...
 
     format!(
         "\
 {prelude}
-{modifiers} {type_string} {name};",
-        modifiers = modifiers.join(" "),
+{access} {type_string} {name} {{ get; {setter}; }}",
+        access = field.parent().access_modifier(),
         name = field.field_name(field_type),
```
InsertCreativityHere, line 43: like this:
```suggestion
        access = field.parent().access_modifier(),
        name = field.field_name(field_type),
        field = if field.is_cs_readonly() { "init" } else { "set" },
```

InsertCreativityHere (approved):
Looks good!

tools/slicec-cs/src/member_util.rs
```diff
@@ -34,20 +33,13 @@ pub fn field_declaration(field: &Field, field_type: FieldType) -> String {
         writeln!(prelude, "[{obsolete}]");
     }
 
-    // Fields inherit the visibility of their parents.
-    let mut modifiers = vec![field.parent().access_modifier()];
-
-    // Check if this field, or it's parent struct, are marked with `cs::readonly`.
```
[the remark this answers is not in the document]
bernardnormier: Moved

PR #3871, opened 2023-12-13
Rework generated unknown, remove top-level Discriminant property

### LINKED ISSUE #3869: The Unknown enumerator should extend the base enumeration
For an unchecked enum with fields we generate something like

```
[Dunet.Union]
public abstract partial record class Shape
{
    ....
    public partial record class Triangle(uint Side1, uint Side2, uint Side3) : Shape {...}
    public partial record class Unknown {...}
}

public static class ShapeSliceDecoderExtensions
{
    public static Shape DecodeShape(this ref SliceDecoder decoder)
    {
        return decoder.DecodeVarInt32() switch
        {
            ....
            int value => new Shape.Unknown(value, decoder.DecodeSequence<byte>())
        };
}
```

When I saw this generated code, I was wondering how can it work, `Decode` returns an `Unknown` which doesn't extend `Shape`. The answer is that Dunet generates another partial that adds `Unknown : Shape`.

While this works, I think it would be clear if we do the same for all the enumerators, either always explicitly extend the parent record, or always rely on the Dunet generator to add this in the generated partial.

### LINKED ISSUE #3870: Remove Discriminant property in generated enum with fields
We currently generate a public Discriminant property in the base record class for enum with fields.

I propose to remove this property, but keep the public Discriminant const in the known enumerators and the Discriminant property in the Unknown enumerator (for unchecked enum).

For "int" enum, the int value of the enumerator is naturally important. However, for enum with fields, it's just an encoding detail. While it's doable to expose this value in C# (as currently done), we probably can't do this as cleanly in other languages. These discriminants are also comparable to tag values, and we currently don't provide any way for a C# applications to retrieve these tag values.

bernardnormier:
Fixes #3869 
Fixes #3870

### DIFF
--- a/tests/ZeroC.Slice.Tests/EnumWithFieldsTests.cs
+++ b/tests/ZeroC.Slice.Tests/EnumWithFieldsTests.cs
@@ -62,7 +62,7 @@ public class EnumWithFieldsTests
         // Assert
         Assert.That(decoded, Is.InstanceOf<Shape.Unknown>());
         Assert.That(decoder.Consumed, Is.EqualTo(encoder.EncodedByteCount));
-        Assert.That(decoded.Discriminant, Is.EqualTo(RevisedShape.Square.Discriminant));
+        Assert.That(((Shape.Unknown)decoded).Discriminant, Is.EqualTo(RevisedShape.Square.Discriminant));
     }
 
     [Test]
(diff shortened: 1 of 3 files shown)


InsertCreativityHere (approved):
Looks good.

pepone (approved):
Looks good

PR #3874, opened 2023-12-15
Remove field name escaping

### LINKED ISSUE #3859: Field name mangling
slicec-cs includes logic to mangle (change) field names when they conflict with object methods like `Equals`.

However, for an unknown reason, we disable this logic for struct fields, even though C# record structs have Equals and more. So slicec-cs produces bad code for:
```slice
struct ManglingTest {
    Equals: string
}
``` 

```
C:\Users\bernard\source\repos\icerpc-csharp\tests\ZeroC.Slice.Tests\generated\StructTests.cs(265,19): error CS0108: 'ManglingTest.Equals' hides inherited member 'ValueType.Equals(
object?)'. Use the new keyword if hiding was intended. [C:\Users\bernard\source\repos\icerpc-csharp\tests\ZeroC.Slice.Tests\ZeroC.Slice.Tests.csproj]
C:\Users\bernard\source\repos\icerpc-csharp\tests\ZeroC.Slice.Tests\generated\StructTests.cs(265,19): error CS0102: The type 'ManglingTest' already contains a definition for 'Equa
ls' [C:\Users\bernard\source\repos\icerpc-csharp\tests\ZeroC.Slice.Tests\ZeroC.Slice.Tests.csproj]
```

bernardnormier:
This PR removes the field name escaping (or mangling), where we would add a Slice prefix to a mapped field name if it matched `Equals` or other various common methods. 

It was not really correct since we didn't apply this logic to struct fields.

After this PR, a user with a field named `Equals` needs to remap it with `cs::identifier`. Otherwise, he'll get a C# compilation error.

Fixes #3859

### DIFF
--- a/tools/slicec-cs/src/builders.rs
+++ b/tools/slicec-cs/src/builders.rs
@@ -143,10 +143,8 @@ impl ContainerBuilder {
                 .data_type()
                 .cs_type_string(&field.namespace(), TypeContext::Field, false);
 
-            self.fields.push(format!(
-                "{type_string} {name}",
-                name = field.field_name(FieldType::Class),
-            ));
+            self.fields
+                .push(format!("{type_string} {name}", name = field.field_name(),));
         }
 
         self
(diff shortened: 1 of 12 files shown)


externl (approved):
Looks good!

pepone (approved):
Looks good. Much simpler now!

PR #3882, opened 2024-01-22
Update slicec-cs to generate code for Result<Success, Failure>

### LINKED ISSUE #3875: Add Support for `Result<T, E>` to C#
Now that the parser supports this new builtin type, we need to add support for it in C#.

For reference see: https://github.com/icerpc/slicec/issues/687 & https://github.com/icerpc/slicec/issues/648

bernardnormier:
This PR maps the new Slice built-in generic type Result<Success, Failure> to C#.

It also updates Dunet from version 1.10 to 1.11. Note that Dunet is now a runtime dependency (for the Dunet.Union attribute).

Fixes #3875.

### DIFF
--- a/src/ZeroC.Slice/Result.cs
+++ b/src/ZeroC.Slice/Result.cs
@@ -6,7 +6,7 @@ namespace ZeroC.Slice;
 /// type of Slice operations.</summary>
 /// <typeparam name="TSuccess">The success type.</typeparam>
 /// <typeparam name="TFailure">The failure type.</typeparam>
-/// <remarks>The Slice built-in generic type Result maps to this class in C#.</remarks>
+/// <remarks>The Slice Result type (a built-in generic type) maps to this generic class in C#.</remarks>
 [Dunet.Union]
 public abstract partial record class Result<TSuccess, TFailure>
 {
--- a/src/ZeroC.Slice/Result.cs
+++ b/src/ZeroC.Slice/Result.cs
@@ -1,0 +1,18 @@
+// Copyright (c) ZeroC, Inc.
+
+namespace ZeroC.Slice;
+
+/// <summary>A discriminated union that represents either a success or a failure. It is typically used as the return
+/// type of Slice operations.</summary>
+/// <typeparam name="TSuccess">The success type.</typeparam>
+/// <typeparam name="TFailure">The failure type.</typeparam>
+/// <remarks>The Slice built-in generic type Result maps to this class in C#.</remarks>
+[Dunet.Union]
+public abstract partial record class Result<TSuccess, TFailure>
+{
+    /// <summary>Represents a successful <see cref="Result{TSuccess, TFailure}" />.</summary>
+    public partial record class Success(TSuccess Value);
+
+    /// <summary>Represents a failed <see cref="Result{TSuccess, TFailure}" />.</summary>
+    public partial record class Failure(TFailure Value);
+}
--- a/tools/slicec-cs/src/decoding.rs
+++ b/tools/slicec-cs/src/decoding.rs
@@ -148,6 +148,7 @@ fn decode_member(member: &impl Member, namespace: &str, encoding: Encoding) -> C
                 name = enum_ref.cs_identifier(Case::Pascal),
             );
         }
+        TypeRefs::ResultType(result_type_ref) => code.write(&decode_result(result_type_ref, namespace, encoding)),
         TypeRefs::CustomType(custom_type_ref) => {
             write!(
                 code,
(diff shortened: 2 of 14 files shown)


src/ZeroC.Slice/Result.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) ZeroC, Inc.
+
+namespace ZeroC.Slice;
+
+/// <summary>A discriminated union that represents either a success or a failure. It is typically used as the return
+/// type of Slice operations.</summary>
+/// <typeparam name="TSuccess">The success type.</typeparam>
+/// <typeparam name="TFailure">The failure type.</typeparam>
+/// <remarks>The Slice built-in generic type Result maps to this class in C#.</remarks>
```
InsertCreativityHere, line 9: I'm not sure what this comment is trying to say because of how many adjectives there are; it's not clear where they stop.
It'd be more readable if they were clearly delineated, say with parenthesis:

I think this is what you're going for:
```suggestion
/// <remarks>Result (the Slice built-in generic type) maps to this class in C#.</remarks>
```

tools/slicec-cs/src/slicec_ext/type_ref_ext.rs
```diff
@@ -107,3 +108,15 @@ fn dictionary_type_to_string(dictionary_ref: &TypeRef<Dictionary>, namespace: &s
             )
     }
 }
+
+// Helper method to convert a result type into a string
```
InsertCreativityHere, line 112: Might as well make this a doc comment so it'll show up on hover.
```suggestion
/// Helper method to convert a result type into a string
```

tools/slicec-cs/src/decoding.rs
```diff
@@ -397,6 +415,42 @@ fn decode_stream_parameter(type_ref: &TypeRef, namespace: &str, encoding: Encodi
     }
 }
 
+fn decode_result_field(type_ref: &TypeRef, namespace: &str, encoding: Encoding) -> CodeBlock {
+    if type_ref.is_optional {
+        let mut decode_func_body = decode_func_body(type_ref, namespace, encoding);
+
+        // TODO: it's lame to have to do this here. We should provide a better API.
+        if matches!(type_ref.concrete_type(), Types::Sequence(_) | Types::Dictionary(_)) {
```
InsertCreativityHere, line 423: What would be the better API you'd expect? In psuedo-code.

tools/slicec-cs/src/decoding.rs
```diff
@@ -397,6 +415,42 @@ fn decode_stream_parameter(type_ref: &TypeRef, namespace: &str, encoding: Encodi
...
+        // Ditto
+        if matches!(type_ref.concrete_type(), Types::Sequence(_) | Types::Dictionary(_)) {
+            write!(
+                decode_func,
+                " as {}",
+                type_ref.cs_type_string(namespace, TypeContext::Nested, false),
+            );
+        }
+
+        decode_func.indent(); // TODO: this doesn't work
```
InsertCreativityHere, line 449: @externl, do you remember how the indentation works?

InsertCreativityHere (approved):
Looks good to me!

tools/slicec-cs/src/decoding.rs
externl: Ident returns a new indexed code block, it does not modify the existing one.

src/ZeroC.Slice/Result.cs
externl: I agree, this sentence is hard to read.

tools/slicec-cs/src/decoding.rs
bernardnormier: Easy: we should provide a Rust function that creates a CodeBlock with the full decoding logic, including this cast. Not repeat this cast code n times.

We even have different variations of this code, with a cast and with an `as` (like here).

tools/slicec-cs/src/decoding.rs
externl:
``` 
pub fn indent(&mut self) -> &mut Self {
    self.content = self.content.replace('\n', "\n    ");
    self
}

Comment (externl) [reply tools/slicec-cs/src/decoding.rs:449]: We never indent the first line. I suspect this is the issue. I guess we could name it better. The logic for this is to work in the more common usage (at least in our code so far).
```rust
format!("
foo(
    {bar},
    b,
    c
)
", bar.indent()); 
```

externl, line 449: What specifically doesn't look correct? We generally only call indent when something is about to be used as it could need to be indented multiple times depending on where it's being written.

tools/slicec-cs/src/decoding.rs
```diff
@@ -227,6 +228,23 @@ decoder.DecodeDictionary(
...
+    let decode_success = decode_result_field(success_type, namespace, encoding);
+    let decode_failure = decode_result_field(failure_type, namespace, encoding);
+
+    format!(
+        "\
+    decoder.DecodeResult(
+        {decode_success},
+        {decode_failure})"
+    )
+    .into()
```
externl, line 245:
```suggestion
    let decode_success = decode_result_field(success_type, namespace, encoding).indent();
    let decode_failure = decode_result_field(failure_type, namespace, encoding).indent();

    format!(
        "\
decoder.DecodeResult(
    {decode_success},
    {decode_failure})"
)
.into()
```

src/ZeroC.Slice/Result.cs
bernardnormier: Reworked this sentence.

tools/slicec-cs/src/decoding.rs
bernardnormier: The suggestion does not compile.

pepone (approved):

PR #3884, opened 2024-01-22
Fix Broken Code Generation for Nested Optional Collections (Fixes #3881)

### LINKED ISSUE #3881: Bad generated code for Sequence<Sequence<string>?> field
The generated code for:
```slice
compact struct SequenceHolder {
    value: Sequence<Sequence<string>?>
}
```

does not compile. Both the decoding constructor and the Encode method are syntactically incorrect.

InsertCreativityHere:
This small PR fixes the bug found in #3881, and fixes an identical bug that was causing dictionaries to break too.
In both cases we weren't correctly handled collections of optional collections. We perform a cast on the inner collection, and these casts weren't taking nullability into account when they needed to.

It also adds some 'tests', which just ensures that the generated code compiles.
Apart from the small test added in this PR, we have no tests for nested dictionaries in Slice.
We should probably add some.

@bernardnormier believes we can avoid these casts altogether by moving them into the factory.
But for now I'd like to get the bug fixed and some basic tests in place before we worry about cleaning things up.

### DIFF
--- a/tools/slicec-cs/src/cs_attributes/cs_identifier.rs
+++ b/tools/slicec-cs/src/cs_attributes/cs_identifier.rs
@@ -27,9 +27,9 @@ impl CsIdentifier {
                 report_unexpected_attribute(self, span, Some(&note), diagnostics);
             }
 
-            Attributables::SliceFile(_)
-            | Attributables::TypeAlias(_)
-            | Attributables::TypeRef(_) => report_unexpected_attribute(self, span, None, diagnostics),
+            Attributables::SliceFile(_) | Attributables::TypeAlias(_) | Attributables::TypeRef(_) => {
+                report_unexpected_attribute(self, span, None, diagnostics)
+            }
 
             _ => {}
         }
--- a/tools/slicec-cs/src/decoding.rs
+++ b/tools/slicec-cs/src/decoding.rs
@@ -198,7 +198,7 @@ pub fn decode_dictionary(dictionary_ref: &TypeRef<Dictionary>, namespace: &str, 
         write!(
             decode_value,
             " as {}",
-            value_type.cs_type_string(namespace, TypeContext::Nested, false),
+            value_type.cs_type_string(namespace, TypeContext::Nested, true),
         );
     }
 
@@ -239,7 +239,7 @@ pub fn decode_sequence(sequence_ref: &TypeRef<Sequence>, namespace: &str, encodi
         write!(
             code,
             "({}[])",
-            element_type.cs_type_string(namespace, TypeContext::Nested, true),
+            element_type.cs_type_string(namespace, TypeContext::Nested, false),
         );
     };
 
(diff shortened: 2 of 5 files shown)


tools/slicec-cs/src/decoding.rs
```diff
@@ -239,7 +239,7 @@ pub fn decode_sequence(sequence_ref: &TypeRef<Sequence>, namespace: &str, encodi
         write!(
             code,
             "({}[])",
-            element_type.cs_type_string(namespace, TypeContext::Nested, true),
+            element_type.cs_type_string(namespace, TypeContext::Nested, false),
```
InsertCreativityHere, line 242: This bool sets `ignore_optional`.

Before, it was true, so the cast type would be missing a `?` when it needed one.
Changing this to `false` means we add the `?` for optional types now.

tools/slicec-cs/src/decoding.rs
```diff
@@ -198,7 +198,7 @@ pub fn decode_dictionary(dictionary_ref: &TypeRef<Dictionary>, namespace: &str,
         write!(
             decode_value,
             " as {}",
-            value_type.cs_type_string(namespace, TypeContext::Nested, false),
+            value_type.cs_type_string(namespace, TypeContext::Nested, true),
```
InsertCreativityHere, line 201: This bool sets `ignore_optional`.

Because we're using an `as` cast, we should be ignoring the optionality and letting the type-inference engine handle it.
This is because nullable reference types can't be used in `as` expressions.

tools/slicec-cs/src/cs_attributes/cs_identifier.rs
```diff
@@ -27,9 +27,9 @@ impl CsIdentifier {
                 report_unexpected_attribute(self, span, Some(&note), diagnostics);
             }
 
-            Attributables::SliceFile(_)
-            | Attributables::TypeAlias(_)
-            | Attributables::TypeRef(_) => report_unexpected_attribute(self, span, None, diagnostics),
+            Attributables::SliceFile(_) | Attributables::TypeAlias(_) | Attributables::TypeRef(_) => {
```
InsertCreativityHere, line 30: Unrelated formatting change.

bernardnormier:
> @bernardnormier believes we can avoid these casts altogether by moving them into the factory. But for now I'd like to get the bug fixed and some basic tests in place before we worry about cleaning things up.

After thinking more about it, I think it's cleaner/clearer to leave the factory alone and cast the return value.

tools/slicec-cs/src/decoding.rs
bernardnormier: The inconsistency between sequences and dictionaries is odd. I'd rather use a cast for both.

bernardnormier (approved):

InsertCreativityHere:
I agree the inconsistency is weird.
I'm going to merge this PR as is just to fix the issue, and do a few rounds of cleanup on `slicec-cs`.
I'll fix the inconsistency in one of those!

PR #3886, opened 2024-01-23
Generate doc comment for @param on enumerator fields

bernardnormier:
This PR updates slicec-cs to generate doc-comments (`<param name="..." ... />`) from @param doc comments on Slice enumerators.

### DIFF
--- a/tools/slicec-cs/src/generators/enum_generator.rs
+++ b/tools/slicec-cs/src/generators/enum_generator.rs
@@ -197,9 +197,6 @@ fn enumerators_as_nested_records(enum_def: &Enum) -> CodeBlock {
         let mut builder = ContainerBuilder::new("public partial record class", "Unknown");
 
         builder
-            .add_field("int Discriminant".to_owned())
-            .add_field("global::System.ReadOnlyMemory<byte> Fields".to_owned())
-            .add_base(enum_def.escape_identifier())
             .add_comment(
                 "summary",
                 format!(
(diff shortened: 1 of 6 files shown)


tools/slicec-cs/src/slicec_ext/member_ext.rs
```diff
@@ -40,14 +42,30 @@ impl FieldExt for Field {
...
+            enumerator.comment().and_then(|comment| {
+                // If it does, search the comment for a '@param' tag with this field's identifier and return it.
+                comment
+                    .params
+                    .iter()
+                    .find(|param_tag| param_tag.identifier.value == self.identifier())
+                    .map(|param_tag| format_comment_message(&param_tag.message, &self.namespace()))
+            })
+        } else {
+            None
```
InsertCreativityHere, line 58: I'd just panic here. If we call this function on anything other than an `Enumerator`, we're doing something wrong. Panicing will alert us to this fact, whereas returning `None` gives the impression this is fine to call on non-enumerator fields.
```suggestion
            panic!("Called 'formatted_param_doc_comment' on field outside of enumerator!");
```

tools/slicec-cs/src/generators/enum_generator.rs
```diff
@@ -197,28 +197,16 @@ fn enumerators_as_nested_records(enum_def: &Enum) -> CodeBlock {
...
-                "Discriminant",
-                "The discriminant of this unknown enumerator.",
-            )
-            .add_comment_with_attribute(
-                "param",
-                "name",
-                "Fields",
-                "The encoded fields of this unknown enumerator.",
-            )
+            .add_field("Discriminant", "int", Some("The discriminant of this enumerator."))
```
InsertCreativityHere, line 207: The previous doc comment had 'unknown' in it's description.
```suggestion
            .add_field("Discriminant", "int", Some("The discriminant of this unknown enumerator."))
```
Feel free to ignore if this was intentional, but just know we kept the 'unknown' in the doc comment underneath this one.

tools/slicec-cs/src/slicec_ext/member_ext.rs
```diff
@@ -31,6 +31,8 @@ impl<T: Member> MemberExt for T {
 pub trait FieldExt {
     /// Check if this field, or its parent struct, are marked with `cs::readonly`.
     fn is_cs_readonly(&self) -> bool;
+
+    fn formatted_param_doc_comment(&self) -> Option<String>;
```
InsertCreativityHere, line 35: It'd be nice to copy the doc comment on this.
And if you went with the `panic`, also mention this should only be called on enumerators, because those are the only place fields support `@param`.

InsertCreativityHere (approved):
Looks good.

I believe with this PR, `slicec` and `slicec-cs` are finally back in full parity with each other!

pepone (approved):

externl (approved):

tools/slicec-cs/src/slicec_ext/member_ext.rs
bernardnormier: Added doc-comment.

PR #3890, opened 2024-01-29
Remove `TypeContext`

InsertCreativityHere:
This PR Removes the need for ‘TypeContext’ from `slicec-cs`. It’s purely refactoring; there should be no changes to the logic. We used this enum for 2 purposes:

## Purpose 1

`cs_type_string` would change the mapping of sequences and dictionaries based on this field. 
Now, instead of 1 function which matches 3 possible TypeContexts, we have 3 separate functions:
`cs_type_string(… TypeContext::Field) -> field_type_string `
`cs_type_string(… TypeContext::IncomingParam) -> incoming_type_string`
` cs_type_string(… TypeContext::OutgoingParam) -> outgoing_type_string`

`incoming_type_string` and `outgoing_type_string` just delegate to `field_type_string`, for types other than seuqence or dictionary.

## Purpose 2

 Various parameter-code used it to tell whether we were on the dispatch side or not.

Now, we just pass a bool named `is_dispatch` which we already had in some functions anyways.
`IncomingParam -> false`, `OutgoingParam -> true`
Usually these matches matched `TypeContext::Field` to `unreachable!`.
This change lets us remove all these dead branches.

Additionally, in `encoding.rs` we used `TypeContext` to tell whether we generated for fields or parameters.
Instead of TypeContext, these functions now also take a bool: `is_outgoing_param`
`OutgoingParam -> true` and `Field -> false`.
Because this is all encoding code, we never encounter `IncomingParam`.

### DIFF
--- a/tools/slicec-cs/src/builders.rs
+++ b/tools/slicec-cs/src/builders.rs
@@ -10,7 +10,6 @@ use crate::slicec_ext::*;
 use slicec::code_block::CodeBlock;
 use slicec::grammar::{Class, Commentable, Encoding, Entity, Operation, *};
 use slicec::supported_encodings::SupportedEncodings;
-use slicec::utils::code_gen_util::TypeContext;
 
 pub trait Builder {
     fn build(&self) -> CodeBlock;
@@ -142,9 +141,7 @@ impl ContainerBuilder {
 
     pub fn add_fields(&mut self, fields: &[&Field]) -> &mut Self {
         for field in fields {
-            let type_string = field
-                .data_type()
-                .cs_type_string(&field.namespace(), TypeContext::Field, false);
+            let type_string = field.data_type().field_type_string(&field.namespace(), false);
 
             self.add_field(
                 &field.field_name(),
@@ -321,7 +318,7 @@ impl FunctionBuilder {
         self
     }
 
-    pub fn add_operation_parameters(&mut self, operation: &Operation, context: TypeContext) -> &mut Self {
+    pub fn add_operation_parameters(&mut self, operation: &Operation, is_dispatch: bool) -> &mut Self {
         let parameters = operation.parameters();
 
         // Find an index such that all parameters after it are optional (but not streamed)
@@ -338,10 +335,10 @@ impl FunctionBuilder {
         };
 
         for (index, parameter) in parameters.iter().enumerate() {
-            let parameter_type = parameter.cs_type_string(&operation.namespace(), context, false);
+            let parameter_type = parameter.cs_type_string(&operation.namespace(), false, is_dispatch);
             let parameter_name = parameter.parameter_name();
 
-            let default_value = if context == TypeContext::OutgoingParam && (index >= trailing_optional_parameters_index) {
+            let default_value = if is_dispatch && (index >= trailing_optional_parameters_index) {
                 match parameter.data_type.concrete_typeref() {
                     // Sequences of fixed-size numeric types are mapped to `ReadOnlyMemory<T>` and have to use
                     // 'default' as their default value. Other optional types are mapped to nullable types and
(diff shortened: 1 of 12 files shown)


tools/slicec-cs/src/builders.rs
```diff
@@ -366,34 +363,26 @@ impl FunctionBuilder {
...
+                Some("null"),
+                Some("The invocation features.".to_owned()),
+            );
+        } else {
+            self.add_parameter(
+                "IceRpc.Features.IFeatureCollection",
+                &escape_parameter_name(&parameters, "features"),
+                None,
+                Some("The dispatch features.".to_owned()),
+            );
```
externl, line 379: This looks backwards.

tools/slicec-cs/src/generators/proxy_generator.rs
```diff
@@ -406,7 +406,7 @@ fn proxy_base_operation_impl(operation: &Operation, namespace: &str) -> CodeBloc
     let mut builder = FunctionBuilder::new("public", &return_task, &async_name, FunctionType::ExpressionBody);
     builder.set_inherit_doc(true);
     builder.add_obsolete_attribute(operation);
-    builder.add_operation_parameters(operation, TypeContext::OutgoingParam);
+    builder.add_operation_parameters(operation, true);
```
externl, line 409: If it's outgoing, and the bool is `is_dispatch`, this seems backwards.

tools/slicec-cs/src/builders.rs
externl: IMO using a bool makes this ever error prone. What about using a two discriminant enum?

InsertCreativityHere: Yes, I picked a better name for this bool. I was too mechanistic in my translation.

tools/slicec-cs/src/slicec_ext/type_ref_ext.rs
```diff
@@ -3,14 +3,14 @@
...
 
 pub trait TypeRefExt {
     /// Is this type known to map to a C# value type?
     fn is_value_type(&self) -> bool;
 
-    /// The C# mapped type for this type reference.
-    fn cs_type_string(&self, namespace: &str, context: TypeContext, ignore_optional: bool) -> String;
+    fn field_type_string(&self, namespace: &str, ignore_optional: bool) -> String;
+    fn incoming_type_string(&self, namespace: &str, ignore_optional: bool) -> String;
+    fn outgoing_type_string(&self, namespace: &str, ignore_optional: bool) -> String;
```
bernardnormier, line 13: I would rename these functions incoming_parameter_type_string / outgoing_parameter_type_string

bernardnormier: I would add:
incoming_stream_parameter_type_string / outgoing_stream_parameter_string, where the normal functions are "no stream".

bernardnormier: I would replace ignore_optional by a function that removes `?` at the end of the type string, and is called when needed.

tools/slicec-cs/src/slicec_ext/member_ext.rs
```diff
@@ -63,24 +62,28 @@ impl FieldExt for Field {
...
 
     /// Returns the message of the `@param` tag corresponding to this parameter from the operation it's part of.
     /// If the operation has no doc comment, or a matching `@param` tag, this returns `None`.
     fn formatted_param_doc_comment(&self) -> Option<String>;
 }
 
 impl ParameterExt for Parameter {
-    fn cs_type_string(&self, namespace: &str, context: TypeContext, ignore_optional: bool) -> String {
-        let type_str = self.data_type().cs_type_string(namespace, context, ignore_optional);
+    fn cs_type_string(&self, namespace: &str, ignore_optional: bool, is_outgoing: bool) -> String {
```
bernardnormier, line 73: I would replace this function by a pair of TypeRefExt functions.

InsertCreativityHere:
Closed in favor of #3893

PR #3893, opened 2024-01-30
Split up `TypeRefExt::cs_type_string`

InsertCreativityHere:
This PR is a first step towards refactoring our `cs_type_string` functions:

Currently, `TypeRefExt::cs_type_string` takes a `TypeContext`, and has 3 different behaviors depending the type-context.
This PR splits this function into 3: `field_type_string`, `incoming_parameter_type_string`, and `outgoing_parameter_type_string`.
The behavior is identical to before, I've just split each context's match case into a standalone function.

Note that `ParameterExt` has it's own function also named `cs_type_string`.
I will deal with this in my next PR, but wanted to split it up to keep these easy to review.
`ParameterExt::cs_type_string` took a `bool: ignore_optional`, but this was _always_ false, so I removed the parameter and hard-coded false. Otherwise this function was left alone.

### DIFF
--- a/tools/slicec-cs/src/decoding.rs
+++ b/tools/slicec-cs/src/decoding.rs
@@ -8,7 +8,7 @@ use crate::slicec_ext::*;
 use convert_case::Case;
 use slicec::code_block::CodeBlock;
 use slicec::grammar::*;
-use slicec::utils::code_gen_util::{get_bit_sequence_size, TypeContext};
+use slicec::utils::code_gen_util::get_bit_sequence_size;
 
 /// Compute how many bits are needed to decode the provided members, and if more than 0 bits are needed,
 /// this generates code that creates a new `BitSequenceReader` with the necessary capacity.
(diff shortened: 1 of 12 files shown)


externl (approved):

externl:
Looks good, but it fails, so there's a bug somewhere

InsertCreativityHere:
I forgot an exclamation point...

tools/slicec-cs/src/decoding.rs
```diff
@@ -75,7 +73,7 @@ pub fn default_activator(encoding: Encoding) -> &'static str {
 fn decode_member(member: &impl Member, namespace: &str, encoding: Encoding) -> CodeBlock {
     let mut code = CodeBlock::default();
     let data_type = member.data_type();
-    let type_string = data_type.cs_type_string(namespace, TypeContext::IncomingParam, true);
+    let type_string = data_type.incoming_parameter_type_string(namespace, true);
```
bernardnormier, line 76: Should we rename this function "decode_parameter"?

bernardnormier (approved):

tools/slicec-cs/src/decoding.rs
InsertCreativityHere: No, this function is used for both fields and parameters.
But the fact we were passing `IncomingParam` is actually bogus.
We only use the `type_string` variable for structs and classes... which don't have a special mapping as parameters.

PR #3894, opened 2024-01-30
Inline 'EncodingBlockBuilder'.

InsertCreativityHere:
The `EncodingBlockBuilder` is only used by structs now, since exceptions are Slice1 only now.

Additionally, while the idea of the builder is to add different encoding-blocks at runtime, in truth, we always know what each block should be. So there's no need to dynamically add different blocks.

This PR simplifies the code by deleting `EncodingBlockBuilder` and inlining a simplified form of it's logic where we use it in the `struct_generator`. I'm sad to see it go.


tools/slicec-cs/src/generators/struct_generator.rs
```diff
@@ -128,3 +121,55 @@ pub fn generate_struct(struct_def: &Struct) -> CodeBlock {
 
     builder.build()
 }
+
+pub fn generate_encoding_blocks(
```
externl, line 125: I think a comment would be useful or what an encoding block actually is.

externl (approved):

tools/slicec-cs/src/generators/struct_generator.rs
```diff
@@ -128,3 +121,55 @@ pub fn generate_struct(struct_def: &Struct) -> CodeBlock {
...
+{slice2_block}
+}}
+",
+                    slice1_block = slice1_block.indent(),
+                    slice2_block = slice2_block.indent(),
+                )
+                .into()
+            } else {
+                unreachable!("it's impossible to have an empty Slice2 block with a non empty Slice1 block");
+            }
```
ReeceHumphreys, line 172:
```suggestion
            match (slice1_block.is_empty(), slice2_block.is_empty()) {
                // Both slices are equal
                (false, false) if slice1_block.to_string() == slice2_block.to_string() => slice2_block,

                // Slice1 is empty, Slice2 is not
                (true, false) => format!(
                    "\
if ({encoding_source}.Encoding != SliceEncoding.Slice1) // Slice2 only
{{
{slice2_block}
}}
                ",
                    slice2_block = slice2_block.indent(),
                )
                .into(),

                // Neither Slice1 nor Slice2 is empty
                (false, false) => format!(
                    "\
if ({encoding_source}.Encoding == SliceEncoding.Slice1)
{{
{slice1_block}
}}
else // Slice2
{{
{slice2_block}
}}
                ",
                    slice1_block = slice1_block.indent(),
                    slice2_block = slice2_block.indent(),
                )
                .into(),

                // Other cases (Slice1 is not empty, Slice2 is)
                // or both are empty - which seems to be treated as an error in the original code.
                _ => unreachable!("it's impossible to have an empty Slice2 block

```

I think this would be a bit clearer of a way to write this. Following all of the ifs can be a bit trickY! (Also this suggestion probably needs additional formatting - was just to communicate the idea)

ReeceHumphreys (approved):
Looks good

InsertCreativityHere:
I simplified the implementation, since only branch was ever actually hit.
@externl and @ReeceHumphreys, could you give it another look over?

externl (approved):

PR #3904, opened 2024-01-31
Hardcode 'sliceP_' Prefix.

InsertCreativityHere:
We have the function `parameter_name_with_prefix`, which accepts an arbitrary prefix and computes the
'parameter string representation' of a member, with that prefix.

This prefix is always `sliceP_` and this PR simplifies the logic by hard-coding this value.
It also adds doc comments to the 2 functions it touches.

### DIFF
--- a/tools/slicec-cs/src/slicec_ext/member_ext.rs
+++ b/tools/slicec-cs/src/slicec_ext/member_ext.rs
@@ -9,7 +9,7 @@ use slicec::utils::code_gen_util::TypeContext;
 
 pub trait MemberExt {
     fn parameter_name(&self) -> String;
-    fn parameter_name_with_prefix(&self, prefix: &str) -> String;
+    fn parameter_name_with_prefix(&self) -> String;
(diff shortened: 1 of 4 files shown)


tools/slicec-cs/src/slicec_ext/member_ext.rs
```diff
@@ -18,9 +18,9 @@ impl<T: Member> MemberExt for T {
         escape_keyword(&self.cs_identifier(Case::Camel))
     }
 
-    fn parameter_name_with_prefix(&self, prefix: &str) -> String {
-        let name = prefix.to_owned() + &self.cs_identifier(Case::Camel);
-        escape_keyword(&name)
+    /// Returns this parameter's C# identifier (in camel case) with a `sliceP_` prefix.
+    fn parameter_name_with_prefix(&self) -> String {
```
bernardnormier, line 22: escape_keyword seems clearly unnecessary here.

bernardnormier (approved):

PR #3907, opened 2024-01-31
Merge `request_decode_func` and `return_value_decode_func`

InsertCreativityHere:
This PR merges `request_decode_func` and `return_value_decode_func` into a single function, since they were actually identical functions. This new function is named: `decode_non_streamed_parameters_func`.

It was fairly small, and after this change, `decode_non_streamed_parameters` was only called in one spot, in this function, so I also went ahead and inlined the function there.

### DIFF
--- a/tools/slicec-cs/src/decoding.rs
+++ b/tools/slicec-cs/src/decoding.rs
@@ -514,10 +514,10 @@ pub fn decode_non_streamed_parameters_func(non_streamed_parameters: &[&Parameter
             let body_content = code.indent();
             format!(
                 "\
-    (ref SliceDecoder decoder) =>
-    {{
-        {body_content}
-    }}",
+(ref SliceDecoder decoder) =>
+{{
+    {body_content}
+}}",
             )
             .into()
         }
--- a/tools/slicec-cs/src/decoding.rs
+++ b/tools/slicec-cs/src/decoding.rs
@@ -484,53 +484,44 @@ pub fn decode_non_streamed_parameters_func(non_streamed_parameters: &[&Parameter
         }
 
         // Otherwise we return a full multi-line lambda function for decoding the parameters.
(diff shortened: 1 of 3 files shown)


tools/slicec-cs/src/decoding.rs
```diff
@@ -468,39 +468,60 @@ fn decode_func_body(type_ref: &TypeRef, namespace: &str, encoding: Encoding) ->
...
+            format!(
+                "\
+    (ref SliceDecoder decoder) =>
+    {{
+        {body_content}
+    }}",
+            )
+            .into()
+        }
+    }
```
externl, line 524: This formatting looks dubious. It should be left indented all the way. You should review the generated code to see that it looks correct.

externl (approved):
See comment

tools/slicec-cs/src/decoding.rs
```diff
@@ -468,39 +468,60 @@ fn decode_func_body(type_ref: &TypeRef, namespace: &str, encoding: Encoding) ->
...
+            }
+            writeln!(code, "return {};", non_streamed_parameters.to_argument_tuple());
+
+            let body_content = code.indent();
+            format!(
+                "\
+    (ref SliceDecoder decoder) =>
+    {{
+        {body_content}
+    }}",
```
InsertCreativityHere, line 520:
```suggestion
(ref SliceDecoder decoder) =>
{{
    {body_content}
}}",
```

InsertCreativityHere (commented):
Looks good

tools/slicec-cs/src/decoding.rs
InsertCreativityHere: Yes, good catch!
It got auto-indented when I copy/pasted this function, and I didn't see it.

PR #3915, opened 2024-02-05
Use casts instead of 'as' conversions for decoding.

InsertCreativityHere:
Currently, in the decoding logic, we use a mix of casts and `as` conversions.
(Casts for decoding sequences, and `as` for dictionaries and results).

This PR changes the latter two to also use casts.
And in the process, consolidates this logic into a new function: `decode_type_with_cast`.

### DIFF
--- a/tools/slicec-cs/src/decoding.rs
+++ b/tools/slicec-cs/src/decoding.rs
@@ -167,13 +167,7 @@ fn decode_dictionary(dictionary_ref: &TypeRef<Dictionary>, namespace: &str, enco
 
     // decode key
     let decode_key = decode_func(key_type, namespace, encoding);
-
-    // decode value
-    let mut decode_value = decode_func(value_type, namespace, encoding);
-    if matches!(value_type.concrete_type(), Types::Sequence(_) | Types::Dictionary(_)) {
-        write!(decode_value, " as {}", value_type.field_type_string(namespace, true));
-    }
-
+    let decode_value = decode_type_with_cast(value_type, namespace, encoding, false);
     let dictionary_type = dictionary_ref.incoming_parameter_type_string(namespace, true);
     let decode_key = decode_key.indent();
     let decode_value = decode_value.indent();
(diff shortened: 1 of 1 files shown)


tools/slicec-cs/src/decoding.rs
```diff
@@ -204,8 +198,8 @@ fn decode_result(result_type_ref: &TypeRef<ResultType>, namespace: &str, encodin
     let success_type = &result_type_ref.success_type;
     let failure_type = &result_type_ref.failure_type;
 
-    let decode_success = decode_result_field(success_type, namespace, encoding);
-    let decode_failure = decode_result_field(failure_type, namespace, encoding);
+    let decode_success = decode_type_with_cast(success_type, namespace, encoding, success_type.is_optional).indent();
```
externl, line 201: Maybe remove the with cast here, it's only used for sequences and dictionaries. What was wrong with the previous name?

tools/slicec-cs/src/decoding.rs
```diff
@@ -382,27 +376,25 @@ fn decode_stream_parameter(type_ref: &TypeRef, namespace: &str, encoding: Encodi
     }
 }
 
-fn decode_result_field(type_ref: &TypeRef, namespace: &str, encoding: Encoding) -> CodeBlock {
-    let mut decode_func = match type_ref.is_optional {
-        true => decode_func_body(type_ref, namespace, encoding),
-        false => decode_func(type_ref, namespace, encoding),
+/// Used by `decode_dictionary` and `decode_result` to generate a decoding action that includes a cast for collections.
```
externl, line 379: I would remove where it's used. Eventually we'll use it somewhere else and never update this comment.

externl (approved):

tools/slicec-cs/src/decoding.rs
InsertCreativityHere: The previous name was specific to results. Now this logic is re-used for both dictionaries and results.

This function is identical to `decode_func`, it just adds a cast out-front.
So I can't think of a better name than `decode_func_with_cast`.

But I expanded the doc comment explaining when and why to call this function.

tools/slicec-cs/src/decoding.rs
InsertCreativityHere: I completely rewrote the doc comment.

externl (approved):

PR #3916, opened 2024-02-05
Simplify 'encode_action_body'.

InsertCreativityHere:
This PR simplifies the logic of `encode_action_body` and fixes a bug with
the encoding of tagged custom types (which map to C# value types) in Slice1 mode (quite niche).

----

For example, take this snippet:
```slice
class Hello {
    tag(1) d1: CustomType?
}
```
We currently generate this:
```
encoder.StartSlice(SliceTypeId);
if (this.D1 is MyCustomType d1_)
{
    encoder.EncodeTagged(1, TagFormat.FSize, d1_, (ref SliceEncoder encoder, MyCustomType value) => CustomTypeSliceEncoderExtensions.EncodeNullableCustomType(ref encoder, value));
}
encoder.EndSlice(true);
```
Notably, this calls `EncodeNullableCustomType` which is incorrect. We know here that `value` is non-null.
Even the type of the lambda is `MyCustomType`, with no `?` suffix.

This PR fixes this to call the non-nullable function:
```
encoder.StartSlice(SliceTypeId);
if (this.D1 is MyCustomType d1_)
{
    encoder.EncodeTagged(1, TagFormat.FSize, d1_, (ref SliceEncoder encoder, MyCustomType value) => CustomTypeSliceEncoderExtensions.EncodeCustomType(ref encoder, value));
}
encoder.EndSlice(true);
```

### DIFF
--- a/tools/slicec-cs/src/encoding.rs
+++ b/tools/slicec-cs/src/encoding.rs
@@ -344,7 +344,6 @@ fn encode_action_body(
     encoding: Encoding,
     is_tagged: bool,
 ) -> CodeBlock {
-    let mut code = CodeBlock::default();
     let is_optional_and_untagged = type_ref.is_optional && !is_tagged;
 
     let value = match (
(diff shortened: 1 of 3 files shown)


tools/slicec-cs/src/encoding.rs
```diff
@@ -362,62 +361,44 @@ fn encode_action_body(
...
-                write!(code, "encoder.EncodeNullableClass(value)");
+                "encoder.EncodeNullableClass(value)".into()
             } else {
-                write!(code, "encoder.EncodeClass(value)");
+                "encoder.EncodeClass(value)".into()
             }
         }
-        TypeRefs::Primitive(primitive_ref) => write!(code, " encoder.Encode{}({value})", primitive_ref.type_suffix()),
+
+        // TODO: does anyone think we need this space? V
```
externl, line 370: I don't think so
```suggestion
```

externl (approved):

InsertCreativityHere:
Part of this PR got spun off into #3916.
So, to save hassle on merge conflicts, I'm going to wait to merge this PR until that one, since this PR will be much simpler to fix the conflicts on.

PR #3922, opened 2024-02-09
Fix server namespaces

### LINKED ISSUE #3910: Review and fix namespace names in server examples
The namespace name in the server examples is supposed to be <ExampleName>Server, and it was initially correct.

However, we renamed many examples without changing this name. For example, Logger used to be GreeterLog, and the namespace remains GreeterLogServer.

bernardnormier:
This PR fixes the server namespaces in the examples.

FIxes #3910.


ReeceHumphreys (approved):

pepone (approved):
Looks good!

PR #3924, opened 2024-02-12
Remove `ignore_optional` Flag from `xxx_type_string` Functions

InsertCreativityHere:
This PR does what the title says.
In the handful of places where we actually set this to `true`, we now have a new function: `remove_optional_modifier_from`.
We get the fully typed string (optional and all), and if we don't want the optional part, we call this function on it.

### DIFF
--- a/tools/slicec-cs/src/slicec_ext/mod.rs
+++ b/tools/slicec-cs/src/slicec_ext/mod.rs
@@ -22,6 +22,16 @@ pub use primitive_ext::PrimitiveExt;
 pub use slice_encoding_ext::EncodingExt;
 pub use type_ref_ext::TypeRefExt;
 
+/// If the provided type string is optional (it ends with a '?' character), that character is removed.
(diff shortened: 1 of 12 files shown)


tools/slicec-cs/src/slicec_ext/mod.rs
```diff
@@ -22,6 +22,15 @@ pub use primitive_ext::PrimitiveExt;
 pub use slice_encoding_ext::EncodingExt;
 pub use type_ref_ext::TypeRefExt;
 
+/// If the provided type string is optional (it ends with a '?' character), that character is removed.
+/// This does not affect nested types that are optional.
+pub fn remove_optional_modifier_from(mut type_string: String) -> String {
+    if type_string.ends_with('?') {
+        type_string.pop();
```
externl, line 29: There's no trim_end in rust?

externl (approved):

PR #3925, opened 2024-02-12
Simplify `decode_sequence` function.

InsertCreativityHere:
This PR untangles and simplifies the `decode_sequence` function.
Before this PR, it was (probably) the longest function in `slicec-cs` at 153 lines long. Now it's only 63 (and easier to read too).

I don't expect the changes to this function to be reviewable, but I'm _pretty_ sure no logic has been altered. So it's fine! 😉

What is worth reviewing are the smaller changes it makes to `FunctionCallBuilder`:
- Now you can specify type parameters for calling generic functions
- We no longer print arguments on multiple lines if there are 0 arguments.

### DIFF
--- a/tools/slicec-cs/src/decoding.rs
+++ b/tools/slicec-cs/src/decoding.rs
@@ -260,7 +260,7 @@ new {sequence_type}(
 
     if has_cs_type_attribute {
         match element_type.concrete_type() {
-            Types::Primitive(primitive) if primitive.fixed_wire_size().is_some() && !(element_type.is_optional && (encoding != Encoding::Slice1 || encoding == Encoding::Slice1)) => {
+            Types::Primitive(primitive) if primitive.fixed_wire_size().is_some() && !uses_bit_sequence => {
                 builder.set_type_argument(remove_optional_modifier_from(incoming_element_type_string));
                 builder.add_argument_if(*primitive == Primitive::Bool, "checkElement: SliceDecoder.CheckBoolValue");
                 wrap_code_in_custom_constructor(builder)
--- a/tools/slicec-cs/src/decoding.rs
+++ b/tools/slicec-cs/src/decoding.rs
@@ -213,7 +213,7 @@ decoder.DecodeResult(
 // TODO try and untangle this function.
 fn decode_sequence(sequence_ref: &TypeRef<Sequence>, namespace: &str, encoding: Encoding) -> CodeBlock {
     let element_type = &sequence_ref.element_type;
-    let incoming_element_type_string = element_type.incoming_parameter_type_string(namespace);
+    let element_type_string = element_type.field_type_string(namespace);
     let has_cs_type_attribute = sequence_ref.has_attribute::<CsType>();
 
     let uses_bit_sequence = element_type.is_optional && encoding != Encoding::Slice1;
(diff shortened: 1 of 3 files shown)


externl:
You should do a before and after diff of the generated code.

tools/slicec-cs/src/slicec_ext/enum_ext.rs
```diff
@@ -4,7 +4,7 @@ use crate::slicec_ext::primitive_ext::PrimitiveExt;
 use slicec::grammar::*;
 
 pub trait EnumExt {
-    fn get_underlying_cs_type(&self) -> String;
+    fn get_underlying_cs_type(&self) -> String; // TODO This could become 'str'!
```
externl, line 7: Might be better to open an issue to update this so it doesn't get lost or forgotten about.

tools/slicec-cs/src/decoding.rs
```diff
@@ -210,161 +210,70 @@ decoder.DecodeResult(
...
-    {})",
-                    decode_func(element_type, namespace, encoding, false).indent(),
-                )
+        }
+        _ => {
+            if has_cs_type_attribute {
+                uses_sequence_factory = true;
+                builder.add_argument(format!("sequenceFactory: (size) => new {sequence_type}(size)"));
             }
+            builder.add_argument(decode_func(element_type, namespace, encoding, false).indent());
```
externl, line 254: Is this indent still correct?

InsertCreativityHere:
I ran a clean build of main, and copied the `src`, `examples`, and `tests` directories into a new folder and committed them.
Then I ran a clean build of this PR, and copied the same folders.

`git diff` says this is the only change:
```diff
index 6fc3d3b..331f9d3 100644
--- b/tests/ZeroC.Slice.Tests/generated/SequenceDecodingTests.cs            // Main
+++ a/tests/ZeroC.Slice.Tests/generated/SequenceDecodingTests.cs            // My PR
@@ -31,7 +31,8 @@ public partial record struct BoolS
     /// <param name="decoder">The Slice decoder.</param>
     public BoolS(ref SliceDecoder decoder)
     {
+        this.Values = decoder.DecodeSequence<bool>(
+            checkElement: SliceDecoder.CheckBoolValue);
-        this.Values = decoder.DecodeSequence<bool>(checkElement: SliceDecoder.CheckBoolValue);
     }

     /// <summary>Encodes the fields of this struct with a Slice encoder.</summary>
```
> Well, technically, a bunch of hashes also changed in various `obj` folders and `.cache` files. But who cares.

This change makes sense. Before, we just hardcoded this line as a special case.
Now that we're using a builder, it will always play it safe and put arguments on their own line.

I'm actually surprised there aren't more changes, since I altered the function builder to not emit newlines for empty function calls.
I guess we never generate empty function calls.

tools/slicec-cs/src/slicec_ext/enum_ext.rs
InsertCreativityHere: If you'd prefer it. It just felt pretty small to me, and I've already started implementing it : v)

externl: 🔧

externl (approved):
LGTM

tools/slicec-cs/src/slicec_ext/enum_ext.rs
InsertCreativityHere: https://github.com/icerpc/icerpc-csharp/issues/3926

pepone (approved):

tools/slicec-cs/src/decoding.rs
InsertCreativityHere: At least it doesn't produce a difference in the generated code.
I was pretty careful to avoid changing behavior.

PR #3931, opened 2024-02-20
Update version to '0.3.1-preview1'.

InsertCreativityHere:
This PR updates the version, and pulls in 2 commits that are on main, which would be nice to have on 0.3.x as well:
fixes to the language extension settings.

This will cause some merge conflicts with my other PR, which I'll deal with.

### DIFF
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -70,5 +70,9 @@
     "rust",
     "softwareTerms",
     "en_US"
+  ],
+  "enableFiletypes": [
+    "ice",
+    "slice"
   ]
 }
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -72,7 +72,6 @@
     "en_US"
   ],
   "enableFiletypes": [
-    "ice",
     "slice"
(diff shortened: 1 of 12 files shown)


.vscode/cspell.json
```diff
@@ -70,5 +70,8 @@
     "rust",
     "softwareTerms",
     "en_US"
+  ],
+  "enableFiletypes": [
+    "slice"
   ]
```
pepone, line 76: This seems unrelated to the PR, is it required?

pepone (approved):

.vscode/cspell.json
InsertCreativityHere: You got here before I edited the comment!

> This PR updates the version, and pulls in 2 commits that are on main, which would be nice to have on 0.3.x as well:
fixes to the language extension settings.

InsertCreativityHere: I was going to rebase this PR to pull them along, but keep the commits seperate.

bernardnormier (approved):

PR #3932, opened 2024-02-20
Update version to '0.4.0-preview1'.

InsertCreativityHere:


tools/slicec-cs/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "slicec-cs"
-version = "0.3.0"
+version = "0.4.0"
```
bernardnormier, line 3: It would be clearer for this version to be 0.4.0-preview1.

bernardnormier (approved):

pepone (approved):
Looks good, we have to upgrade the version in icerpc-ice-csharp-interop repository to match this update.

See https://github.com/icerpc/icerpc-ice-csharp-interop/blob/fda69a315fe7e14659260ac9f0511aec782e33fd/build/Versions.props#L6

tools/slicec-cs/Cargo.toml
InsertCreativityHere: Some time ago, and I looked into this - no one really does pre-release versions in Rust.
These are all actively developed master branches, and they just put the version in there:

Tonic:
https://github.com/hyperium/tonic/blob/18a2b30922460be02829706cf9dd0cd1ec6a19c1/tonic/Cargo.toml#L23

LALRPOP:
https://github.com/lalrpop/lalrpop/blob/0b244afd19f6da8f5c095e8c780a48f80abe21bf/Cargo.toml#L17

Quinn:
https://github.com/quinn-rs/quinn/blob/dcc8048974ce9b1ca6b365019149b5586ed88f4a/quinn/Cargo.toml#L3

RustDesk:
https://github.com/rustdesk/rustdesk/blob/7b317619acdf26853e05c5e398f5bdc19fee776c/Cargo.toml#L3

ChatGPT (bad example, because they're still on version '0.0.0'):
https://github.com/lencx/ChatGPT/blob/1c97623156411de8e00496119376e930261b15e9/src-tauri/Cargo.toml#L3

zkSync:
https://github.com/matter-labs/zksync-era/blob/ce6c120c31b270739eecb102f6742237e6c9d004/core/lib/zksync_core/Cargo.toml#L3

There **is** a standard for doing pre-releases, but it uses `-beta.0` instead of `preview1`.
And even more importantly, in my searching, I only found this one crate using it:
https://github.com/tauri-apps/tauri/blob/3fb414b61ad7cfce67751230826fddfb39effec5/core/tauri/Cargo.toml#L3

InsertCreativityHere:
I opened PRs to update the version, but obviously I'll wait on those until this is merged in.

PR #3952, opened 2024-03-14
Enable CI on 0.3.x branch

pepone:
This PR fixes CI job on 0.3.x branch, to run for push/pull_request on this branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,6 @@
 name: ci
 
-on:
-  push:
-    branches: ["main", "0.3.x"]
-  pull_request:
-    branches: ["main", "0.3.x"]
+on: [push, pull_request]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -2,9 +2,9 @@ name: ci
 
 on:
   push:
-    branches: ["main", "0.2.x"]
+    branches: ["main", "0.3.x"]
```
externl, line 5: We should probably remove main

externl (approved):

externl:
Git just does this https://github.com/git/git/blob/master/.github/workflows/main.yml#L3C1-L3C25

pepone:
The problem with removing the branch, is that is most likely to run twice one from the PR, and another from the push to the forked repository.

PR #3955, opened 2024-03-25
QUIC transport template option added to `IceRpc-Slice-Server` and `IceRpc-Slice-Client` (#3954)

### LINKED ISSUE #3954: Add quic option to templates
We want to add a quic option to the template so that you can easily setup your new application to use QUIC instead of slic+tcp.

ReeceHumphreys:
Closes #3954 

This PR adds a new `--transport` flag to the `IceRpc-Slice-Server` and `IceRpc-Slice-Client` templates. By default this flag is set to `slic` but you can set it to `quic`. When set to `quic` it will modify the output of the template to produce a QUIC version of the template. For ease of review I have also attached the [output](https://github.com/icerpc/icerpc-csharp/files/14748795/tmp.zip) from running:

```shell
dotnet new icerpc-slice-server -o slic-server
dotnet new icerpc-slice-client -o slic-client
dotnet new icerpc-slice-server -o quic-server --transport quic
dotnet new icerpc-slice-client -o quic-client --transport quic  
```

The only part that is kinda lame right now is I could not find a way to get the template to copy in the certs from the top level `certs/` directory. As such, we now have a duplicate of the `server.p12` and `cacert.der` inside the template folder which is kinda lame.  @pepone do you know of any solution for this?


src/IceRpc.Templates/Templates/IceRpc-Slice-Client/IceRpc-Slice-Client.csproj
```diff
@@ -13,6 +13,9 @@
     <PackageReference Include="IceRpc.Slice" Version="0.4.0-preview1" />
     <PackageReference Include="IceRpc.Deadline" Version="0.4.0-preview1" />
     <PackageReference Include="IceRpc.Logger" Version="0.4.0-preview1" />
+    <!--#if(transport=="quic")-->
```
bernardnormier, line 16: This PR should be for the 0.3.x branch, not the main branch.

src/IceRpc.Templates/Templates/IceRpc-Slice-Client/.template.config/template.json
```diff
@@ -1,57 +1,147 @@
...
+    "preferNameDirectory": true,
+    "defaultName": "Client",
+    "symbols": {
+        "skipRestore": {
+            "type": "parameter",
+            "datatype": "bool",
+            "description": "If specified, skips the automatic restore of the project on create.",
+            "defaultValue": "false"
+        },
+        "transport": {
```
bernardnormier, line 27: My preference would be to have a simpler --quic option. Not a transport choice.

src/IceRpc.Templates/Templates/IceRpc-Slice-Client/.template.config/template.json
```diff
@@ -1,57 +1,147 @@
...
+            "type": "parameter",
+            "description": "The transport to use for IceRPC connections.",
+            "datatype": "choice",
+            "choices": [
+                {
+                    "choice": "quic",
+                    "description": "Use QUIC transport"
+                },
+                {
+                    "choice": "slic",
```
bernardnormier, line 37: Assuming we keep the transport choice, we should not expose / talk about Slic here (note the spelling for Slic).

The transport is actually tcp / TCP. See for example https://github.com/icerpc/icerpc-csharp/tree/main/examples/slice/TcpFallback. It's "TCP fallback" not "QUIC fallback". Slic is only of interest to advanced users.

src/IceRpc.Templates/Templates/IceRpc-Slice-Client/.template.config/template.json
```diff
@@ -1,57 +1,147 @@
...
+            "modifiers": [
                 {
-                    "rename":{
-                        "editorconfig":".editorconfig"
+                    "rename": {
+                        "editorconfig": ".editorconfig"
+                    }
+                },
+                {
+                    "condition": "(transport == 'quic')",
```
bernardnormier, line 54: Could we instead use ifdefs in a common source file?

I think this would be more maintainable. With separate source files for TCP (Slic) and QUIC, it's more likely they'll get out of sync.

pepone (commented):
We should update CI to test with the new parameter too

See: https://github.com/icerpc/icerpc-csharp/blob/main/.github/actions/test-templates/action.yaml

Regarding the certificates, seems that you can include them in the NuGet package for the templates:

```
diff --git a/src/IceRpc.Templates/IceRpc.Templates.csproj b/src/IceRpc.Templates/IceRpc.Templates.csproj
index 2a084925..70b40d1b 100644
--- a/src/IceRpc.Templates/IceRpc.Templates.csproj
+++ b/src/IceRpc.Templates/IceRpc.Templates.csproj
@@ -18,6 +18,7 @@
             <Pack>true</Pack>
         </Content>
         <Compile Remove="**/*" />
+        <Content Include="../../certs/*" Pack="true" PackagePath="/certs" />
```

Then in the `template.json` add a new source entry like

```
+        {
+            "source": "../../certs/",
+            "include": [
+                "cacert.der",
+                "client.p12"
+            ],
+            "target": "./certs"
         }
```

src/IceRpc.Templates/Templates/IceRpc-Slice-Client/.template.config/template.json
```diff
@@ -1,57 +1,147 @@
...
+                    "rename": {
+                        "slic/Program.cs": "Program.cs"
                     }
                 }
             ]
         }
     ],
-    "primaryOutputs":[
+    "SpecialCustomOperations": {
+        "**/*.xaml": {
```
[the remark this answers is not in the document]
ReeceHumphreys: This is for the conditional inside of the `.csproj`, I used this Stack overflow for reference https://stackoverflow.com/questions/60056191/add-optional-content-in-dotnet-new-templates-in-non-c-sharp-files!

src/IceRpc.Templates/Templates/IceRpc-Slice-Client/.template.config/template.json
```diff
@@ -1,57 +1,147 @@
...
+                            "#endif",
+                            "<!--#endif"
+                        ],
+                        "trim": "true",
+                        "wholeLine": "true"
+                    }
+                }
+            ]
+        },
+        "editorconfig": {
```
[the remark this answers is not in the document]
ReeceHumphreys: I tried that and I thought I couldn't get it to work for whatever reason. I'll give it another crack though

.gitignore
```diff
@@ -92,3 +92,6 @@ docfx/api
 nuget.config
 NuGet.config
 *.nupkg
+
+# CSharp Dev Kit
+.mono
```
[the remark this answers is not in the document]
ReeceHumphreys: The new C# Dev Kit extension for VSCode keeps creating this folder. I have almost committed it a few different times so I just added this to the `.gitignore`

src/IceRpc.Templates/Templates/IceRpc-Slice-Client/.template.config/template.json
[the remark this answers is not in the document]
pepone: See https://github.com/dotnet/templating/wiki/Conditional-processing-and-comment-syntax#msbuild-files

Supports `.*proj` files, which should work for `.csproj`.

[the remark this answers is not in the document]
ReeceHumphreys: Awesome! So I would just use it like this?

```
<TargetFramework Condition="'$(transport)' == 'quic'">
... 
</TargetFramework>
```

src/IceRpc.Templates/Templates/IceRpc-Slice-Client/.template.config/template.json
ReeceHumphreys: I prefer the verbosity of `--transport quic` it makes it crystal clear exactly what the options are quic or tcp with it defaulting to tcp. Im okay with changing it to `--quic` if everyone thinks that is cleaner though.

ReeceHumphreys:
> Regarding the certificates, seems that you can include them in the NuGet package for the templates:

I tried doing this morning and was running into issues manually installing the templates using `dotnet new install path/to/the/template/I/want` as the cert now exists in the NuGet but not permanently in the folder. I think we should just leave it as is for now and can always refine it a bit more in the future!

@bernardnormier brought up the idea of adding more options in the future like an SSL option for TCP!

pepone:
> I tried doing this morning and was running into issues manually installing the templates using dotnet new install path/to/the/template/I/want

Another option would be to look into add a built step to copy them, instead of keeping the copies in git.

src/IceRpc.Templates/Templates/IceRpc-Slice-Client/IceRpc-Slice-Client.csproj
```diff
@@ -7,12 +7,18 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <!-- Copy the PDBs from the NuGet packages to get file names and line numbers in stack traces. -->
     <CopyDebugSymbolFilesFromPackages>true</CopyDebugSymbolFilesFromPackages>
+    <!--#if(transport=="quic") -->
```
bernardnormier, line 10: Can we use the same spacing as in the C# code?

bernardnormier (approved):

ReeceHumphreys:
> The updates look good, with a few considerations:
> 
> * Have we considered providing the same feature with the Protobuf templates?
> * It wouldn't hurt to include a README file with the templates, clarifying that the provided certificates are for testing purposes.

Definitely! The main goal of this PR was to demonstrate how it could be done and to do it for slice client and server so I can refactor my blog to use these new templates. I'll open issues for these improvements.

PR #3961, opened 2024-03-27
Separate example and assembly version and update Cargo deps

externl:
This PR separates the "version number" used for the examples and the IceRPC assemblies. It also updates the slices cargo dep.

Additionally it moves the testing of the examples to a separate step job independent of the src build.

### DIFF
--- a/build/IceRpc.Examples.props
+++ b/build/IceRpc.Examples.props
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     <PropertyGroup>
-        <!-- The version of the IceRPC NuGet packages used by the examples -->
+        <!-- The version of the IceRPC NuGet packages used by the docfx examples. This version is already published on nuget.org. -->
         <Version Condition="'$(Version)' == ''">0.3.0</Version>
         <WarningLevel>4</WarningLevel>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
--- a/build/IceRpc.Examples.props
+++ b/build/IceRpc.Examples.props
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <Import Project="$(MSBuildThisFileDirectory)IceRpc.Version.props" />
     <PropertyGroup>
+        <!-- The version of the IceRPC NuGet packages used by the examples -->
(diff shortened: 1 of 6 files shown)


pepone:
How would we add examples for new features if examples are built against the released version? How would we know if an update breaks an example?

externl:
> How would we add examples for new features if examples are built against the released version? How would we know if an update breaks an example?

- To me, the most important thing is that these examples use the published NuGets, no building required.
- These 0.X.x branches are about patch fixes, we probably shouldn't adding new features. If we really need to I'm sure we can come up with a solution.
- We could run an additional build of the examples with a published src build by doing `dotnet build /p:Version=...` after a local publish.
-  We really shouldn't be making changes that break examples, and we shouldn't use these examples as a way to enforce "binary compatibility".

pepone:
How would we update the version when we want to release?

Do we have to first release the next version for example, v0.3.1, and after publishing the new packages we upgrade the examples version, or are you thinking of a different workflow?

externl:
> Do we have to first release the next version for example, v0.3.1, and after publishing the new packages we upgrade the examples version

Exactly this.

bernardnormier:
> How would we add examples for new features if examples are built against the released version? 

Easy: we don't add new features (at the very least new features that affect the existing examples) on released branches.

> How would we know if an update breaks an example?

It is desirable to configure the example build to use the latest preview if available (typically available from a local build). I assume it's doable but not the default.

pepone (approved):
Looks good, one concern is conflicts when merging to main. In main we need to build examples with a non released version, maybe using different CI files is simpler as most of the work is handled by separate actions.

externl:
Yes, when we merge to main we need to make sure not to take these files.

externl:
Is main far behind 0.3.x?

build/IceRpc.Examples.props
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <Import Project="$(MSBuildThisFileDirectory)IceRpc.Version.props" />
     <PropertyGroup>
+        <!-- The version of the IceRPC NuGet packages used by the examples -->
```
bernardnormier, line 4: I would add:
This version is already published on nuget.org.

bernardnormier (approved):

pepone:
> Is main far behind 0.3.x?

No but, we want to test with a different version in main, don't we? Or do we  plan to keep examples in main building against latest released version, would be difficult for new eexamples?

externl:
> No but, we want to test with a different version in main, don't we? Or do we plan to keep examples in main building against latest released version, would be difficult for new examples?

For main, I think we should look into publishing NuGet packages to a registry that allows replacing a package. Hopefully we can do this with GitHub packages. When we can up the package nightly on for every commit to main. Then we can build the examples which those packages.

PR #3962, opened 2024-03-27
QUIC transport template option added to `IceRpc-Protobuf-Server` and `IceRpc-Protobuf-Client`

ReeceHumphreys:
Sister PR to #3955 
Closes #3956 

This PR adds the QUIC transport option to the Protobuf templates .

PR #3967, opened 2024-03-27
Add README.md to `certs/` in `IceRpc.Templates`

### LINKED ISSUE #3957: Include README files for the .NET Templates
> It wouldn't hurt to include a README file with the templates, clarifying that the provided certificates are for testing purposes.

_Originally posted by @pepone in https://github.com/icerpc/icerpc-csharp/pull/3955#pullrequestreview-1961299352_

ReeceHumphreys:
Closes #3957 

I checked the `grpc`, `blazor`, and `web` .NET templates and none of them include a README.md at the root of the generated output. However, I do think it is very important to tell users that these certs are not to be used in production. As such, I added a README.md to the `certs` directory inside the templates instead. This way, you only see the README when generating the QUIC version too.

I will update this PR to do the same for Protobuf Templates once #3962 is merged.


src/IceRpc.Templates/Templates/IceRpc-Slice-Client/certs/README.md
```diff
@@ -0,0 +1,4 @@
+# IceRPC SSL test certificates
+
+The QUIC transport version of this template uses the SSL certificate found in this directory. These
```
bernardnormier, line 3: I would not refer to "template" in this README.md.

Users won't see this README when browsing the contents of the NuGet package. They will see it when browsing the file for _their_ project that they happen to create with one of our templates.

ReeceHumphreys: Good point! I will update the wording.

bernardnormier (approved):

pepone (approved):

PR #3968, opened 2024-03-27
Pull in latest 'slicec' and fix 'compile_all_test_slice'.

InsertCreativityHere:
This PR updates the dependencies of `slicec-cs`, including the latest `slicec`.

The only change of relevance in `slicec` was changing how the `SliceFile`s are stored in a `CompilationState`.
This let us perform some slight simplifications here in `slicec-cs`.

----

This PR also fixes the `compile_all_test_slice` test, which has probably been broken for a very long time.
1) We moved where the test files were stored from `IceRpc.Tests` to both `IceRpc.Slice.Tests` and `Zeroc.Slice.Tests`.
So, this test was 'passing' because it found `0` files, and so, obviously `number_of_errors == 0`... which is what the test checked.

2) We were doing something weird/wrong with the slice_options, because instead of creating a new one for each test-run. We were modifying an existing one in place. So instead of only compiling 1 test file, we were compiling `N`. So after fixing issue 1, everything failed.
I fixed the test so everything is much more isolated now. We only compile a single source file at a time.
And we assert that the parsed version of the file has the same path as the one we expect to be testing.

### DIFF
--- a/tools/slicec-cs/src/cs_compile.rs
+++ b/tools/slicec-cs/src/cs_compile.rs
@@ -30,7 +30,7 @@ pub fn cs_validator(compilation_state: &mut CompilationState) {
 fn check_for_unique_names(compilation_state: &mut CompilationState) {
     let mut file_map = std::collections::HashMap::new();
 
-    for slice_file in compilation_state.files.values().filter(|file| file.is_source) {
+    for slice_file in compilation_state.files.iter().filter(|file| file.is_source) {
         if let Some(old_path) = file_map.insert(&slice_file.filename, &slice_file.relative_path) {
             Diagnostic::new(Error::IO {
                 action: "generate code for",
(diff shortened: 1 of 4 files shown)


tools/slicec-cs/src/cs_compile.rs
```diff
@@ -81,30 +81,48 @@ mod test {
     #[test]
     fn compile_all_test_slice() {
```
ReeceHumphreys, line 82: It seems that integrating this check as a test within the Rust code might not be the most appropriate approach. There isn't a specific unit of behavior that it's validating for `slicec-cs`. Rather, it appears better suited as a step within a GitHub action. The idea would be to invoke `slicec-cs` on all slice files, and if anything other than an exit code of 0 is returned, the action should fail. Thoughts, @externl?

externl: This test is primarily about helping to generating code coverage. So while it's not an AAA test it's still (IMO) quite useful.

externl (approved):

tools/slicec-cs/src/cs_compile.rs
InsertCreativityHere: Just for context, it is _technically_ in the wrong place.
Integration tests like these _should_ be in a top level `tests` folder.
But we spent so much time writing unit tests, we just got used to putting the tests directly in the code,
even though by convention, sure, you should only do that for unit tests.

I don't have any vested stake in this code, but I do think that if we decide to do your suggestion, it should be in a separate PR than this one.

PR #3971, opened 2024-03-28
Run `test-templates` action in release workflow

### LINKED ISSUE #3934: Test project templates with release workflow
The release workflow test the packages by building the examples, but this doesn't cover the template packages.

We should test the project templates too with this workflow.

ReeceHumphreys:
Closes #3934

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -116,16 +116,18 @@ jobs:
       with:
         name: packages
         path: packages
-    - name: Install packages to local NuGet repository
+    - name: Install NuGet packages and dotnet new templates
       working-directory: packages
       run: |
         dotnet nuget push *.nupkg --source ~/.nuget/packages
+        dotnet new install IceRpc.Templates*.nupkg
       if: runner.os == 'macOS' || runner.os == 'Linux'
       shell: bash
-    - name: Install packages to local NuGet repository
+    - name: Install NuGet packages and dotnet new templates
       working-directory: packages
       run: |
         dotnet nuget push *.nupkg --source $env:USERPROFILE/.nuget/packages
+        dotnet new install IceRpc.Templates*.nupkg
       if: runner.os == 'Windows'
       shell: powershell
     - name: Set version from tag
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -149,3 +149,6 @@ jobs:
         foreach ($example in $examples) { dotnet build $example.FullName }
       if: runner.os == 'Windows'
       shell: powershell
+    - name: 🧪 Test Templates
+      run: |
+        uses: ./.github/actions/test-templates
```
pepone, line 154: We need to fix test-templates, it currently install the templates from the current build. Here we want to use the package we download from the release artifacts. Maybe the simpler is to move the install to a separate step for each use case.

ReeceHumphreys: Done, I just removed the step from the `test-templates` that installs it and added a separate step in both `ci.yml` and `release.yml` that installs the templates there!

.github/workflows/release.yml
```diff
@@ -149,3 +149,10 @@ jobs:
         foreach ($example in $examples) { dotnet build $example.FullName }
       if: runner.os == 'Windows'
       shell: powershell
+    - name: 🔗 Install Templates
+      run: |
+        dotnet new install $env:USERPROFILE/.nuget/packages/icerpc.templates
+      shell: bash
```
pepone, line 155: This is not quite correct. `$env:USERPROFILE/` is a powershell expression and we are setting shell bash. Unlike `CI` workflow this runs in Windows, macOS and Linux. I think the simpler is to install the step that install NuGet packages to also install templates, something like:

```
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -116,16 +116,18 @@ jobs:
       with:
         name: packages
         path: packages
-    - name: Install packages to local NuGet repository
+    - name: Install NuGet packages and dotnet new templates
       working-directory: packages
       run: |
         dotnet nuget push *.nupkg --source ~/.nuget/packages
+        dotnet new install IceRpc.Templates*.nupkg
       if: runner.os == 'macOS' || runner.os == 'Linux'
       shell: bash
-    - name: Install packages to local NuGet repository
+    - name: Install NuGet packages and dotnet new templates
       working-directory: packages
       run: |
         dotnet nuget push *.nupkg --source $env:USERPROFILE/.nuget/packages
+        dotnet new install IceRpc.Templates*.nupkg
       if: runner.os == 'Windows'
       shell: powershell
```

ReeceHumphreys: Oh okay wasn't sure when I saw it if it was some special action syntax I wasn't aware of. Will fix.

pepone (approved):
Looks good!

PR #3979, opened 2024-03-29
Merge 0.3.x into main

externl:
This PR merges 0.3.x into main. To merge this I need to be careful and create a merge commit. One main diverges further away from 0.3.x we should instead cherry-pick fixes onto 0.3.x instead of merging upward.


build/IceRpc.Examples.props
```diff
@@ -1,9 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <Import Project="$(MSBuildThisFileDirectory)IceRpc.Version.props" />
     <PropertyGroup>
+        <!-- The version of the IceRPC NuGet packages used by the examples. It must be available from nuget.org. -->
```
bernardnormier, line 4: Need to update the comment.

Maybe something like:
It must be available from nuget.org once this branch is released.

bernardnormier (approved):

PR #3980, opened 2024-04-02
Fix clone of authentication options

pepone:
The Clone extensions methods in SslAuthenticationOptionsExtensions are missing some new options.


bernardnormier (approved):
We should fix this on 0.3.x first.

externl (approved):
Do you think this is why the deployed server isn't providing the chain? Seems pretty brittle if we have to update this when they make changes.

pepone:
> Do you think this is why the deployed server isn't providing the chain?

No, I have another test for that (and a fix).

PR #3982, opened 2024-04-06
Update examples and project template to set ServerCertificateContext

### LINKED ISSUE #3981: Upgrade examples to use ServerCertificateContext
According to https://github.com/dotnet/runtime/issues/100530#issuecomment-2034141821 the recommended way to pass the certificates is to use the server certificate context. We should upgrade the examples to do this.

pepone:
Fix #3981


externl (approved):

pepone:
Unfortunately, this doesn't work on Windows, the key imported by CreateFromPenFile is not usable by SCHannel

see https://github.com/dotnet/runtime/issues/86328

One solution to work on all platforms would be:

```
// The X509 certificate used by the server.
using var serverCertificate = new X509Certificate2(serverCert);

// Create a collection with the server certificate and any intermediate certificates. This is used by
// ServerCertificateContext to provide the certificate chain to the peer.
var intermediates = new X509Certificate2Collection();
intermediates.ImportFromPemFile(serverFullChain);

// Create the authentication options using the test server certificate.
var serverAuthenticationOptions = new SslServerAuthenticationOptions()
{
    ServerCertificateContext = SslStreamCertificateContext.Create(serverCertificate, intermediates)
};
```

Do you think this is worth the added complexity? Using `ServerCertificateContext` ensures that the server provides the certificate chain to the client not just the leaf certificate. In this example with the test certificates, it doesn't matter because there are no intermediate certificates, and the peer root CA is never sent, and must not be used.

externl:
I think we probably should just because people will likely end up copying our code.

bernardnormier:
First question: why is this PR adding .pem files?

Are these .pem files in addition to the existing .p12 and .der files? Are the p12 and der files still used?

pepone:
> First question: why is this PR adding .pem files?

We use PEM files to load the certificate chain, used to set the `ServerCertificateContext`.

> Are these .pem files in addition to the existing .p12 and .der files? Are the p12 and der files still used?

On Windows, we can either load the server certificate from the `.p12` or load from a PEM and export as mentioned in https://github.com/dotnet/runtime/issues/86328

If we export the files as in the mentioned bug report, we can remove  `.p12` files, otherwise, we need them.

bernardnormier (approved):
If we use different files (.pem vs .p12) for the examples and the tests, it would be clearer to have separate certs directories for each.

PR #3991, opened 2024-05-22
Add new KeepAliveInterval option in QuicClientTransportOptions

### LINKED ISSUE #3353: Missing Quic keep alive option to enable keep alive
There's no option to enable keep alive on a Quic connection. As a consequence the Quic connection might idle timeout if there's no activity and even if an icerpc request is pending. This is quite bad IMO.

Should we consider adding a workaround to keep alive the Quic connection? One option would be to open an unidirectional stream to send a byte regularly (implemented in the Quic transport).

See also https://github.com/dotnet/runtime/issues/87478. Hopefully this will motivate the Quic developers for adding this option.

bernardnormier:
This PR adds the new KeepAliveInterval option to QuicClientTransportOptions. It's implemented only with .NET 9.0.

Fixes #3353.

### DIFF
--- a/src/IceRpc.Transports.Quic/QuicTransportOptions.cs
+++ b/src/IceRpc.Transports.Quic/QuicTransportOptions.cs
@@ -9,9 +9,11 @@ public record class QuicTransportOptions
 {
     /// <summary>Gets or sets the idle timeout. This timeout is used to monitor the transport connection health. If no
     /// data is received within the idle timeout period, the transport connection is aborted.</summary>
-    /// <value>The idle timeout. Defaults to <see cref="TimeSpan.Zero" /> meaning that the underlying QUIC
-    /// implementation default will be used.</value>
-    public TimeSpan IdleTimeout { get; set; } = TimeSpan.Zero;
+    /// <value>The idle timeout. Defaults to 30 seconds. <see cref="TimeSpan.Zero" /> means "use the default value
+    /// provided by the underlying implementation".</value>
+    /// <remarks>The idle timeout is negotiated by QUIC during connection establishment. The actual idle timeout value
+    /// for a connection can be lower than the value you've set here.</remarks>
+    public TimeSpan IdleTimeout { get; set; } = TimeSpan.FromSeconds(30);
 }
 
 /// <summary>The options class for configuring <see cref="QuicClientTransport"/>.</summary>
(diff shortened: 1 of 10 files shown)


src/IceRpc.Transports.Quic/QuicTransportOptions.cs
```diff
@@ -9,9 +9,11 @@ public record class QuicTransportOptions
...
     /// <summary>Gets or sets the idle timeout. This timeout is used to monitor the transport connection health. If no
     /// data is received within the idle timeout period, the transport connection is aborted.</summary>
-    /// <value>The idle timeout. Defaults to <see cref="TimeSpan.Zero" /> meaning that the underlying QUIC
-    /// implementation default will be used.</value>
-    public TimeSpan IdleTimeout { get; set; } = TimeSpan.Zero;
+    /// <value>The idle timeout. Defaults to 30 seconds. <see cref="TimeSpan.Zero" /> means "use the default value
+    /// provided by the underlying implementation".</value>
+    /// <remarks>The idle timeout is negotiated by QUIC during connection establishment. The actual idle timeout value
+    /// for a connection can be lower than the value you've set here.</remarks>
+    public TimeSpan IdleTimeout { get; set; } = TimeSpan.FromSeconds(30);
```
bernardnormier, line 16: Changed the default from Zero to 30 seconds - which is the msquic default (= what Zero meant). It's clearer to spell out the actual default, and it's highly desirable to use the same default in all IceRPC implementation.

tests/IceRpc.Compressor.Tests/CompressorInterceptorTests.cs
```diff
@@ -43,7 +43,7 @@ public class CompressorInterceptorTests
             new BrotliStream(outStream, CompressionMode.Decompress) :
             new DeflateStream(outStream, CompressionMode.Decompress);
         var decompressedPayload = new byte[4096];
-        await decompressedStream.ReadAsync(decompressedPayload);
+        await decompressedStream.ReadAtLeastAsync(decompressedPayload, _payload.Length);
```
bernardnormier, line 46: These are tiny unrelated fixes for analyzer warnings introduced in .NET 9.0.

tests/IceRpc.Quic.Tests/Transports/QuicIdleTimeoutTests.cs
```diff
@@ -60,14 +60,56 @@ public async Task Quic_connection_idle_after_idle_timeout([Values]bool configure
...
+        var data = new byte[] { 0x1, 0x2, 0x3 };
+        await sut.Local.Output.WriteAsync(data);
+        ReadResult readResult = await sut.Remote.Input.ReadAsync();
+
+        // Act/Assert
+
+        // Without the keep-alive interval, the idle timer aborts the connection after 500ms.
+        using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(2));
+
+#if NET9_0_OR_GREATER
```
bernardnormier, line 103: This assumes the unit tests are built with the same version of .NET as the IceRpc.Transport.Quic assembly (a safe assumption).

externl (approved):

tests/IceRpc.Quic.Tests/Transports/QuicIdleTimeoutTests.cs
```diff
@@ -60,14 +60,56 @@ public async Task Quic_connection_idle_after_idle_timeout([Values]bool configure
...
+
+        // Simulate a request
+        var data = new byte[] { 0x1, 0x2, 0x3 };
+        await sut.Local.Output.WriteAsync(data);
+        ReadResult readResult = await sut.Remote.Input.ReadAsync();
+
+        // Act/Assert
+
+        // Without the keep-alive interval, the idle timer aborts the connection after 500ms.
+        using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(2));
```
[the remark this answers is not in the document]
bernardnormier: You're right, I reworked the test and removed the cancellation token source.

Note that with QUIC, canceling a read is fatal to the stream. You can't cancel reading and then start reading again. We may be doing the same in Slic.

pepone (approved):

PR #3995, opened 2024-05-22
Additional fixes for .NET 9 builds

pepone:
This PR includes additional fixes for .NET 9 builds

### DIFF
--- a/build/IceRpc.Version.props
+++ b/build/IceRpc.Version.props
@@ -5,8 +5,9 @@
         <!-- The Protobuf version used by IceRpc.Protobuf -->
         <ProtobufVersion>26.1</ProtobufVersion>
         <NuGetProtobufVersion>3.$(ProtobufVersion)</NuGetProtobufVersion>
-        <!-- The target framework used by assemblies -->
-        <IceRpcTargetFramework>net8.0</IceRpcTargetFramework>
-        <IceRpcNetRuntimeVersion>8.0.*</IceRpcNetRuntimeVersion>
+        <!-- The target framework used by assemblies. We target .NET 8 by default, for building with .NET 9 preview update
+             the IceRpcTargetFramework to `net9.0` and IceRpcNetRuntimeVersion to `9.0.*-preview.*`  -->
+        <IceRpcTargetFramework>net9.0</IceRpcTargetFramework>
+        <IceRpcNetRuntimeVersion>9.0.*-preview.*</IceRpcNetRuntimeVersion>
     </PropertyGroup>
 </Project>
--- a/build/IceRpc.Version.props
+++ b/build/IceRpc.Version.props
@@ -5,9 +5,12 @@
         <!-- The Protobuf version used by IceRpc.Protobuf -->
         <ProtobufVersion>26.1</ProtobufVersion>
         <NuGetProtobufVersion>3.$(ProtobufVersion)</NuGetProtobufVersion>
-        <!-- The target framework used by assemblies. We target .NET 8 by default, for building with .NET 9 preview update
-             the IceRpcTargetFramework to `net9.0` and IceRpcNetRuntimeVersion to `9.0.*-preview.*`  -->
-        <IceRpcTargetFramework>net9.0</IceRpcTargetFramework>
-        <IceRpcNetRuntimeVersion>9.0.*-preview.*</IceRpcNetRuntimeVersion>
+        <!--
+            The target framework used by assemblies.
+            We target .NET 8 by default. To build with .NET 9 preview, update the
+            IceRpcTargetFramework to `net9.0` and IceRpcNetRuntimeVersion to `9.0.*-preview.*`.
+        -->
+        <IceRpcTargetFramework>net8.0</IceRpcTargetFramework>
+        <IceRpcNetRuntimeVersion>8.0.*</IceRpcNetRuntimeVersion>
     </PropertyGroup>
 </Project>
--- a/build/IceRpc.Version.props
+++ b/build/IceRpc.Version.props
@@ -5,7 +5,8 @@
         <!-- The Protobuf version used by IceRpc.Protobuf -->
         <ProtobufVersion>26.1</ProtobufVersion>
         <NuGetProtobufVersion>3.$(ProtobufVersion)</NuGetProtobufVersion>
-        <TargetFramework>net8.0</TargetFramework>
-        <NetRuntimeVersion>8.0.*</NetRuntimeVersion>
+        <!-- The target framework used by assemblies -->
+        <IceRpcTargetFramework>net8.0</IceRpcTargetFramework>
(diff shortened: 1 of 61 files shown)


build/IceRpc.Version.props
```diff
@@ -5,7 +5,8 @@
         <!-- The Protobuf version used by IceRpc.Protobuf -->
         <ProtobufVersion>26.1</ProtobufVersion>
         <NuGetProtobufVersion>3.$(ProtobufVersion)</NuGetProtobufVersion>
-        <TargetFramework>net8.0</TargetFramework>
-        <NetRuntimeVersion>8.0.*</NetRuntimeVersion>
+        <!-- The target framework used by assemblies -->
+        <IceRpcTargetFramework>net8.0</IceRpcTargetFramework>
```
pepone, line 9: Rather than setting the TargetFramework directly here, we use `IceRpcTargetFramework` variable, this allows importing the files into other projects without changing the TargetFramework.

build/IceRpc.Version.props
```diff
@@ -5,7 +5,8 @@
...
+        <IceRpcNetRuntimeVersion>8.0.*</IceRpcNetRuntimeVersion>
```
pepone, line 10: I rename this variable for consistency

src/Directory.Build.props
```diff
@@ -2,6 +2,10 @@
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     <Import Project="$(MSBuildThisFileDirectory)../build/IceRpc.Src.props" />
     <PropertyGroup>
+        <!-- The target framework to build assemblies in src directory, the default value $(IceRpcTargetFramework) is
+             set in ./build/IceRpc.Version.props
+             -->
+        <TargetFramework>$(IceRpcTargetFramework)</TargetFramework>
```
pepone, line 8: We build all projects under `src` with this target framework.

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
```diff
@@ -64,7 +64,9 @@
       </None>
       <None Include="$(IntermediateOutputPath)/$(AssemblyName).dll" Pack="true" PackagePath="tasks/" Visible="false" />
       <!-- Use the compilers from this source build. -->
-      <None Include="../IceRpc.ProtocGen/bin/$(Configuration)/net8.0/*" Exclude="../IceRpc.ProtocGen/bin/$(Configuration)/net8.0/*.exe">
+      <None
```
pepone, line 67: This was an issue before because `net8.0` was hardcoded here. We cannot use `TargetFramework` because for this project `TargetFramework` must remain set to `netstandard2.0`.

src/Directory.Build.props
bernardnormier: Except for the template projects that set `<TargetFramework>net8.0</TargetFramework>` and therefore overwrite this setting?

It's not clear to me how we handle the templates.

build/IceRpc.Version.props
bernardnormier: What should we set it to when using net9.0? I tried 9.0.* and it didn't work; 8.0.* seems to work but it's odd.

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.props
```diff
@@ -16,7 +16,7 @@
                  the source build. -->
             <PropertyGroup>
                 <IceRpcProtocPath>$(MSBuildThisFileDirectory)obj/tools/</IceRpcProtocPath>
-                <IceRpcProtocGenPath>$(MSBuildThisFileDirectory)../IceRpc.ProtocGen/bin/$(Configuration)/net8.0/</IceRpcProtocGenPath>
+                <IceRpcProtocGenPath>$(MSBuildThisFileDirectory)../IceRpc.ProtocGen/bin/$(Configuration)/$(IceRpcTargetFramework)/</IceRpcProtocGenPath>
```
bernardnormier, line 19: It's not clear to me where IceRpcTargetFramework comes from. I don't see any import or Directory.Build.props with imports.

We should not count on the user of this props file to somehow import IceRpc.Version.props first.

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
bernardnormier: Here, we import IceRpc.Version.props at the top of the file, so it's clear where it comes from.

src/Directory.Build.props
pepone: We don't handle templates yes, as templates are just packaged but not built. We need to add a template parameter if we want to support multiple frameworks with the templates.

build/IceRpc.Version.props
pepone: The problem is that the expression `9.0.*` doesn't include preview versions. It should be `9.0.*-preview*`. I will add a comment on how to set this.

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.props
pepone: We don't, there are two branches here, and we count on `IceRpcTargetFramework` to be defined with the source builds.

For example, with a source build this condition is true:
```
<When Condition="Exists('$(MSBuildThisFileDirectory)../IceRpc.ProtocGen')">
````

For builds that use the NuGet package, everything is in the package's tools directory and there are no framework directories.

src/Directory.Build.props
pepone: Added a Framework parameter that allows you to choose net8.0|net9.0 target frameworks.

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.props
pepone: Added comments to clarify this section of the props file

build/IceRpc.Version.props
pepone: Added comment wit example usage for .NET 9

externl:
What is the standard way to configure a .NET project and its build targets? Can anyone point to another .NET project that does this sort of thing that we're doing (Ie. using variables for specifying the TargetFramework? I took a look and didn't find anything. 

It's not like we're going to be shipping two DLLS, one targeting .NET 8 and one for .NET 9. My understand is that when .NET 9 is released we can do a multi-target build, I'm fine with that. 

I really don't think we should have a setting for this for this for the examples, for a preview version of .NET.

No one else is doing this, presumably we should be able to do this same as other popular projects and not make things so complicated.

PR #4000, opened 2024-05-24
Upgrade dev container image for .NET 9 development

pepone:
This PR updates the devcontainer in main for .NET 9 developing


.devcontainer/Dockerfile
```diff
@@ -1,18 +1,20 @@
...
+    && apt-get install -y libmsquic git libicu72 build-essential \
+    && curl https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh -L \
+    && chmod +x dotnet-install.sh \
+    && ./dotnet-install.sh -v latest -c 8.0 \
+    && ./dotnet-install.sh -v latest -q preview -c 9.0 \
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-get clean \
+    && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y \
+    && echo 'export DOTNET_ROOT=$HOME/.dotnet' >> ~/.bashrc \
+    && echo 'export PATH=$PATH:$DOTNET_ROOT:$DOTNET_ROOT/tools:$HOME/.cargo/bin' >> ~/.bashrc
```
externl, line 20: I would make this a separate step. Was there an issue using ENV?

.devcontainer/Dockerfile
```diff
@@ -1,18 +1,20 @@
...
-    && sudo apt-get clean \
-    && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-
-# Set environment variables
-ENV PATH=/home/vscode/.cargo/bin:$PATH
+    && apt-get install -y libmsquic git libicu72 build-essential \
+    && curl https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh -L \
+    && chmod +x dotnet-install.sh \
+    && ./dotnet-install.sh -v latest -c 8.0 \
+    && ./dotnet-install.sh -v latest -q preview -c 9.0 \
```
externl, line 15: I would make downloading the install script and running it a separate RUN step.

externl (commented):
We should use separate steps.

.devcontainer/Dockerfile
pepone: Reworked to use ENV

.devcontainer/Dockerfile
pepone: Split the RUN command into several steps

.devcontainer/Dockerfile
```diff
@@ -1,18 +1,33 @@
...
-    && wget -q https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb \
-    && sudo dpkg -i packages-microsoft-prod.deb \
     && rm packages-microsoft-prod.deb \
-    && sudo apt update \
-    && sudo apt-get install -y libmsquic \
-    && sudo rm -rf /var/lib/apt/lists/* \
-    && sudo apt-get clean \
-    && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
```
externl, line 29: It's best practice to do this any time you apt install something. It's not too important here since we don't ship this container so size doesn't really matter.

ReeceHumphreys (approved):

.devcontainer/Dockerfile
pepone: Right the files would remain in the base layer. Maybe is just simpler to not call `apt-get clean` as this is a dev image and this extra size isn't that important.

externl: Sure, I'm fine to remove it for this dev container.

PR #4006, opened 2024-06-05
Telemetry added to `IceRpc.Slice.Tools` and `IceRpc.Protobuf.Tools`

ReeceHumphreys:
This PR introduces a simple telemetry client to the `IceRpc.Slice.Tools` and `IceRpc.Protobuf.Tools` build pipeline. This initial proof of concept will require additional cleanup and refinement in subsequent PRs.

### Notes
- To fix the build CI we need to do a release of `slicec` as I had to temporarily point the `Cargo.toml` for `slicec-cs` to GitHub to access the new hashing methods for slice files. The CI seems to not like this way of defining dependencies despite it being fine locally.
- We should format the `.props` and `.targets` files. It seems that some are formatted and others are not which caused my local formatter to make changes when saving the files I edited.

### Future Planned PRs

- **Telemetry Task Trigger:** The method used to trigger the `TelemetryTask` could be improved. Assistance from someone more familiar with the C# build system would be beneficial such as @pepone .
- **Telemetry Client Integration:** Eventually, the telemetry client might be moved into a protoc plugin (and a slicec plugin once support is added).
- **Telemetry Client and Slice Definition:** Both the telemetry client and the corresponding slice definition need further refinement.
- **Server Deployment:** After deploying the telemetry server, we can remove some of the certificate configuration from the client and replace `localhost` with the server's address.
- **NuGet Bundling:** Assistance from @pepone is needed to ensure the NuGet bundling works correctly. While a local build with `./build.sh --publish` confirms that the telemetry client is triggered, further validation is required to ensure everything functions during a production deployment build.

---

**Edit**: Opening this as a draft while I add a README.md, do my own final checks, and open the telemetry server PR.
**Edit 2**: The CI failing will be resolved once we update slicec-cs to not use the slicec version on GitHub. No idea why the CI does not like it so much.


tools/Common/IceRpc.Telemetry.Internal/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+    RemoteCertificateValidationCallback = (sender, certificate, chain, errors) =>
+    {
+        using var customChain = new X509Chain();
+        customChain.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
+        customChain.ChainPolicy.DisableCertificateDownloads = true;
+        customChain.ChainPolicy.TrustMode = X509ChainTrustMode.CustomRootTrust;
+        customChain.ChainPolicy.CustomTrustStore.Add(rootCA);
+        return customChain.Build((X509Certificate2)certificate!);
+    }
+};
```
ReeceHumphreys, line 26: Once we have the telemetry server deployed, we can just use the default `SslClientAuthenticationOptions` and will not need to add a test cert to the Trust store.

tools/Common/IceRpc.Telemetry.Internal/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using IceRpc;
+using IceRpc.Retry;
+using IceRpc.Telemetry.Internal;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+
+const int timeout = 3000; // The timeout for the RPC call in milliseconds.
+const int maxAttempts = 3; // The maximum number of attempts to retry the RPC call.
+const string uri = "icerpc://localhost"; // The URI of the server.
```
ReeceHumphreys, line 11: Tried to just pick reasonable numbers. Fine with changing these values.

tools/Common/IceRpc.Telemetry.Internal/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+    // Upload the telemetry to the server.
+    await reporter.UploadAsync(new Telemetry(args));
+
+    // Shutdown the connection.
+    await connection.ShutdownAsync();
+}
+catch (Exception)
+{
+    // Consume all exceptions.
+}
```
ReeceHumphreys, line 51: Even with the `protected override void LogEventsFromTextOutput` in the `TelemetryTask`, any exceptions will show up as warnings in the build console. As such, I just consume any exceptions. Is there a way to 'disable exceptions' via configuration like how you can enable or disable nullable?

tools/Common/IceRpc.Telemetry.Internal/README.md
```diff
@@ -0,0 +1,40 @@
...
+optional, and you may opt-out if you do not like to share any information.
+
+This package includes the `IceRpc.Telemetry.Internal` client. This client is a
+IceRPC C# client that sends anonymous telemetry data over a secure connection
+to the IceRPC telemetry server during the compilation of Slice and Protobuf
+files. This data includes:
+
+- The version of IceRPC that is being used.
+- The operating system of the host machine.
+- The processor count of the host machine.
```
ReeceHumphreys, line 14: Not sure if 'host machine' is the correct term.

tools/Common/IceRpc.Telemetry.Internal/slice/Reporter.slice
```diff
@@ -0,0 +1,33 @@
...
+
+/// Represents a simple telemetry reporter that can upload telemetry data to a
+/// remote server.
+interface Reporter {
+    /// Uploads telemetry data.
+    /// @param telemetry: The telemetry data to upload.
+    [oneway] upload(telemetry: Telemetry)
+}
+
+struct Telemetry {
```
ReeceHumphreys, line 12: I think these names and the `Telemetry` struct could be improved in future PRs.

tools/IceRpc.Protobuf.Tools/ProtocTask.cs
```diff
@@ -132,3 +140,17 @@ protected override void LogEventsFromTextOutput(string singleLine, MessageImport
...
+    public static string ToHexString(byte[] bytes)
+    {
+        var sb = new StringBuilder(bytes.Length * 2);
+        foreach (byte b in bytes)
+        {
+            sb.Append(b.ToString("x2"));
+        }
+        return sb.ToString();
+    }
+}
```
ReeceHumphreys, line 156: Would ideally use `Convert.ToHexString` but this is not available in `ToolTasks` due to being net standard 2.0

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -37,9 +37,21 @@ public class SliceCCSharpTask : ToolTask
     [Required]
     public string WorkingDirectory { get; set; } = "";
 
+    // <summary>If verbose output should be enabled.</summary>
+    [Required]
+    public bool Verbose { get; set; } = false;
```
ReeceHumphreys, line 42: Could not think of a better name for this additional output.

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
```diff
@@ -35,9 +35,18 @@
...
       <PackageReference Include="Google.Protobuf.Tools" Version="$(NuGetProtobufVersion)" />
+      <PackageReference Include="System.Text.Json" Version="8.0.3" />
    </ItemGroup>
    <ItemGroup>
       <Compile Include="../Common/IceRpc.CaseConverter.Internal/Converter.cs" />
+      <None Include="../Common/IceRpc.Telemetry.Internal/bin/$(Configuration)/net8.0/*" Exclude="../IceRpc.Telemetry.Internal/bin/$(Configuration)/net8.0/*.exe">
+        <PackagePath>tools/</PackagePath>
+        <Pack>true</Pack>
+      </None>
+      <None Include="../Common/IceRpc.Telemetry.Internal/certs/">
```
pepone, line 46: This is not correct, "Include" must refer to an actual file. This prevents loading the project in Visual Studio.

ReeceHumphreys: Is there a better temporary solution? The certs are not going to be needed once the telemetry server is actually deployed. Purely exist for development and testing currently.

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -145,4 +158,48 @@ protected override void LogEventsFromTextOutput(string singleLine, MessageImport
...
+        using (var process = Process.Start(startInfo))
+        {
+            if (process == null)
+            {
+                Log.LogError("Failed to start the Slice compiler process.");
+                return -1;
+            }
+
+            string output = process.StandardOutput.ReadToEnd();
+            process.WaitForExit();
```
bernardnormier, line 184: I find this synchronous waiting pretty odd.

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
bernardnormier: The best solution is to not include this property at all. When you run slicec-cs via this Task, you always get "verbose" output.

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -145,4 +158,48 @@ protected override void LogEventsFromTextOutput(string singleLine, MessageImport
...
+
+            if (Verbose == true)
```
bernardnormier, line 186: In C# and most other language, this is written as simply:

```csharp
if (Verbose)
```

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -145,4 +158,48 @@ protected override void LogEventsFromTextOutput(string singleLine, MessageImport
...
+            {
+                try
+                {
+                    var jsonDoc = System.Text.Json.JsonDocument.Parse(output);
+                    OutputHash = jsonDoc.RootElement.GetProperty("hash").GetString();
+                    UpdatedFiles = jsonDoc.RootElement.GetProperty("updated_files").GetBoolean();
+                }
+                catch (Exception)
+                {
+                    // Don't fail the build if we can't parse the output
```
bernardnormier, line 196: Why not?

Something is very wrong if slicec-cs emits unexpected output to stdout.

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -145,4 +158,48 @@ protected override void LogEventsFromTextOutput(string singleLine, MessageImport
...
+            }
+
+            string output = process.StandardOutput.ReadToEnd();
+            process.WaitForExit();
+
+            if (Verbose == true)
+            {
+                try
+                {
+                    var jsonDoc = System.Text.Json.JsonDocument.Parse(output);
```
bernardnormier, line 190: It's odd that you parse the output even when the process exited with an error.

bernardnormier (commented):
I would suggest to first open a PR focused on the changes the Slice compiler and SliceCCsharp task.

tools/Common/IceRpc.Telemetry.Internal/README.md
```diff
@@ -0,0 +1,40 @@
...
+- The amount of memory that the client has.
```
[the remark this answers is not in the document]
ReeceHumphreys: I think it doesn't hurt us to collect. Could give us insights when combined with the other telemetry as to the type of development environments our users are using.

I would also be fine with not collecting it.

ReeceHumphreys:
Going to spin off the `slicec-cs` + the relevant `IceRpc.Slice.Tools` part for the verbose output into a separate PR!

tools/Common/IceRpc.Telemetry.Internal/README.md
```diff
@@ -0,0 +1,40 @@
+# Telemetry
+
+`IceRpc.Protobuf.Tools` and `IceRpc.Slice.Tools` collect anonymous
+telemetry data about general usage. Participation in this anonymous program is
+optional, and you may opt-out if you do not like to share any information.
+
+This package includes the `IceRpc.Telemetry.Internal` client. This client is a
```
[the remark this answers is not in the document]
ReeceHumphreys: Do you have any suggestions for the naming of the namespace and assembly? Could not think of anything good.

tools/Common/IceRpc.Telemetry.Internal/README.md
```diff
@@ -0,0 +1,40 @@
...
+- The operating system of the host machine.
+- The processor count of the host machine.
+- The amount of memory that the client has.
+- The source of the telemetry data (either `IceRpc.Slice.Tools or
+ `IceRpc.Protobuf.Tools`).
+- The hashed slice files or protobuf files. This is used to determine if
+the compiled files are known to the IceRPC telemetry server, such as the
+example slice files or protobuf files provided in the icerpc-csharp
+repository.
+- If the build requires recompiling any Slice or Protobuf files,
```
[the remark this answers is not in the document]
ReeceHumphreys: Not sure which part you are referencing

> If the build requires recompiling any Slice or Protobuf files

If it's this line then we can actually figure out if we need to generate code (i.e if any of the files were updated) for both slice and protobuf. However, after discussing with @pepone I think we dont need to capture this information since we can infer it via if the hashes changed.

pepone (commented):
I would structure this a bit differently:

```
tools
    Common
        IceRpc.Telemetry
            TelemetryTask.cs - The telemetry task to invoke the telemetry reporter
    IceRpc.Protobuf.Tools
    IceRpc.Slice.Tools
    IceRpc.Telemetry.Reporter - IceRPC program to upload the telemetry data
```

in `IceRpc.Protobuf.Tools.csproj` and `IceRpc.Slice.Tools.csproj` include `TelemetryTask.cs`

```
<Compile Include="../Common/IceRpc.Telemetry/TelemetryTask.cs" />
```

We can also use the same telemetry task to compute the hash and call the reporter in both cases Slice/Protobuf

in `IceRpc.Protobuf.Tools.targets`
```
<Target Name="ReportTelemetry" AfterTargets="ProtoCompile" Condition="'$(TelemetryEnabled)' == 'yes'">
    <TelemetryTask
        WorkingDirectory="$(MSBuildProjectDirectory)"
        Sources="@(ProtoFile)"/>
</Target>
```

in `IceRpc.Slice.Tools.targets`
```
<Target Name="ReportTelemetry" AfterTargets="SliceC" Condition="'$(TelemetryEnabled)' == 'yes'">
    <TelemetryTask
        WorkingDirectory="$(MSBuildProjectDirectory)"
        Sources="@(SliceFile)"/>
</Target>
```

For the version I think we should just get it from the task assembly. The IceRPC version is the same for the tasks and for the compilers as we distribute them as part of the same package.

We also don't need to update the compiler to compute these hash, this can be done in the task for both Slice and Protobuf files.

InsertCreativityHere (commented):
I think it might be worthwhile to open a blog discussion about what we want to collect,
because I'm not convinced that this is the right set of information to be collecting.

1) `IceRpcVersion` and OperatingSystem is fine.
But, just out of curiosity, how specific is the operating system?
Do you have an example output you could paste? Not important, I'm just curious.

2) `ProcessorCount` and `Memory` seem a little weird to me.
I'm not strictly against it, but what reason do you see for collecting it?
Like, what future decisions would having this information be relevant to?

3) I see that `source` is related to `protoc` and `slicec`, but what is this actually hold?
Again, do you have a concrete example? Also, I think the name needs to be something way less vague than `source`.

4) Again, I'm not strictly against it, but what's the point of collecting whether the user "updated their files"? What decisions will this information help inform? Personally, I don't see it. Also... why is it nullable? At a statistical level, what what's the difference in information content between seeing `false` vs `null`? What does that tell us?

5) `slicec-cs` also outputs an `exit_code` field to the telemetry... but I don't see anywhere we're capturing it. Again, I'm curious what the reason for collecting that information is... But if we want to keep it in `slicec-cs`, we should at least capture it in the telemetry interceptor thing.

InsertCreativityHere:
Whoops, I only just saw your comment about splitting up the PRs.

tools/slicec-cs/Cargo.toml
```diff
@@ -14,7 +14,9 @@ rust-version = "1.70"
 clap = { version = "~4.3.20", features = ["derive"] }
 convert_case = "0.6.0"
 in_definite = "1.0.0"
-slicec = "0.3.1"
+serde_json = "1.0.117"
+sha2 = "0.10.8"
+slicec = { git = "ssh://git@github.com/icerpc/slicec.git", branch = "main" }
```
InsertCreativityHere, line 19: @ReeceHumphreys
To fix CI, try changing this to:
```suggestion
slicec = { git = "https://github.com/icerpc/slicec.git", branch = "main" }
```
The HTTPS doesn't require any kind of authenteication, whereas using `ssh` does require an SSH key to be configured with a GitHub account (which maybe we don't have configured with CI).

Totally possible there's something else wrong here, but that's a good first thing to check!

tools/Common/IceRpc.Telemetry.Internal/Program.cs
bernardnormier: No, and you can't compare nullable with exceptions.

tools/Common/IceRpc.Telemetry.Internal/README.md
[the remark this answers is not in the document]
bernardnormier: I'd rather not collect any data for no reason. 

Naturally, the struct is extensible so we can always add more information if/when we think collecting it is meaningful.

tools/Common/IceRpc.Telemetry.Internal/README.md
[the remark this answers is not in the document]
bernardnormier: What about namespace/assembly `IceRpc.BuildTelemetry`?

And I wouldn't put it in Common but directly in tools/src next to the other assemblies. Common is for common source files referenced by other projects.

InsertCreativityHere:
After thinking for a little bit, I have a potentially radical proposal:
**I think we should make every field in the Slice `Telemetry` struct tagged.**

In the future, we'll probably add extra fields, and they'll _have_ to be tagged anyways for backwards compatibility.
And who knows, maybe some of the information which we think is useful to collect now, may end up not being useful later, and then it'd be nice to remove it from the struct. Tagging everything gives us maximum flexibility, and I think that's useful for this Slice definition.

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -145,4 +158,48 @@ protected override void LogEventsFromTextOutput(string singleLine, MessageImport
...
+        {
+            FileName = pathToTool,
+            Arguments = commandLineCommands,
+            RedirectStandardOutput = true,
+            UseShellExecute = false,
+            CreateNoWindow = true,
+            WorkingDirectory = GetWorkingDirectory()
+        };
+
+        using (var process = Process.Start(startInfo))
```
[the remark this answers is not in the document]
pepone: Yes, using declarations were introduced in C# 8.0 https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-version-history#c-version-80.

The tasks assemblies use C# 10.0

ReeceHumphreys:
XML formatting in this PR https://github.com/icerpc/icerpc-csharp/pull/4007

ReeceHumphreys:
`slicec-cs` metrics output PR https://github.com/icerpc/icerpc-csharp/pull/4008

tools/slicec-cs/src/main.rs
```diff
@@ -57,38 +59,59 @@ pub fn main() {
         }
     }
 
-    std::process::exit(i32::from(compilation_state.emit_diagnostics(slice_options)))
+    // Emit diagnostics and totals.
+    let exit_code = i32::from(compilation_state.emit_diagnostics(slice_options));
+
+    // If we're not in dry-run mode and verbose mode is enabled, output verbose compilation
+    // information.
+    if cs_options.verbose && !slice_options.dry_run {
```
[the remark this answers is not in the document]
ReeceHumphreys: It felt unnecessary but I see no reason not too!

README.md
```diff
@@ -199,6 +199,12 @@ Ice server, and you can call services hosted by an IceRPC server from an Ice cli
 
 [IceRPC for Ice users] provides all the details.
 
+## Telemetry
+
+IceRPC collects completely anonymous telemetry data during the build process. Participation in this anonymous program is
```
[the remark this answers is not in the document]
ReeceHumphreys: My plan is to hash the ip addresses or something like that. We don't actually care what your actual IP is at the end of the day. We just want a way to differentiate users and to be able to tell if the traffic is coming from us!

tools/Common/IceRpc.Telemetry.Internal/Program.cs
bernardnormier: How do retry work here? The only operation that we call is oneway but non-idempotent.

tools/Common/IceRpc.Telemetry.Internal/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+
+try
+{
+    // Create a client connection
+    await using var connection = new ClientConnection(new Uri(uri), clientAuthenticationOptions);
+
+    // Create an invocation pipeline with the retry and deadline interceptors.
+    Pipeline pipeline = new Pipeline()
+        .UseRetry(new RetryOptions { MaxAttempts = maxAttempts })
+        .UseDeadline(defaultTimeout: TimeSpan.FromMilliseconds(timeout))
```
bernardnormier, line 36: I think the interceptors should be installed the other way around. The order matters.

ReeceHumphreys:
Okay, most of the relevant review feedback for this PR has now been implemented! The only stuff for future PRs should be refining the slice definitions and data we collect, removing the testing certificate code once the telemetry server is deployed, and potentially removing the hashing code once we decide if we want to collect it or not.

The goal of this PR should just be to get the underlying structure in for the build telemetry client.

README.md
```diff
@@ -2,7 +2,7 @@
   <img src="https://github.com/icerpc/icerpc-csharp/raw/main/.github/assets/icerpc-banner.svg" height="100" width="100" />
 </p>
 
-# IceRPC for C#
```
ReeceHumphreys, line 5: This is something we should be doing in all of our README.md when we have `C#`. Most markdown linters will complain about it.

tools/IceRpc.BuildTelemetry.Reporter/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+- The IceRPC version.
+- The system's operating system and version.
+- The source of the telemetry data (either `IceRpc.Slice.Tools or
+ `IceRpc.Protobuf.Tools`).
+- A SHA256 hash which is computed from the Slice files or Protobuf files.
+
+This data is used to help the IceRPC team understand how the tools are being
+used and to help prioritize future development efforts. The data is stored in a
+secure database and not shared with third parties.
```
InsertCreativityHere, line 20: grammar I think
```suggestion
secure database and is not shared with any third parties.
```

tools/Common/IceRpc.BuildTelemetry/TelemetryTask.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// Gets or sets the compilation hash.
+    /// </summary>
+    [Required]
+    public string CompilationHash { get; set; } = "";
+
+    /// <summary>
+    /// Gets or sets the source.
+    /// </summary>
+    [Required]
+    public string Source { get; set; } = "";
```
InsertCreativityHere, line 23: I'm still not a super big fan of the term "source", it's just so vague, and there's no further explanation about it.
Is this the IP address? A full copy of their source code? I know it has something to do with "Slice" vs "Protobuf" but that's it.

README.md
```diff
@@ -199,6 +199,11 @@ Ice server, and you can call services hosted by an IceRPC server from an Ice cli
 
 [IceRPC for Ice users] provides all the details.
 
+## Telemetry
+
+IceRPC collects completely anonymous telemetry data during the build process. Participation in this anonymous program is optional, and you may opt-out if you do not wish to share any information. Detailed information about the telemetry
+data collected and how to opt-out can be found in the [Telemetry] README.
```
InsertCreativityHere, line 205: 1) This is way over the 120 character limit.
2) To me, it feels like it's over-harping on how anonymous we are: "completely anonymous" "this anonymous 
program" I'd just mention it once at the beginning and leave it at that
3) "information about the telemetry data collected" is definitely correct English, but idk, it sounds a little weird to me.

Just to demonstrate (because you'd need to fix the line length anyways):
```suggestion
IceRPC collects anonymous telemetry data during the build process. Participation in this program is optional, and you may opt-out if you do not wish to share any information. Detailed information about what data is
collected and how to opt-out can be found in the [Telemetry] README.
```

README.md
```diff
@@ -231,3 +236,4 @@ in-memory transport for testing). Future releases may add additional transports.
 [Slic]: https://docs.icerpc.dev/icerpc/slic-transport/overview
 [Slice]: https://docs.icerpc.dev/slice2
 [System.IO.Pipelines]: https://learn.microsoft.com/en-us/dotnet/standard/io/pipelines
+[Telemetry]: tools/Common/IceRpc.Telemetry.Internal/README.md
```
InsertCreativityHere, line 239: Looks like this path changed as well:
```suggestion
[Telemetry]: tools/IceRpc.BuildTelemetry.Reporter/README.md
```

InsertCreativityHere (commented):
So far so good, taking a break before looking at the C# code.

tools/Common/IceRpc.BuildTelemetry/TelemetryTask.cs
ReeceHumphreys: This property will go away when we split the Slice definition into a Slice struct and Protobuf struct so I would prefer to just leave it as is for now!

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -62,6 +69,19 @@
       Sources="%(_ProtoFile.Identity)"
       Condition="'%(_ProtoFile.UpToDate)' != 'true'"
     />
+    <OutputHashTask Sources="%(_ProtoFile.Identity)" CompilationHash="$(Hash)" Condition="'$(DisableTelemetry)' != 'true'">
```
pepone, line 72:
```suggestion
    <OutputHashTask Sources=""@(ProtoFile)" Condition="'$(DisableTelemetry)' != 'true'">
```

tools/IceRpc.BuildTelemetry.Reporter/TelemetryExtensions.cs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Diagnostics;
```
pepone, line 3: Remove unused import
```suggestion
```

tools/IceRpc.BuildTelemetry.Reporter/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+        customChain.ChainPolicy.CustomTrustStore.Add(rootCA);
+        return customChain.Build((X509Certificate2)certificate!);
+    }
+};
+
+try
+{
+    // Create a cancellation token source to cancel the telemetry upload if it
+    // takes too long.
+    using var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(timeout));
```
pepone, line 30: Why not use the deadline interceptor?

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
```diff
@@ -36,9 +36,22 @@
     <PackageReference Include="Microsoft.Build.Framework" Version="17.8.3" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="17.8.3" />
     <PackageReference Include="Google.Protobuf.Tools" Version="$(NuGetProtobufVersion)" />
+    <PackageReference Include="System.Text.Json" Version="8.0.3" />
```
pepone, line 39: Is this required in Protobuf tools?

tools/IceRpc.Protobuf.Tools/OutputHashTask.cs
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) ZeroC, Inc.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System;
+using System.Collections.Generic;
```
pepone, line 6: Unused import

```suggestion
```

tools/IceRpc.Protobuf.Tools/OutputHashTask.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        using SHA256 sha256 = SHA256.Create();
+        string aggregatedHash = Sources
+            .Select(source =>
+            {
+                byte[] fileBytes = File.ReadAllBytes(source.GetMetadata("FullPath"));
+                byte[] hashBytes = sha256.ComputeHash(fileBytes);
+                return HexStringConverter.ToHexString(hashBytes);
+            })
+            .Aggregate((current, next) => current + next);
```
pepone, line 35: This doesn't seem correct. It computes a hash for each source file, and aggregates the hex strings concatenating them.

tools/IceRpc.Protobuf.Tools/ProtocTask.cs
```diff
@@ -132,3 +140,17 @@ protected override void LogEventsFromTextOutput(string singleLine, MessageImport
     /// <inheritdoc/>
     protected override void LogToolCommand(string message) => Log.LogMessage(MessageImportance.Normal, message);
 }
+
+/// <summary>Converts a byte array to a hexadecimal string.</summary>
+public static class HexStringConverter
```
pepone, line 145: It doesn't belong here, it should be a private or local method in OutputHashTask

tools/IceRpc.BuildTelemetry.Reporter/Program.cs
bernardnormier: We want to use this token for the invocation + the shutdown.

tools/IceRpc.BuildTelemetry.Reporter/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    // Upload the telemetry to the server.
+    await reporter.UploadAsync(new Telemetry(args), cancellationToken: cts.Token);
+
+    // Shutdown the connection.
+    await connection.ShutdownAsync(cts.Token);
+}
+catch (Exception)
+{
+    // Consume all exceptions.
```
bernardnormier, line 46: You can / should remove `(Exception)` in this case.

tools/IceRpc.BuildTelemetry.Reporter/README.md
```diff
@@ -0,0 +1,35 @@
+# Telemetry
```
bernardnormier, line 1: Build telemetry

tools/IceRpc.BuildTelemetry.Reporter/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+`IceRpc.Protobuf.Tools` and `IceRpc.Slice.Tools` collect anonymous
+telemetry data about general usage. Participation in this anonymous program is
+optional, and you may [opt-out](#opting-out) if you'd prefer to not share any
+information.
+
+This package includes the `IceRpc.BuildTelemetry.Reporter` client. This client is an
+IceRPC C# client that sends anonymous telemetry data over a secure connection
```
bernardnormier, line 9: build telemetry data ...

tools/IceRpc.BuildTelemetry.Reporter/README.md
```diff
@@ -0,0 +1,35 @@
...
+secure database and is not shared with any third parties.
+
+## Opting Out
+
+To opt out of the telemetry program, add the following property to your
+csharp project file:
+
+```xml
+<PropertyGroup>
+ <DisableTelemetry>true</DisableTelemetry>
```
bernardnormier, line 30: Should be DisableBuildTelemetry.

tools/IceRpc.BuildTelemetry.Reporter/README.md
```diff
@@ -0,0 +1,35 @@
...
+</PropertyGroup>
+```
+
+This will prevent the telemetry client from sending any data to the IceRPC
```
bernardnormier, line 34: build telemetry

tools/IceRpc.BuildTelemetry.Reporter/TelemetryExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="Telemetry" /> struct using the specified command-line arguments.
+    /// </summary>
+    /// <param name="args">The command-line arguments.</param>
+    public Telemetry(string[] args)
+    {
+        // Determine the IceRPC version using the assembly version
+        var assembly = Assembly.GetAssembly(typeof(Telemetry));
+        string version = assembly?.GetName()?.Version?.ToString() ?? "unknown";
```
bernardnormier, line 18: How can assembly be null here? It's our own assembly.

tools/IceRpc.BuildTelemetry.Reporter/TelemetryExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+            .Skip(1)
+            .FirstOrDefault() ?? "unknown";
+
+        // Parse the compilation hash
+        string compilationHash = args
+            .SkipWhile(arg => arg != "--hash")
+            .Skip(1)
+            .FirstOrDefault() ?? "unknown";
+
+        IceRpcVersion = version;
```
bernardnormier, line 32: Rather than report the version of the BuildTelemetry assembly - not all that interesting - should we report the version of slicec / protoc that we're using?

tools/IceRpc.BuildTelemetry.Reporter/TelemetryExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// Initializes a new instance of the <see cref="Telemetry" /> struct using the specified command-line arguments.
+    /// </summary>
+    /// <param name="args">The command-line arguments.</param>
+    public Telemetry(string[] args)
+    {
+        // Determine the IceRPC version using the assembly version
+        var assembly = Assembly.GetAssembly(typeof(Telemetry));
+        string version = assembly?.GetName()?.Version?.ToString() ?? "unknown";
+
+        // Parse the source
```
bernardnormier, line 20: I don't know what this `source` is, even after reading the code below.

We need a clearer name.

tools/IceRpc.BuildTelemetry.Reporter/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+    using var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(timeout));
+
+    // Create a client connection
+    await using var connection = new ClientConnection(new Uri(uri), clientAuthenticationOptions);
+
+    // Create a reporter proxy with this client connection.
+    var reporter = new ReporterProxy(connection);
+
+    // Upload the telemetry to the server.
+    await reporter.UploadAsync(new Telemetry(args), cancellationToken: cts.Token);
```
bernardnormier, line 39: I am not a fan of this `Telemetry(args)` construction.

I think it would be cleaner to have Slice and Protobuf-specific Programs, where each constructs its own (Slice|Protobuf)BuildTelemetryReport struct "by hand".

tools/Common/IceRpc.BuildTelemetry/TelemetryTask.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace IceRpc.BuildTelemetry;
+
+/// <summary>
+/// A custom MSBuild task to run telemetry commands.
+/// </summary>
+public class TelemetryTask : ToolTask
```
bernardnormier, line 11: I would create separate tasks for (in) IceRpc.Protobuf.Tools and IceRpc.Slice.Tools - simpler and cleaner. Same for the actual client code.

tools/IceRpc.BuildTelemetry.Reporter/TelemetryExtensions.cs
ReeceHumphreys: The version of the BuildTelemetry is set to the same as the IceRPC version so its really the version of IceRPC that is being used.

We could also include slicec / protoc version though!

tools/IceRpc.BuildTelemetry.Reporter/Program.cs
ReeceHumphreys: I think this type of refactor should be done in a follow up! It works good enough for now plus we will eventually want to refactor this to all work for the plugin system anyway.

tools/IceRpc.BuildTelemetry.Reporter/TelemetryExtensions.cs
ReeceHumphreys: Added a comment

tools/Common/IceRpc.BuildTelemetry/TelemetryTask.cs
ReeceHumphreys: In the interest of moving forward with this I think the current structure is fine. It does its job and is clean enough. This code is going to be refactored once we have the plugin system in the future so I don't want to spend too much time obsessing over having the cleanest possible telemetry client. It should be treat as an "Alpha" version of the telemetry at this point I think.

tools/IceRpc.Slice.Tools/IceRpc.Slice.Tools.csproj
```diff
@@ -48,6 +48,18 @@
...
   <ItemGroup>
+    <Compile Include="../Common/IceRpc.BuildTelemetry/TelemetryTask.cs" />
+    <None
+      Include="../IceRpc.BuildTelemetry.Reporter/bin/$(Configuration)/net8.0/*"
+      Exclude="../IceRpc.BuildTelemetry.Reporter/bin/$(Configuration)/net8.0/*.exe"
+    >
+      <PackagePath>tools/</PackagePath>
+      <Pack>true</Pack>
+    </None>
+    <None Include="../IceRpc.BuildTelemetry.Reporter/certs/cacert.der">
```
pepone, line 59: Add a TODO to remove this

tools/IceRpc.Protobuf.Tools/OutputHashTask.cs
```diff
@@ -0,0 +1,63 @@
...
+    public ITaskItem[] Sources { get; set; } = Array.Empty<ITaskItem>();
+
+    /// <summary>The computed SHA-256 hash of the Protobuf files.</summary>
+    [Output]
+    public string? CompilationHash { get; set; }
+
+    /// <inheritdoc/>
+    public override bool Execute()
+    {
+        using SHA256 sha256 = SHA256.Create();
```
pepone, line 27:
```suggestion
        using var sha256 = SHA256.Create();
```

tools/IceRpc.BuildTelemetry.Reporter/slice/Reporter.slice
```diff
@@ -0,0 +1,60 @@
...
+/// Represents a simple telemetry reporter that can upload telemetry data to a
+/// remote server.
+interface Reporter {
+    /// Uploads telemetry data.
+    /// @param telemetry: The telemetry data to upload.
+    [oneway] upload(telemetry: BuildTelemetry)
+}
+
+/// Represents the source of the telemetry data.
+enum Source {
```
bernardnormier, line 15: I find the name 'source' confusing. I would use TargetLanguage instead--meaning the programming language for which the build is generating code.

Or just skip this entirely since it's all Alpha anyway.

tools/IceRpc.BuildTelemetry.Reporter/slice/Reporter.slice
```diff
@@ -0,0 +1,60 @@
...
+    /// The operating system that the client is running on.
+    OperatingSystem: string
+
+    /// The source of the telemetry data.
+    Source: Source
+
+    /// The hashed slice files.
+    CompilationHash: string
+
+    /// Whether the compilation used any Slice1 files.
```
bernardnormier, line 43: It's not clear to me what this means.

Is it:
 - whether any of the source + reference files contains Slice1?
 That's what Austin wants but makes little sense to me as our Reference files contain Slice1
 - whether any of the source files contains Slice1?

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -62,6 +69,19 @@
...
     />
+    <OutputHashTask Sources="@(ProtoFile)" Condition="'$(DisableBuildTelemetry)' != 'true'">
+      <Output PropertyName="Hash" TaskParameter="CompilationHash" />
+    </OutputHashTask>
+
+    <!-- Run telemetry -->
+    <TelemetryTask
+      WorkingDirectory="$(TelemetryScriptPath)"
+      CompilationHash="$(Hash)"
+      Condition="'$(DisableBuildTelemetry)' != 'true'"
```
bernardnormier, line 80: I am not a fan of expressions like DisableXxx != true.

tools/IceRpc.BuildTelemetry.Reporter/slice/Reporter.slice
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) ZeroC, Inc.
+
+[cs::namespace("IceRpc.BuildTelemetry.Reporter")]
```
bernardnormier, line 3: I don't find this correct.

The module should be `IceRpc::BuildTelemetry` and we should not use any `cs::namespace`.

tools/IceRpc.BuildTelemetry.Reporter/slice/Reporter.slice
```diff
@@ -0,0 +1,60 @@
...
+enum Source {
+    /// The data is coming from icerpc-csharp
+    CSharp(dotnetRuntime: string)
+
+    // Additional sources can be added along with corresponding data once
+    // needed.
+}
+
+/// Represents the telemetry data that can be uploaded.
+enum BuildTelemetry {
```
bernardnormier, line 24: The names BuildTelemetry/SliceTelemetry/ProtobufTelemetry make it sound they are similar types, but they are not.

I would rename SliceTelemetry and ProtobufTelemetry to SliceTelemetryData / ProtobufTelemetryData.

tools/Common/IceRpc.BuildTelemetry/TelemetryTask.cs
bernardnormier: I don't see the difficulty in splitting this task into two tasks. It should be a matter of a couple minutes. And it would make the code simpler since you can eliminate Idl.

This common task ties the Slice compilation and the Protobof compilation, which I don't think is desirable.

pepone (approved):
This is a good starting point and can be refined in future PRs.

Good job, @ReeceHumphreys!

ReeceHumphreys:
Thanks @pepone! This build system stuff is definitely not my expertise haha.

bernardnormier (commented):
> I would structure this a bit differently:

I only see this comment now. In my view, the common task is not a good idea. It's simpler and cleaner to have separate tasks for Slice and Protobuf.

tools/slicec-cs/src/main.rs
```diff
@@ -61,7 +62,13 @@ pub fn main() {
     // If the telemetry flag is set, output additional compilation information.
     if cs_options.telemetry {
         let hash = SliceFile::compute_sha256_hash(&compilation_state.files);
-        println!(r#"{{ "hash": "{hash}" }}"#);
+        let contains_slice1 = compilation_state
```
InsertCreativityHere, line 65: Fine for now as a proof of concept, just pointing to
https://zeroc.atlassian.net/wiki/spaces/~austin/pages/1595539457/Telemetry+Information+Worth+Collecting?focusedCommentId=1610022918
so we can have a proper discussion about this, instead of mixing it into review comments!

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -95,10 +99,18 @@ protected override void LogEventsFromTextOutput(string singleLine, MessageImport
...
+            {
+                // Messages from stdout
+                var jsonDoc = System.Text.Json.JsonDocument.Parse(singleLine);
+                CompilationHash = jsonDoc.RootElement.GetProperty("hash").GetString();
+                ContainsSlice1 = jsonDoc.RootElement.GetProperty("contains_slice1").GetString();
+                return;
+            }
+            catch (System.Text.Json.JsonException)
+            {
+                Log.LogError(singleLine);
```
InsertCreativityHere, line 112: Should we print part of the `JsonException` here too? Just to provide some details about what actually failed for when we're trying to debug this 5 years from now : vP

src/Directory.Build.props
```diff
@@ -5,6 +5,7 @@
     <!-- Include PDB in the built .nupkg -->
     <AllowedOutputExtensionsInPackageBuildOutputFolder
     >$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <DisableBuildTelemetry>true</DisableBuildTelemetry>
```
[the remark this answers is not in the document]
InsertCreativityHere: Personally, I prefer the current name. Having it be a 'negative' phrase It makes it very clear that this is opt-out, and that by _default_, we're sending build telemetry.

Whereas naming it something positive (IMO) implicitly makes it seem more like it's opt-in. At least without having to go read some defaults somewhere.

You are completely correct that most don't have negative prefixes, but it's also not completely taboo, since there are 2 which start with negative prefixes.

----

Either way, I think this can be left for a second phase of refinement.

InsertCreativityHere (approved):
Looks like a good start!

I didn't look too closely at the C# code/build-system stuff, since I'm not super familiar with it.

tools/IceRpc.BuildTelemetry.Reporter/README.md
```diff
@@ -0,0 +1,35 @@
+# Build Telemetry
+
+`IceRpc.Protobuf.Tools` and `IceRpc.Slice.Tools` collect anonymous
+build telemetry data about general usage. Participation in this anonymous
+program is optional, and you may [opt-out](#opting-out) if you'd prefer to not
+share any information.
```
InsertCreativityHere, line 6: Personally, I feel like we're overusing `anonymous` here. At best it's just redundant, at worst, stressing how anonymous it is makes it seem suspicious IMO.
```suggestion
`IceRpc.Protobuf.Tools` and `IceRpc.Slice.Tools` collect anonymous
build telemetry data about general usage. Participation in this program is
optional, and you may [opt-out](#opting-out) if you'd prefer to not share any 
information.
```

tools/IceRpc.BuildTelemetry.Reporter/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+This package includes the `IceRpc.BuildTelemetry.Reporter` client. This client is an
+IceRPC C# client that sends anonymous telemetry data over a secure connection
```
InsertCreativityHere, line 9: And here I think we're overusing 'client'. We say it 3 times in 2 very short sentences.
Personally, I'd get rid of the 2nd one, it's the most redundant here.
```suggestion
This package includes the `IceRpc.BuildTelemetry.Reporter` client. This is an
IceRPC C# client that sends anonymous telemetry data over a secure connection
```

tools/IceRpc.BuildTelemetry.Reporter/README.md
```diff
@@ -0,0 +1,35 @@
...
+csharp project file:
+
+```xml
+<PropertyGroup>
+ <DisableBuildTelemetry>true</DisableBuildTelemetry>
+</PropertyGroup>
+```
+
+This will prevent the build telemetry client from sending any data to the IceRPC
+telemetry server.
```
InsertCreativityHere, line 35: I find this a little weird sounding. It suggests that we still compute/collect the data, just that we don't send it to our servers.

I think something like this is simpler/closer to what actually happens (right?):
```suggestion
Setting this property to `true` completely disables the computation and collection of telemetry.
```
Sorry for the nitpicking, I just want to reduce the chance of people being put off by our telemetry due to misunderstandings.

tools/IceRpc.BuildTelemetry.Reporter/slice/Reporter.slice
InsertCreativityHere: Technically what Austin wants is:
"what compilation modes are being used in the user's files"
_after filtering out the built-in-slice files_!

There's no benefit to including these built-in files in any of our telemetry, so we should always be filtering them out (but that logic can wait until we agree on what to collect of course).

ReeceHumphreys: This is if any of the users slice files are slice1

PR #4007, opened 2024-06-06
Format XML with prettier

ReeceHumphreys:
After looking at a few different XML parsers, I think this integration from prettier seems to be the best.
- I looked into the Red Hat one, which is a popular VSCode extension, but there does not seem to be a way to get the formatting from it as a command line tool or in a GitHub action.
- I also looked into `xmllint` but its formatting wasn't that much of an improvement from what we have.
- Prettier seems to do the best job at formatting, and it's nice and easy to run from the command line. It will be integrated into the CI/CD easily. If this formatting is okay, I will update this PR to update the actions.


ReeceHumphreys:
Will look into CI failing. Regardless, this PR is fine to review as it's just about if you guys are fine with this formatting style!

ReeceHumphreys:
Since some of the MSBuild stuff is whitespace sensitive the latest commit does the formatting in two passes

```
prettier --write '**/*.{props,csproj,targets}' --plugin=@prettier/plugin-xml --ignore-path .prettierignore   
prettier --write 'tools/**/*.{props,targets}' --plugin=@prettier/plugin-xml --config tools/.prettierrc --ignore-path /dev/null
```

Where the pass over the tools directory uses a config file that removes the whitespace formatting. Not ideal but should fix the builds.

pepone (approved):
Looks good, do we need to update `.vscode/extensions.json` to add this formatting plug-in?

bernardnormier:
It would be good to open the projects with Visual Studio (_not VS code) to make sure VS does not immediately reformat all the project files.

ReeceHumphreys:
@bernardnormier Just opened the `IceRpc.sln` on my mac with visual studio and nothing changed! I have 0 familiarity with Visual Studio so im not sure if this is sufficient to show that the formatting is preserved. @pepone could you verify if Visual Studio leaves the files with the new formatting?

pepone:
> could you verify if Visual Studio leaves the files with the new formatting?

Yes, formatting remains the same after opening the solution in Visual Studio 2022 on Windows.

PR #4008, opened 2024-06-06
`metrics` flag for `slicec-cs` added

ReeceHumphreys:
This PR is the cleaned up version of the additional output from `slicec-cs` from https://github.com/icerpc/icerpc-csharp/pull/4006. Notably, I did remove the updated files information. After discussion with @pepone I don't actually think we need this data point.

### DIFF
--- a/tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
+++ b/tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
@@ -95,6 +95,7 @@ public class SliceCCSharpTask : ToolTask
     {
         if (messageImportance == MessageImportance.Low)
         {
+            // Messages from stdout
             var jsonDoc = System.Text.Json.JsonDocument.Parse(singleLine);
             OutputHash = jsonDoc.RootElement.GetProperty("hash").GetString();
             return;
--- a/tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
+++ b/tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
@@ -39,7 +39,7 @@ public class SliceCCSharpTask : ToolTask
 
     /// <inheritdoc/>
     protected override string ToolName =>
-    RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "slicec-cs.exe" : "slicec-cs";
+        RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "slicec-cs.exe" : "slicec-cs";
 
     /// <summary>The computed SHA-256 hash of the Slice files.</summary>
     [Output]
--- a/tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
+++ b/tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
@@ -67,7 +67,7 @@ public class SliceCCSharpTask : ToolTask
             builder.AppendTextUnquoted(option);
         }
         builder.AppendSwitch("--diagnostic-format=json");
-        builder.AppendSwitch("--metrics");
+        builder.AppendSwitch("--telemetry");
         builder.AppendFileNamesIfNotNull(
             Sources.Select(item => item.GetMetadata("FullPath").ToString()).ToArray(),
             " ");
@@ -95,7 +95,8 @@ public class SliceCCSharpTask : ToolTask
     {
         if (messageImportance == MessageImportance.Low)
         {
-            // Ignore messages from stdout, messageImportance is set to MessageImportance.Low for messages from stdout.
+            var jsonDoc = System.Text.Json.JsonDocument.Parse(singleLine);
+            OutputHash = jsonDoc.RootElement.GetProperty("hash").GetString();
             return;
         }
 
@@ -150,40 +151,4 @@ public class SliceCCSharpTask : ToolTask
 
     /// <inheritdoc/>
     protected override void LogToolCommand(string message) => Log.LogMessage(MessageImportance.Normal, message);
-
-    /// <inheritdoc/>
-    protected override int ExecuteTool(string pathToTool, string responseFileCommands, string commandLineCommands)
-    {
-        var startInfo = new ProcessStartInfo
-        {
-            FileName = pathToTool,
-            Arguments = commandLineCommands,
-            RedirectStandardOutput = true,
-            UseShellExecute = false,
-            CreateNoWindow = true,
-            WorkingDirectory = GetWorkingDirectory()
-        };
-
-        using var process = Process.Start(startInfo);
-
-        if (process is null)
-        {
-            Log.LogError("Failed to start the Slice compiler process.");
-            return -1;
-        }
-
-        string output = process.StandardOutput.ReadToEnd();
-        process.WaitForExit();
-
-        if (process.ExitCode != 0)
-        {
-            Log.LogError($"Slice compiler failed with exit code {process.ExitCode}.");
-            return process.ExitCode;
-        }
-
-        var jsonDoc = System.Text.Json.JsonDocument.Parse(output);
-        OutputHash = jsonDoc.RootElement.GetProperty("hash").GetString();
-
-        return process.ExitCode;
-    }
 }
--- a/tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
+++ b/tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
@@ -39,7 +39,11 @@ public class SliceCCSharpTask : ToolTask
 
     /// <inheritdoc/>
     protected override string ToolName =>
-        RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "slicec-cs.exe" : "slicec-cs";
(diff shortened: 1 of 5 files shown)


tools/slicec-cs/src/cs_options.rs
```diff
@@ -18,6 +18,10 @@ pub struct CsOptions {
     #[arg(long = "rpc", value_enum, default_value_t = RpcProvider::IceRpc, ignore_case = true)]
     pub rpc_provider: RpcProvider,
 
+    /// Enable additional compilation information output.
+    #[arg(long)]
+    pub metrics: bool,
```
bernardnormier, line 23: Since it's for build telemetry, I would not use a different term here.

tools/slicec-cs/src/main.rs
```diff
@@ -57,7 +58,15 @@ pub fn main() {
         }
     }
 
-    std::process::exit(i32::from(compilation_state.emit_diagnostics(slice_options)))
+    // Emit diagnostics and totals.
+    let exit_code = i32::from(compilation_state.emit_diagnostics(slice_options));
+
+    // If the metrics flag is set mode is enabled, output additional compilation information.
```
bernardnormier, line 64: The comment needs to be fixed.

It's unclear why we include the exit code in the output. Presumably the caller always gets the exit code and is free to do whatever it wants with it.

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -145,4 +150,40 @@ protected override void LogEventsFromTextOutput(string singleLine, MessageImport
 
     /// <inheritdoc/>
     protected override void LogToolCommand(string message) => Log.LogMessage(MessageImportance.Normal, message);
+
+    /// <inheritdoc/>
+    protected override int ExecuteTool(string pathToTool, string responseFileCommands, string commandLineCommands)
```
bernardnormier, line 155: What is the default ExecuteTool implementation that we override here?

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -39,7 +39,11 @@ public class SliceCCSharpTask : ToolTask
 
     /// <inheritdoc/>
     protected override string ToolName =>
-        RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "slicec-cs.exe" : "slicec-cs";
+    RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "slicec-cs.exe" : "slicec-cs";
```
bernardnormier, line 42: Formatting bug.

ReeceHumphreys:
I already addressed all of @InsertCreativityHere's feedback and plan on trying to get this merged tomorrow so I am removing him as a reviewer.

tools/slicec-cs/src/cs_options.rs
ReeceHumphreys: We use it for telemetry but in the future it could include additional information. Information that other tooling may want to use such as the language server. As such, I thought keeping the term more neutral makes more sense.

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
ReeceHumphreys: https://github.com/dotnet/msbuild/blob/b8e5cce93e2e664eaf470c0fed9f5b84b8591e1e/src/Utilities/ToolTask.cs#L715

pepone: I'm not sure that we need to override this method. The base implementation does many things see:
 https://github.com/dotnet/msbuild/blob/b8e5cce93e2e664eaf470c0fed9f5b84b8591e1e/src/Utilities/ToolTask.cs#L715

Why don't we process the messages in https://github.com/icerpc/icerpc-csharp/blob/0153bc8323107da94d1fc9a360f59d737647b480/tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs#L93

MSBuild calls LogEventsFromTextOutput to process each line sent to stderr or stdout.  The trick here is that we either get all the output on a single line, accumulate it and retry.

tools/slicec-cs/src/cs_options.rs
bernardnormier: All this code is temporary: slicec-cs will be refactored and once this is done, we should move the build telemetry collection to a slicec plugin.

Why not simply output this hash all the time? It's not like anybody executes slicec-cs "by hand".

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
bernardnormier: This sounds like a good idea, but I don't understand the difficulty. 

Does slicec-cs output anything else to stdout? Warnings and errors go to stderr, don't they?

pepone: I think that with this override `LogEventsFromTextOutput` would not be called, and errors would not be properly logged, to be confirmed.

tools/slicec-cs/src/cs_options.rs
ReeceHumphreys: Very true that's a fair point! No point in trying to make this overly future proof.

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
pepone: > This sounds like a good idea, but I don't understand the difficulty.

It shouldn't be to difficult. We are already using this method to process error messages. It is simpler when the compiler emits a complete JSON message in each line, and the task can parse it right away and decide what to do with it.

tools/slicec-cs/src/cs_options.rs
ReeceHumphreys: Thinking more about this I actually think we shouldn't just output the hash at this time. We don't fully know all the telemetry we want to collect in this initial version yet. One additional part of the output I was considering adding was if the user is using any `encoding = Slice1` files. This would be included as part of the JSON output then. Im cool with renaming it to telemetry though!

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -41,6 +41,10 @@ public class SliceCCSharpTask : ToolTask
     protected override string ToolName =>
         RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "slicec-cs.exe" : "slicec-cs";
 
+    /// <summary>The computed SHA-256 hash of the Slice files.</summary>
+    [Output]
+    public string? OutputHash { get; set; }
```
externl, line 46:
```suggestion
    public string? ComputedHash { get; set; }
```

externl (approved):

pepone (approved):
The code looks good, I'm still not sure about the usefulness of CompilationHash

tools/slicec-cs/src/main.rs
```diff
@@ -57,7 +58,15 @@ pub fn main() {
         }
     }
 
-    std::process::exit(i32::from(compilation_state.emit_diagnostics(slice_options)))
+    // If the metrics flag is set output additional compilation information.
```
bernardnormier, line 61: Update the comment.

bernardnormier (approved):

tools/slicec-cs/src/main.rs
```diff
@@ -34,7 +35,7 @@ pub fn main() {
     let slice_options = &cs_options.slice_options;
 
     let mut compilation_state = slicec::compile_from_options(slice_options, cs_patcher, cs_validator);
-
+    let hash = SliceFile::compute_sha256_hash(&compilation_state.files);
```
InsertCreativityHere, line 38: Could we move this down into the `cs_options.telemetry` if-check?
1) That's the only place where it's used, and it's better to put code where it's used
2) If it telemetry isn't enabled, we shouldn't bother computing it (since we do nothing with it).

I'd also like to keep the space between `let mut compilation_state` and the `if !compilation_state`.
These are both very big steps of compiler execution, and don't really relate to each other directly.

tools/slicec-cs/src/cs_options.rs
```diff
@@ -18,6 +18,10 @@ pub struct CsOptions {
     #[arg(long = "rpc", value_enum, default_value_t = RpcProvider::IceRpc, ignore_case = true)]
     pub rpc_provider: RpcProvider,
 
+    /// Enable additional compilation information output.
```
InsertCreativityHere, line 21: Sorry to be a little nitpicky, but I also think that this comment should be reworded.
It sounds like a hold-over from when this was `--verbose`.

If this was internal, I wouldn't care, but `clap` uses these doc comments to generate help messages, suggestions, etc.
So they are directly user-facing.

InsertCreativityHere (approved):
Looks good.

----

I'm also not convinced about the compilation hash's usefulness.
But this PR gets the structure in place, and that's what's most important right now.

For context, see: https://zeroc.atlassian.net/wiki/spaces/~austin/pages/1595539457/Telemetry+Information+Worth+Collecting

PR #4009, opened 2024-06-09
Make decorator non-resettable on the last attempt

bernardnormier:
This is a small enhancement to the Retry interceptor - it makes the decorator non-resettable on the last attempt to recycle the request memory as soon as possible.

### DIFF
--- a/src/IceRpc.Retry/RetryInterceptor.cs
+++ b/src/IceRpc.Retry/RetryInterceptor.cs
@@ -43,9 +43,9 @@ namespace IceRpc.Retry;
 public class RetryInterceptor : IInvoker
 {
     private readonly ILogger _logger;
-    private readonly IInvoker _next;
     private readonly int _maxAttempts;
     private readonly int _maxPayloadSize;
+    private readonly IInvoker _next;
 
     /// <summary>Constructs a retry interceptor.</summary>
     /// <param name="next">The next invoker in the invocation pipeline.</param>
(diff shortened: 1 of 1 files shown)


src/IceRpc.Retry/RetryInterceptor.cs
```diff
@@ -43,8 +43,9 @@ namespace IceRpc.Retry;
 public class RetryInterceptor : IInvoker
 {
     private readonly ILogger _logger;
+    private readonly int _maxAttempts;
```
bernardnormier, line 46: I think the preferred pattern is / should be to _not_ keep a reference to options (which the caller could in theory change) but instead copy options into individual fields. That's what we do already for Ice and IceRpcProtocolConnection.

src/IceRpc.Retry/RetryInterceptor.cs
```diff
@@ -53,21 +54,22 @@ public class RetryInterceptor : IInvoker
...
+        _maxPayloadSize = options.MaxPayloadSize;
         _logger = logger;
     }
 
     /// <inheritdoc/>
     public async Task<IncomingResponse> InvokeAsync(OutgoingRequest request, CancellationToken cancellationToken)
     {
-        if (request.PayloadContinuation is not null)
+        // This interceptor does not support retrying requests with a payload continuation.
+        if (request.PayloadContinuation is not null || _maxAttempts == 1)
```
bernardnormier, line 66: If max attempts is 1, there is no retrying and this interceptor shouldn't install a decorator.

externl (approved):

PR #4013, opened 2024-07-05
Simplify Slice exceptions and Slice activator

bernardnormier:
This PR simplifies Slice exceptions and the Slice activator (used to create class and exception instances from a Slice type ID).

Slice exceptions were more complicated than needed because for a while we wanted to use them with Slice2 too, but we abandoned this idea. So exceptions are:
 - Slice1-only
 - usually transmitted over the ice protocol
 - conceptually similar to the error component of a Slice2 Result

This means the exception message (not transmitted by Slice1) and the inner exception (not meaningful) don't matter much. Note that the error component of a Slice2 Result does have these either.

So this PR removes the message and innerException parameters from the generated exception constructors. As a result, both generated classes and exceptions rely on the same parameterless constructor for decoding.

This PR also adds a small test to demonstrate how users can create custom messages for generated exception (by overriding Message).

### DIFF
--- a/slice/IceRpc/Internal/IceDefinitions.slice
+++ b/slice/IceRpc/Internal/IceDefinitions.slice
@@ -97,19 +97,20 @@ enum ReplyStatus {
     /// The target object does not exist.
     ObjectNotExistException = 2
 
-    /// The target object does not support the facet.
+    /// The target object does not support the facet (treated like ObjectNotExistException by IceRPC).
     FacetNotExistException = 3
 
     /// The target object does not support the operation.
     OperationNotExistException = 4
 
-    /// The reply message carries an unknown Ice local exception.
+    /// The dispatch failed with an Ice local exception (not applicable with IceRPC).
     UnknownLocalException = 5
 
-    /// The reply message carries an unknown Ice user exception.
+    /// The dispatch failed with a Slice user exception does not conform to the exception specification of the Slice
+    // operation (not applicable with IceRPC).
     UnknownUserException = 6
 
-    /// The reply message carries an unknown exception.
+    /// The dispatch failed with some other exception.
     UnknownException = 7
 }
 
--- a/slice/IceRpc/Internal/IceDefinitions.slice
+++ b/slice/IceRpc/Internal/IceDefinitions.slice
@@ -106,7 +106,7 @@ enum ReplyStatus {
     /// The dispatch failed with an Ice local exception (not applicable with IceRPC).
     UnknownLocalException = 5
 
-    /// The dispatch failed with a Slice user exception does not conform to the exception specification of the Slice
+    /// The dispatch failed with a Slice user exception that does not conform to the exception specification of the Slice
     // operation (not applicable with IceRPC).
     UnknownUserException = 6
 
--- a/slice/IceRpc/Internal/IceDefinitions.slice
+++ b/slice/IceRpc/Internal/IceDefinitions.slice
@@ -106,8 +106,8 @@ enum ReplyStatus {
     /// The dispatch failed with an Ice local exception (not applicable with IceRPC).
     UnknownLocalException = 5
 
-    /// The dispatch failed with a Slice user exception that does not conform to the exception specification of the Slice
-    // operation (not applicable with IceRPC).
+    /// The dispatch failed with a Slice user exception that does not conform to the exception specification of the
+    /// operation (not applicable with IceRPC).
     UnknownUserException = 6
 
     /// The dispatch failed with some other exception.
--- a/src/ZeroC.Slice/Internal/Activator.cs
+++ b/src/ZeroC.Slice/Internal/Activator.cs
@@ -19,11 +19,7 @@ internal class Activator : IActivator
 
     private readonly IReadOnlyDictionary<string, Lazy<ActivateObject>> _dict;
 
-    public object? CreateClassInstance(string typeId) =>
-        _dict.TryGetValue(typeId, out Lazy<ActivateObject>? factory) ?
-            factory.Value() : null;
-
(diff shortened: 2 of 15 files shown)


tests/IceRpc.Slice.Tests/ExceptionTests.slice
```diff
@@ -25,7 +25,6 @@ exception EmptyException {}
 
 interface SliceExceptionOperations {
     opThrowsMultipleExceptions() throws (MyException, EmptyException)
-
```
externl, line 28: I think it's a bit cleaner to have them all separated by a newline, or all by none.

tools/slicec-cs/src/builders.rs
```diff
@@ -336,22 +331,23 @@ impl FunctionBuilder {
...
-                match parameter.data_type.concrete_typeref() {
-                    // Sequences of fixed-size numeric types are mapped to `ReadOnlyMemory<T>` and have to use
-                    // 'default' as their default value. Other optional types are mapped to nullable types and
-                    // can use 'null' as the default value, which makes it clear what the default is.
-                    TypeRefs::Sequence(sequence_ref)
-                        if sequence_ref.has_fixed_size_primitive_elements()
-                            && !sequence_ref.has_attribute::<CsType>() =>
-                    {
-                        Some("default")
+            let default_value =
```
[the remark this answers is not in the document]
externl: More than likely, the rust formatter is updated frequently.

externl (approved):

tools/slicec-cs/src/generators/class_generator.rs
```diff
@@ -95,7 +95,8 @@ pub fn generate_class(class_def: &Class) -> CodeBlock {
 
     // parameterless constructor for decoding, generated only if the preceding constructors are not parameterless
     if non_nullable_fields.len() + non_nullable_base_fields.len() > 0 {
-        let mut decode_constructor = FunctionBuilder::new(access, "", &class_name, FunctionType::BlockBody);
+        // The constructor needs to be public for System.Activator.CreateInstance.
```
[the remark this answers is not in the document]
bernardnormier: I actually don't want to remove the constructors, but still rework this code. See #4014.

slice/IceRpc/Internal/IceDefinitions.slice
```diff
@@ -97,19 +97,20 @@ enum ReplyStatus {
...
 
     /// The target object does not support the operation.
     OperationNotExistException = 4
 
-    /// The reply message carries an unknown Ice local exception.
+    /// The dispatch failed with an Ice local exception (not applicable with IceRPC).
     UnknownLocalException = 5
 
-    /// The reply message carries an unknown Ice user exception.
+    /// The dispatch failed with a Slice user exception does not conform to the exception specification of the Slice
```
pepone, line 109:
```suggestion
    /// The dispatch failed with a Slice user exception that does not conform to the exception specification of the Slice
```

src/ZeroC.Slice/Internal/Activator.cs
```diff
@@ -100,28 +89,6 @@ internal Activator Get(Assembly assembly)
...
-                null) ?? throw new InvalidOperationException($"Cannot find Slice decoding constructor for '{type}'.");
-
-            ParameterExpression messageParam = Expression.Parameter(typeof(string), "message");
-            ParameterExpression innerExceptionParam = Expression.Parameter(typeof(Exception), "innerException");
-
-            Expression expression = isException ?
-                Expression.New(constructor, messageParam, innerExceptionParam) :
-                Expression.New(constructor);
-
-            return Expression.Lambda<ActivateObject>(expression, messageParam, innerExceptionParam).Compile();
```
pepone, line 123: Using `Expression.Lambda`  is much faster than  `Activator.CreateInstance`

InsertCreativityHere (approved):
Looks good!

src/ZeroC.Slice/Internal/Activator.cs
bernardnormier: It's surprising, and more importantly, does it matter?

slice/IceRpc/Internal/IceDefinitions.slice
```diff
@@ -97,19 +97,20 @@ enum ReplyStatus {
...
+    // operation (not applicable with IceRPC).
```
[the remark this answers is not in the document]
bernardnormier: Fixed.

tests/IceRpc.Slice.Tests/ExceptionTests.slice
bernardnormier: Fixed

src/ZeroC.Slice/Internal/Activator.cs
pepone: > It's surprising,

One uses compiled code and the other uses reflection.

>  does it matter?

It depends, this would be noticeable for applications decoding a larger number of classes. So probably not very common.

It is a tradeoff of simpler code to this extra bit of performance for class decoding.

bernardnormier: Here is benchmark:
https://andrewlock.net/benchmarking-4-reflection-methods-for-calling-a-constructor-in-dotnet/

If we keep the "compiled expression" approach, we also need to keep LazyObject.

Note that ActivatorCreateInstance is not as slow as actual Reflection.

pepone: I'm fine with using the simpler CreateInstance, I just wrote the comment so that we were aware of the performance differences.

pepone (approved):

PR #4018, opened 2024-07-12
Collect additional build telemetry data

ReeceHumphreys:
This PR adds a couple fields that was discussed earlier in the week to the build telemetry.


tools/Common/IceRpc.BuildTelemetry/TelemetryTask.cs
```diff
@@ -56,6 +71,17 @@ protected override string GenerateCommandLineCommands()
...
+            commandLine.AppendSwitch("--contains-slice2");
+            commandLine.AppendSwitch(ContainsSlice2);
+            commandLine.AppendSwitch("--src-file-count");
+            commandLine.AppendSwitch(SourceFileCount);
+            commandLine.AppendSwitch("--ref-file-count");
+            commandLine.AppendSwitch(ReferenceFileCount);
+        }
+        else if (Idl == "Protobuf")
+        {
+            commandLine.AppendSwitch("--src-file-count");
```
ReeceHumphreys, line 83: Rather than add another flag for Protobuf file count (probably `--file-count`), I am just using the `--src-file-count` flag that we use for Slice.

ReeceHumphreys:
The CI is failing for reasons unrelated to this PR. Is related to this advisory https://github.com/advisories/GHSA-hh2w-p6rv-4g7w.

tools/Common/IceRpc.BuildTelemetry/TelemetryTask.cs
```diff
@@ -27,6 +27,21 @@ public class TelemetryTask : ToolTask
     /// </summary>
     public string? ContainsSlice1 { get; set; } = "";
 
+    /// <summary>
+    /// Gets or sets the if the compilation contained any Slice2 files.
```
pepone, line 31: That sounds a bit odd what about

```
Gets or sets a value indicating whether the compilation contained any Slice2 files.
```

tools/Common/IceRpc.BuildTelemetry/TelemetryTask.cs
```diff
@@ -56,6 +71,17 @@ protected override string GenerateCommandLineCommands()
         {
             commandLine.AppendSwitch("--contains-slice1");
             commandLine.AppendSwitch(ContainsSlice1);
+            commandLine.AppendSwitch("--contains-slice2");
+            commandLine.AppendSwitch(ContainsSlice2);
```
pepone, line 75: I think we can use 

```
 commandLine.AppendSwitchIfNotNull("--contains-slice2", ContainsSlice2);
```

tools/IceRpc.BuildTelemetry.Reporter/SystemHelpers.cs
```diff
@@ -0,0 +1,46 @@
...
+    internal static string GetArchitecture()
+    {
+        return RuntimeInformation.ProcessArchitecture.ToString();
+    }
+
+    internal static bool IsCi()
+    {
+        // If the CI environment variable is set, we are running in a CI environment.
+        return Environment.GetEnvironmentVariable("CI") != null;
+    }
```
pepone, line 45: Use arrow functions 
```
internal static string  GetArchitecture() =>  RuntimeInformation.ProcessArchitecture.ToString();
```

tools/IceRpc.BuildTelemetry.Reporter/SystemHelpers.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            return "macOS";
+        }
+        else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+        {
+            return "Linux";
+        }
+        else
+        {
+            return "Unknown OS";
```
pepone, line 30: Maybe return `RuntimeInformation.OSDescription` here, or maybe we should collect both always.

tools/IceRpc.BuildTelemetry.Reporter/SystemHelpers.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// <returns> The architecture of the current process.</returns>
+    internal static string GetArchitecture()
+    {
+        return RuntimeInformation.ProcessArchitecture.ToString();
+    }
+
+    internal static bool IsCi()
+    {
+        // If the CI environment variable is set, we are running in a CI environment.
+        return Environment.GetEnvironmentVariable("CI") != null;
```
pepone, line 44: DotNet reproducible builds have a more complete check for this, you might want to check

https://github.com/dotnet/reproducible-builds/blob/main/src/DotNet.ReproducibleBuilds/DotNet.ReproducibleBuilds.props

tools/Common/IceRpc.BuildTelemetry/TelemetryTask.cs
```diff
@@ -27,6 +27,21 @@ public class TelemetryTask : ToolTask
     /// </summary>
     public string? ContainsSlice1 { get; set; } = "";
 
+    /// <summary>
+    /// Gets or sets the if the compilation contained any Slice2 files.
+    /// </summary>
+    public string? ContainsSlice2 { get; set; } = "";
```
pepone, line 33: Why not a `bool` or `bool?` if not required?

tools/IceRpc.BuildTelemetry.Reporter/SystemHelpers.cs
ReeceHumphreys: The `RuntimeInformation.OSDescription` is too high of cardinality:

```
Darwin 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:14:59 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T8122
```

We would never really be able to use of filter or anything based on this as every single telemetry data sent would be different.

ReeceHumphreys: It contains a timestamp, if we were to store it we would have to strip some of that data out of it.

pepone: As far as I know, the timestamp is for the OS build, there are not that many OS builds.

tools/Common/IceRpc.BuildTelemetry/TelemetryTask.cs
```diff
@@ -23,9 +23,24 @@ public class TelemetryTask : ToolTask
     public string Idl { get; set; } = "";
 
     /// <summary>
-    /// Gets or sets the if the compilation contained any Slice1 files.
+    /// Gets or sets a value indicating whether the compilation contained any Slice1 files.
     /// </summary>
-    public string? ContainsSlice1 { get; set; } = "";
+    public bool ContainsSlice1 { get; set; } = false;
```
bernardnormier, line 28: You don't need to and should not set a property to its default value. I am surprised you didn't get a warning. Same for the next property.

tools/Common/IceRpc.BuildTelemetry/TelemetryTask.cs
```diff
@@ -23,9 +23,24 @@ public class TelemetryTask : ToolTask
...
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the compilation contained any Slice2 files.
+    /// </summary>
+    public bool ContainsSlice2 { get; set; } = false;
+
+    /// <summary>
+    /// Gets or sets the number of source files in the Slice compilation.
+    /// </summary>
+    public string? SourceFileCount { get; set; } = "";
```
bernardnormier, line 38: Here and for the next property, you should simply keep the default (null).

Also, it's pretty odd for a count to be a string.

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -47,7 +47,19 @@ public class SliceCCSharpTask : ToolTask
 
     /// <summary>Whether the Slice compilation contained any Slice1 files.</summary>
     [Output]
-    public string? ContainsSlice1 { get; set; }
+    public bool ContainsSlice1 { get; set; } = false;
```
bernardnormier, line 50: remove = false

tools/IceRpc.Protobuf.Tools/OutputHashTask.cs
```diff
@@ -21,6 +21,10 @@ public class OutputHashTask : Task
     [Output]
     public string? CompilationHash { get; set; }
 
+    /// <summary>The number of protobuf files in the compilation.</summary>
+    [Output]
+    public string? FileCount { get; set; }
```
bernardnormier, line 26: Should this be an int?

tools/IceRpc.BuildTelemetry.Reporter/ProtobufTelemetry.cs
```diff
@@ -10,12 +10,16 @@ public partial record struct ProtobufTelemetryData
     /// </summary>
     /// <param name="version">The version of the IceRPC.</param>
     /// <param name="compilationHash">The SHA-256 hash of the Slice files.</param>
-    public ProtobufTelemetryData(string version, string compilationHash)
+    /// <param name="fileCount">The number of source files in the Slice compilation.</param>
+    public ProtobufTelemetryData(string version, string compilationHash, int fileCount)
```
bernardnormier, line 14: The file name should match the struct name. Here: ProtobufTelemetryData.cs.

tools/IceRpc.BuildTelemetry.Reporter/SliceTelemetry.cs
```diff
@@ -11,13 +11,21 @@ public partial record struct SliceTelemetryData
     /// <param name="version">The version of the IceRPC.</param>
     /// <param name="compilationHash">The SHA-256 hash of the Slice files.</param>
     /// <param name="containsSlice1">Whether the Slice compilation contains Slice1 files.</param>
-    public SliceTelemetryData(string version, string compilationHash, bool containsSlice1)
+    /// <param name="containsSlice2">Whether the Slice compilation contains Slice2 files.</param>
+    /// <param name="sourceFileCount">The number of source files in the Slice compilation.</param>
+    /// <param name="referenceFileCount">The number of reference files in the Slice compilation.</param>
+    public SliceTelemetryData(string version, string compilationHash, bool containsSlice1, bool containsSlice2, int sourceFileCount, int referenceFileCount)
```
bernardnormier, line 17: File name should be SliceTelemetryData.cs.

bernardnormier: As far as I can tell, we only call this constructor once (in Program). Would it be simpler / cleaner to move this code to the construction point and not add this helper ctor?

bernardnormier (approved):

tools/IceRpc.BuildTelemetry.Reporter/SliceTelemetry.cs
ReeceHumphreys: Renamed the file, I think for now having it separate will make it easier for future refactors.

pepone (approved):

PR #4022, opened 2024-07-28
Use required fields in classes and exceptions

### LINKED ISSUE #4014: Improve mapping for classes and exceptions with required properties
The C# mapping for a simple class Person:
```slice
class Person {
    name: string
    age: int32
    tag(1) phoneNumber: string?
}
```

is currently:

```csharp
public partial class Person : SliceClass
{
    public string Name { get; set; }

    public int Age { get; set; }

    public string? PhoneNumber { get; set; }

    /// <summary>Constructs a new instance of <see cref="Person" />.</summary>
    public Person(
        string name,
        int age,
        string? phoneNumber)
    {
        this.Name = name;
        this.Age = age;
        this.PhoneNumber = phoneNumber;
    }

    /// <summary>Constructs a new instance of <see cref="Person" />.</summary>
    public Person(
        string name,
        int age)
    {
        this.Name = name;
        this.Age = age;
    }

    /// <summary>Constructs a new instance of <see cref="Person"/> for the Slice decoder.</summary>
    [global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
    public Person()
    {
        this.Name = default!;
    }

    // encode/decode logic
}
```

We provide:
 - a primary constructor to initialize all properties (including nullable properties and value types that have a reasonable default)
 - (optionally) another constructor to initialize all non-tagged properties
 - a hidden parameterless constructor used for decoding. This paramerless ctor assigns `default!` to all properties with a non-nullable reference type.

This structure is dictated in large part by the Slice1 decoding logic for classes and exceptions, where we decode the fields of the most derived slice first and then go up the inheritance chain.

## Proposal:

 - Always make the parameterless constructor public / non-hidden, with an empty body (no more `= default!`)
 - Add `required` to all properties with a non-nullable reference type
 - When there is at least one required property, add `[SetsRequiredMembers]` to the generated constructor(s) other than the parameterless constructor.

This gives users more flexibility while making the generated code cleaner.

bernardnormier:
This PR updates the C# mapping for Slice classes and exceptions to use 'required' fields for non-nullable reference types.

Fixes #4014.

### DIFF
--- a/tools/slicec-cs/src/generators/class_generator.rs
+++ b/tools/slicec-cs/src/generators/class_generator.rs
@@ -70,41 +70,39 @@ pub fn generate_class(class_def: &Class) -> CodeBlock {
 
     let constructor_summary = format!(r#"Constructs a new instance of <see cref="{class_name}" />."#);
 
-    // The primary constructor (may be parameterless)
-    class_builder.add_block(constructor(
-        &class_name,
-        access,
-        constructor_summary.clone(),
-        &namespace,
-        &fields,
-        &base_fields,
-    ));
+    if !class_def.all_fields().is_empty() {
+        // parameterless constructor.
+        // The constructor needs to be public for System.Activator.CreateInstance.
+        let mut parameterless_constructor = FunctionBuilder::new("public", "", &class_name, FunctionType::BlockBody);
+        parameterless_constructor.add_comment("summary", constructor_summary.clone());
+        class_builder.add_block(parameterless_constructor.build());
 
-    // Secondary constructor for all fields minus those with optional types.
-    // This constructor is only generated if necessary
-    if non_nullable_fields.len() + non_nullable_base_fields.len() < fields.len() + base_fields.len() {
+        // The primary constructor.
         class_builder.add_block(constructor(
             &class_name,
             access,
-            constructor_summary,
+            constructor_summary.clone(),
             &namespace,
-            &non_nullable_fields,
-            &non_nullable_base_fields,
+            &fields,
+            &base_fields,
         ));
-    }
 
-    // parameterless constructor for decoding, generated only if the preceding constructors are not parameterless
-    if non_nullable_fields.len() + non_nullable_base_fields.len() > 0 {
-        // The constructor needs to be public for System.Activator.CreateInstance.
-        let mut decode_constructor = FunctionBuilder::new("public", "", &class_name, FunctionType::BlockBody);
-        decode_constructor.add_never_editor_browsable_attribute();
-        decode_constructor.add_comment(
-            "summary",
-            format!(r#"Constructs a new instance of <see cref="{class_name}"/> for the Slice decoder."#),
-        );
-        decode_constructor.set_body(initialize_required_fields(&fields));
-        class_builder.add_block(decode_constructor.build());
+        // Secondary constructor for all fields minus those with optional types.
+        // This constructor is only generated if necessary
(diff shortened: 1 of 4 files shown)


tools/slicec-cs/src/generators/exception_generator.rs
```diff
@@ -49,23 +49,21 @@ pub fn generate_exception(exception_def: &Exception) -> CodeBlock {
...
-                    ),
+                    format!(r#"Constructs a new instance of <see cref="{}"/>."#, &exception_name),
                 )
-                .set_body(initialize_required_fields(fields))
                 .build(),
         );
+
+        exception_class_builder.add_block(one_shot_constructor(exception_def));
+
+        // TODO: generate secondary constructor like for classes.
```
bernardnormier, line 64: It's unfortunate that class_generator and exception_generator are so different, when the logic and generated code are almost identical. Fixing this inconsistency is outside the scope of this PR.

tools/slicec-cs/src/generators/class_generator.rs
```diff
@@ -70,41 +70,39 @@ pub fn generate_class(class_def: &Class) -> CodeBlock {
...
-        // The constructor needs to be public for System.Activator.CreateInstance.
-        let mut decode_constructor = FunctionBuilder::new("public", "", &class_name, FunctionType::BlockBody);
-        decode_constructor.add_never_editor_browsable_attribute();
-        decode_constructor.add_comment(
-            "summary",
-            format!(r#"Constructs a new instance of <see cref="{class_name}"/> for the Slice decoder."#),
-        );
-        decode_constructor.set_body(initialize_required_fields(&fields));
-        class_builder.add_block(decode_constructor.build());
+        // Secondary constructor for all fields minus those with optional types.
```
bernardnormier, line 90: The comment should use "nullable" for consistency with terminology in the code below.

tools/slicec-cs/src/generators/class_generator.rs
```diff
@@ -70,41 +70,39 @@ pub fn generate_class(class_def: &Class) -> CodeBlock {
...
+        // This constructor is only generated if necessary
+        if (non_nullable_fields.len() + non_nullable_base_fields.len() > 0)
```
bernardnormier, line 92: Is there a more elegant way to write this expression?

externl (approved):

tools/slicec-cs/src/generators/class_generator.rs
```diff
@@ -123,6 +121,10 @@ fn constructor(
 
     let mut builder = FunctionBuilder::new(access, "", escaped_name, FunctionType::BlockBody);
 
+    if fields.iter().any(|f| f.is_required()) || base_fields.iter().any(|f| f.is_required()) {
```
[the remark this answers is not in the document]
bernardnormier: This doesn't work, we don't pass all fields to constructor.

tools/slicec-cs/src/slicec_ext/member_ext.rs
```diff
@@ -45,6 +48,12 @@ impl FieldExt for Field {
             .any(|a| a.downcast::<CsReadonly>().is_some())
     }
 
+    fn is_required(&self) -> bool {
+        return !self.data_type().is_optional
+            && !self.data_type().is_value_type()
+            && !matches!(self.parent().concrete_entity(), Entities::Struct(_));
```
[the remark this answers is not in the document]
bernardnormier: See #4023.

PR #4024, opened 2024-07-29
Synchronize class_generator and exception_generator

bernardnormier:
Slice class and Slice exceptions are very similar, so it makes sense for class generator and exception generator to be very similar.

This PR synchronizes these 2 implementations, starting with the existing class_generator implementation.

### DIFF
--- a/tools/slicec-cs/src/generators/class_generator.rs
+++ b/tools/slicec-cs/src/generators/class_generator.rs
@@ -8,19 +8,19 @@ use crate::decoding::decode_fields;
 use crate::encoding::encode_fields;
 use crate::member_util::*;
 use crate::slicec_ext::*;
-use slicec::grammar::{Class, Encoding, Field};
+use slicec::grammar::{Class, Encoding, Field, Member};
 
 // Keep this file in sync with exception_generator.rs
 
-pub fn generate_class(def: &Class) -> CodeBlock {
-    let class_name = def.escape_identifier();
-    let namespace = def.namespace();
+pub fn generate_class(class_def: &Class) -> CodeBlock {
+    let class_name = class_def.escape_identifier();
+    let namespace = class_def.namespace();
 
-    let fields = def.fields();
-    let base_fields = def.base_class().map_or(vec![], Class::all_fields);
-    let all_fields = def.all_fields();
+    let fields = class_def.fields();
+    let base_fields = class_def.base_class().map_or(vec![], Class::all_fields);
+    let all_fields = class_def.all_fields();
 
-    let access = def.access_modifier();
+    let access = class_def.access_modifier();
 
     let mut non_nullable_fields = fields.clone();
     non_nullable_fields.retain(|f| !f.data_type.is_optional);
@@ -30,18 +30,18 @@ pub fn generate_class(def: &Class) -> CodeBlock {
 
     let mut builder = ContainerBuilder::new(&format!("{access} partial class"), &class_name);
 
-    if let Some(summary) = def.formatted_doc_comment_summary() {
+    if let Some(summary) = class_def.formatted_doc_comment_summary() {
         builder.add_comment("summary", summary);
     }
 
     builder
-        .add_generated_remark("class", def)
-        .add_comments(def.formatted_doc_comment_seealso())
-        .add_type_id_attribute(def)
-        .add_compact_type_id_attribute(def)
-        .add_obsolete_attribute(def);
+        .add_generated_remark("class", class_def)
+        .add_comments(class_def.formatted_doc_comment_seealso())
+        .add_type_id_attribute(class_def)
+        .add_compact_type_id_attribute(class_def)
+        .add_obsolete_attribute(class_def);
 
-    if let Some(base) = def.base_class() {
+    if let Some(base) = class_def.base_class() {
         builder.add_base(base.escape_scoped_identifier(&namespace));
     } else {
         builder.add_base("SliceClass".to_owned());
@@ -61,7 +61,7 @@ pub fn generate_class(def: &Class) -> CodeBlock {
         format!("private static readonly string SliceTypeId = typeof({class_name}).GetSliceTypeId()!;").into(),
     );
 
-    if def.compact_id.is_some() {
+    if class_def.compact_id.is_some() {
         builder.add_block(
                 format!(
                     "private static readonly int _compactSliceTypeId = typeof({class_name}).GetCompactSliceTypeId()!.Value;"
@@ -76,7 +76,7 @@ pub fn generate_class(def: &Class) -> CodeBlock {
         // parameterless constructor.
         // The constructor needs to be public for System.Activator.CreateInstance.
         let mut parameterless_constructor = FunctionBuilder::new("public", "", &class_name, FunctionType::BlockBody);
-        parameterless_constructor.add_comment("summary", constructor_summary.clone());
+        parameterless_constructor.add_comment("summary", &constructor_summary);
         builder.add_block(parameterless_constructor.build());
 
         // The primary constructor.
@@ -105,7 +105,7 @@ pub fn generate_class(def: &Class) -> CodeBlock {
     }
     // else, we rely on the default parameterless constructor.
 
-    builder.add_block(encode_and_decode(def));
+    builder.add_block(encode_and_decode(class_def));
 
     builder.build()
 }
@@ -118,8 +118,6 @@ fn constructor(
     fields: &[&Field],
     base_fields: &[&Field],
 ) -> CodeBlock {
-    let mut code = CodeBlock::default();
-
     let mut builder = FunctionBuilder::new(access, "", escaped_name, FunctionType::BlockBody);
 
     if fields.iter().any(|f| f.is_required()) || base_fields.iter().any(|f| f.is_required()) {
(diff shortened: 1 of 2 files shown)


tools/slicec-cs/src/generators/class_generator.rs
```diff
@@ -10,43 +10,45 @@ use crate::member_util::*;
...
 
-pub fn generate_class(class_def: &Class) -> CodeBlock {
-    let class_name = class_def.escape_identifier();
-    let namespace = class_def.namespace();
+// Keep this file in sync with exception_generator.rs
 
-    let fields = class_def.fields();
-    let base_fields = class_def.base_class().map_or(vec![], Class::all_fields);
-    let all_fields = class_def.all_fields();
+pub fn generate_class(def: &Class) -> CodeBlock {
```
bernardnormier, line 15: Renamed a few parameters/variables in this file to make it more neutral.

tools/slicec-cs/src/generators/exception_generator.rs
```diff
@@ -1,41 +1,53 @@
 // Copyright (c) ZeroC, Inc.
 
-use crate::builders::{AttributeBuilder, Builder, CommentBuilder, ContainerBuilder, FunctionBuilder, FunctionType};
+use crate::builders::{
```
bernardnormier, line 3: Now a copy-paste of class_generator, with a few updates like:
 - Class => Exception
 - no compact type ID.

tools/slicec-cs/src/generators/exception_generator.rs
```diff
@@ -44,118 +56,144 @@ pub fn generate_exception(exception_def: &Exception) -> CodeBlock {
...
+        builder.add_block(constructor(
+            &class_name,
+            access,
+            constructor_summary.clone(),
+            &namespace,
+            &fields,
+            &base_fields,
+        ));
+
+        // Secondary constructor for all fields minus those that are nullable.
```
bernardnormier, line 82: Prior to this PR, exceptions didn't have this secondary constructor.

tools/slicec-cs/src/generators/exception_generator.rs
```diff
@@ -44,118 +56,144 @@ pub fn generate_exception(exception_def: &Exception) -> CodeBlock {
...
+    let encode_class = FunctionBuilder::new("protected override", "void", "EncodeCore", FunctionType::BlockBody)
+        .add_parameter("ref SliceEncoder", "encoder", None, None)
+        .set_body({
+            let mut code = CodeBlock::default();
+
+            code.writeln(
+                &FunctionCallBuilder::new("encoder.StartSlice")
+                    .add_argument("SliceTypeId")
+                    .build(),
+            );
```
InsertCreativityHere, line 160: I know you're just copy-pasting... but I think `Exception` did it better than `Class` here.
This whole thing could just be:
```
code.writeln("encoder.StartSlice(SliceTypeId);");
```
`FunctionCallBuilder` is really only useful when you're dynamically adding different numbers or types or arguments or something. In this case, there's no point to using it, instead of just writing the string literals directly.

tools/slicec-cs/src/generators/exception_generator.rs
```diff
@@ -44,118 +56,144 @@ pub fn generate_exception(exception_def: &Exception) -> CodeBlock {
...
-        exception_class_builder.add_block(one_shot_constructor(exception_def));
-
-        // TODO: generate secondary constructor like for classes.
+    let constructor_summary = format!(r#"Constructs a new instance of <see cref="{class_name}" />."#);
+
+    if !all_fields.is_empty() {
+        // parameterless constructor.
+        // The constructor needs to be public for System.Activator.CreateInstance.
+        let mut parameterless_constructor = FunctionBuilder::new("public", "", &class_name, FunctionType::BlockBody);
+        parameterless_constructor.add_comment("summary", constructor_summary.clone());
```
InsertCreativityHere, line 69: The clone should be unnecessary here, you can probably just pass a borrow:
```suggestion
        parameterless_constructor.add_comment("summary", &constructor_summary);
```
This should be updated in the class generator as well.

tools/slicec-cs/src/generators/exception_generator.rs
```diff
@@ -44,118 +56,144 @@ pub fn generate_exception(exception_def: &Exception) -> CodeBlock {
...
+
+    builder.add_base_parameters(&base_fields.iter().map(|m| m.parameter_name()).collect::<Vec<String>>());
 
-    for field in &all_fields {
-        ctor_builder.add_parameter(
-            &field.data_type().field_type_string(namespace),
-            field.parameter_name().as_str(),
+    for field in base_fields.iter().chain(fields.iter()) {
+        builder.add_parameter(
+            &field.data_type.field_type_string(namespace),
```
InsertCreativityHere, line 125: Here, I think `Exception` had it right again, this should be:
```suggestion
            &field.data_type().field_type_string(namespace),
```
ie. like calling a property-get, vs accessing the field directly.

Does it really matter right now? Honestly, no.
In the future, I'd like to a pass over `slicec` to fix the visibility of things, since right now, pretty much everything is public, even when it maybe shouldn't be. And when I (hopefully) do that, these kinds of accesses won't be allowed.

tools/slicec-cs/src/generators/exception_generator.rs
```diff
@@ -44,118 +56,144 @@ pub fn generate_exception(exception_def: &Exception) -> CodeBlock {
...
 
-    // ctor impl
-    let mut ctor_body = CodeBlock::default();
-    for field in exception_def.fields() {
-        let field_name = field.field_name();
-        let parameter_name = field.parameter_name();
+    builder.set_body({
+        let mut code = CodeBlock::default();
+        for field in fields {
+            writeln!(code, "this.{} = {};", field.field_name(), field.parameter_name(),);
```
InsertCreativityHere, line 135: Looks like a bug introduced by rustfmt. Obviously there shouldn't be a trailing comma here.
```suggestion
            writeln!(code, "this.{} = {};", field.field_name(), field.parameter_name());
```

tools/slicec-cs/src/generators/exception_generator.rs
```diff
@@ -44,118 +56,144 @@ pub fn generate_exception(exception_def: &Exception) -> CodeBlock {
...
+        }
+        code
+    });
 
-        writeln!(ctor_body, "this.{field_name} = {parameter_name};");
-    }
+    code.add_block(builder.build());
+
+    code
```
InsertCreativityHere, line 142: The `code` variable is being shadowed in this section of code.

Thankfully, there's an easy fix. The outer `code` is completely useless...
At the top of this function we create `let mut code = CodeBlock::default();`.
And all we ever do is add a block here at the bottom, and return it.

But this is pointless. `builder.build()` already returns a `CodeBlock`. We should just return `builder.build()`, and delete this wrapper `CodeBlock`. Note that the original `Exception` code didn't do this.

tools/slicec-cs/src/generators/exception_generator.rs
```diff
@@ -44,118 +56,144 @@ pub fn generate_exception(exception_def: &Exception) -> CodeBlock {
...
+fn encode_and_decode(def: &Exception) -> CodeBlock {
+    let mut code = CodeBlock::default();
+
+    let fields = def.fields();
+    let has_base = def.base_exception().is_some();
+
+    let encode_class = FunctionBuilder::new("protected override", "void", "EncodeCore", FunctionType::BlockBody)
+        .add_parameter("ref SliceEncoder", "encoder", None, None)
+        .set_body({
+            let mut code = CodeBlock::default();
```
InsertCreativityHere, line 154: More shadowed `code` variables...
Here the outer code is actually useful, so we shouldn't delete it.
But, we should absolutely change one of their names to avoid shadowing here.

tools/slicec-cs/src/generators/exception_generator.rs
InsertCreativityHere: Note that this is what we do in the `decoding` code right below this:
```
code.writeln("decoder.StartSlice();");
```
No `FunctionCallBuilder`s necessary.

InsertCreativityHere (approved):
The synchronization looks good... but...
There are some parts of the code which I think `Exception` did better than `Class`, and I think instead of overwriting this with the lamer `Class` logic, we should preserve it.

pepone (approved):

tools/slicec-cs/src/generators/exception_generator.rs
bernardnormier: I let you fix this in a follow-up PR. Don't forget to update both class_generator and exception_generator.

tools/slicec-cs/src/generators/exception_generator.rs
bernardnormier: I prefer to keep the same code as in class_generator.

tools/slicec-cs/src/generators/class_generator.rs
bernardnormier: Fixed.

InsertCreativityHere (approved):
Still looks good!

Once you've merged I'll open a small PR to fix the `code` name then.

PR #4026, opened 2024-07-30
Add 'required' to non-nullable struct fields

### LINKED ISSUE #4023: Generate required fields for mapped structs?
As a follow-up to #4022, we should consider generating required for struct fields too.

See:
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/required-members

bernardnormier:
Fixes #4023.

### DIFF
--- a/tests/IceRpc.Locator.Tests/LocationResolverTests.cs
+++ b/tests/IceRpc.Locator.Tests/LocationResolverTests.cs
@@ -23,7 +23,7 @@ public class LocationResolverTests
                 TimeSpan.FromSeconds(refreshThreshold),
                 ttl: Timeout.InfiniteTimeSpan);
 
-        (ServiceAddress? resolved, bool _) = await resolver.ResolveAsync(new Location(), refreshCache: true, default);
+        (ServiceAddress? resolved, bool _) = await resolver.ResolveAsync(default, refreshCache: true, default);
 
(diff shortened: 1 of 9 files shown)


tests/IceRpc.Locator.Tests/LocationResolverTests.cs
```diff
@@ -23,7 +23,7 @@ public class LocationResolverTests
                 TimeSpan.FromSeconds(refreshThreshold),
                 ttl: Timeout.InfiniteTimeSpan);
 
-        (ServiceAddress? resolved, bool _) = await resolver.ResolveAsync(new Location(), refreshCache: true, default);
+        (ServiceAddress? resolved, bool _) = await resolver.ResolveAsync(default, refreshCache: true, default);
```
pepone, line 26: How does this work for the required property `Value`?

pepone (approved):

tests/IceRpc.Locator.Tests/LocationResolverTests.cs
bernardnormier: For structs, "default" bypasses the "required" check, while "new T()" doesn't.

See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/required-members#struct-defaults

externl (approved):

PR #4027, opened 2024-08-25
Add new connection cache test

bernardnormier:
This PR adds a new connection cache test to verify that the connection cache correctly selects an (the) alt server when the primary server is busy because it reached MaxConnections.

This test includes two test cases - for ice and icerpc.

A concern however is:
 - the ConnectionCache implementation relies on specific error codes to "move next": 
   https://github.com/icerpc/icerpc-csharp/blob/b1f65d8d66b671c9948c7bfb04c3881243ba6a96/src/IceRpc/ConnectionCache.cs#L494

 - the MaxConnections test suggests we can get IceRpcError.IceRpcError with the ice protocol
 https://github.com/icerpc/icerpc-csharp/blob/b1f65d8d66b671c9948c7bfb04c3881243ba6a96/tests/IceRpc.Tests/ServerTests.cs#L84

(this MaxConnections test uses the default transport (tcp), while all the connection cache tests use coloc)

### DIFF
--- a/tests/IceRpc.Tests/ConnectionCacheTests.cs
+++ b/tests/IceRpc.Tests/ConnectionCacheTests.cs
@@ -108,6 +108,80 @@ public sealed class ConnectionCacheTests
         await cache.ShutdownAsync();
     }
 
+    /// <summary>Verifies that the connection cache uses the alt-server when the primary server is busy.</summary>
+    [TestCase("icerpc")]
+    [TestCase("ice")]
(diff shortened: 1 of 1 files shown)


pepone:
> the MaxConnections test suggests we can get IceRpcError.IceRpcError with the ice protocol

See https://github.com/icerpc/icerpc-csharp/pull/2462

pepone (approved):

externl (approved):

PR #4033, opened 2024-08-26
Rename property to enable/disable build telemetry

### LINKED ISSUE #4015: Rename `DisableBuildTelemetry` to  `IceRpcBuildTelemetry`
I would simplify and just use `IceRpcBuildTelemetry`. If you look at boolean properties here: https://learn.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-properties?view=vs-2022, most of them don't have a Disable prefix.

Also, IceRpcBuildTelemetry makes it clearer what it's about.

_Originally posted by @bernardnormier in https://github.com/icerpc/icerpc-csharp/pull/4006#discussion_r1664208085_

Opening an issue for further discussion on this as there does not seem to be a consensus on if it should be renamed.

bernardnormier:
This PR renames two build telemetry properties:

 - DisableBuildTelemetry => IceRpcBuildTelemetry
 - TelemetryScriptPath => IceRpcTelemetryScriptPath

Fixes #4015

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -5,7 +5,7 @@
     <!-- Include PDB in the built .nupkg -->
     <AllowedOutputExtensionsInPackageBuildOutputFolder
     >$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
-    <DisableBuildTelemetry>true</DisableBuildTelemetry>
+    <IceRpcBuildTelemetry>false</IceRpcBuildTelemetry>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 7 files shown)


externl (approved):

InsertCreativityHere (approved):

InsertCreativityHere:
I still believe in my previous position, but I also think that the difference in understandability is near-negligible. So: "looks good!"
> Personally, I prefer the current name. Having it be a 'negative' phrase It makes it very clear that this is opt-out, and that by default, we're sending build telemetry. Whereas naming it something positive (IMO) implicitly makes it seem more like it's opt-in. At least without having to go read some defaults somewhere.

pepone (approved):

PR #4039, opened 2024-09-05
Add docfx workflow

externl:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/checkout@v4
+      - name: Build
+        uses: ./.github/actions/build
+
+      - name: Build Docs
+        run: |
+            dotnet tool install --global docfx
+            ./build.sh --doc
+
+      - name: Upload Docs
```
pepone, line 25: Uploading this isn't very useful.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,37 @@
...
+  docfx:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+
+        uses: actions/checkout@v4
+      - name: Build
+        uses: ./.github/actions/build
+
```
pepone, line 19:
```suggestion
      - name: Checkout
        uses: actions/checkout@v4

      - name: Build
        uses: ./.github/actions/build

```

pepone (approved):

PR #4047, opened 2024-09-10
Upgrade NuGet dependencies

pepone:
This PR upgrades NuGet dependencies.

### DIFF
--- a/.github/actions/test/action.yaml
+++ b/.github/actions/test/action.yaml
@@ -7,5 +7,5 @@ runs:
     run: cargo test --manifest-path tools/slicec-cs/Cargo.toml
     shell: bash
   - name: 🧪 Test
-    run: dotnet test --no-build --verbosity normal --blame-hang-timeout 10m
+    run: dotnet test --no-build --verbosity normal --blame-hang-timeout 8m
     shell: bash
--- a/.github/actions/test/action.yaml
+++ b/.github/actions/test/action.yaml
@@ -7,5 +7,5 @@ runs:
     run: cargo test --manifest-path tools/slicec-cs/Cargo.toml
     shell: bash
   - name: 🧪 Test
-    run: dotnet test --no-build --verbosity normal
+    run: dotnet test --no-build --verbosity normal --blame-hang-timeout 10m
     shell: bash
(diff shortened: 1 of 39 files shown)


.github/actions/test/action.yaml
```diff
@@ -7,5 +7,5 @@ runs:
     run: cargo test --manifest-path tools/slicec-cs/Cargo.toml
     shell: bash
   - name: 🧪 Test
-    run: dotnet test --no-build --verbosity normal
+    run: dotnet test --no-build --verbosity normal --blame-hang-timeout 8m
```
pepone, line 10: `--blame-hang-timeout` is used to terminate the tests and dump core after the timeout, the NUnit timeout attribute is now deprecated.

tests/IceRpc.Tests.Common/AssertTraceListener.cs
```diff
@@ -12,18 +12,18 @@ public class AssertTraceListener : DefaultTraceListener
...
     {
         if (detailMessage is null || detailMessage.Length == 0)
         {
-            Assert.Fail(message);
+            Assert.Fail(message ?? "");
         }
         else
         {
-            Assert.Fail(message, detailMessage);
+            Assert.Fail($"{message}\n{detailMessage}");
```
pepone, line 26: The overload with two parameters was removed.

tests/IceRpc.Compressor.Tests/AssemblyInfo.cs
```diff
@@ -1,3 +0,0 @@
-// Copyright (c) ZeroC, Inc.
-
-[assembly: NUnit.Framework.Timeout(8000)]
```
bernardnormier, line 3: When I run "dotnet test", I want a test that takes more than 8s to fail. This shouldn't be a CI-only behavior.

This also seems unrelated to this dependency upgrade.

externl (approved):

bernardnormier (approved):
Should we also update Dunet?

tests/IceRpc.Compressor.Tests/AssemblyInfo.cs
pepone: See also https://docs.nunit.org/articles/nunit/writing-tests/attributes/cancelafter.html

Timeout is deprecated, and the new CancelAfter requires to upgrade all test to deal with an extra cancelation token.

When not using CI we can just attach a debugger to the hanging test

PR #4051, opened 2024-09-11
Improve parsing of JSON messages in Slice tools

pepone:
This PR reworks the parsing of JSON messages in Slice tools. All messages from slicec are supposed to be JSON, and JsonException are not handled explicitly. If parsing a message throws JsonException MSBuild will log the exception.


externl (approved):

InsertCreativityHere (approved):
Looks good!

Issue #3976, opened 2024-03-28
Template testing CI action doesn't work on Windows


pepone:
We have a CI action to test templates in https://github.com/icerpc/icerpc-csharp/blob/99a76dcf00f241723cf8256cc9559e08ca96715f/.github/actions/test-templates/action.yaml#L1

This action uses bash and cannot be used by the release workflow on Windows machines. We should either fix this action, or add a similar action for windows and update the release workflow to use it on Windows.

pepone:
We should also fix the action to test with the different template parameters, currently it only tests the default parameters.

PR #4053, opened 2024-09-11
Run CI workflow on macOS, Windows, and Linux

externl:
We don't run template testing on Windows. See #3976


bernardnormier (approved):

.github/actions/test/action.yaml
```diff
@@ -7,5 +7,5 @@ runs:
     run: cargo test --manifest-path tools/slicec-cs/Cargo.toml
     shell: bash
   - name: 🧪 Test
-    run: dotnet test --no-build --verbosity normal --blame-hang-timeout 8m
+    run:  dotnet test --no-build --verbosity normal --blame-hang-timeout 10s tests/IceRpc.Tests/IceRpc.Tests.csproj
```
InsertCreativityHere, line 10: Is this extra space intentional? It's fine if you want it for some reason. Just double-checking.
```suggestion
    run: dotnet test --no-build --verbosity normal --blame-hang-timeout 10s tests/IceRpc.Tests/IceRpc.Tests.csproj
```

externl: Not intentional, just a bad copy and paste.

.github/actions/test/action.yaml
```diff
@@ -7,5 +7,13 @@ runs:
     run: cargo test --manifest-path tools/slicec-cs/Cargo.toml
     shell: bash
   - name: 🧪 Test
-    run: dotnet test --no-build --verbosity normal --blame-hang-timeout 8m
+    # See https://github.com/microsoft/vstest/issues/2080#issuecomment-539879345
+    run: dotnet test --no-build --verbosity normal --blame-hang-timeout 10s --logger:"console;noprogress=true"
```
InsertCreativityHere, line 11: How long does a single test usually take? 10s feels on the short side to me, but to be fair, I don't really have a scale for how quickly we expect these tests to complete...

InsertCreativityHere (approved):
Looks good to me!

.github/actions/test/action.yaml
externl: It's for a single unit test. So it should be much faster than 10s

.github/workflows/ci.yml
```diff
@@ -6,10 +6,19 @@ on:
...
 jobs:
   build_and_test:
-    runs-on: ubuntu-latest
-    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 20
```
pepone, line 21: Should we revert this to 10 minutes?

externl:
In the interest of getting this PR merged I've removed macOS from the matrix.

pepone (approved):

Issue #4049, opened 2024-09-10
Build telemetry failure makes build fail


pepone:
If the build telemetry task fails it causes the whole build to fail, this doesn't look desirable.

We set `ContinueOnError` but doesn't seems to have the desired effect 

https://github.com/icerpc/icerpc-csharp/blob/9edecc0ccaddee829828f2da64ce150b692d340f/tools/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets#L84

I guess this can be because warning as errors in our builds...

We can also consider providing some logging for this task, currently it doesn't emit any logs

```
Restore complete (0.2s)
    Determining projects to restore...
    All projects are up-to-date for restore.
You are using a preview version of .NET. See: https://aka.ms/dotnet-support-policy
  Interop.Tests succeeded with 1 warning(s) (0.5s) → tests/Interop.Tests/bin/Debug/net8.0/Interop.Tests.dll
    /Users/bernard/.nuget/packages/icerpc.slice.tools/0.4.0-preview1/build/IceRpc.Slice.Tools.targets(76,5): warning MSB6006: "dotnet" exited with code 1.
```

externl:
Do we know why the telemetry task is failing?

pepone:
In this case the working directory was not correctly set with the new package, see #4048

bernardnormier:
It would be good to provide some diagnostic on warning/error. Otherwise, you have no idea what's wrong and that you need to (for example) disable build telemetry to get a clean build.

bernardnormier:
Fixed by #4054.

InsertCreativityHere:
This isn't completely solved yet.
If the JSON parsing of the telemetry data (from `slicec-cs`) fails, then it causes the whole build to fail.

Specifically, I intentionally added an extra comma in the middle of the JSON output in `main.rs`.
So, I don't think this is a realistic problem, and no, we shouldn't guard against "we broke slicec-cs".

All I'm saying is that some runtime errors in the build telemetry still cause the build to fail.
And if this random JSON parsing error causes that, I'm sure that there are still others too.

```
PS D:\Code\Workspace\icerpc-csharp> .\build.cmd

//...
// building the Rust crate
//...

dotnet build -nr:false  --configuration Debug
Restore complete (1.4s)
You are using a preview version of .NET. See: https://aka.ms/dotnet-support-policy
[...]

PR #4054, opened 2024-09-11
Report errors from build telemetry

pepone:
This PR updates the telemetry reporter to report errors, that are then treated as warning and allow the build to continue:

For example, with a disconnected network:

```
C:\Users\jose\source\repos\icerpc-csharp\examples\slice\Greeter\Server>dotnet build
Restore complete (0.3s)
You are using a preview version of .NET. See: https://aka.ms/dotnet-support-policy
  Server succeeded with 1 warning(s) (1.4s) → bin\Any CPU\Debug\net8.0\Server.dll
    C:\Users\jose\.nuget\packages\icerpc.slice.tools\0.4.0-preview1\build\IceRpc.Slice.Tools.targets(76,5): warning MSB3073: The command "dotnet IceRpc.BuildTelemetry.Reporter.dll --idl slice --hash=0e7817f6cb2092323a7ee81a42bd5301177621e819f3178a04c4522cd8fae240 --contains-slice1=False --contains-slice2=True --src-file-count=1 --ref-file-count=11" exited with code -532462766.

Build succeeded with 1 warning(s) in 2.1s
```

or with detailed logging enabled `/v:d`

```
C:\Users\jose\source\repos\icerpc-csharp\examples\slice\Greeter\Server>dotnet build /v:d
Restore complete (0.3s)
    Determining projects to restore...
    All projects are up-to-date for restore.
You are using a preview version of .NET. See: https://aka.ms/dotnet-support-policy
  Server succeeded with 1 warning(s) (1.4s) → bin\Any CPU\Debug\net8.0\Server.dll
    Unhandled exception. IceRpc.IceRpcException: An IceRpc call failed with error 'IceRpcError'.
     ---> System.Net.Sockets.SocketException (11001): No such host is known.
       at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
       at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource.GetResult(Int16 token)
       at System.Net.Sockets.Socket.<ConnectAsync>g__WaitForConnectWithCancellation|285_0(AwaitableSocketAsyncEventArgs saea, ValueTask connectTask, CancellationToken cancellationToken)
       at IceRpc.Transports.Tcp.Internal.TcpClientConnection.ConnectAsyncCore(CancellationToken cancellationToken) in C:\Users\jose\source\repos\icerpc-csharp\src\IceRpc\Transports\Tcp\Internal\TcpConnection.cs:line 332
       --- End of inner exception stack trace ---
       at IceRpc.Transports.Tcp.Internal.TcpClientConnection.ConnectAsyncCore(CancellationToken cancellationToken) in C:\Users\jose\source\repos\icerpc-csharp\src\IceRpc\Transports\Tcp\Internal\TcpConnection.cs:line 358
       at IceRpc.Transports.Slic.Internal.SlicConnection.<>c__DisplayClass56_0.<<ConnectAsync>g__PerformConnectAsync|0>d.MoveNext() in C:\Users\jose\source\repos\icerpc-csharp\src\IceRpc\Transports\Slic\Internal\SlicConnection.cs:line 149
    --- End of stack trace from previous location ---
       at IceRpc.Transports.Slic.Internal.SlicConnection.<DisposeAsync>g__PerformDisposeAsync|59_0() in C:\Users\jose\source\repos\icerpc-csharp\src\IceRpc\Transports\Slic\Internal\SlicConnection.cs:line 478
       at Program.<Main>$(String[] args) in C:\Users\jose\source\repos\icerpc-csharp\tools\IceRpc.BuildTelemetry.Reporter\Program.cs:line 93
       at Program.<Main>(String[] args)
    C:\Users\jose\.nuget\packages\icerpc.slice.tools\0.4.0-preview1\build\IceRpc.Slice.Tools.targets(76,5): warning MSB3073: The command "dotnet IceRpc.BuildTelemetry.Reporter.dll --idl slice --hash=0e7817f6cb2092323a7ee81a42bd5301177621e819f3178a04c4522cd8fae240 --contains-slice1=False --contains-slice2=True --src-file-count=1 --ref-file-count=11" exited with code -532462766.

Build succeeded with 1 warning(s) in 2.2s
```


tools/IceRpc.BuildTelemetry.Reporter/Program.cs
```diff
@@ -2,98 +2,92 @@
...
-        _ => throw new ArgumentException($"Unknown IDL: {idl}") // This should never happen
-    };
-
-    // Upload the telemetry to the server.
-    await reporter.UploadAsync(buildTelemetry, cancellationToken: cts.Token);
-
-    // Shutdown the connection.
-    await connection.ShutdownAsync(cts.Token);
-}
-catch
```
pepone, line 96: Just removed this catch block that was ignoring all exceptions

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -75,13 +75,11 @@
     </OutputHashTask>
 
     <!-- Run build telemetry -->
-    <BuildTelemetryTask
+    <Exec
```
pepone, line 78: We can use the standard [Exec](https://learn.microsoft.com/en-us/visualstudio/msbuild/exec-task?view=vs-2022) task to call the reporter, no need for a custom task.

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -75,13 +75,11 @@
...
       WorkingDirectory="$(IceRpcBuildTelemetryScriptPath)"
-      CompilationHash="$(Hash)"
-      SourceFileCount="$(FileCount)"
+      Command="dotnet IceRpc.BuildTelemetry.Reporter.dll --idl protobuf --hash=$(Hash) --src-file-count=$(FileCount)"
       Condition="'$(IceRpcBuildTelemetry)' != 'false'"
-      ContinueOnError="true"
-    />
+      ContinueOnError="true" />
```
pepone, line 82: Alternatively we can use `IgnoreExitCode` parameter from Exec task. The difference is that errors would be reported as regular messages instead of warnings.

externl (approved):
Looks good to me

tools/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets
```diff
@@ -73,16 +73,11 @@
...
     <!-- Run build telemetry -->
-    <BuildTelemetryTask
+    <Exec
       WorkingDirectory="$(IceRpcBuildTelemetryScriptPath)"
-      CompilationHash="$(Hash)"
-      ContainsSlice1="$(ContainsSlice1)"
-      ContainsSlice2="$(ContainsSlice2)"
-      SourceFileCount="$(SourceFileCount)"
-      ReferenceFileCount="$(ReferenceFileCount)"
+      Command="dotnet IceRpc.BuildTelemetry.Reporter.dll --idl slice --hash=$(Hash) --contains-slice1=$(ContainsSlice1) --contains-slice2=$(ContainsSlice2) --src-file-count=$(SourceFileCount) --ref-file-count=$(ReferenceFileCount)"
```
InsertCreativityHere, line 78: Is there a reason why we use `--switch=value` for everything except for the `idl`?
(there's no equal sign for that one).

I guess it doesn't make a difference whether you put an equals or not, but the old `BuildTelemetryTask` didn't place any `=` between switches and their values.

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -75,13 +75,11 @@
     </OutputHashTask>
 
     <!-- Run build telemetry -->
-    <BuildTelemetryTask
+    <Exec
       WorkingDirectory="$(IceRpcBuildTelemetryScriptPath)"
-      CompilationHash="$(Hash)"
-      SourceFileCount="$(FileCount)"
+      Command="dotnet IceRpc.BuildTelemetry.Reporter.dll --idl protobuf --hash=$(Hash) --src-file-count=$(FileCount)"
```
InsertCreativityHere, line 80: Same here with no equals after the `idl` switch.

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
InsertCreativityHere: I think that `IgnoreExitCode` would be preferable.

Maybe our collective sentiment has changed, but I think that a telemetry failure should have no impact on whether your build succeeds or not; and emitting warnings would still cause a failure if warnings are being treated as errors (something that seems to be at least slightly common).

InsertCreativityHere (approved):
Looks _much_ simpler!

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
bernardnormier: Presumably ContinueOnError=true means we keep going on error.

Keeping in mind this process (slice compilation) is under our control here.

tools/IceRpc.BuildTelemetry.Reporter/Program.cs
```diff
@@ -2,98 +2,92 @@
...
 using System.Reflection;
 
 var timeout = TimeSpan.FromSeconds(3); // The timeout for the complete telemetry upload process.
 
 const string uri = "icerpc://telemetry.icerpc.dev"; // The URI of the server.
 
-try
+// Create a cancellation token source to cancel the telemetry upload if it
+// takes too long.
+using var cts = new CancellationTokenSource(timeout);
```
bernardnormier, line 15: I find this code oddly written. We use timeout only here, no need for `var timeout`.

tools/IceRpc.BuildTelemetry.Reporter/Program.cs
```diff
@@ -2,98 +2,92 @@
...
+
+// Create a client connection
+var clientAuthenticationOptions = new SslClientAuthenticationOptions();
```
bernardnormier, line 18: No need for this extra variable.

tools/IceRpc.BuildTelemetry.Reporter/Program.cs
```diff
@@ -2,98 +2,92 @@
...
+await using var connection = new ClientConnection(new Uri(uri), clientAuthenticationOptions);
+
+// Create a reporter proxy with this client connection.
+var reporter = new ReporterProxy(connection);
+
+// Parse the IDL
+string? idl = args
```
bernardnormier, line 25: From the expression below, doesn't sound like in can be null.

tools/IceRpc.BuildTelemetry.Reporter/Program.cs
```diff
@@ -2,98 +2,92 @@
...
+    .Skip(1)
+    .FirstOrDefault(), out int sourceFileCountResult) ? sourceFileCountResult : 0;
+
+// Parse the ref-file-count argument
+int referenceFileCount = int.TryParse(args
+    .SkipWhile(arg => arg != "--ref-file-count")
+    .Skip(1)
+    .FirstOrDefault(), out int referenceFileCountResult) ? referenceFileCountResult : 0;
+
+BuildTelemetry buildTelemetry = idl.ToLower() switch
```
bernardnormier, line 65: Given how trivial this program is, I would prefer to have separate versions for IceRpc.Protobuf.Tools and IceRpc.Slice.Tools. For a follow-up PR.

Then no need for an "idl" argument to the program.

Also note that most (maybe all) arguments are IDL-specific. --contains-slice1/--contains-slice2/--ref-file-count is Slice-only, which is not clear with the current setup.

tools/IceRpc.BuildTelemetry.Reporter/Program.cs
```diff
@@ -2,98 +2,92 @@
 
 using IceRpc;
 using IceRpc.BuildTelemetry;
+using Microsoft.Extensions.Logging;
```
bernardnormier, line 5: Where do we use Logging?

bernardnormier (approved):

tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
InsertCreativityHere: Maybe I misunderstood, but #4049 made it sound like `ContinueOnError` was insufficient.
Even though, "keep going on error" is definitely the impression you get from it's name.

tools/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets
pepone: Removed `=`

tools/IceRpc.BuildTelemetry.Reporter/Program.cs
pepone: Agree, I just removed the try/catch would rewrite this in a follow up PR where we split the program in protobuf/slice versions.

Issue #4055, opened 2024-09-11
Rename `SourceFileCount` in `IceRpc.Protobuf.Tools`


externl:
I think that you're correct and "FileCount" is what we should call this,
but it looks like in `IceRpc.Protobuf.Tools`, we call this `SourceFileCount` and use `--src-file-count`:
https://github.com/icerpc/icerpc-csharp/blob/50f97c80cd47367112e037876fc41b3b08d9e200/tools/IceRpc.Protobuf.Tools/BuildTelemetryTask.cs#L19-L22
So it's a little inconsistent. Maybe after this PR, we should fix the name in the `Protobuf.Tools` to match this one,
for consistency's sake.

_Originally posted by @InsertCreativityHere in https://github.com/icerpc/telemetry/pull/4#discussion_r1755369923_

pepone:
Fixed in #4060

PR #4060, opened 2024-09-12
Split build telemetry reporter

pepone:
This PR splits the telemetry reporter in Protobuf and Slice versions.


tools/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.props
```diff
@@ -27,7 +27,6 @@
       <PropertyGroup>
         <IceRpcProtocPath>$(MSBuildThisFileDirectory)../tools/</IceRpcProtocPath>
         <IceRpcProtocGenPath>$(MSBuildThisFileDirectory)../tools/</IceRpcProtocGenPath>
-        <IceRpcBuildTelemetryScriptPath>$(MSBuildThisFileDirectory)../tools/</IceRpcBuildTelemetryScriptPath>
```
pepone, line 30: We need the other Path variables because they are different when building with sources and when building with the NuGet package. Build telemetry is always disabled for source builds, and no need to use variable for the path.

tools/IceRpc.ProtobufBuildTelemetry.Reporter/slice/Reporter.slice
pepone, line 1: Would be nice to move this file to a shared location, can we move it to the top-level Slice?

tools/IceRpc.SliceBuildTelemetry.Reporter/Program.cs
```diff
@@ -0,0 +1,102 @@
...
+        // Create a reporter proxy with this client connection.
+        var reporter = new ReporterProxy(connection);
+
+        // Determine the IceRPC version using the assembly version.
+        var assembly = Assembly.GetAssembly(typeof(SliceTelemetryData));
+        string version = assembly!.GetName().Version!.ToString();
+
+        var sliceTelemetryData = new SliceTelemetryData(
+            version,
+            RuntimeInformation.OSDescription,
```
pepone, line 56: `OSDescription` is more verbose than the previous `SystemHelpers.GetOperatingSystem()`. I think it makes more sense to normalize this data in the server.

externl:
What's the benefit of splitting the report into two reporters. Just seems like more to maintain to me.

externl (approved):

pepone:
> What's the benefit of splitting the report into two reporters. 

@bernardnormier suggested we do that because the options are IDL dependent, and the program is too simple to care about reusing.

tools/IceRpc.ProtobufBuildTelemetry.Reporter/slice/Reporter.slice
bernardnormier: yes, we should create a tools/slice directory.

tools/IceRpc.ProtobufBuildTelemetry.Reporter/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        const string uri = "icerpc://telemetry.icerpc.dev";
+        using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(3));
+        await using var connection = new ClientConnection(new Uri(uri), new SslClientAuthenticationOptions());
+
+        // Create a reporter proxy with this client connection.
+        var reporter = new ReporterProxy(connection);
+
+        // Determine the IceRPC version using the assembly version.
+        var assembly = Assembly.GetAssembly(typeof(SliceTelemetryData));
```
bernardnormier, line 36: Should use ProtobufTelemetryData since this is the Protobuf tool.

bernardnormier (approved):
Looks good.

bernardnormier:
Does this PR fix #4055?

pepone:
> Does this PR fix https://github.com/icerpc/icerpc-csharp/issues/4055?

Yes, this fixes #4055

pepone:
Moved the Slice definitions to `tools/slice/BuildTelemetry.slice` updated the doc comments. I also updated the struct fields to use camelCase.

PR #4063, opened 2024-09-13
Fix project template testing

### LINKED ISSUE #3976: Template testing CI action doesn't work on Windows

pepone:
Fixes #3976

### DIFF
--- a/.github/actions/test-templates/action.yaml
+++ b/.github/actions/test-templates/action.yaml
@@ -20,7 +20,7 @@ runs:
 
   - name: 🧪 Test IceRpc + Protobuf Client Template .NET 9
     run: |
-      dotnet new icerpc-protobuf-client -o templates-test/tcp/ProtobufClient -F net9.0
+      dotnet new icerpc-protobuf-client -o templates-test/net9.0/tcp/ProtobufClient -F net9.0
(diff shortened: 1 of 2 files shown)


.github/actions/test-templates/action.yaml
```diff
@@ -3,39 +3,163 @@ name: Test Templates
 runs:
   using: "composite"
   steps:
+
+  # Protobuf Client template
+
   - name: 🧪 Test IceRpc + Protobuf Client Template
-    run: mkdir -p templates-test/ProtobufClient && cd templates-test/ProtobufClient && dotnet new icerpc-protobuf-client && dotnet build /p:TreatWarningsAsErrors=true
+    run: |
+      dotnet new icerpc-protobuf-client -o templates-test/tcp/ProtobufClient
```
pepone, line 11: using `dotnet new -o` to create the directories solves the issue about the action not working on Windows

externl (approved):

bernardnormier (approved):

PR #4064, opened 2024-09-13
Add Rpc metadata for Slice items

pepone:
This PR adds a Rpc metadata for Slice items that corresponds to the slicec-cs rpc parameter.


tools/IceRpc.Slice.Tools/IceRpc.Slice.Tools.props
```diff
@@ -49,6 +49,7 @@
   <ItemDefinitionGroup>
     <SliceFile>
       <OutputDir>$(MSBuildProjectDirectory)/generated</OutputDir>
+      <Rpc>icerpc</Rpc>
```
bernardnormier, line 52: This ItemDefinitionsGroup is where we specify the default values?

tools/IceRpc.Slice.Tools/README.md
```diff
@@ -84,12 +84,13 @@ The [ZeroC.Slice][zeroc-slice] NuGet package follows this pattern and provides d
...
+| Name              | Default   | Description                                                                                                                         |
+|-------------------|-----------|-------------------------------------------------------------------------------------------------------------------------------------|
+| AdditionalOptions |           | Specifies additional options to pass to [`slicec-cs`][slicec-cs].                                                                   |
+| OutputDir         | generated | Sets the output directory for the generated code. This metadata corresponds to the `--output-dir` option of `slicec-cs`.            |
+| Pack              | `false`   | Specifies whether or not to include the items (Slice files) in the NuGet package.                                                   |
+| PackagePath       | slice     | Sets the target path in the NuGet package. Used only when Pack is `true`.                                                           |
+| Rpc               | icerpc    | Specifies the RPC framework to generate code for `icerpc` or `none`. This metadata corresponds to the `--rpc` option of `slicec-cs` |
```
bernardnormier, line 93: This does not sound correct. What about:
Specifies the RPC framework to generate code for. Possible values: `icerpc`, `none`. This metadata corresponds...

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -64,6 +70,21 @@ public class SliceCCSharpTask : ToolTask
 
     private readonly JsonSerializerOptions _jsonSerializeOptions = new() { PropertyNameCaseInsensitive = true };
 
+    protected override bool ValidateParameters()
+    {
+        var validRpcValues = new[] { "icerpc", "none" };
+
+        if (!validRpcValues.Contains(Rpc, StringComparer.OrdinalIgnoreCase))
+        {
+            Log.LogError($"Invalid Rpc value '{Rpc}'. Valid options are 'icerpc' and 'none'.");
```
bernardnormier, line 79: Should be Valid values are...

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
```diff
@@ -26,6 +26,12 @@ public class SliceCCSharpTask : ToolTask
     /// to <c>-R</c> slicec-cs compiler option.</summary>
     public string[] References { get; set; } = Array.Empty<string>();
 
+    /// <summary>
+    /// The RPC provider to generate code for, corresponds to the <c>--rpc</c> slicec-cs compiler option.
+    /// </summary>
+    [Required]
+    public string Rpc { get; set; } = "";
```
bernardnormier, line 33: The `[Required]` attribute does not suppress the need to have a default value for this property?

tools/IceRpc.Slice.Tools/SliceFile.FileProperties.xaml
```diff
@@ -85,6 +85,15 @@
         </StringProperty.DataSource>
     </StringProperty>
 
+    <EnumProperty Name="Rpc"
+                  DisplayName="Rpc Provider"
```
bernardnormier, line 89: I would use RPC Provider for the display name

tools/IceRpc.Slice.Tools/SliceFile.ProjectProperties.xaml
```diff
@@ -33,6 +33,15 @@
         </StringProperty.DataSource>
     </StringProperty>
 
+    <EnumProperty Name="Rpc"
+                  DisplayName="Rpc Provider"
```
bernardnormier, line 37: Same here.

Also, do we need in both SliceFile.FileProperties.xaml and SliceFile.ProjectProperties.xaml?

bernardnormier (approved):

tools/IceRpc.Slice.Tools/SliceFile.ProjectProperties.xaml
pepone: One is per file, and the other is per project. This is to allow setting things from Visual Studio UI

tools/IceRpc.Slice.Tools/IceRpc.Slice.Tools.props
pepone: Yes these are the defaults for all SliceFile items

tools/IceRpc.Slice.Tools/SliceCCSharpTask.cs
pepone: No, this only cause an error if the task caller doesn't provide the parameter.

Without the value I get:

```
warning CS8618: Non-nullable property 'Rpc' must contain a non-null value when exiting constructor
```

PR #4070, opened 2024-09-16
Use SSL with Protobuf and Slice templates

pepone:
This PR updates templates to setup SSL with tcp transport.


bernardnormier (approved):
Looks good.

PR #4080, opened 2024-09-16
Update release workflow to build the reporter before packaging tools

### LINKED ISSUE #4079: Fix tools packaging
The IceRpc.Slice.Tools and IceRpc.Protobuf.Tools package for version 0.4.0 do not include the build telemetry reporter.

pepone:
Fix #4079

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,8 +4,8 @@ on:
   push:
     # Run on release and preview tags
     tags:
-      - "v[0-9]+.[0-9]+.[0-9]+"
-      - "v[0-9]+.[0-9]+.[0-9]+-preview[0-9]+"
+      - "v[0-9]+\\.[0-9]+\\.[0-9]+(\\.[0-9]+)?"
+      - "v[0-9]+\\.[0-9]+\\.[0-9]+(\\.[0-9]+)?-preview[0-9]+"
 
 jobs:
   build-compiler:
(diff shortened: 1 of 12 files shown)


.github/workflows/release.yml
```diff
@@ -80,11 +80,16 @@ jobs:
           # Copy the compiler to the expected location to avoid rebuilding it when creating the NuGet packages
           New-Item -ItemType Directory -Path "$env:GITHUB_WORKSPACE\tools\slicec-cs\target\release"
           Copy-Item "slicec-cs-x86_64-pc-windows-msvc\slicec-cs.exe" -Destination "$env:GITHUB_WORKSPACE\tools\slicec-cs\target\release"
+      - name: Build IceRPC Tools
+        working-directory: tools
+        run: dotnet build --configuration Release
+        env:
+          SLICEC_CS_STAGING_PATH: ${{ github.workspace }}\tools\slicec-cs\staging
+      - name: Build IceRPC
+        run: dotnet build --configuration Release --output .
```
pepone, line 89: We need to build IceRpc solution before creating the tool packages, otherwise the reporters would be missing in the tools packages.

build.sh
```diff
@@ -89,7 +89,7 @@ publish()
     global_packages=$(dotnet nuget locals -l global-packages)
     global_packages=${global_packages/global-packages: /""}
     run_command rm "-rf" "$global_packages/zeroc.slice/$version" "$global_packages/icerpc/$version" "$global_packages"/icerpc.*/"$version"
-    run_command dotnet "nuget" "push" "tools/**/$dotnet_config/*.$version.nupkg" "--source" "$global_packages"
+    run_command dotnet "nuget" "push" "tools/**/$dotnet_config/*.$version*.nupkg" "--source" "$global_packages"
```
bernardnormier, line 92: Is `*` missing from the line below, in case we update some src packages, like for example IceRpc.Templates?

bernardnormier (approved):

build.sh
pepone: Windows script doesn't handle this versions well, need to fix it.

PR #4085, opened 2024-09-18
Use github.ref_name instead of hardcoded branch name

externl:
This PR updates the coverage and doc workflows to use `github.ref_name` instead of a hard coded branch name for the s3 path. 

This way there's nothing to update in this part of the workflows when we create a new branch of off `main`.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -36,9 +36,9 @@ jobs:
           license: ${{ secrets.REPORT_GENERATOR_LICENSE }}
 
       - name: Sync Coverage Report to S3
-        run: aws s3 sync ./coveragereport s3://${{ secrets.AWS_S3_BUCKET }}/csharp/main/coverage --delete
+        run: aws s3 sync ./coveragereport s3://${{ secrets.AWS_S3_BUCKET }}/csharp/${{ github.ref_name }}/coverage --delete
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yml
```diff
@@ -36,9 +36,9 @@ jobs:
...
 
       - name: Sync Coverage Report to S3
-        run: aws s3 sync ./coveragereport s3://${{ secrets.AWS_S3_BUCKET }}/csharp/main/coverage --delete
+        run: aws s3 sync ./coveragereport s3://${{ secrets.AWS_S3_BUCKET }}/csharp/${{ github.ref_name }}/coverage --delete
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION: us-east-1
-        if: github.ref == 'refs/heads/main' && github.event_name != 'pull_request'
+        if: github.event_name != 'pull_request'
```
externl, line 44: we only run the workflow on push or prs on the branch specified at the top of the file

pepone: What is the `event_name` that triggers the build when you push to an open PR?

pepone: Why no `if: github.event_name == 'push'` it makes clear we are only interested in push events.

pepone (approved):

.github/workflows/coverage.yml
externl: The original version I found on stack overflow. I think the original intention for using `!= 'pull_request'` is that you could trigger the workflow manually if enabled.

externl: > What is the `event_name` that triggers the build when you push to an open PR?

Not sure, I suspect still `pull_request`

PR #4114, opened 2024-11-15
Update version to 0.4.2-preview1

bernardnormier:
on 0.4.x branch


bernardnormier:
Once merged, we need to do the corresponding update in icerpc-ice-csharp-interop/0.4.x.

pepone (approved):

PR #4126, opened 2025-11-19
Tagged parameters no longer get a default value

### LINKED ISSUE #4121: Do not default trailing parameters
We currently give default values (usually `null`) to trailing parameters with optional types in mapped C# proxy methods.

See https://github.com/icerpc/icerpc-csharp/blob/f109b98c1cb69755ccca7ddf3cc247982e7afb8c/tools/slicec-cs/src/builders.rs#L349

This undocumented behavior is not desirable: having an optional type should not mean "forgettable" or defaulted. See https://github.com/zeroc-ice/ice/issues/3970.

Note:
we should most likely split this "add_operation_parameters" function in two, one for proxies and another one for services. Currently:
 - TypeContext::OutgoingParam => proxy code
 - TypeContext::IncomingParam => dispatch code

and I see numerous match blocks / if blocks in this function based on the "context".

bernardnormier:
Tagged parameters are no longer mapped to C# parameters with default values.

Fixes #4121.

### DIFF
--- a/tools/slicec-cs/src/builders.rs
+++ b/tools/slicec-cs/src/builders.rs
@@ -315,14 +315,10 @@ impl FunctionBuilder {
 
     pub fn add_operation_parameters(&mut self, operation: &Operation, context: TypeContext) -> &mut Self {
         let parameters = operation.parameters();
-
-        for (_, parameter) in parameters.iter().enumerate() {
-            let parameter_type = parameter.cs_type_string(&operation.namespace(), context);
-            let parameter_name = parameter.parameter_name();
-
+        for parameter in parameters.iter() {
             self.add_parameter(
-                &parameter_type,
-                &parameter_name,
+                &parameter.cs_type_string(&operation.namespace(), context),
+                &parameter.parameter_name(),
                 None,
                 parameter.formatted_param_doc_comment(),
             );


tests/IceRpc.Slice.Tests/OperationTests.cs
```diff
@@ -592,44 +592,6 @@ public void Skip_tagged_parameters()
         Assert.That(() => proxy.OpAsync(10, 1, 12), Throws.Nothing);
     }
 
-    /// <summary>Verifies that tagged parameters has a default value that is equivalent to a non set tagged parameter.
```
bernardnormier, line 595: Removed tests for the mapping deleted by this PR.

tools/slicec-cs/src/builders.rs
```diff
@@ -317,45 +316,14 @@ impl FunctionBuilder {
...
-        let trailing_optional_parameters_index = match parameters
-            .iter()
-            .rposition(|p| !p.data_type.is_optional || p.is_streamed)
-        {
-            Some(last_index) => last_index + 1,
-            None => 0,
-        };
-
-        for (index, parameter) in parameters.iter().enumerate() {
+        for (_, parameter) in parameters.iter().enumerate() {
```
InsertCreativityHere: My Rust is a little Rusty, but I'm pretty sure you can just do this, since `Vec` implements `IntoIterator`:
```suggestion
        for parameter in operation.parameters() {
```

tools/slicec-cs/src/builders.rs
```diff
@@ -316,46 +315,11 @@ impl FunctionBuilder {
...
-                        {
-                            Some("default")
-                        }
-                        _ => Some("null"),
-                    }
-                } else {
-                    None
-                };
-
+        for parameter in parameters.iter() {
```
InsertCreativityHere, line 318: `iter` is superfluous here. We only needed it before to be able to call `enumerate`.


```suggestion
        let parameters = operation.parameters();
        for parameter in &parameters {
```

PR #4138, opened 2025-11-21
Improve build telemetry error handling

### LINKED ISSUE #4073: Improve error reporting for build telemetry reporter
Currently, when the build telemetry reporter code throws an exception, the program fails with exit code 1 and .NET prints a stack trace for the exception. The MSBuild task "continues on error" so it doesn't affect the build (it just shows this task failing with exit code 1).

We should improve this error handling as follows:
 - for expected errors, catch the exception/error, print/log a short message and return 0
   For example, when the build telemetry server is unreachable, or the reporting timed out
 - for unexpected errors, catch the exception/error, print/log the exception message + inner exception name/message (if any), but not the full stack trace, and return 1

pepone:
Fix #4073

### DIFF
--- a/tools/IceRpc.ProtobufBuildTelemetry.Reporter/Program.cs
+++ b/tools/IceRpc.ProtobufBuildTelemetry.Reporter/Program.cs
@@ -61,7 +61,7 @@ rootCommand.SetAction(
         }
         catch (OperationCanceledException)
         {
-            Console.Error.WriteLine($"Failed to report Protobuf build telemetry: OperationCanceledException");
+            Console.Error.WriteLine($"Failed to report Protobuf build telemetry: operation canceled");
             return 0;
         }
         catch (IceRpcException ex) when (ex.IceRpcError is IceRpcError.InvocationCanceled or IceRpcError.ServerUnreachable)
--- a/tools/IceRpc.SliceBuildTelemetry.Reporter/Program.cs
+++ b/tools/IceRpc.SliceBuildTelemetry.Reporter/Program.cs
@@ -86,7 +86,7 @@ rootCommand.SetAction(
         }
         catch (OperationCanceledException)
         {
-            Console.Error.WriteLine($"Failed to report Slice build telemetry: OperationCanceledException");
+            Console.Error.WriteLine($"Failed to report Slice build telemetry: operation canceled");
             return 0;
         }
         catch (IceRpcException ex) when (ex.IceRpcError is IceRpcError.InvocationCanceled or IceRpcError.ServerUnreachable)


tools/IceRpc.SliceBuildTelemetry.Reporter/Program.cs
```diff
@@ -76,11 +76,34 @@
...
+        {
+            // Upload the telemetry to the server.
+            await reporter.UploadAsync(new BuildTelemetry.Slice(sliceTelemetryData), cancellationToken: cts.Token);
+
+            // Shutdown the connection.
+            await connection.ShutdownAsync(cts.Token);
+        }
+        catch (OperationCanceledException ex) when (ex.CancellationToken == cts.Token)
+        {
+            Console.Error.WriteLine($"Failed to report slice build telemetry: InvocationCanceled");
```
bernardnormier: I would capitalize both Slice and Protobuf. Also, this InvocationCanceled is confusing: the exception is OperatonCanceledException, and there is a separate IceRpcError.InvocationCanceled.

Furthermore, is the when (ex.CancellationToken == cts.Token) necessary here? That's the only cancellation token we use, so if we got an OperationCanceledException where ex.CancellationToken != cts.Token, it would arguably represent a bug in UpLoadAsync or ShutdownAsync.

pepone: > Furthermore, is the when (ex.CancellationToken == cts.Token) necessary here?

It is not clear what cancellation token is provided to the action, but you are right the distinction probably doesn't matter here

tools/IceRpc.ProtobufBuildTelemetry.Reporter/Program.cs
```diff
@@ -51,11 +51,34 @@
...
+            await reporter.UploadAsync(new BuildTelemetry.Protobuf(protobufTelemetryData), cancellationToken: cts.Token);
 
-        // Shutdown the connection.
-        await connection.ShutdownAsync(cts.Token);
+            // Shutdown the connection.
+            await connection.ShutdownAsync(cts.Token);
+        }
+        catch (OperationCanceledException)
+        {
+            Console.Error.WriteLine($"Failed to report Protobuf build telemetry: OperationCanceledException");
```
externl: I don't see a reason to put the exception type in the message. `OperationCanceledException` sounds like a bug. Really user likely hit control-c.

```suggestion
            Console.Error.WriteLine($"Failed to report Protobuf build telemetry: operation cancelled");
```

tools/IceRpc.SliceBuildTelemetry.Reporter/Program.cs
```diff
@@ -76,11 +76,34 @@
...
+        {
+            // Upload the telemetry to the server.
+            await reporter.UploadAsync(new BuildTelemetry.Slice(sliceTelemetryData), cancellationToken: cts.Token);
+
+            // Shutdown the connection.
+            await connection.ShutdownAsync(cts.Token);
+        }
+        catch (OperationCanceledException)
+        {
+            Console.Error.WriteLine($"Failed to report Slice build telemetry: OperationCanceledException");
```
externl: ditto.

tools/IceRpc.ProtobufBuildTelemetry.Reporter/Program.cs
bernardnormier: `canceled` of course

PR #4139, opened 2025-11-21
Use azure code signing to sign Windows slice compilers

### LINKED ISSUE #4120: Update release workflow to use Azure Trusted Signing
We should update the release workflow to use the Azure Trusted Signing to sign .NET assemblies.

pepone:
Fix #4120

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,17 +36,13 @@ jobs:
         with:
           cargo-build-args: --release
           target: ${{ matrix.target }}
-      - name: Sign Compiler Artifact
+      - name: Sign slicec-cs compiler on Windows
         if: runner.os == 'Windows'
         uses: azure/trusted-signing-action@v0
-        env:
-          AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
-          AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
-          AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
         with:
-          azure-tenant-id: ${{ env.AZURE_TENANT_ID }}
-          azure-client-id: ${{ env.AZURE_CLIENT_ID }}
-          azure-client-secret: ${{ env.AZURE_CLIENT_SECRET }}
+          azure-tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          azure-client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          azure-client-secret: ${{ secrets.AZURE_CLIENT_SECRET }}
           endpoint: https://eus.codesigning.azure.net/
           trusted-signing-account-name: zeroc
           certificate-profile-name: zeroc-ice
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -36,6 +36,24 @@ jobs:
...
+          AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
+          AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+          AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
+        with:
+          azure-tenant-id: ${{ env.AZURE_TENANT_ID }}
+          azure-client-id: ${{ env.AZURE_CLIENT_ID }}
+          azure-client-secret: ${{ env.AZURE_CLIENT_SECRET }}
+          endpoint: https://eus.codesigning.azure.net/
+          trusted-signing-account-name: zeroc
+          certificate-profile-name: zeroc-ice
```
bernardnormier, line 48: Should we create a separate profile for slicec?

.github/workflows/release.yml
```diff
@@ -36,6 +36,24 @@ jobs:
         with:
           cargo-build-args: --release
           target: ${{ matrix.target }}
+      - name: Sign Compiler Artifact
```
bernardnormier: Can we use a clearer name? E.g. Sign slicec compiler on Windows

build/IceRpc.Src.props
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <SignAssembly>true</SignAssembly>
```
bernardnormier, line 4: I am confused. We no longer sign the assemblies?

.github/workflows/release.yml
pepone: I tried but seems our subscription allows only one.

build/IceRpc.Src.props
pepone: We were not signing them, there is no sign key.

See also https://learn.microsoft.com/en-gb/dotnet/standard/assembly/strong-named

> For .NET Core and .NET 5+, strong-named assemblies do not provide material benefits. The runtime never validates the strong-name signature, nor does it use the strong-name for assembly binding.

My understanding is that strong name assemblies are not useful for .NET, this has some use case for .NET Framework that we no longer support.

.github/workflows/release.yml
```diff
@@ -36,6 +36,24 @@ jobs:
...
+        if: runner.os == 'Windows'
+        uses: azure/trusted-signing-action@v0
+        env:
+          AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
+          AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+          AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
+        with:
+          azure-tenant-id: ${{ env.AZURE_TENANT_ID }}
+          azure-client-id: ${{ env.AZURE_CLIENT_ID }}
+          azure-client-secret: ${{ env.AZURE_CLIENT_SECRET }}
```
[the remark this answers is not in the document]
pepone: fixed as suggested.

.github/workflows/release.yml
bernardnormier: We can use zeroc-ice, that's fine.

build/IceRpc.Src.props
bernardnormier: So this `<SignAssembly>true</SignAssembly>` was doing nothing?

What about authenticode-signing? We're not authenticode-signing the assemblies, only slicec.exe?

pepone: only slicec-cs.exe now

pepone: We should also look into sign the NuGet packages, https://learn.microsoft.com/en-us/nuget/create-packages/sign-a-package

But the integration with Azure Trusted Signing is a work in progress https://github.com/NuGet/NuGetGallery/issues/10027

PR #4140, opened 2025-11-21
Small fix to Slic tests

bernardnormier:


tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
```diff
@@ -583,9 +584,20 @@ public async Task Connection_with_idle_timeout_is_not_aborted_when_idle([Values]
...
+        }
+
         await sut.Client.CloseAsync(MultiplexedConnectionCloseError.NoError, default);
-        Assert.That(async () => await acceptStreamTask, Throws.InstanceOf<IceRpcException>());
+
+        // Successful completion.
+        Assert.That(
+            async () => await acceptStreamTask,
+            Throws.InstanceOf<IceRpcException>()
+                .With.Property("IceRpcError").EqualTo(IceRpcError.ConnectionClosedByPeer));
```
bernardnormier, line 600: We should definitely avoid a plain:
```cs
 Assert.That(async () => await acceptStreamTask, Throws.InstanceOf<IceRpcException>());
```

for a successful completion.

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
```diff
@@ -555,10 +555,11 @@ public async Task Connection_with_no_idle_timeout_is_not_aborted_when_idle()
         Assert.That(async () => await acceptStreamTask, Throws.InstanceOf<IceRpcException>());
     }
 
-    /// <summary>Verifies that setting the idle timeout doesn't abort the connection if it's idle.</summary>
+    /// <summary>Verifies that setting the idle timeout doesn't abort the connection if there is no application-level
+    /// activity.</summary>
     [Test]
     [NonParallelizable]
-    public async Task Connection_with_idle_timeout_is_not_aborted_when_idle([Values] bool serverIdleTimeout)
+    public async Task Connection_with_idle_timeout_is_not_aborted_when_inactive([Values] bool serverIdleTimeout)
```
pepone, line 562: One puzzling bit of this tests is, the name and the comment say it verifies the connection, but the implementation checks an accept stream task is not completed?

pepone (approved):

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
bernardnormier: It's a Slic multiplexed connection:

https://github.com/icerpc/icerpc-csharp/blob/main/src/IceRpc/Transports/IMultiplexedConnection.cs

You see that's the connection is live or dead based on your ability to create new streams (client) and accept new streams from the peer (server).

PR #4141, opened 2025-11-21
Switch to macos-26 runners

bernardnormier:
This PR "upgrades" CI to use macos-26 instead of macos-15 on macos.

The main difficulty was (apparently) a bug in macos 26.0 where TCP does not honor the listen backlog parameter. The work-around is to skip the affected tests on macos 26.0. Works fine on macos 26.1.

### DIFF
--- a/tests/IceRpc.Tests/Transports/Tcp/TcpTransportTests.cs
+++ b/tests/IceRpc.Tests/Transports/Tcp/TcpTransportTests.cs
@@ -161,7 +161,7 @@ public class TcpTransportTests
     [Test]
     public async Task Configure_server_connection_listen_backlog()
     {
-        // TODO: the listen backlog does not work on macos 26.0, so we skip this test.
+        // The listen backlog does not work on macos 26.0, so we skip this test. Works on macos 26.1.
         if (OperatingSystem.IsMacOS() &&
             Environment.OSVersion.Version.Major == 26 &&
             Environment.OSVersion.Version.Minor == 0)
--- a/tests/IceRpc.Conformance.Tests/Transports/DuplexConnectionConformanceTests.cs
+++ b/tests/IceRpc.Conformance.Tests/Transports/DuplexConnectionConformanceTests.cs
@@ -18,8 +18,9 @@ public abstract class DuplexConnectionConformanceTests
     [Test]
     public async Task Connect_cancellation()
     {
-        // TODO: the listen backlog does not work on macos 26.0, so we skip this test.
-        if (OperatingSystem.IsMacOS() &&
+        // The listen backlog does not work on macos 26.0 with TCP, so we skip this test. Works on macos 26.1.
+        if (UsesListenBacklog &&
+            OperatingSystem.IsMacOS() &&
             Environment.OSVersion.Version.Major == 26 &&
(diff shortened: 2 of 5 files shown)


.github/actions/build/action.yaml
```diff
@@ -13,26 +13,8 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: 📦 NuGet Cache
-      uses: actions/cache@v4
-      with:
-        path: ~/.nuget/packages
-        key: ${{ runner.os }}-nuget-${{ hashFiles('**/packages.lock.json') }}
```
bernardnormier, line 20: CI on macos fails every time on hashFiles. No idea what's wrong, so I removed these steps.

tests/IceRpc.Deadline.Tests/IceRpc.Deadline.Tests.csproj
```diff
@@ -10,7 +10,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.0.1" />
```
bernardnormier, line 13: This is unrelated to the macos-26 support. Added by @externl.

.github/workflows/ci.yml
```diff
@@ -16,10 +16,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-24.04, macos-15, windows-2025]
+        os: [ubuntu-24.04, macos-26, windows-2025]
     runs-on: ${{ matrix.os }}
     timeout-minutes: 15
     steps:
+      - name: Dump System Information
```
bernardnormier, line 23: Not that all interesting.

.github/actions/build/action.yaml
bernardnormier: Also fails with my other PR that still uses macos-15 runners.

pepone (approved):

externl (approved):

tests/IceRpc.Tests/Transports/Tcp/TcpTransportTests.cs
```diff
@@ -192,10 +203,11 @@ public async Task Configure_server_connection_listen_backlog()
         }
 
         // Assert
-        Assert.That(connections, Has.Count.GreaterThanOrEqualTo(18));
+        Assert.That(connections, Has.Count.GreaterThanOrEqualTo(backlog));
+
         // The OS may allow a few more connections than specified by the ListenBacklog. This test ensures that
         // Socket.Listen was called with the ListenBacklog value set in TcpServerTransportOptions.
-        Assert.That(connections, Has.Count.LessThanOrEqualTo(50));
+        Assert.That(connections, Has.Count.LessThan(hardLimit));
```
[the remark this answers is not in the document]
bernardnormier: If we reach the hard limit in the while loop, we want to fail, not succeed.

PR #4143, opened 2025-11-21
Create tests report using dotnet-trx

pepone:
This adds dotnet-trx to the test action, see https://github.com/devlooped/dotnet-trx

### DIFF
--- a/.github/actions/test/action.yaml
+++ b/.github/actions/test/action.yaml
@@ -13,9 +13,15 @@ runs:
       dotnet test --no-build --verbosity normal --blame-hang-timeout 10s --logger "trx" --results-directory TestResults
     shell: bash
 
-  - name: Create test report
-    run: dnx dotnet-trx --yes -- -p TestResults
+  - name: Install dotnet-trx tool
+    run: dotnet tool install -g dotnet-trx
     shell: bash
+    if: always()
+
+  - name: Create test report
+    run: trx --path TestResults
+    shell: bash
+    if: always()
 
   - name: Upload blame hang dumps
     uses: actions/upload-artifact@v4


.github/actions/test/action.yaml
```diff
@@ -13,6 +13,10 @@ runs:
       dotnet test --no-build --verbosity normal --blame-hang-timeout 10s --logger "trx" --results-directory TestResults
     shell: bash
 
+  - name: Create test report
+    run: dnx dotnet-trx --yes -- -p TestResults
```
externl: What's the output file type?

.github/actions/test/action.yaml
```diff
@@ -13,6 +13,10 @@ runs:
       dotnet test --no-build --verbosity normal --blame-hang-timeout 10s --logger "trx" --results-directory TestResults
     shell: bash
```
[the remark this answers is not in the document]
pepone: dnx is part of .NET 10 sdk used for the build.

bernardnormier (approved):

.github/actions/test/action.yaml
```diff
@@ -13,6 +13,16 @@ runs:
...
 
+  - name: Install dotnet-trx tool
+    run: dotnet tool install -g dotnet-trx
+    shell: bash
+    if: always()
+
+  - name: Create test report
+    run: trx --path TestResults
```
externl, line 22: What file type does this generate that we upload?

pepone: It doesn't upload any files, it emits a report to standard output. It also attach the report as a comment to the summary.

externl (approved):

PR #4145, opened 2025-11-21
Update LangVersion to 14.0

bernardnormier:


src/ZeroC.Slice/SliceDecoderExtensions.cs
```diff
@@ -123,7 +123,7 @@ public static T[] DecodeSequence<T>(this ref SliceDecoder decoder, Action<T>? ch
             int elementSize = Unsafe.SizeOf<T>();
             decoder.IncreaseCollectionAllocation(count * elementSize);
             var value = new T[count];
-            Span<byte> destination = MemoryMarshal.Cast<T, byte>(value);
+            Span<byte> destination = MemoryMarshal.Cast<T, byte>(value.AsSpan());
```
bernardnormier, line 126: Without this AsSpan, we end up with another Cast overload that returns a ReadOnlySpan.

PR #4146, opened 2025-11-21
Change type of _mutex field to Lock

bernardnormier:
This PR changes the type of _mutex fields (and occasionally mutex variables) to (System.Threading.)Lock. 

Requires .NET 9.0 and C#13 or better.

### DIFF
--- a/examples/protobuf/Thermostat/Device/ThermoBot.cs
+++ b/examples/protobuf/Thermostat/Device/ThermoBot.cs
@@ -60,6 +60,7 @@ internal partial class ThermoBot : IThermoControlService
     {
         while (true)
         {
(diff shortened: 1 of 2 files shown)


examples/protobuf/Thermostat/Device/ThermoBot.cs
pepone, line 64: having the await inside the lock seems indeed wrong.

bernardnormier: There is no await inside the lock. It was a yield. Fixed.

pepone (approved):

externl (approved):

Issue #4151, opened 2025-11-23
Reenable Nuget and Cargo caching


bernardnormier:
I removed it as part of #4141 because it was failing on macos.

bernardnormier:
Fixed by #4152.

PR #4152, opened 2025-11-24
Enable cache

pepone:
This PR re-enables cargo and NuGet package caches see #4151. 

It doesn't improve the build times, not clear that it is worth to have this cache.

### DIFF
--- a/.github/actions/build/action.yaml
+++ b/.github/actions/build/action.yaml
@@ -20,7 +20,7 @@ runs:
           ~/.cargo/registry/index/
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
-          target/
+          tools/slicec-cs/target/
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
         restore-keys: |
           ${{ runner.os }}-cargo-


src/Directory.Build.props
```diff
@@ -6,6 +6,8 @@
     <AllowedOutputExtensionsInPackageBuildOutputFolder
     >$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
     <IceRpcBuildTelemetry>false</IceRpcBuildTelemetry>
+    <!-- Enable package lock file to ensure consistent package versions across builds -->
+    <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
```
pepone, line 10: See https://devblogs.microsoft.com/dotnet/enable-repeatable-package-restores-using-a-lock-file/

this ensure that package.lock.json files are generated, the cache action create a hash with these files to know if dependencies change.

bernardnormier:
I don't understand how this works:

- where are the lock files for the tests coming from since there is no Directory.Build.props for the tests?
- why did the Cargo caching fail before on macos since Cargo.lock was already checked in?

pepone:
The failures seems related to this bug https://github.com/actions/runner-images/issues/13341

Seems they rollback the affected images.

pepone:
> where are the lock files for the tests coming from since there is no Directory.Build.props for the tests?

Fixed I generated them locally, but forget to update the test props.

> why did the Cargo caching fail before on macos since Cargo.lock was already checked in?

This is a bug in the action, as mentioned above.

PR #4154, opened 2025-11-24
Update slicec-cs version

pepone:


tools/slicec-cs/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "slicec-cs"
-version = "0.4.0"
+version = "0.5.0-preview1"
```
externl, line 3: I think the canonical semver version is supposed to be "0.5.0-preview.1"

pepone: I used this to match the one we have for nuget. Maybe update the NuGet one too.

externl: I think Cargo cares more about semver.

pepone: the current version is a valid semver version

bernardnormier (approved):

tools/slicec-cs/Cargo.toml
externl: I think you're right, however it seems to be convention to do `.x`.

https://semver.org/spec/v2.0.0-rc.1.html#spec-item-10

> A pre-release version MAY be denoted by appending a dash and a series of dot separated identifiers immediately following the patch version.

pepone: I will update it everywhere in a follow up PR, as we use it in a few places.

externl: `preview1` trips up semver version checkers. See `https://jubianchi.github.io/semver-check/#/version/0.5.0-preview1` and `https://jubianchi.github.io/semver-check/#/version/0.5.0-preview.1`

Issue #4164, opened 2025-11-25
DeadlineInterceptorTests:Deadline_interceptor_can_enforce_application_deadline failure (Windows)


bernardnormier:
from https://github.com/icerpc/icerpc-csharp/actions/runs/19673179081/job/56347098741

```
  Failed IceRpc:Deadline:Tests:DeadlineInterceptorTests:Deadline_interceptor_can_enforce_application_deadline [792 ms]
  Error Message:
     Assert.That(caughtException, expression)
  Expected: <System.TimeoutException>
  But was:  null

  Stack Trace:
     at IceRpc.Deadline.Tests.DeadlineInterceptorTests.Deadline_interceptor_can_enforce_application_deadline() in D:\a\icerpc-csharp\icerpc-csharp\tests\IceRpc.Deadline.Tests\DeadlineInterceptorTests.cs:line 162

1)    at IceRpc.Deadline.Tests.DeadlineInterceptorTests.Deadline_interceptor_can_enforce_application_deadline() in D:\a\icerpc-csharp\icerpc-csharp\tests\IceRpc.Deadline.Tests\DeadlineInterceptorTests.cs:line 162
```

pepone:
This failure is curious, in a successful run this test should take a bit more that 100ms, the timeout used to create the deadline. Here in took 792ms to complete, and there was no exception.

If the timeout didn't fire this should have taken approximately 500ms

```
var invoker = new InlineInvoker(async (request, cancellationToken) =>
{
    await Task.Delay(TimeSpan.FromMilliseconds(500), cancellationToken);
    return new IncomingResponse(request, FakeConnectionContext.Instance);
});
```

not clear why the extra 192ms.

One possible explanation for the timeout not firing in time, is a clock adjustment which can affect the computation of the deadline.

bernardnormier:
Probably just flaky Windows runner.

pepone:
Fixed in #4167

PR #4167, opened 2025-11-25
Add TimeProvider parameter to deadline interceptor and middleware

pepone:
Adding a TimerProvider to the deadline interceptor and middleware for improve testing support. See #4164

### DIFF
--- a/tests/IceRpc.Deadline.Tests/DeadlineInterceptorTests.cs
+++ b/tests/IceRpc.Deadline.Tests/DeadlineInterceptorTests.cs
@@ -120,8 +120,8 @@ public sealed class DeadlineInterceptorTests
             token = cancellationToken;
             return Task.FromResult(new IncomingResponse(request, FakeConnectionContext.Instance));
         });
-        var timeProvider = new FakeTimeProvider(new DateTimeOffset(2025, 1, 1, 12, 0, 0, TimeSpan.Zero));
 
+        var timeProvider = new FakeTimeProvider();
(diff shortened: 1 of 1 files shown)


tests/IceRpc.Deadline.Tests/DeadlineInterceptorTests.cs
```diff
@@ -119,12 +120,13 @@ public async Task Deadline_interceptor_does_not_enforce_deadline_by_default()
             token = cancellationToken;
             return Task.FromResult(new IncomingResponse(request, FakeConnectionContext.Instance));
         });
+        var timeProvider = new FakeTimeProvider(new DateTimeOffset(2025, 1, 1, 12, 0, 0, TimeSpan.Zero));
```
bernardnormier: Why not use the default (Jan 1, 2000)?

tests/IceRpc.Deadline.Tests/DeadlineInterceptorTests.cs
```diff
@@ -146,15 +148,17 @@ public void Deadline_interceptor_can_enforce_application_deadline()
             await Task.Delay(TimeSpan.FromMilliseconds(500), cancellationToken);
             return new IncomingResponse(request, FakeConnectionContext.Instance);
         });
+        var timeProvider = new FakeTimeProvider(new DateTimeOffset(2025, 1, 1, 12, 0, 0, TimeSpan.Zero));
```
bernardnormier: Same here.

tests/IceRpc.Deadline.Tests/DeadlineInterceptorTests.cs
pepone: Misread the docs, we should use the default

PR #4169, opened 2025-11-26
Add InitialReceiveWindowSizes QUIC transport option

### LINKED ISSUE #4136: Review new QuicConnectionOptions in .NET 10
.NET 10 adds a number of QuicConnectionOptions, such as HandShakeTimeout and InitialReceiveWindowSizes.

https://learn.microsoft.com/en-us/dotnet/api/system.net.quic.quicconnectionoptions?view=net-10.0

We should check if we want to expose any of them as IceRpc.Transport.Quic.Quic(Client/Server)TransportOptions.

pepone:
Fixes #4136

### DIFF
--- a/src/IceRpc.Transports.Quic/QuicTransportOptions.cs
+++ b/src/IceRpc.Transports.Quic/QuicTransportOptions.cs
@@ -19,11 +19,7 @@ public record class QuicTransportOptions
     /// <summary>Gets or sets the initial receive window sizes for the connection and individual stream types.
     /// </summary>
     /// <value>The initial receive window sizes.</value>
-    public QuicReceiveWindowSizes InitialReceiveWindowSizes
-    {
-        get => field ??= new QuicReceiveWindowSizes();
-        set => field = value;
-    }
+    public QuicReceiveWindowSizes InitialReceiveWindowSizes { get; set; } = new();
 }
 
 /// <summary>The options class for configuring <see cref="QuicClientTransport"/>.</summary>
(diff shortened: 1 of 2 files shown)


src/IceRpc.Transports.Quic/QuicTransportOptions.cs
```diff
@@ -14,6 +15,15 @@ public record class QuicTransportOptions
     /// <remarks>The idle timeout is negotiated by QUIC during connection establishment. The actual idle timeout value
     /// for a connection can be lower than the value you've set here.</remarks>
     public TimeSpan IdleTimeout { get; set; } = TimeSpan.FromSeconds(30);
+
+    /// <summary>Gets or sets the initial receive window sizes for the connection and individual stream types.
+    /// </summary>
+    /// <value>The initial receive window sizes.</value>
+    public QuicReceiveWindowSizes InitialReceiveWindowSizes
```
bernardnormier: I would prefer:
```cs
public QuicReceiveWindowSizes InitialReceiveWindowSizes { get; set; } = new();
```

No need for lazy initialization here.

src/IceRpc.Transports.Quic/QuicClientTransport.cs
```diff
@@ -65,6 +66,7 @@ public IMultiplexedConnection CreateConnection(
             DefaultCloseErrorCode = (int)MultiplexedConnectionCloseError.Aborted,
             DefaultStreamErrorCode = 0,
             IdleTimeout = _quicTransportOptions.IdleTimeout,
+            InitialReceiveWindowSizes = _quicTransportOptions.InitialReceiveWindowSizes,
```
bernardnormier, line 69: If it's a transport options, it must affect both the client and server. Here, you updated only the client options.

pepone: Fixed the listener to apply this setting too.

PR #4170, opened 2025-11-26
Map new Ice ReplyStatus to IceRpc.StatusCode

pepone:
This PR maps the new Ice ReplyStatus code to the corresponding IceRpc.StatusCode

### DIFF
--- a/tests/IceRpc.Tests/IceProtocolConnectionTests.cs
+++ b/tests/IceRpc.Tests/IceProtocolConnectionTests.cs
@@ -22,7 +22,7 @@ public sealed class IceProtocolConnectionTests
                 new InlineDispatcher(
                     (request, cancellationToken) =>
                         new(new OutgoingResponse(request, StatusCode.InvalidData, message: null, invalidDataException))),
-                StatusCode.InternalError,
+                StatusCode.InvalidData,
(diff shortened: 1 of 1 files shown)


tests/IceRpc.Tests/IceProtocolConnectionTests.cs
```diff
@@ -22,7 +22,7 @@ private static IEnumerable<TestCaseData> DispatchFailureSource
                 new InlineDispatcher(
                     (request, cancellationToken) =>
                         new(new OutgoingResponse(request, StatusCode.InvalidData, message: null, invalidDataException))),
-                StatusCode.InternalError,
+                StatusCode.InvalidData,
```
bernardnormier, line 25: We need an extra test for an unmapped StatusCode, for example DeadlineExceeded.

pepone: we already have DeadlineExceeded below:

```
            yield return new TestCaseData(
                new InlineDispatcher(
                    (request, cancellationToken) =>
                        new(new OutgoingResponse(request, StatusCode.DeadlineExceeded, "deadline message"))),
                StatusCode.InternalError,
                "deadline message { Original StatusCode = DeadlineExceeded }");
```

bernardnormier (approved):

PR #4175, opened 2025-11-26
Use System.CommandLine 2.0.0 in thermostat examples

pepone:
This PR updates thermostat examples to use System.CommandLine 2.0.0 , they were using the beta version.

### DIFF
--- a/examples/protobuf/Thermostat/Client/Program.cs
+++ b/examples/protobuf/Thermostat/Client/Program.cs
@@ -28,14 +28,12 @@ var thermostat = new ThermostatClient(pipeline);
 // set <value> changes the set point and exits.
 
 var monitorCommand = new Command("monitor", "Monitor the thermostat (default)");
-monitorCommand.SetAction((parseResult, cancellationToken) => MonitorAsync(cancellationToken));
+monitorCommand.SetAction(MonitorAsync);
 
 var setCommand = new Command("set", "Change the set point");
 var setArgument = new Argument<float>("setPoint");
 setCommand.Arguments.Add(setArgument);
-setCommand.SetAction(
-    (parseResult, cancellationToken) =>
-        ChangeSetPointAsync(parseResult.GetRequiredValue(setArgument), cancellationToken));
+setCommand.SetAction(ChangeSetPointAsync);
 
 var rootCommand = new RootCommand()
 {
@@ -49,8 +47,9 @@ await rootCommand.Parse(args).InvokeAsync();
 // Graceful shutdown
 await connection.ShutdownAsync();
 
-async Task ChangeSetPointAsync(float value, CancellationToken cancellationToken)
+async Task ChangeSetPointAsync(ParseResult parseResult, CancellationToken cancellationToken)
 {
+    float value = parseResult.GetRequiredValue(setArgument);
     try
     {
         await thermostat.ChangeSetPointAsync(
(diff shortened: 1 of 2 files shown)


examples/protobuf/Thermostat/Client/Program.cs
```diff
@@ -28,62 +28,63 @@
 // set <value> changes the set point and exits.
 
 var monitorCommand = new Command("monitor", "Monitor the thermostat (default)");
-monitorCommand.SetHandler(MonitorAsync);
+monitorCommand.SetAction(
+    async (parseResult, cancellationToken) =>
```
bernardnormier: I'd rather keep the local functions. I find it much more readable.

pepone: Reworked to keep the local functions.

examples/protobuf/Thermostat/Client/Program.cs
```diff
@@ -69,20 +80,14 @@ async Task ChangeSetPointAsync(float value)
     Console.ResetColor();
 }
 
-async Task MonitorAsync()
+async Task MonitorAsync(CancellationToken cancellationToken)
 {
```
pepone, line 76: Don't need to use CancelKeyPress we can directly use the provided cancellation token.

examples/protobuf/Thermostat/Client/Program.cs
```diff
@@ -28,38 +28,42 @@
 // set <value> changes the set point and exits.
 
 var monitorCommand = new Command("monitor", "Monitor the thermostat (default)");
-monitorCommand.SetHandler(MonitorAsync);
+monitorCommand.SetAction((parseResult, cancellationToken) => MonitorAsync(cancellationToken));
```
bernardnormier: I meant, keep `monitorCommand.SetAction(MonitorAsync)` and change the signature of `MonitorAsync`.

examples/protobuf/Thermostat/Client/Program.cs
```diff
@@ -28,38 +28,42 @@
...
 
 var setCommand = new Command("set", "Change the set point");
-var argument = new Argument<float>("setPoint");
-setCommand.AddArgument(argument);
-setCommand.SetHandler(ChangeSetPointAsync, argument);
+var setArgument = new Argument<float>("setPoint");
+setCommand.Arguments.Add(setArgument);
+setCommand.SetAction(
```
bernardnormier: Same here.

pepone: And capture `setArgument` in the local function?

bernardnormier: Ok, no, this does not work. We can keep the delegate for this one.

externl (approved):

examples/protobuf/Thermostat/Client/Program.cs
pepone: Seems to work as they are not marked as static.

examples/slice/Thermostat/Client/Program.cs
```diff
@@ -27,61 +27,54 @@
...
     {
         Console.ForegroundColor = ConsoleColor.Red;
-        Console.WriteLine($"Failed to change set point to {setPoint}°F: {exception}");
+        Console.WriteLine($"Failed to change set point to {value}°F: {exception}");
     }
     Console.ResetColor();
 }
 
-async Task MonitorAsync()
+async Task MonitorAsync(ParseResult parseResult, CancellationToken cancellationToken)
```
bernardnormier, line 72: Should parseResult be _ since we don't use it?

bernardnormier (approved):

examples/slice/Thermostat/Client/Program.cs
pepone: I tried but get 

```
Program.cs(75,37): error SA1313: Parameter '_' should begin with lower-case letter (https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1313.md)
```

Keeping it as is.

PR #4180, opened 2025-11-28
Collect additional data from CI tests

pepone:


.github/actions/test/action.yaml
```diff
@@ -7,29 +7,38 @@ runs:
...
+  - name: 🧪 Test .NET
     run: |
       mkdir -p TestResults
-      dotnet test --no-build --verbosity normal --blame-hang-timeout 10s --logger "trx" --results-directory TestResults
+      dotnet test \
+        --no-build \
+        --verbosity normal \
+        --blame \
+        --blame-crash \
+        --blame-hang-timeout 300s \
```
bernardnormier: 300 s (5 minutes) sound quite long. My understanding it's for a single test.

Maybe 10 seconds was too short, and 30 or 60 seconds makes more sense?

bernardnormier (approved):

.github/actions/test/action.yaml
pepone: Updated to 60s

PR #4184, opened 2025-12-01
Update `slicec-cs` Dependencies (Inluding the Latest `slicec`)

InsertCreativityHere:
This PR updates all the dependencies of `slicec-cs`, raises the minimum MSRV to `1.82` (since that's what's required by `slicec` now), and fixes a small lint found by clippy.

I looked through all the dependency's changelogs, and there were no code changes to make, surprisingly.

### DIFF
--- a/tools/slicec-cs/src/generators/dispatch_generator.rs
+++ b/tools/slicec-cs/src/generators/dispatch_generator.rs
@@ -74,6 +74,8 @@ fn request_class(interface_def: &Interface) -> CodeBlock {
         return "".into();
     }
 
+    // Check if any of the base interfaces will already have a 'Request' class generated.
+    // We generate a 'Request' class if and only if one of an interface's operations has a parameter.
     let base_operations = interface_def.all_inherited_operations();
     let has_base_request_class = base_operations.iter().any(|o| !o.parameters().is_empty());
 
@@ -151,6 +153,8 @@ fn response_class(interface_def: &Interface) -> CodeBlock {
         return "".into();
     }
 
+    // Check if any of the base interfaces will already have a 'Response' class generated.
+    // We generate a 'Response' class if and only if one of an interface's operations has a non-streamed return member.
     let base_operations = interface_def.all_inherited_operations();
     let has_base_response_class = base_operations.iter().any(|o| o.has_non_streamed_return_members());
 


tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -74,14 +74,14 @@ fn request_class(interface_def: &Interface) -> CodeBlock {
         return "".into();
     }
 
-    let mut base_operations = interface_def.all_inherited_operations();
-    base_operations.retain(|o| !o.parameters.is_empty());
```
InsertCreativityHere, line 78: Removing elements from a mutable `Vec` is overkill.
All we really care about is checking if any of them have a parameter.

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -74,14 +74,14 @@ fn request_class(interface_def: &Interface) -> CodeBlock {
...
+    let base_operations = interface_def.all_inherited_operations();
+    let has_parameters = base_operations.iter().any(|o| !o.parameters().is_empty());
 
     let mut class_builder = ContainerBuilder::new(
-        if base_operations.is_empty() {
-            "public static class"
-        } else {
+        if has_parameters {
```
InsertCreativityHere: `has_parameters` is the opposite of `is_empty`, so I swapped the order of these braces. (instead of having `if !has_parameters`...)

externl (approved):

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -74,14 +74,14 @@ fn request_class(interface_def: &Interface) -> CodeBlock {
         return "".into();
     }
 
-    let mut base_operations = interface_def.all_inherited_operations();
-    base_operations.retain(|o| !o.parameters.is_empty());
+    let base_operations = interface_def.all_inherited_operations();
+    let has_parameters = base_operations.iter().any(|o| !o.parameters().is_empty());
```
pepone: `has_parameters` isn't a great name here. What we are really checking is if there is a base Request class, which is the case when any of the operations defined in the base interfaces has parameters.

I think `has_base_request_class` a bit verbose but more to the point of what this is about. And maybe a small comment for this logic.

Issue #2053, opened 2022-11-04
Enable CA1849?


bernardnormier:
Should we enable CA1849 in our code base?

It used to be enabled but it's now disabled.
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1849

While it's a pain for Complete/CompleteAsync, I find it otherwise useful to write clean code.

Issue #965, opened 2022-03-24
PipeReader / PipeWriter Complete vs CompleteAsync proposal


bernardnormier:
I propose the following:

- we consider calls to `PipeReader.Complete` and `PipeWriter.Complete` to be modern C# fast non-blocking calls. A decorator or third-party transport implementation must not implement or decorate Complete by blocking.

- we ensure that we never call `PipeWriter.Complete(null)` or `CompleteAsync(null)` on a pipe writer with unflushed bytes and our transport PipeWriter implementation throw `InvalidOperationException` in this situation (as SlicPipeWriter already does)

- we always use the default CompleteAsync in our PipeReader/PipeWriter implementations

- when it makes sense, we use a #pragma to disable the compiler warning to call `Complete` instead of `CompleteAsync` in a Task/ValueTask-returning method.

A good example of where this makes sense is the BinderInterceptor. We should throw `NoEndpointException` synchronously - not asynchronously just because we need to `CompleteAsync` various PipeReader/PipeWriter.

PR #4189, opened 2025-12-13
Remove Json serialize helper from JSON demo

bernardnormier:
This PR simplifies the JSON demo by removing the serialize/deserialize helper.

Instead, the demo serialize and deserialize using the SerializeAsync/DeserializeAsync overloads that accept a PipeWriter resp. a PipeReader.

### DIFF
--- a/examples/json/Greeter/Client/Program.cs
+++ b/examples/json/Greeter/Client/Program.cs
@@ -23,7 +23,7 @@ async Task<string> GreetAsync(string name)
     pipe.Writer.Complete();
 
     // Construct an outgoing request to the icerpc:/greeter service.
-    // The payload is the PiperReader of our pipe.
+    // The payload is the PipeReader of our pipe.
     using var request = new OutgoingRequest(new ServiceAddress(new Uri("icerpc:/greeter")))
     {
         Operation = "greet",


examples/common/PipeReaderJsonSerializer.cs
```diff
@@ -1,54 +0,0 @@
...
-using System.Buffers;
-using System.Diagnostics;
-using System.IO.Pipelines;
-using System.Text.Json;
-
-namespace IceRpc.Json;
-
-/// <summary>A helper class for serializing and deserializing JSON objects into and from <see cref="PipeReader"/>.
-/// </summary>
-internal static class PipeReaderJsonSerializer
```
bernardnormier, line 12: It's not immediately clear to me why we didn't use the PipeReader / PipeWriter overloads provided by JsonSerializer.

examples/json/Greeter/Client/Program.cs
```diff
@@ -15,21 +16,29 @@
 // Create the request to the greeter and then await and decode the response.
 async Task<string> GreetAsync(string name)
 {
+    // Create a PipeReader holding the JSON response message.
+    var pipe = new Pipe();
+    var greetRequest = new GreetRequest { Name = name };
+    await JsonSerializer.SerializeAsync(pipe.Writer, greetRequest);
+    pipe.Writer.Complete();
```
bernardnormier, line 23: It's Complete not CompleteAsync on purpose, which unfortunately generates a lint.

See #965 and #2053. 

In this particular case, it doesn't matter, so we could write `await pipe.Write.CompleteAsync();` ... but I don't like it as it suggests "complete" could block, which is not true.

PR #4194, opened 2026-01-02
QUIC is no longer preview

bernardnormier:
This PR removes the various EnablePreviewFeatures we have for QUIC.

QUIC is supposed to be stable (non-preview) since .NET 9:
https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/quic/quic-overview#quic-in-net

_however_ this PR does not work. You can't build the slice or protobuf Quic example. It generates a warning for the slice/Quic/Client example (but not the Server). For protobuf, both Client and Server generate warnings.

```
bernard@pro14 Client % dotnet build
Restore complete (0.1s)
  Client net10.0 failed with 1 error(s) (0.7s)
    /Users/bernard/builds/icerpc-csharp/examples/slice/Quic/Client/Program.cs(37,33): error CA2252: Using 'QuicClientTransport' requires opting into preview features. See https://aka.ms/dotnet-warnings/preview-features for more information. (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2252)

Build failed with 1 error(s) in 1.0s
```

Update: works when I set the correct version in IceRpc.Examples.props

### DIFF
--- a/build/IceRpc.Examples.props
+++ b/build/IceRpc.Examples.props
@@ -5,7 +5,7 @@
       The version of the IceRPC NuGet packages used by the examples. It must be available from nuget.org once
       this branch is released.
     -->
-    <IceRpcVersion Condition="'$(IceRpcVersion)' == ''">0.5.0-nightly.*</IceRpcVersion>
+    <IceRpcVersion Condition="'$(IceRpcVersion)' == ''">0.6.0-nightly.*</IceRpcVersion>
     <WarningLevel>4</WarningLevel>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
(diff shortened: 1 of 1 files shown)


bernardnormier:
> It generates a warning for the slice/Quic/Client example (but not the Server).

Solved this mystery: there is a bug in slice/Quic/Server, it doesn't use Quic!

bernardnormier:
I think Copilot is half-right: there is a bug in IceRpc.Examples.props:
```
 <IceRpcVersion Condition="'$(IceRpcVersion)' == ''">0.5.0-nightly.*</IceRpcVersion>
```

!!

(Fixed)

PR #4199, opened 2026-01-05
Update examples to use common auth methods

bernardnormier:
This PR updates all the examples to use the common auth helper methods added by #4197.


examples/protobuf/GenericHost/Client/Program.cs
```diff
@@ -30,33 +29,12 @@
             .Bind(hostContext.Configuration.GetSection("Client"))
             .Configure(options =>
             {
-                // Configure the authentication options
                 var rootCA = X509CertificateLoader.LoadCertificateFromFile(
```
bernardnormier, line 32: We should have the DI container dispose the rootCA or the server certificate in the GenericHost examples. I plan to fix this in a follow-up PR.

PR #4201, opened 2026-01-05
Create certificates as singletons in GenericHost examples

bernardnormier:
This PR updates the GenericHost examples to create certificates (root CA and server certificate) as singletons managed (and disposed) by the DI container.


examples/protobuf/GenericHost/Server/Program.cs
```diff
@@ -27,22 +27,22 @@
         string workingDirectory = Directory.GetCurrentDirectory();
         Console.WriteLine(workingDirectory);
 
+        // Load and register the server certificate as a singleton so it stays alive and gets disposed.
+        services.AddSingleton<X509Certificate2>(sp =>
```
bernardnormier, line 31: The type `(<X509Certificate2>)` is not necessary, but making it explicit is clearer since the type returned by `X509CertificateLoader.LoadPkcs12FromFile` is not obvious.

PR #4204, opened 2026-01-05
Add new common auth to non-DI templates

bernardnormier:
The DI templates currently don't use SSL (or QUIC) at all.

### DIFF
--- a/src/IceRpc.Templates/Templates/IceRpc-Protobuf-Client/Program.Authentication.cs
+++ b/src/IceRpc.Templates/Templates/IceRpc-Protobuf-Client/Program.Authentication.cs
@@ -1,0 +1,47 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+
+// Provides helper methods to create the authentication options used by the examples.
+public static partial class Program
+{
+    /// <summary>Creates client authentication options with a custom certificate validation callback that uses the
+    /// specified root CA.</summary>
+    /// <param name="rootCA">The root CA certificate.</param>
+    /// <returns>The client authentication options.</returns>
+    public static SslClientAuthenticationOptions CreateClientAuthenticationOptions(X509Certificate2 rootCA) =>
+        new()
+        {
+            RemoteCertificateValidationCallback = (sender, certificate, chain, errors) =>
+            {
+                if (certificate is X509Certificate2 peerCertificate)
+                {
+                    using var customChain = new X509Chain();
+                    customChain.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
+                    customChain.ChainPolicy.DisableCertificateDownloads = true;
+                    customChain.ChainPolicy.TrustMode = X509ChainTrustMode.CustomRootTrust;
+                    customChain.ChainPolicy.CustomTrustStore.Add(rootCA);
+                    return customChain.Build(peerCertificate);
+                }
+                else
+                {
+                    return false;
+                }
+            }
+        };
+
+    /// <summary>Creates server authentication options for the specified server certificate.</summary>
+    /// <param name="serverCertificate">The server certificate. It must include a private key.</param>
+    /// <returns>The server authentication options.</returns>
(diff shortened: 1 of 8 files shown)


src/IceRpc.Templates/Templates/IceRpc-Protobuf-Client/Program.Authentication.cs
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+
+// Provides helper methods to create the authentication options used by the examples.
+public static partial class Program
```
bernardnormier, line 7: It's a copy of the examples/common file with the same name. Is there a better solution?

src/IceRpc.Templates/Templates/IceRpc-Protobuf-Client/Program.cs
```diff
@@ -14,40 +13,19 @@
...
-        }
-        else
-        {
-            return false;
-        }
-    }
-};
-
-#if (transport == "quic")
 // Create a client connection that logs messages to a logger with category IceRpc.ClientConnection.
```
bernardnormier, line 18: The comment applies to both QUIC and Slic/TCP connections.

src/IceRpc.Templates/Templates/IceRpc-Protobuf-Client/Program.Authentication.cs
pepone: If you want to avoid duplicating this file with each template we can probably copy it when the templates are build. Need to test but I think is doable. We should probably do the same for the certs.

pepone: I tested adding this to `src/IceRpc.Templates/IceRpc.Templates.csproj`

```
<!-- Define shared assets to copy to templates that need them -->
  <ItemGroup>
    <CommonAuth Include="../../examples/common/Program.Authentication.cs" />
    <TemplatesNeedingCommonAuth Include="Templates/IceRpc-Protobuf-Client" />
    <TemplatesNeedingCommonAuth Include="Templates/IceRpc-Protobuf-Server" />
    <TemplatesNeedingCommonAuth Include="Templates/IceRpc-Slice-Client" />
    <TemplatesNeedingCommonAuth Include="Templates/IceRpc-Slice-Server" />
  </ItemGroup>

  <!-- Copy shared assets to templates before pack. Must run before Restore so the files exist
       when the Templates/**/* glob is evaluated. -->
  <Target Name="CopySharedAssets" BeforeTargets="Restore">
    <Copy
      SourceFiles="@(CommonAuth)"
      DestinationFolder="%(TemplatesNeedingCommonAuth.Identity)"
      SkipUnchangedFiles="true" />
  </Target>

  <!-- Clean up copied files after pack -->
  <Target Name="CleanSharedAssets" AfterTargets="Pack">
    <Delete Files="@(TemplatesNeedingCommonAuth->'%(Identity)/Program.Authentication.cs')" />
  </Target>
```

pepone (commented):
We should try copy this file as per comment. https://github.com/icerpc/icerpc-csharp/pull/4204/changes/33254c9eaa55b0652c734b59e5d0679aad7369ab#r2662861111

src/IceRpc.Templates/Templates/IceRpc-Protobuf-Client/Program.Authentication.cs
bernardnormier: Is this compatible with installing the templates using:

```shell
dotnet new install src/IceRpc.Templates
```

pepone: No you need to pack, and then install.

bernardnormier: Currently you can install the templates with `dotnet new install src/IceRpc.Templates`; you don't even need to build anything.

Issue #4192, opened 2025-12-31
Make QUIC the default transport for icerpc


bernardnormier:
Now that QUIC is available on all platforms (see #4191), we should make QUIC the default transport for the icerpc protocol.

I would also keep Slic in the IceRpc core, just not as the default.

bernardnormier:
Fixed by the PRs referenced above and a few additional PRs.

PR #4206, opened 2026-01-06
Make QUIC a built-in transport

bernardnormier:
This PR moves the QUIC transport into the main IceRpc assembly.

It's a preliminary step towards #4192.


src/IceRpc/Transports/Quic/QuicClientTransport.cs
```diff
@@ -4,10 +4,14 @@
 using System.Net;
 using System.Net.Quic;
 using System.Net.Security;
+using System.Runtime.Versioning;
 
 namespace IceRpc.Transports.Quic;
 
 /// <summary>Implements <see cref="IMultiplexedClientTransport"/> using QUIC.</summary>
+[SupportedOSPlatform("linux")]
```
bernardnormier, line 12: It's cleaner to put these attributes on the QUIC-related classes than on the whole assembly. If we put it on the whole assembly, then all dependent assemblies (the various middleware and interceptors) need it as well ... pretty messy.

externl (approved):

pepone (approved):
Looks good.

In a follow up PR we should move the QUIC tests to https://github.com/icerpc/icerpc-csharp/tree/main/tests/IceRpc.Tests/Transports

PR #4209, opened 2026-01-07
Make QUIC the default transport

bernardnormier:
This PR makes QUIC the default multiplexed transport (see #4192): it updates src, the tests, the build telemetry clients, and one example (slice/Greeter).

I plan to update all remaining examples in a follow-up PR. And also update the templates in another PR.

### DIFF
--- a/src/IceRpc/Transports/IMultiplexedServerTransport.cs
+++ b/src/IceRpc/Transports/IMultiplexedServerTransport.cs
@@ -20,7 +20,7 @@ public interface IMultiplexedServerTransport
                 return _quicServerTransport;
             }
             throw new PlatformNotSupportedException(
-                "The default Quic server transport is not supported on this platform.");
+                "The default QUIC server transport is not supported on this platform.");
         }
     }
 


tests/IceRpc.Tests/ClientConnectionTests.cs
```diff
@@ -107,12 +107,19 @@ public async Task Connect_can_be_canceled()
     public async Task Connection_can_reconnect_after_underlying_connection_shutdown()
     {
         // Arrange
-        var server = new Server(NotFoundDispatcher.Instance, new Uri("icerpc://127.0.0.1:0"));
+        var server = new Server(
+            NotFoundDispatcher.Instance,
+            new Uri("icerpc://127.0.0.1:0"),
+            multiplexedServerTransport: new SlicServerTransport(new TcpServerTransport()));
```
bernardnormier, line 113: I fixed a few tests to use Slic over Tcp instead of the new default transport.

tools/IceRpc.ProtobufBuildTelemetry.Reporter/Program.cs
```diff
@@ -33,7 +35,10 @@
         const string uri = "icerpc://build-telemetry.icerpc.dev";
         using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
         cts.CancelAfter(TimeSpan.FromSeconds(3));
-        await using var connection = new ClientConnection(new Uri(uri), new SslClientAuthenticationOptions());
+        await using var connection = new ClientConnection(
```
bernardnormier, line 38: It would be desirable to alter this logic with something like:
 - check if Quic is supported
   if yes, use Quic
   if no, use Slic over Tcp

We can't do that right now as the build telemetry server only listens on Slic over TCP.

externl (approved):

PR #4217, opened 2026-01-08
Rework exception when QUIC is not supported / available

bernardnormier:
This PR updates the behavior when QUIC is not supported or available.

When not supported (because the platform is not macos, linux, or windows), we throw PlatformNotSupportedException. 
We don't ever use another default as it would be confusing.

When not available (you didn't install msquic, you're on Windows 10...), we throw NotSupportedException with another message ("not available on this system"). It's typically not a platform issue but a configuration issue. Just install msquic (etc.) as per the .NET doc.

### DIFF
--- a/src/IceRpc/Transports/Quic/QuicServerTransport.cs
+++ b/src/IceRpc/Transports/Quic/QuicServerTransport.cs
@@ -34,7 +34,7 @@ public class QuicServerTransport : IMultiplexedServerTransport
         MultiplexedConnectionOptions options,
         SslServerAuthenticationOptions? serverAuthenticationOptions)
     {
-        if (!QuicConnection.IsSupported)
+        if (!QuicListener.IsSupported)
         {
             throw new NotSupportedException(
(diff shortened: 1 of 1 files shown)


src/IceRpc.Extensions.DependencyInjection/ClientConnectionServiceCollectionExtensions.cs
```diff
@@ -55,25 +55,23 @@ internal static IServiceCollection TryAddIceRpcClientTransport(this IServiceColl
                 provider => new TcpClientTransport(
                     provider.GetRequiredService<IOptions<TcpClientTransportOptions>>().Value));
 
-        // The default multiplexed transport is QUIC on most platforms.
-        if (OperatingSystem.IsLinux() || OperatingSystem.IsMacOS() || OperatingSystem.IsWindows())
```
bernardnormier, line 59: This was actually quite buggy. The check must be _inside_ TryAddSingleton, not before we call it.

pepone (approved):

Issue #3738, opened 2023-10-21
Remove SliceD methods


bernardnormier:
We should look into removing SliceD methods, like #3737 removes ProtobufD methods.

For example:
```csharp
   [SliceOperation("opWithMultipleParametersAndReturnValues")]
    [global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
    protected static async global::System.Threading.Tasks.ValueTask<IceRpc.OutgoingResponse> SliceDOpWithMultipleParametersAndReturnValuesAsync(
        IMyOperationsAService target,
        IceRpc.IncomingRequest request,
        global::System.Threading.CancellationToken cancellationToken)
    {
        request.CheckNonIdempotent();
        var args = await Request.DecodeOpWithMultipleParametersAndReturnValuesAsync(request, cancellationToken).ConfigureAwait(false);
        var returnValue = await target.OpWithMultipleParametersAndReturnValuesAsync(args.P1, args.P2, request.Features, cancellationToken).ConfigureAwait(false);
        return new IceRpc.OutgoingResponse(request)
        {
            Payload = Response.EncodeOpWithMultipleParametersAndReturnValues(returnValue.R1, returnValue.R2, request.Features.Get<ISliceFeature>()?.EncodeOptions),
            PayloadContinuation = null
        };
    }
```

The idea is for the source generator to generate equivalent code in the DispatchAsync "switch/case". We'd generate Request.DecodeXxx and Response.EncodeXxx with slice-cs as usual.

If needed, we could add additional information in the SliceOperationAttribute.

PR #4224, opened 2026-01-12
Make generated SliceD code more uniform

bernardnormier:
This PR is a preliminary step for #3738: it makes the generated SliceD methods more uniform by moving code to more generated methods in the Request and Response helper classes.

Newly generated methods:
 - Request.DecodeXxxAsync for empty args
 - Response.EncodeXxx for void return
 - Response.EncodeStreamOfxxx for returned stream parameter

### DIFF
--- a/tools/slicec-cs/src/encoding.rs
+++ b/tools/slicec-cs/src/encoding.rs
@@ -562,37 +562,29 @@ int startPos_ = encoder_.EncodedByteCount;",
 // Dispatch-only for now
 pub fn encode_operation_stream(operation: &Operation) -> CodeBlock {
     let namespace = operation.namespace();
-    let return_values = operation.return_members();
     let encoding = operation.encoding.to_cs_encoding();
 
-    match operation.streamed_return_member() {
-        None => "null".into(), // TODO: should never happen
-        Some(stream_return) => {
-            let stream_type = stream_return.data_type();
+    let stream_return = operation
+        .streamed_return_member()
+        .expect("encode_operation_stream called on an operation without a streamed return");
 
-            let stream_arg = if return_values.len() == 1 {
-                "returnValue".to_owned()
-            } else {
-                stream_return.parameter_name()
-            };
+    let stream_type = stream_return.data_type();
+    let stream_arg = stream_return.parameter_name();
 
-            match stream_type.concrete_type() {
-                Types::Primitive(Primitive::UInt8) if !stream_type.is_optional => stream_arg.into(),
-                _ => format!(
-                    "\
+    match stream_type.concrete_type() {
+        Types::Primitive(Primitive::UInt8) if !stream_type.is_optional => stream_arg.into(),
+        _ => format!(
+            "\
 {stream_arg}.ToPipeReader(
     {encode_stream_parameter},
     {use_segments},
     {encoding},
     {encode_options})",
-                    encode_stream_parameter =
-                        encode_stream_parameter(stream_type, &namespace, operation.encoding).indent(),
-                    use_segments = stream_type.fixed_wire_size().is_none(),
-                    encode_options = "encodeOptions",
-                )
-                .into(),
-            }
-        }
+            encode_stream_parameter = encode_stream_parameter(stream_type, &namespace, operation.encoding).indent(),
+            use_segments = stream_type.fixed_wire_size().is_none(),
+            encode_options = "encodeOptions",
+        )
+        .into(),
     }
 }
 
--- a/tools/slicec-cs/src/generators/dispatch_generator.rs
+++ b/tools/slicec-cs/src/generators/dispatch_generator.rs
@@ -482,7 +482,8 @@ request.Features = IceRpc.Features.FeatureCollectionExtensions.With(
         match operation.streamed_return_member() {
             None => "returnValue",
             Some(_) => "returnValue.Payload",
-        }.into()
+        }
+        .into()
     } else {
         dispatch_return_payload(operation)
     };


tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -67,20 +67,16 @@ private static readonly IActivator _defaultActivator =
 }
 
 fn request_class(interface_def: &Interface) -> CodeBlock {
-    let mut operations = interface_def.operations();
-    operations.retain(|o| !o.parameters.is_empty());
+    let operations = interface_def.operations();
```
bernardnormier, line 70: We're now generating a DecodeXxxAsync for every single operation - no special case.

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -103,16 +99,22 @@ fn request_class(interface_def: &Interface) -> CodeBlock {
             FunctionType::ExpressionBody
         };
 
+        let return_type = if parameters.is_empty() {
+            "global::System.Threading.Tasks.ValueTask".to_owned()
```
bernardnormier, line 103: Unfortunately, C# does not provide a Void type like Swift.

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -146,20 +148,16 @@ fn request_class(interface_def: &Interface) -> CodeBlock {
 }
 
 fn response_class(interface_def: &Interface) -> CodeBlock {
-    let mut operations = interface_def.operations();
-    operations.retain(|o| o.has_non_streamed_return_members());
+    let operations = interface_def.operations();
```
bernardnormier, line 151: Like for Request, we now generate an EncodeXxx for every single operation - no special case.

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -425,42 +477,25 @@ await request.DecodeEmptyArgsAsync({encoding}, cancellationToken).ConfigureAwait
...
-                dispatch_and_return,
-                "\
-return new IceRpc.OutgoingResponse(request)
-{{
-    Payload = returnValue.Payload,
-    PayloadContinuation = {payload_continuation}
-}};",
-                payload_continuation = payload_continuation(operation, encoding).indent(),
-            );
+    let payload = if operation.has_attribute::<CsEncodedReturn>() {
```
bernardnormier, line 481: When we apply the Slice attribute `encodedReturn`, the user calls EncodeXxx for the non-stream return values. The stream return (if any) is not affected by `encodedReturn`.

tools/slicec-cs/src/encoding.rs
```diff
@@ -526,22 +540,60 @@ var encoder_ = new SliceEncoder(pipe_.Writer, {encoding}, {class_format});
...
+}
+
+// Dispatch-only for now
+pub fn encode_operation_stream(operation: &Operation) -> CodeBlock {
+    let namespace = &operation.namespace();
+    let return_values = operation.return_members();
+    let encoding = operation.encoding.to_cs_encoding();
+
+    match operation.streamed_return_member() {
+        None => "null".into(), // TODO: should never happen
```
externl: Maybe add a `debug_assert`?

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -184,7 +183,11 @@ fn response_class(interface_def: &Interface) -> CodeBlock {
             "public static",
             "global::System.IO.Pipelines.PipeReader",
             format!("Encode{operation_name}").as_str(),
-            FunctionType::BlockBody,
+            if non_streamed_returns.is_empty() {
+                FunctionType::ExpressionBody
+            } else {
+                FunctionType::BlockBody
+            },
```
pepone, line 190: Here we do this because, encode_operation returns a ExpressionBody when empty? 

It would be clear if encode_operation returned both the FunctionType and the function body. A call it here before we construct the builder.

tools/slicec-cs/src/encoding.rs
```diff
@@ -526,22 +540,60 @@ var encoder_ = new SliceEncoder(pipe_.Writer, {encoding}, {class_format});
...
+    }
+}
+
+// Dispatch-only for now
+pub fn encode_operation_stream(operation: &Operation) -> CodeBlock {
+    let namespace = &operation.namespace();
+    let return_values = operation.return_members();
+    let encoding = operation.encoding.to_cs_encoding();
+
+    match operation.streamed_return_member() {
```
pepone: Can we access stream_return directly without match here?

```
let stream_return = operation
    .streamed_return_member()
    .expect("streamed_return_member must exist");
```
?

pepone (approved):

tools/slicec-cs/src/encoding.rs
InsertCreativityHere: I think we should `expect` here, if this case should never happen,

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -425,42 +478,25 @@ await request.DecodeEmptyArgsAsync({encoding}, cancellationToken).ConfigureAwait
...
-    Payload = returnValue.Payload,
-    PayloadContinuation = {payload_continuation}
-}};",
-                payload_continuation = payload_continuation(operation, encoding).indent(),
-            );
+    let payload = if operation.has_attribute::<CsEncodedReturn>() {
+        match operation.streamed_return_member() {
+            None => "returnValue".into(),
+            Some(_) => "returnValue.Payload".into(),
         }
```
InsertCreativityHere: Can move the `into()` to the outside of the match, since you're applying it to both branches anyways.
`match` is an expression in Rust.
```suggestion
        match operation.streamed_return_member() {
            None => "returnValue",
            Some(_) => "returnValue.Payload",
        }.into()
```

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -498,62 +534,46 @@ catch ({catch_expression})
...
-
-    match non_streamed_return_values.len() {
-        0 => format!("{encoding}.CreateEmptyStructPayload()"),
-        _ => format!(
-            "Response.Encode{operation_name}({args}, request.Features.Get<ISliceFeature>()?.EncodeOptions)",
-            operation_name = operation.escape_identifier(),
-            args = returns.join(", "),
-        ),
+    if !non_streamed_return_values.is_empty() {
+        let mut returns = vec![];
```
InsertCreativityHere, line 544: This block looks very dubious to me.
We create an empty `vec`, push exactly 1 element into it, then call `join` on it?

What is the point of having a `Vec` here? Can't we just have a single `str`?

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -498,62 +534,46 @@ catch ({catch_expression})
...
+                .collect::<Vec<_>>()
+                .join(", "),
+        });
+
+        args = returns.join(", ") + ", ";
     }
+
+    format!(
+        "Response.Encode{operation_name}({args}request.Features.Get<ISliceFeature>()?.EncodeOptions)",
+        operation_name = &operation.escape_identifier()
```
InsertCreativityHere: No point in borrowing here, `escape_identifier` returns an owned `String`, and we're fine to move it.
```suggestion
        operation_name = operation.escape_identifier()
```

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -498,62 +534,46 @@ catch ({catch_expression})
...
-                    encode_stream_parameter =
-                        encode_stream_parameter(stream_type, namespace, operation.encoding).indent(),
-                    use_segments = stream_type.fixed_wire_size().is_none(),
-                    encode_options = "request.Features.Get<ISliceFeature>()?.EncodeOptions",
-                )
-                .into(),
-            }
+            format!(
+                "Response.EncodeStreamOf{operation_name}({stream_arg}, request.Features.Get<ISliceFeature>()?.EncodeOptions)",
+                operation_name = &operation.escape_identifier()
```
InsertCreativityHere: Also no point to borrowing here I'm pretty sure.

```suggestion
                operation_name = operation.escape_identifier()
```

tools/slicec-cs/src/encoding.rs
```diff
@@ -526,22 +540,60 @@ var encoder_ = new SliceEncoder(pipe_.Writer, {encoding}, {class_format});
...
+            class_format = operation.get_class_format(is_dispatch),
+            encode_returns = encode_operation_parameters(operation, is_dispatch, "encoder_"),
+        )
+        .into()
+    }
+}
+
+// Dispatch-only for now
+pub fn encode_operation_stream(operation: &Operation) -> CodeBlock {
+    let namespace = &operation.namespace();
```
InsertCreativityHere: The compiler works this out, but it's conceptually dubious. `namespace` returns an owned `String`, and we immediately borrow it. At this point, it's not clear who owns this string.
Instead, `namespace` should keep the owned copy, and we just borrow it as needed.
```suggestion
    let namespace = operation.namespace();
```

This is cleaner Rust, where ownership semantics is important.
But like I said, the compiler is smart enough to not drop the `String`, even if it's 'orphaned' here.

tools/slicec-cs/src/encoding.rs
```diff
@@ -526,22 +540,60 @@ var encoder_ = new SliceEncoder(pipe_.Writer, {encoding}, {class_format});
...
+                Types::Primitive(Primitive::UInt8) if !stream_type.is_optional => stream_arg.into(),
+                _ => format!(
+                    "\
+{stream_arg}.ToPipeReader(
+    {encode_stream_parameter},
+    {use_segments},
+    {encoding},
+    {encode_options})",
+                    encode_stream_parameter =
+                        encode_stream_parameter(stream_type, namespace, operation.encoding).indent(),
```
InsertCreativityHere: And the borrow moves here.
```suggestion
                        encode_stream_parameter(stream_type, &namespace, operation.encoding).indent(),
```

tools/slicec-cs/src/encoding.rs
```diff
@@ -526,22 +540,60 @@ var encoder_ = new SliceEncoder(pipe_.Writer, {encoding}, {class_format});
...
+    match operation.streamed_return_member() {
+        None => "null".into(), // TODO: should never happen
+        Some(stream_return) => {
+            let stream_type = stream_return.data_type();
+
+            let stream_arg = if return_values.len() == 1 {
+                "returnValue".to_owned()
+            } else {
+                stream_return.parameter_name()
+            };
```
InsertCreativityHere: Also, I'm double-checking, but I think this is unnecessary.
If you have a single return type, I'm pretty sure `slicec` gives it the name `returnValue` already.

Now, separately, maybe it's best not to rely on this obscure detail of `slicec` though.

bernardnormier: Seems to work.

tools/slicec-cs/src/generators/dispatch_generator.rs
bernardnormier: I didn't change this code. Keep in mind the whole code will be rewritten in C# soon.

PR #4226, opened 2026-01-13
Remove generated SliceD methods

### LINKED ISSUE #3738: Remove SliceD methods

bernardnormier:
This PR replaces the generated SliceD methods by an enhanced source generator.

The source generator relies on:
 - additional information passed to the SliceOperation attribute (e.g. is this an idempotent operation?)
 - new DispatchOperationAsync helper extension methods in IceRpc.Slice

Fixes #3738

### DIFF
--- a/src/IceRpc.Slice/SliceOperationAttribute.cs
+++ b/src/IceRpc.Slice/SliceOperationAttribute.cs
@@ -16,7 +16,7 @@ public sealed class SliceOperationAttribute : Attribute
     public string Name { get; }
 
     /// <summary>Gets a value indicating whether the return value needs to be compressed.</summary>
-    /// <value><see langword="true"/> if the return values needs to be compressed; otherwise, <see langword="false"/>.
+    /// <value><see langword="true"/> if the return value needs to be compressed; otherwise, <see langword="false"/>.
     /// </value>
     public bool CompressReturn { get; init; }
 


tests/IceRpc.Compressor.Tests/IceRpc.Compressor.Tests.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="../../tools/IceRpc.Slice.Tools/IceRpc.Slice.Tools.props" />
   <PropertyGroup>
+    <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
```
bernardnormier: It's good to see the generated code for now. We could turn this off / comment is out later.

src/IceRpc.Slice/SliceOperationAttribute.cs
```diff
@@ -12,9 +13,30 @@ public sealed class SliceOperationAttribute : Attribute
...
+    /// <summary>Gets a value indicating whether the return value needs to be compressed.</summary>
+    /// <value><see langword="true"/> if the return values needs to be compressed; otherwise, <see langword="false"/>.
+    /// </value>
+    public bool CompressReturn { get; init; }
+
+    /// <summary>Gets a value indicating whether the non-stream portion of the return value is pre-encoded by the
+    /// application.</summary>
+    /// <value><see langword="true"/> if the non-stream portion of the return value is pre-encoded; otherwise,
+    /// <see langword="false"/>.</value>
+    public bool EncodedReturn { get; init; }
```
bernardnormier, line 27: We can't deduce this info from the return type - at least not all the time.

A service method that returns a `ValueTask<PipeReader>` can correspond to either:

```slice
op() -> stream uint8
```

or an encoded return with no stream.

src/IceRpc.Slice/IncomingRequestExtensions.cs
```diff
@@ -117,6 +119,138 @@ public static ValueTask DecodeEmptyArgsAsync(
...
+    /// <param name="request">The incoming request.</param>
+    /// <param name="decodeArgs">A function that decodes the empty arguments from the request payload.</param>
+    /// <param name="method">The user-provided implementation of the operation.</param>
+    /// <param name="encodeReturnValue">A function that encodes the return value into a PipeReader.</param>
+    /// <param name="encodeReturnValueStream">A function that encodes the stream portion of the return value.</param>
+    /// <param name="inExceptionSpecification">A function that returns <see langword="true" /> when the provided Slice
+    /// exception conforms to the exception specification; otherwise, <see langword="false" />.</param>
+    /// <param name="cancellationToken">A cancellation token that receives the cancellation requests.</param>
+    /// <returns>A value task that holds the outgoing response.</returns>
+    public static async ValueTask<OutgoingResponse> DispatchOperationAsync<TReturnValue>(
```
bernardnormier, line 171: We need 4 methods instead of 1 to handle "void" args and return properly. The logic is the same in all.

src/IceRpc.Slice/SliceOperationAttribute.cs
```diff
@@ -12,9 +13,30 @@ public sealed class SliceOperationAttribute : Attribute
 {
     /// <summary>Gets the operation name.</summary>
     /// <value>The operation name.</value>
-    public string Value { get; }
+    public string Name { get; }
```
bernardnormier, line 16: I want to rename SliceOperation => SliceServiceMethod and Name => OperationName in a follow-up PR.

tests/IceRpc.Compressor.Tests/IceRpc.Compressor.Tests.csproj
pepone: We can moved this up to the DirectoryBuild.props instead of adding it to each project

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -107,7 +115,120 @@ private static string GenerateContainer(string header, string body)
...
+                }
+            }
+            else
+            {
+                // Splatting required.
+                var nonStreamReturnNames = new List<string>(serviceMethod.ReturnFieldNames);
+                if (serviceMethod.ReturnStream)
+                {
+                    nonStreamReturnNames.RemoveAt(serviceMethod.ReturnFieldNames.Length - 1);
+                }
```
pepone, line 193: Maybe is clear if the service method also provides this as a computed property `NonStreamReturnFieldNames`

src/IceRpc.Slice.Generators/Internal/Parser.cs
```diff
@@ -188,7 +192,145 @@ private IReadOnlyList<ServiceMethod> GetServiceMethods(INamedTypeSymbol interfac
...
+                    lastFieldType = methodReturnTypeArg;
+                }
+
+                if (SymbolEqualityComparer.Default.Equals(lastFieldType.OriginalDefinition, _asyncEnumerableSymbol))
+                {
+                    streamReturn = true;
+                }
+                else if (SymbolEqualityComparer.Default.Equals(lastFieldType.OriginalDefinition, _pipeReaderSymbol))
+                {
+                    streamReturn = !encodedReturn || returnCount > 1;
```
pepone, line 310: is this for stream of bytes?

src/IceRpc.Slice.Generators/Internal/Parser.cs
```diff
@@ -188,7 +192,145 @@ private IReadOnlyList<ServiceMethod> GetServiceMethods(INamedTypeSymbol interfac
...
+                    // else, the single parameter is the encoded return, and there is no stream
+                }
+            }
+            // else: It's ValueTask (non-generic), returnCount stays 0
```
pepone: and streamReturn stays false.

src/IceRpc.Slice.Generators/Internal/Parser.cs
pepone, line 310: Wouldn't be simple to also pass streamReturn in the attribute, and simplify this?

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -383,198 +379,39 @@ fn operation_declaration(operation: &Operation) -> CodeBlock {
...
-    let parameters = operation.parameters();
-    let return_parameters = operation.return_members();
-
-    let mut check_and_decode = CodeBlock::default();
-
-    if !operation.is_idempotent {
-        check_and_decode.writeln("request.CheckNonIdempotent();");
-    }
+fn operation_attribute(operation: &Operation) -> String {
+    let mut attribute = format!(r#"SliceOperation("{name}""#, name = operation.identifier());
```
InsertCreativityHere: In `slicec-cs` we have a convention of `name` meaning the _mapped_ name, and `ident` meaning the Slice identifier... But here I don't think this named intermediary does anything useful.

Rust allows you to omit these; really only useful for complex interpolation, or existing variable names.

```suggestion
    let mut attribute = format!(r#"SliceOperation("{}""#, operation.identifier());
```

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -383,198 +379,39 @@ fn operation_declaration(operation: &Operation) -> CodeBlock {
...
 
     if operation.compress_return() {
-        check_and_decode.writeln(
-            "\
-request.Features = IceRpc.Features.FeatureCollectionExtensions.With(
-    request.Features,
-    IceRpc.Features.CompressFeature.Compress);
-            ",
-        )
+        attribute.push_str(", CompressReturn = true");
```
InsertCreativityHere: Nowadays, Strings implement the `AddAssign` trait:
```suggestion
        attribute += ", CompressReturn = true";
```
Note that `+=` and `push_str` are equivalent. But I think `+=` is more natural looking unless you're working with older rust.

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -383,198 +379,39 @@ fn operation_declaration(operation: &Operation) -> CodeBlock {
...
-    {dispatch_and_return}
-}}
-catch ({catch_expression})
-{{
-    return request.CreateSliceExceptionResponse(sliceException, {encoding});
-}}",
-            dispatch_and_return = dispatch_and_return.indent(),
-        );
+    if operation.has_attribute::<CsEncodedReturn>() {
+        attribute.push_str(", EncodedReturn = true");
```
InsertCreativityHere: Same:

```suggestion
        attribute += ", EncodedReturn = true";
```

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -383,198 +379,39 @@ fn operation_declaration(operation: &Operation) -> CodeBlock {
...
-            _ => non_streamed_return_values
-                .iter()
-                .map(|r| format!("returnValue.{}", &r.field_name()))
-                .collect::<Vec<_>>()
-                .join(", "),
-        });
-
-        args = returns.join(", ") + ", ";
+    if operation.is_idempotent {
+        attribute.push_str(", Idempotent = true");
```
InsertCreativityHere:
```suggestion
        attribute += ", Idempotent = true";
```

src/IceRpc.Slice.Generators/Internal/Emitter.cs
bernardnormier: I like the consistency with the (in) parameters.

src/IceRpc.Slice.Generators/Internal/Parser.cs
bernardnormier: Yes, a stream of bytes in mapped to a PipeReader, just like an encoded return.

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -383,198 +379,39 @@ fn operation_declaration(operation: &Operation) -> CodeBlock {
...
+        let exceptions = operation
+            .exception_specification
+            .iter()
+            .map(|ex| format!("typeof({})", ex.escape_scoped_identifier(&operation.namespace())))
+            .collect::<Vec<_>>()
+            .join(", ");
+
+        attribute.push_str(&format!(
+            ", ExceptionSpecification = new System.Type[] {{ {exceptions} }}",
+            exceptions = exceptions,
```
InsertCreativityHere: Pretty sure this line is redundant. There's already a variable in scope named `exceptions`.

```suggestion
```

src/IceRpc.Slice.Generators/Internal/Parser.cs
bernardnormier: I prefer to keep the SliceOperation properties to a minimum: it's only what we strictly need and can't deduce from the other info like the signature.

That's also what we do for Protobuf: we figure out if the service method is unary or streaming (...) based on the signature, not based on properties of the attribute.

With the EncodedReturn property, we can figure out reliably when a return value field corresponds to a stream or not.

tools/slicec-cs/src/generators/dispatch_generator.rs
```diff
@@ -383,198 +379,39 @@ fn operation_declaration(operation: &Operation) -> CodeBlock {
...
+        ));
     }
+
+    attribute.push(')');
+    attribute
```
InsertCreativityHere: Simpler to do:
```suggestion
    attribute + ")"
```
Generally, if you're just doing normal string manipulation, it's best to avoid `char` in Rust.
Strings in Rust are `UTF-8`, whereas `char` are _more or less_ a `UTF-32` scalar value.

Not saying your code doesn't work, just that it's generally safer and more efficient to use strings, and not need to convert between 32 and 8 bit codepoints.

tests/IceRpc.Compressor.Tests/IceRpc.Compressor.Tests.csproj
InsertCreativityHere: I think it's good to be able to see the generated code for our tests. No harm either.

src/IceRpc.Slice.Generators/Internal/ServiceMethod.cs
```diff
@@ -2,21 +2,69 @@
 
 namespace IceRpc.Slice.Generators.Internal;
 
-/// <summary>Represents an RPC operation annotated with the <c>IceRpc.Slice.SliceOperationAttribute</c> attribute.
+/// <summary>Represents an RPC operation annotated with the <c>IceRpc.Slice.SliceDOperationAttribute</c> attribute.
```
InsertCreativityHere: Typo?
```suggestion
/// <summary>Represents an RPC operation annotated with the <c>IceRpc.Slice.SliceOperationAttribute</c> attribute.
```

src/IceRpc.Slice.Generators/Internal/Parser.cs
pepone: > That's also what we do for Protobuf: we figure out if the service method is unary or streaming (...) based on the signature, not based on properties of the attribute.

For protobuf we didn't have the option of rewriting the generated code.

> With the EncodedReturn property, we can figure out reliably when a return value field corresponds to a stream or not.

The logic is tied to a specific implementation, whether the method returns a stream or not is an Slice concept.

InsertCreativityHere (approved):

PR #4231, opened 2026-01-15
Add CodeBuilder helper classes,  update IceRpc.Slice source generator

### LINKED ISSUE #4225: Port CodeBlock and Builders to C#
We want to port the Rust CodeBlock and builders to C#.

The corresponding C# assembly must usable by the C# source generators.

bernardnormier:
This PR adds a new ZeroC.CodeBuilder library. 

It's a port of the CodeBlock and Builder classes in slicec-cs (written in Rust), largely performed by Claude.

This PR also updates the IceRpc.Slice.Generators emitter to use CodeBuilder for code formatting. I'll update the IceRpc.Protobuf.Generators emitter in a follow-up PR.

Fixes #4225.

### DIFF
--- a/src/IceRpc.Slice.Generators/Internal/Emitter.cs
+++ b/src/IceRpc.Slice.Generators/Internal/Emitter.cs
@@ -21,12 +21,12 @@ internal class Emitter
         // We need to implement IDispatcher all the time, even when there is a base class that itself implements
         // IDispatcher.
         CodeBlock container = new ContainerBuilder($"partial {serviceClass.Keyword}", serviceClass.Name)
-                .AddBase("IceRpc.IDispatcher")
-                .AddComment(
-                    "summary",
-                    "Implements <see cref=\"IceRpc.IDispatcher\" /> for the Slice interface(s) implemented by this class.")
-                .AddBlock(GenerateDispatch(serviceClass))
-                .Build();
+            .AddBase("IceRpc.IDispatcher")
+            .AddComment(
+                "summary",
+                @"Implements <see cref=""IceRpc.IDispatcher"" /> for the Slice interface(s) implemented by this class.")
+            .AddBlock(GenerateDispatch(serviceClass))
+            .Build();
 
         ContainerDefinition? containerDefinition = serviceClass;
         while (containerDefinition.Enclosing is ContainerDefinition enclosing)
@@ -44,36 +44,26 @@ internal class Emitter
 
     private static CodeBlock GenerateDispatch(ServiceClass serviceClass)
     {
-        FunctionBuilder dispatchFunctionBuilder =
-            new FunctionBuilder(
-                access: serviceClass.HasBaseServiceClass ?
-                    "public override" :
-                    serviceClass.IsSealed ? "public" : "public virtual",
-                "global::System.Threading.Tasks.ValueTask<IceRpc.OutgoingResponse>",
-                "DispatchAsync",
-                FunctionType.BlockBody)
-            .AddComment(
-                "summary",
-                @$"Dispatches an incoming request to a method of {serviceClass.Name} based on the operation name
-carried by the request.")
-            .AddParameter(
-                "IceRpc.IncomingRequest",
-                "request",
-                docComment: "The incoming request.")
-            .AddParameter(
-                "global::System.Threading.CancellationToken",
-                "cancellationToken",
-                docComment: "A cancellation token that receives the cancellation requests.")
-            .AddComment("returns", "The outgoing response.")
-            .AddComment(
-                "exception",
-                "cref",
-                "IceRpc.DispatchException",
-                @"Thrown if the operation name carried by the request does not correspond to any method implemented by this class.
-The exception status code is <see cref=""IceRpc.StatusCode.NotImplemented"" /> in this case.");
+        string methodModifier = serviceClass.HasBaseServiceClass
+            ? "public override"
+            : serviceClass.IsSealed ? "public" : "public virtual";
 
-        dispatchFunctionBuilder.SetBody(GenerateDispatchBody(serviceClass));
-        return dispatchFunctionBuilder.Build();
+        return @$"
+/// <summary>Dispatches an incoming request to a method of {serviceClass.Name} based on the operation name
+/// carried by the request.</summary>
+/// <param name=""request"">The incoming request.</param>
+/// <param name=""cancellationToken"">A cancellation token that receives the cancellation requests.</param>
+/// <returns>The outgoing response.</returns>
+/// <exception cref=""IceRpc.DispatchException"">Thrown if the operation name carried by the request does not
+/// correspond to any method implemented by this class. The exception status code is
+/// <see cref=""IceRpc.StatusCode.NotImplemented"" /> in this case.</exception>
+{methodModifier} global::System.Threading.Tasks.ValueTask<IceRpc.OutgoingResponse> DispatchAsync(
+    IceRpc.IncomingRequest request,
+    global::System.Threading.CancellationToken cancellationToken)
+{{
+    {GenerateDispatchBody(serviceClass).Indent()}
+}}
+";
     }
 
     private static CodeBlock GenerateDispatchBody(ServiceClass serviceClass)
@@ -118,14 +108,9 @@ The exception status code is <see cref=""IceRpc.StatusCode.NotImplemented"" /> i
         }
     }
 
-    private static CodeBlock GenerateDispatchCase(ServiceMethod serviceMethod)
-    {
-        var codeBlock = new CodeBlock();
-        codeBlock.WriteLine(@$"case ""{serviceMethod.OperationName}"":
-{GenerateDispatchCaseBody(serviceMethod)}");
-
-        return codeBlock;
-    }
+    private static CodeBlock GenerateDispatchCase(ServiceMethod serviceMethod) =>
+        @$"case ""{serviceMethod.OperationName}"":
+{GenerateDispatchCaseBody(serviceMethod)}";
 
     private static CodeBlock GenerateDispatchCaseBody(ServiceMethod serviceMethod)
     {
--- a/src/ZeroC.CodeBuilder/CodeBlock.cs
+++ b/src/ZeroC.CodeBuilder/CodeBlock.cs
@@ -10,8 +10,8 @@ public sealed class CodeBlock
     /// <summary>Gets the raw content of the code block.</summary>
     public string Content => _content.ToString();
 
-    /// <summary>Gets a value indicating whether the code block is empty or contains only whitespace.</summary>
-    public bool IsEmpty => _content.Length == 0 || string.IsNullOrWhiteSpace(_content.ToString());
+    /// <summary>Gets a value indicating whether the code block is empty.</summary>
+    public bool IsEmpty => _content.Length == 0;
 
     /// <summary>Combines multiple code blocks into a single code block, separated by newlines.</summary>
     /// <param name="blocks">The blocks to combine.</param>
@@ -94,10 +94,7 @@ public sealed class CodeBlock
         }
     }
 
-    /// <summary>Writes a newline to the code block.</summary>
-    public void WriteLine() => _content.AppendLine();
-
     /// <summary>Writes the specified value followed by a newline to the code block.</summary>
     /// <param name="value">The value to write.</param>
-    public void WriteLine(string value) => _content.AppendLine(value);
+    public void WriteLine(string value) => Write($"{value}\n");
 }
--- a/src/ZeroC.CodeBuilder/ContainerBuilder.cs
+++ b/src/ZeroC.CodeBuilder/ContainerBuilder.cs
@@ -119,17 +119,9 @@ public sealed class ContainerBuilder : IBuilder, IAttributeBuilder<ContainerBuil
 
         var bodyContent = CodeBlock.FromBlocks(_contents);
 
-        if (bodyContent.IsEmpty)
-        {
-            code.WriteLine("{");
-            code.WriteLine("}");
-        }
-        else
-        {
-            code.WriteLine("{");
-            code.WriteLine($"    {bodyContent.Indent()}");
-            code.WriteLine("}");
-        }
+        code.WriteLine("{");
+        code.WriteLine($"    {bodyContent.Indent()}");
+        code.WriteLine("}");
 
         return code;
     }
--- a/src/ZeroC.CodeBuilder/FunctionBuilder.cs
+++ b/src/ZeroC.CodeBuilder/FunctionBuilder.cs
@@ -68,11 +68,9 @@ public sealed class FunctionBuilder : IBuilder, IAttributeBuilder<FunctionBuilde
 
     /// <summary>Adds the EditorBrowsable(Never) attribute.</summary>
     /// <returns>This builder instance for method chaining.</returns>
-    public FunctionBuilder AddNeverEditorBrowsableAttribute()
-    {
-        return AddAttribute(
+    public FunctionBuilder AddNeverEditorBrowsableAttribute() =>
+        AddAttribute(
             "global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)");
-    }
 
     /// <summary>Adds a parameter to the function.</summary>
     /// <param name="paramType">The parameter type.</param>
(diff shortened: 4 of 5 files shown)


src/IceRpc.Slice.Generators/Internal/StringExtensions.cs
```diff
@@ -21,10 +21,4 @@ internal static string ReplaceLineEndings(this string contents)
         }
         return contents;
     }
-
-    internal static string TrimTrailingWhiteSpaces(this string contents) =>
```
bernardnormier, line 25: Unused

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -1,139 +1,147 @@
...
-
     internal string Emit(ServiceClass serviceClass, CancellationToken cancellationToken)
     {
         // Stop if we're asked to.
         cancellationToken.ThrowIfCancellationRequested();
 
-        string methodModifier =
-            serviceClass.HasBaseServiceClass ? "public override" :
-            serviceClass.IsSealed ? "public" : "public virtual";
+        CodeBlock codeBlock = Preamble();
```
bernardnormier, line 14: The code is very different: looking at the diff is not that helpful.

src/ZeroC.CodeBuilder/FunctionBuilder.cs
```diff
@@ -0,0 +1,203 @@
...
+        switch (_functionType)
+        {
+            case FunctionType.Declaration:
+                code.WriteLine(";");
+                break;
+
+            case FunctionType.ExpressionBody:
+                if (_body.IsEmpty)
+                {
+                    code.WriteLine(" => {};");
```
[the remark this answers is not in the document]
bernardnormier: Added missing space.

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -1,139 +1,147 @@
...
+    private static CodeBlock GenerateDispatchBody(ServiceClass serviceClass)
+    {
         if (serviceClass.ServiceMethods.Count > 0)
         {
-            dispatchImplementation = "";
+            var cases = new CodeBlock();
             foreach (ServiceMethod serviceMethod in serviceClass.ServiceMethods)
             {
-                dispatchImplementation += GenerateDispatchCase(serviceMethod);
+                cases.AddBlock(GenerateDispatchCase(serviceMethod).Indent());
```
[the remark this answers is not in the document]
bernardnormier: It's intentional. Added comment.

src/ZeroC.CodeBuilder/FunctionCallBuilder.cs
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) ZeroC, Inc.
+
+namespace ZeroC.CodeBuilder;
+
+/// <summary>A builder for creating function call expressions.</summary>
+public sealed class FunctionCallBuilder : IBuilder
```
bernardnormier, line 6: Once all the Rust codegen code is ported to C#, we should rename all these functions methods. Functions is not the correct term here.

src/ZeroC.CodeBuilder/CodeBlock.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    /// <summary>Writes the specified value to the code block if it is not empty or whitespace.</summary>
+    /// <param name="value">The value to write.</param>
+    public void Write(string value)
+    {
+        if (!string.IsNullOrWhiteSpace(value))
+        {
+            _content.Append(value);
+        }
+    }
```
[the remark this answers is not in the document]
bernardnormier: I am guessing this was ported straight for Rust, and I do indeed find this behavior strange. What's the use-case ?

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -221,13 +244,28 @@ private string GenerateDispatchCase(ServiceMethod serviceMethod)
...
+#nullable enable
+
+#pragma warning disable CS0612 // Type or member is obsolete
+#pragma warning disable CS0618 // Type or member is obsolete
+#pragma warning disable CS0619 // Type or member is obsolete
+
+using IceRpc.Slice;
+using ZeroC.Slice;
+
+";
```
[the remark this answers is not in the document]
bernardnormier: Added missing ReplaceLineEndings().

src/ZeroC.CodeBuilder/CodeBlock.cs
```diff
@@ -0,0 +1,103 @@
...
+
+/// <summary>Represents a block of generated code that can be manipulated and formatted.</summary>
+public sealed class CodeBlock
+{
+    /// <summary>Gets the raw content of the code block.</summary>
+    public string Content => _content.ToString();
+
+    /// <summary>Gets a value indicating whether the code block is empty or contains only whitespace.</summary>
+    public bool IsEmpty => _content.Length == 0 || string.IsNullOrWhiteSpace(_content.ToString());
+
```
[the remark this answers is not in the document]
bernardnormier: It's another method that doesn't make a lot of sense. When is the content ever whitespace-only?

src/ZeroC.CodeBuilder/ICommentBuilder.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <returns>The builder instance for method chaining.</returns>
+    T AddComment(string tag, string content);
+
+    /// <summary>Adds an XML documentation comment with an attribute.</summary>
+    /// <param name="tag">The comment tag.</param>
+    /// <param name="attributeName">The attribute name.</param>
+    /// <param name="attributeValue">The attribute value.</param>
+    /// <param name="content">The comment content.</param>
+    /// <returns>The builder instance for method chaining.</returns>
+    T AddComment(string tag, string attributeName, string attributeValue, string content);
```
pepone, line 22: I think is preferable to have a single AddComment that takes the comment tag. But we can refactor later.

Thinking this will help if we evolve CommentTag to take multiple attributes:

```
builder.AddComment(
    new CommentTag("param", "The parameter description")
        .AddAttribute("name", "value")
        .AddAttribute("cref", "SomeType"));
```

src/ZeroC.CodeBuilder/FunctionBuilder.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        return this;
+    }
+
+    /// <summary>Adds the SetsRequiredMembers attribute.</summary>
+    /// <returns>This builder instance for method chaining.</returns>
+    public FunctionBuilder AddSetsRequiredMembersAttribute()
+    {
+        return AddAttribute("global::System.Diagnostics.CodeAnalysis.SetsRequiredMembers");
+    }
```
pepone: We should use expression-bodied members `=>` here and other single line functions.

src/ZeroC.CodeBuilder/FunctionCallBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            AddArgument(argument);
+        }
+        return this;
+    }
+
+    /// <summary>Adds an argument to the function call if it is not null.</summary>
+    /// <param name="argument">The optional argument to add.</param>
+    /// <returns>This builder instance for method chaining.</returns>
+    public FunctionCallBuilder AddArgumentIfPresent(string? argument)
```
pepone: We should rename this to `IfNotNull` sounds better for  the C# version.

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -1,139 +1,148 @@
...
+
+        if (serviceClass.ContainingNamespace is not null)
+        {
+            codeBlock.AddBlock($"namespace {serviceClass.ContainingNamespace};");
+        }
+
+        // We need to implement IDispatcher all the time, even when there is a base class that itself implements
+        // IDispatcher.
+        CodeBlock container = new ContainerBuilder($"partial {serviceClass.Keyword}", serviceClass.Name)
+                .AddBase("IceRpc.IDispatcher")
```
pepone: indentation looks bogus.

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -1,139 +1,148 @@
...
+                .AddComment(
+                    "summary",
+                    "Implements <see cref=\"IceRpc.IDispatcher\" /> for the Slice interface(s) implemented by this class.")
```
pepone: should we prefer `"""` raw strings?

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -1,139 +1,148 @@
...
+            .AddComment(
+                "exception",
+                "cref",
+                "IceRpc.DispatchException",
+                @"Thrown if the operation name carried by the request does not correspond to any method implemented by this class.
+The exception status code is <see cref=""IceRpc.StatusCode.NotImplemented"" /> in this case.");
+
+        dispatchFunctionBuilder.SetBody(GenerateDispatchBody(serviceClass));
+        return dispatchFunctionBuilder.Build();
+    }
```
pepone: I'm not convinced that we should do it this way, what about using templates for things that are mostly static, and keep the builders for things that have more dynamic nature.

```cs
private static CodeBlock GenerateDispatch(ServiceClass serviceClass)
{
    string methodModifier = serviceClass.HasBaseServiceClass
        ? "public override"
        : serviceClass.IsSealed ? "public" : "public virtual";

    return $"""
        /// <summary>Dispatches an incoming request to a method of {serviceClass.Name} based on the operation name
        /// carried by the request.</summary>
        /// <param name="request">The incoming request.</param>
        /// <param name="cancellationToken">A cancellation token that receives the cancellation requests.</param>
        /// <returns>The outgoing response.</returns>
        {methodModifier} global::System.Threading.Tasks.ValueTask<IceRpc.OutgoingResponse> DispatchAsync(
            global::System.Threading.CancellationToken cancellationToken)
        {{
            {GenerateDispatchBody(serviceClass).Indent()}
        }}
        """;
}
```

I think the template here helps see what is getting generated.

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -1,139 +1,148 @@
...
+            return "return new(new IceRpc.OutgoingResponse(request, IceRpc.StatusCode.NotImplemented));";
         }
-
-        generated += container.Trim();
-        generated += "\n";
-        return generated.ReplaceLineEndings();
     }
 
-    private static string GenerateContainer(string header, string body)
+    private static CodeBlock GenerateDispatchCase(ServiceMethod serviceMethod)
```
pepone: This can use the implicit string conversion:

```
    private static CodeBlock GenerateDispatchCase(ServiceMethod serviceMethod)
    {
        return @$"case ""{serviceMethod.OperationName}"":
{GenerateDispatchCaseBody(serviceMethod)}";
    }
```

src/ZeroC.CodeBuilder/CodeBlock.cs
```diff
@@ -0,0 +1,103 @@
...
+namespace ZeroC.CodeBuilder;
+
+/// <summary>Represents a block of generated code that can be manipulated and formatted.</summary>
+public sealed class CodeBlock
+{
+    /// <summary>Gets the raw content of the code block.</summary>
+    public string Content => _content.ToString();
+
+    /// <summary>Gets a value indicating whether the code block is empty or contains only whitespace.</summary>
+    public bool IsEmpty => _content.Length == 0 || string.IsNullOrWhiteSpace(_content.ToString());
```
pepone: Maybe we can improve this to not call to string, but set IsEmpty to false internally when we write non empty string, or add a non empty code block. Shouldn't be difficult and avoid converting to string to just check if is empty.

src/ZeroC.CodeBuilder/CodeBlock.cs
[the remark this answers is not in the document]
pepone: not clear why we have this.

src/ZeroC.CodeBuilder/CodeBlock.cs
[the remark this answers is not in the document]
externl: This is actually quite common in the rust compiler. For instance when we do something like:

```
cb.writeln("startSomething();");
cb.writeln(generate_something_that_may_be_empty());
cb.writeln("sendSomething();");
```

src/ZeroC.CodeBuilder/CodeBlock.cs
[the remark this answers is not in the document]
externl: It's so we don't write whitespace only lines. This is by design. We never want to write whitespace only lines. 

We often call write from `WriteLine` or when merging CodeBlocks by calling `Write("\n{var}\n")` or `Write("{var\n")`.  These can empty in various circumstances.

src/ZeroC.CodeBuilder/ContainerBuilder.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            code.WriteLine("{");
+            code.WriteLine("}");
+        }
+        else
+        {
+            code.WriteLine("{");
+            code.WriteLine($"    {bodyContent.Indent()}");
+            code.WriteLine("}");
+        }
```
externl: Only the else case is needed here. CodeBlock will deal with the empty bodyContent line.

src/ZeroC.CodeBuilder/FunctionBuilder.cs
```diff
@@ -0,0 +1,203 @@
...
+                    code.WriteLine($" =>\n    {_body.Indent()};");
+                }
+                break;
+
+            case FunctionType.BlockBody:
+                if (_body.IsEmpty)
+                {
+                    code.WriteLine("");
+                    code.WriteLine("{");
+                    code.WriteLine("}");
```
externl: This is buggy. `code.WriteLine("");` does nothing.

externl: This if else should be reduced to:

```
                    code.WriteLine("\n{");
                    code.WriteLine($"    {_body.Indent()}");
                    code.WriteLine("}");
```
or

```
                    code.WriteLine($"\n{{\n    {_body.Indent()}}}");
```

src/ZeroC.CodeBuilder/CodeBlock.cs
bernardnormier: I am going to remove IsEmpty until we understand the use-case.

src/IceRpc.Slice.Generators/Internal/Emitter.cs
```diff
@@ -1,139 +1,148 @@
...
+                "global::System.Threading.CancellationToken",
+                "cancellationToken",
+                docComment: "A cancellation token that receives the cancellation requests.")
+            .AddComment("returns", "The outgoing response.")
+            .AddComment(
+                "exception",
+                "cref",
+                "IceRpc.DispatchException",
+                @"Thrown if the operation name carried by the request does not correspond to any method implemented by this class.
+The exception status code is <see cref=""IceRpc.StatusCode.NotImplemented"" /> in this case.");
```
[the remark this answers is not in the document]
bernardnormier: This is fine.

src/ZeroC.CodeBuilder/FunctionCallBuilder.cs
bernardnormier: Fixed as suggested

src/IceRpc.Slice.Generators/Internal/Emitter.cs
bernardnormier: Yes, fixed

src/IceRpc.Slice.Generators/Internal/Emitter.cs
bernardnormier: Fixed as suggested. The exception was missing above.

src/ZeroC.CodeBuilder/CodeBlock.cs
externl: We use the Rust version in several places.

bernardnormier: Actually we use it so I kept it with a simpler implementation.

src/ZeroC.CodeBuilder/FunctionBuilder.cs
bernardnormier: Fixed

src/ZeroC.CodeBuilder/ContainerBuilder.cs
bernardnormier: Fixed

src/ZeroC.CodeBuilder/FunctionBuilder.cs
bernardnormier: Fixed

PR #4233, opened 2026-01-16
Don't mark Protobuf services or service methods Obsolete

bernardnormier:
This PR fixes the ProtoGen generated code to _not_ mark deprecated services and service methods Obsolete in C#.

Obsolete only makes senses for the caller (Client). See ClientGenerator.cs.


tools/IceRpc.ProtocGen/ServiceGenerator.cs
```diff
@@ -38,11 +38,6 @@ public static string GenerateInterface(ServiceDescriptor service)
     /// <param name=""features"">The dispatch features.</param>
     /// <param name=""cancellationToken"">A cancellation token that receives the cancellation requests.</param>
     /// <returns>{returnTypeDocComment}</returns>";
```
bernardnormier, line 40: Will refactor to use CodeBuilder in a follow-up PR.

PR #4235, opened 2026-01-17
Update ProtocGen to use CodeBuilder

bernardnormier:
This PR refactors ProtocGen to use CodeBuilder.

Other than formatting changes, the generated code should be identical.

### DIFF
--- a/tools/IceRpc.ProtocGen/Program.cs
+++ b/tools/IceRpc.ProtocGen/Program.cs
@@ -61,7 +61,7 @@ using IceRpc.Protobuf;
         new CodeGeneratorResponse.Types.File
         {
             Name = $"{Path.GetFileNameWithoutExtension(descriptor.Name).ToPascalCase()}.IceRpc.cs",
-            Content = codeBlock.Content.ReplaceLineEndings()
+            Content = codeBlock.ToString().ReplaceLineEndings()
         });
 }
 


src/ZeroC.CodeBuilder/FunctionBuilder.cs
```diff
@@ -155,7 +149,8 @@ public CodeBlock Build()
                 }
                 else
                 {
-                    code.WriteLine($" =>\n    {_body.Indent()};");
+                    code.WriteLine(@$" =>
+        {_body.Indent()};");
```
bernardnormier, line 153: Extra indent: this seems more correct, but I let the experts weigh in.

src/ZeroC.CodeBuilder/FunctionCallBuilder.cs
```diff
@@ -64,7 +64,7 @@ public FunctionCallBuilder AddArguments(IEnumerable<string> arguments)
     /// <summary>Sets whether arguments should be placed on separate lines.</summary>
     /// <param name="argumentsOnNewline">True to place arguments on new lines.</param>
     /// <returns>This builder instance for method chaining.</returns>
-    public FunctionCallBuilder ArgumentsOnNewline(bool argumentsOnNewline)
+    public FunctionCallBuilder ArgumentsOnNewline(bool argumentsOnNewline = true)
```
bernardnormier, line 67: In C#, Newline is spelled NewLine:
https://learn.microsoft.com/en-us/dotnet/api/system.environment.newline?view=net-10.0

Need to fix in a follow-up PR.

src/ZeroC.CodeBuilder/IAttributeBuilder.cs
```diff
@@ -12,3 +12,23 @@ public interface IAttributeBuilder<T>
     /// <returns>The builder instance for method chaining.</returns>
     T AddAttribute(string attribute);
 }
+
+/// <summary>Extension methods for <see cref="IAttributeBuilder{T}"/>.</summary>
+public static class AttributeBuilderExtensions
```
bernardnormier, line 17: It's likely other methods should move from builder-class methods to builder-interface extensions.

tools/IceRpc.ProtocGen/ServiceDescriptorExtensions.cs
```diff
@@ -1,11 +0,0 @@
-// Copyright (c) ZeroC, Inc.
-
-using Google.Protobuf.Reflection;
-
-namespace IceRpc.ProtocGen;
-
-internal static class ServiceDescriptorExtensions
```
bernardnormier, line 7: Unused

tools/IceRpc.ProtocGen/StringExtensions.cs
```diff
@@ -1,33 +0,0 @@
-// Copyright (c) ZeroC, Inc.
-
-using System.Text;
-
-namespace IceRpc.ProtocGen;
-
-internal static class StringExtensions
```
bernardnormier, line 7: No longer need with CodeBuilder.

src/ZeroC.CodeBuilder/FunctionBuilder.cs
pepone: What about:

```
                    code.WriteLine(" =>");
                    code.WriteLine($"    {_body.Indent()};");
```

pepone (approved):

src/ZeroC.CodeBuilder/FunctionBuilder.cs
bernardnormier: There are 8 spaces before the `{_body.Indent()}`.

externl: Is 8 spaces not too many?

externl: I wonder is there's just a missing indent call else where. Pretty sure this should be only be 4 spaces. I would expect the body to need two indent() calls as well here.

PR #4242, opened 2026-01-21
Move tools C# projects to src and IceRpc.slnx

bernardnormier:
This PR moves all C# tools project to the src directory and the IceRpc.slnx solution.

There are two main updates to the project files:

- project references to the Tools projects with `ReferenceOutputAssembly="false"`, for example:

```
<ProjectReference Include="../IceRpc.Slice.Tools/IceRpc.Slice.Tools.csproj" ReferenceOutputAssembly="false" />
```

- when we detect a source build (in ...Tools.props), don't assume the Tools assembly is already built

### DIFF
--- a/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
+++ b/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
@@ -8,7 +8,6 @@
     <Title>IceRPC Protobuf Tools</Title>
     <PackageId>$(AssemblyName)</PackageId>
     <PackageIcon>icerpc-icon.png</PackageIcon>
-    <Copyright>Copyright (c) ZeroC, Inc.</Copyright>
     <Description>Provides tools to generate C# code from Protobuf definitions; includes support for MSBuild projects.</Description>
     <DevelopmentDependency>true</DevelopmentDependency>
     <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
--- a/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.props
+++ b/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.props
@@ -20,8 +20,6 @@
       <PropertyGroup>
         <IceRpcProtocPath>$(MSBuildThisFileDirectory)obj/tools/</IceRpcProtocPath>
         <IceRpcProtocGenPath>$(MSBuildThisFileDirectory)../IceRpc.ProtocGen/bin/$(Configuration)/net10.0/</IceRpcProtocGenPath>
-        <!-- Disable build telemetry when using an IceRPC source builds -->
-        <IceRpcBuildTelemetry>false</IceRpcBuildTelemetry>
 
         <!-- Use local build of IceRpc.Protobuf.Tools -->
         <IceRpcProtobufToolsTaskAssembliesPath>$(MSBuildThisFileDirectory)bin/$(Configuration)/net10.0/</IceRpcProtobufToolsTaskAssembliesPath>
--- a/src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.props
+++ b/src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.props
@@ -16,8 +16,6 @@
       <PropertyGroup>
         <SliceCompilerPath Condition="'$(Configuration)' == 'Debug'">$(MSBuildThisFileDirectory)../../tools/slicec-cs/target/debug</SliceCompilerPath>
         <SliceCompilerPath Condition="'$(Configuration)' == 'Release'">$(MSBuildThisFileDirectory)../../tools/slicec-cs/target/release</SliceCompilerPath>
-        <!-- Disable build telemetry when using an IceRPC source builds -->
-        <IceRpcBuildTelemetry>false</IceRpcBuildTelemetry>
 
         <!-- Use tools assembly built from source -->
         <SliceToolsTaskAssembliesPath>$(MSBuildThisFileDirectory)bin/$(Configuration)/net10.0/</SliceToolsTaskAssembliesPath>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -3,8 +3,8 @@
   <Import Project="$(MSBuildThisFileDirectory)../build/IceRpc.Src.props" />
   <PropertyGroup>
     <!-- Include PDB in the built .nupkg -->
-    <AllowedOutputExtensionsInPackageBuildOutputFolder
-    >$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <!-- Disable build telemetry -->
     <IceRpcBuildTelemetry>false</IceRpcBuildTelemetry>
     <!-- Enable package lock file to ensure consistent package versions across builds -->
     <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
--- a/tests/Directory.Build.props
+++ b/tests/Directory.Build.props
@@ -7,6 +7,7 @@
     <GlobalAnalyzerConfigFiles Include="$(MSBuildThisFileDirectory)../build/CodeAnalysis.Tests.globalconfig" />
   </ItemGroup>
   <PropertyGroup>
+     <!-- Turn off build telemetry -->
     <IceRpcBuildTelemetry>false</IceRpcBuildTelemetry>
     <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
     <!-- Enable package lock file to ensure consistent package versions across builds -->


src/IceRpc.Slice.Tools/.editorconfig
```diff
@@ -9,3 +9,6 @@ dotnet_diagnostic.CA1062.severity = none
 # CA1819 PropertiesShouldNotReturnArrays
 # The MSBuild task model requires the use of arrays for properties that are collections.
 dotnet_diagnostic.CA1819.severity = none
+
+# CS1591 Missing XML comment for publicly visible type or member
```
bernardnormier, line 13: That's for public class we encode in JSON. We should add doc-comment in a follow-up PR.

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.csproj
```diff
@@ -5,17 +5,12 @@
     <AssemblyName>IceRpc.Slice.Tools</AssemblyName>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
     <TargetFramework>net10.0</TargetFramework>
-    <LangVersion>12.0</LangVersion>
-    <AnalysisMode>AllEnabledByDefault</AnalysisMode>
     <Title>IceRPC Slice Tools</Title>
```
bernardnormier, line 8: It's likely the csproj should be further simplified. Now that it's in src, it inherits many common properties.

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.props
```diff
@@ -11,40 +11,27 @@
     >$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString().ToLower())</IceRpcOSArch>
   </PropertyGroup>
   <Choose>
-    <When Condition="Exists('$(MSBuildThisFileDirectory)../slicec-cs')">
+    <When Condition="Exists('$(MSBuildThisFileDirectory)../../tools/slicec-cs')">
       <!-- Use the Slice compiler from this source build -->
       <PropertyGroup>
-        <SliceCompilerPath
```
bernardnormier, line 17: This line splitting made it hard to read.

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.props
```diff
@@ -11,40 +11,27 @@
...
-        >$(MSBuildThisFileDirectory)../slicec-cs/target/debug</SliceCompilerPath>
-        <SliceCompilerPath
-          Condition="'$(Configuration)' == 'Release'"
-        >$(MSBuildThisFileDirectory)../slicec-cs/target/release</SliceCompilerPath>
+        <SliceCompilerPath Condition="'$(Configuration)' == 'Debug'">$(MSBuildThisFileDirectory)../../tools/slicec-cs/target/debug</SliceCompilerPath>
+        <SliceCompilerPath Condition="'$(Configuration)' == 'Release'">$(MSBuildThisFileDirectory)../../tools/slicec-cs/target/release</SliceCompilerPath>
         <!-- Disable build telemetry when using an IceRPC source builds -->
         <IceRpcBuildTelemetry>false</IceRpcBuildTelemetry>
+
+        <!-- Use tools assembly built from source -->
```
bernardnormier, line 20: Even if it's not built yet, which is critical here.

src/ZeroC.CodeBuilder/CodeBlock.cs
```diff
@@ -27,7 +27,13 @@ public static CodeBlock FromBlocks(IEnumerable<CodeBlock> blocks)
...
     /// <summary>Converts a string to a <see cref="CodeBlock"/>.</summary>
     /// <param name="content">The string content.</param>
-    public static implicit operator CodeBlock(string content) => new(content);
+    /// <returns>A new <see cref="CodeBlock"/>.</returns>
+    public static implicit operator CodeBlock(string content) => FromString(content);
+
+    /// <summary>Creates a <see cref="CodeBlock"/> from a string.</summary>
+    /// <param name="content">The string content.</param>
+    /// <returns>A new <see cref="CodeBlock"/>.</returns>
+    public static CodeBlock FromString(string content) => new(content);
```
bernardnormier, line 36: The code analysis / linting in tools didn't work properly ... but the src one does.

.github/actions/build/action.yaml
```diff
@@ -60,11 +60,6 @@ runs:
         cargo build --manifest-path tools/slicec-cs/Cargo.toml ${{ inputs.cargo-build-args }}
       shell: bash
 
-    - name: 🔨 Build IceRPC Tools
```
bernardnormier, line 63: It's not clear to me why these actions don't use the build scripts.

Is it because the build scripts are temporary?

IceRpc.slnx
```diff
@@ -13,10 +13,13 @@
     <Project Path="src/IceRpc.Metrics/IceRpc.Metrics.csproj" />
```
pepone, line 13: We should sort these files, in a follow-up PR.

src/IceRpc.Locator/IceRpc.Locator.csproj
```diff
@@ -10,6 +10,7 @@
   <ItemGroup>
     <ProjectReference Include="../IceRpc/IceRpc.csproj" ExactVersion="true" />
     <ProjectReference Include="../IceRpc.Slice/IceRpc.Slice.csproj" ExactVersion="true" />
+    <ProjectReference Include="../IceRpc.Slice.Tools/IceRpc.Slice.Tools.csproj" ReferenceOutputAssembly="false" />
```
pepone, line 13: I think we can add this in the imported `.targets` like in the example from 

https://github.com/rainersigwald/build-task-in-solution-demo/blob/ddd8f884901017089c196d98742452df80a31cf5/Task/DemoFunctionality.targets#L12-L17

In our case we need a Condition to do it only for source builds.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
```diff
@@ -5,15 +5,8 @@
...
-    <AnalysisMode>AllEnabledByDefault</AnalysisMode>
     <Title>IceRPC Protobuf Tools</Title>
-    <Authors>ZeroC, Inc.</Authors>
     <PackageId>$(AssemblyName)</PackageId>
-    <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
-    <PackageProjectUrl>https://github.com/icerpc/icerpc-csharp</PackageProjectUrl>
-    <RepositoryUrl>https://github.com/icerpc/icerpc-csharp</RepositoryUrl>
-    <RepositoryType>git</RepositoryType>
     <PackageIcon>icerpc-icon.png</PackageIcon>
     <Copyright>Copyright (c) ZeroC, Inc.</Copyright>
```
pepone: We don't have this Copyright in other src/ projects

```
jose@mac src % find . -name "*.csproj" | xargs fgrep Copyr
./IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj:    <Copyright>Copyright (c) ZeroC, Inc.</Copyright>
```

build/Common.props adds it.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.props
```diff
@@ -19,10 +19,12 @@
       -->
       <PropertyGroup>
         <IceRpcProtocPath>$(MSBuildThisFileDirectory)obj/tools/</IceRpcProtocPath>
-        <IceRpcProtocGenPath
-        >$(MSBuildThisFileDirectory)../IceRpc.ProtocGen/bin/$(Configuration)/net10.0/</IceRpcProtocGenPath>
+        <IceRpcProtocGenPath>$(MSBuildThisFileDirectory)../IceRpc.ProtocGen/bin/$(Configuration)/net10.0/</IceRpcProtocGenPath>
         <!-- Disable build telemetry when using an IceRPC source builds -->
         <IceRpcBuildTelemetry>false</IceRpcBuildTelemetry>
```
pepone: We are setting this in `src/DirectoryBuild.props` too, we should decide which approach is more correct but I think we don't need both.

Probably this, as it applies when using the tools source build independently of the project where you use it.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.props
```diff
@@ -19,10 +19,12 @@
       -->
       <PropertyGroup>
         <IceRpcProtocPath>$(MSBuildThisFileDirectory)obj/tools/</IceRpcProtocPath>
```
pepone, line 21: I know I added this properties originally but seems didn't have time to write a comment, we should in a follow up PR :)

src/IceRpc.ProtobufBuildTelemetry.Reporter/IceRpc.ProtobufBuildTelemetry.Reporter.csproj
```diff
@@ -14,6 +14,7 @@
     <SliceFile Include="../IceRpc.SliceBuildTelemetry.Reporter/BuildTelemetry.slice" />
     <PackageReference Include="System.CommandLine" Version="2.0.0" />
     <ProjectReference Include="../IceRpc.Slice/IceRpc.Slice.csproj" />
+    <ProjectReference Include="../IceRpc.Slice.Tools/IceRpc.Slice.Tools.csproj" ReferenceOutputAssembly="false" />
```
pepone, line 17: Ditto we better move this the .targets file.

src/IceRpc.Locator/IceRpc.Locator.csproj
bernardnormier: For a follow-up PR

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
bernardnormier: Removed.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.props
bernardnormier: src/DirectoryBuild.props is sufficient. Removed.

.github/actions/build/action.yaml
pepone: We can probably simplify the workflows.

PR #4244, opened 2026-01-22
Move Tools project reference to .targets

bernardnormier:
This PR includes two small follow-up to #4242:

- project references to the Tools projects are now in the Tools .targets instead of directly in the referencing .csproj
- each folder of IceRpc.slnx is now sorted alphabetically

### DIFF
--- a/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
+++ b/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
@@ -28,6 +28,7 @@
     Runtime="NET"
   />
   <ItemGroup Condition="'$(IceRpcProtobufToolsSourceBuild)' == 'true'">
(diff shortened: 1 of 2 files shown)


src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -27,6 +27,9 @@
     AssemblyFile="$(IceRpcProtobufToolsTaskAssembliesPath)IceRpc.Protobuf.Tools.dll"
     Runtime="NET"
   />
+  <ItemGroup Condition="'$(IceRpcProtobufToolsSourceBuild)' == 'true'">
```
pepone, line 30: I will add a comment here like in the MSbuild examples

```
 <!-- In source builds add a ProjectReference to ensure that the task gets built before it needs to be called. -->
```

pepone (approved):

PR #4249, opened 2026-01-25
Convert Protobuf build telemetry to protoc plug-in

bernardnormier:
This PR converts the build telemetry for Protobuf to use a protoc plug-in, instead of a "normal" program.

It also updates the IceRpcBuildTelemetry MSBuild property to be a boolean property. If you set it to a non-boolean value (say 'foo'), you'll get a build error. Previously, any value other than 'false' was treated as 'true'.

### DIFF
--- a/src/IceRpc.Protobuf.Tools/ProtocTask.cs
+++ b/src/IceRpc.Protobuf.Tools/ProtocTask.cs
@@ -165,7 +165,7 @@ public class ProtocTask : ToolTask
         }
         else
         {
-            Log.LogMessage(singleLine);
+            Log.LogError(singleLine);
         }
     }
 


src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
```diff
@@ -15,18 +15,13 @@
     <NoWarn>NU5100</NoWarn>
   </PropertyGroup>
   <ItemGroup>
+    <ProjectReference Include="../IceRpc.ProtocGen/IceRpc.ProtocGen.csproj" ReferenceOutputAssembly="false" />
```
bernardnormier, line 18: We want to build these projects _before_ IceRpc.Protobuf.Tools since tasks in .targets rely on these projects' assemblies.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.props
```diff
@@ -21,6 +24,7 @@
         <IceRpcProtobufToolsSourceBuild>true</IceRpcProtobufToolsSourceBuild>
         <IceRpcProtocPath>$(MSBuildThisFileDirectory)obj/tools/</IceRpcProtocPath>
         <IceRpcProtocGenPath>$(MSBuildThisFileDirectory)../IceRpc.ProtocGen/bin/$(Configuration)/net10.0/</IceRpcProtocGenPath>
+        <IceRpcProtocBuildTelemetryPath>$(MSBuildThisFileDirectory)../IceRpc.ProtocGen.BuildTelemetry/bin/$(Configuration)/net10.0/</IceRpcProtocBuildTelemetryPath>
```
bernardnormier, line 27: Indentation looks off.

src/IceRpc.Protobuf.Tools/ProtocTask.cs
```diff
@@ -130,13 +156,21 @@ protected override void LogEventsFromTextOutput(string singleLine, MessageImport
...
         {
             Debug.Assert(singleLine is not null);
-            string[] parts = singleLine.Split([':'], 4);
-            string fileName = parts[0];
-            int lineNumber = int.Parse(parts[1], CultureInfo.InvariantCulture);
-            int columnNumber = int.Parse(parts[2], CultureInfo.InvariantCulture);
-            string errorMessage = parts[3];
-
-            Log.LogError("", "", "", fileName, lineNumber, columnNumber, -1, -1, errorMessage);
+            if (singleLine.Contains(':', StringComparison.Ordinal))
```
bernardnormier: The previous code didn't handle properly non-file-specific error message from protoc. This code should be further improved - what if the error message has a single `:` for some reason?

src/IceRpc.ProtocGen.BuildTelemetry/Program.cs
```diff
@@ -0,0 +1,143 @@
...
+
+if (fileCount > 0)
+{
+    string compilationHash = Convert.ToHexString(SHA256.HashData(hashBytes.ToArray())).ToLowerInvariant();
+
+    // Determine the IceRPC version using the assembly version.
+    var assembly = Assembly.GetAssembly(typeof(Program));
+    string toolVersion = assembly!.GetName().Version!.ToString();
+
+    var protobufTelemetryData = new ProtobufTelemetryData(
```
bernardnormier, line 63: All this is code is just like the previous Protobuf build telemetry reported. In particular:
 - same Slice definitions
 - no QUIC

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
pepone: We need to update the "Build roadmap" in BUILDING.md

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
```diff
@@ -52,10 +47,17 @@
       <PackagePath />
     </None>
     <None Include="$(IntermediateOutputPath)/$(AssemblyName).dll" Pack="true" PackagePath="tasks/" Visible="false" />
-    <!-- Use the compilers from this source build. -->
+    <!-- Include the protoc plug-ins from this source build. -->
```
pepone, line 50: Maybe is clear to create a folder in tools for each plug-in.

src/IceRpc.ProtocGen.BuildTelemetry/Program.cs
```diff
@@ -0,0 +1,143 @@
...
+IEnumerable<byte> hashBytes = [];
+
+foreach (FileDescriptor descriptor in descriptors)
+{
+    if (!request.FileToGenerate.Contains(descriptor.Name))
+    {
+        // This descriptor represents a reference file for which we are not generating code.
+        continue;
+    }
+    hashBytes = hashBytes.Concat(SHA256.HashData(descriptor.SerializedData.Memory.Span));
```
pepone: Previously we read the file from disk, I guess this is more correct we compute the hash over the serialized data we read from stdin. Is interesting that there is separate SerializedData for each descriptor, something we will not have with Slice.

src/IceRpc.ProtocGen.BuildTelemetry/Program.cs
```diff
@@ -0,0 +1,143 @@
...
+    {
+        string message = ex.Message;
+        if (ex.InnerException is not null)
+        {
+            message += $"\n  Inner exception: {ex.InnerException.Message}";
+        }
+        responseContent = $"Failed to report build telemetry to {uri}: {message}";
+    }
+
+    // We return a single file containing the result of the build telemetry reporting.
```
pepone, line 115: Something to keep in mind for `IceRpc.ProtoGen` but I will comment here:

I think we are not reporting errors correctly, See comments in CodeGeneratorResponse  [error field](https://github.com/protocolbuffers/protobuf/blob/a904af9cc8597eea47b3d4c657dd4f36eefa1145/src/google/protobuf/compiler/plugin.proto#L92).

I guess is fine to not report errors in this build telemetry plugin, we don't want the compilation to fail if the telemetry server is down. But we should review how this is done in `IceRpc.ProtoGen`.

src/IceRpc.Protobuf.Tools/ProtocTask.cs
pepone: I don't see where we emit error that use the ":", probably copied from another task in the initial version but I doubt this is required here.

For example BuildTelemetry doesn't emit any per-file errors, so this is most likely never used.

bernardnormier: It's errors from `protoc`. It's not build telemetry specific.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
bernardnormier: I've simplified the mermaid diagram.

src/IceRpc.Protobuf.Tools/ProtocTask.cs
pepone: I see, I misread this.

We should try passing `--error_format=msvs` to protoc which should emit errors that are already in the MSBuild format, for a follow-up PR.

src/IceRpc.ProtocGen.BuildTelemetry/Program.cs
bernardnormier: What is not correct in this plug-in?

If there is a serious parsing problem, we'll get an exception and a message on stderr => that's the correct way to report serious errors to protoc.

Otherwise, we return 0 and write a response to stdout. Also what protoc expects.

If we fill the error field of the response, protoc considers the code generation failed. We should not do that in this plug-in.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj
bernardnormier: I think it's fine as is.

src/IceRpc.ProtocGen.BuildTelemetry/Program.cs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+var response = new CodeGeneratorResponse();
+
+if (fileCount > 0)
+{
+    string compilationHash = Convert.ToHexString(SHA256.HashData(hashBytes.ToArray())).ToLowerInvariant();
+
+    // Determine the IceRPC version using the assembly version.
+    var assembly = Assembly.GetAssembly(typeof(Program));
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested.

src/IceRpc.ProtocGen.BuildTelemetry/Program.cs
```diff
@@ -0,0 +1,143 @@
...
+
+foreach (FileDescriptor descriptor in descriptors)
+{
+    if (!request.FileToGenerate.Contains(descriptor.Name))
+    {
+        // This descriptor represents a reference file for which we are not generating code.
+        continue;
+    }
+    hashBytes = hashBytes.Concat(SHA256.HashData(descriptor.SerializedData.Memory.Span));
+    fileCount++;
```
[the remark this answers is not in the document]
bernardnormier: Reworked this code.

src/IceRpc.ProtocGen.BuildTelemetry/protoc-gen-icerpc-build-telemetry.bat
```diff
@@ -0,0 +1,3 @@
+@echo OFF
+
+call dotnet %~dp0\IceRpc.ProtocGen.BuildTelemetry.dll -- %*
```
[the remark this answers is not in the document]
bernardnormier: Not a concern as far as I can tell.

src/IceRpc.Protobuf.Tools/ProtocTask.cs
```diff
@@ -61,15 +69,15 @@ public class ProtocTask : ToolTask
     /// <inheritdoc/>
```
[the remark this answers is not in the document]
bernardnormier: Removed.

PR #4258, opened 2026-02-11
Add Game of the Goose example

pepone:
A multiplayer board game example that demonstrates real-time state synchronization using IceRPC streams.

The server pushes a snapshot followed by incremental event updates to each client, showcasing bidirectional communication, streaming, discriminated unions, and request context.

### DIFF
--- a/examples/slice/GameOfTheGoose/Client/Program.cs
+++ b/examples/slice/GameOfTheGoose/Client/Program.cs
@@ -137,8 +137,11 @@ string FormatEvent(GameEvent gameEvent) => gameEvent switch
     GameEvent.ExtraTurn extra =>
         $"{state.Players[extra.Player].Name} gets an extra turn (goose)",
 
+    GameEvent.Trapped trapped =>
+        $"{state.Players[trapped.Player].Name} landed on a trap (next turn will be skipped)",
+
     GameEvent.SkipTurn skip =>
-        $"{state.Players[skip.Player].Name}'s turn will be skipped (trap)",
+        $"{state.Players[skip.Player].Name}'s turn is skipped (trap)",
 
     GameEvent.TurnAdvanced turned =>
         turned.NextPlayer == myPlayerIndex
--- a/examples/slice/GameOfTheGoose/Server/Game.cs
+++ b/examples/slice/GameOfTheGoose/Server/Game.cs
@@ -106,7 +106,7 @@ internal class Game
             if (Board.SkipTurnSpaces.Contains(playerNewPosition))
             {
                 _skipTurnPlayers.Add(playerIndex);
-                events.Add(new GameEvent.SkipTurn(playerIndex));
+                events.Add(new GameEvent.Trapped(playerIndex));
             }
             events.Add(new GameEvent.PlayerMoved(playerIndex, oldPosition, playerNewPosition));
             player.Position = playerNewPosition;
--- a/examples/slice/GameOfTheGoose/slice/Game.slice
+++ b/examples/slice/GameOfTheGoose/slice/Game.slice
@@ -62,6 +62,10 @@ enum GameEvent
     /// @param player: The index of the player who landed on the goose space.
     ExtraTurn(player: uint8)
 
+    /// A player landed on a trap space and their next turn will be skipped.
+    /// @param player: The index of the player who landed on the trap.
+    Trapped(player: uint8)
+
     /// A player's turn is being skipped because they previously landed on a trap space.
     /// @param player: The index of the player whose turn is skipped.
     SkipTurn(player: uint8)


examples/slice/GameOfTheGoose/slice/Game.slice
```diff
@@ -0,0 +1,159 @@
...
+interface GameSession
+{
+    /// Rolls the dice for the current player and applies the result to the game board.
+    rollDice() -> Result<uint8, GameError>
+}
+
+// A proxy to a service that implements GameSession; it's encoded as a service address.
+[cs::type("GameOfTheGoose.GameSessionProxy")]
+custom GameSessionProxy
+
```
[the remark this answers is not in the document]
pepone: Removed.

examples/slice/GameOfTheGoose/Client/GameState.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using GameOfTheGoose;
+
+namespace GameOfTheGooseClient;
+
+/// <summary>Thread-safe wrapper around GameState that synchronizes access from the main thread and the background
+/// event listener.</summary>
+internal class GameStateManager
+{
+    private GameState _state;
```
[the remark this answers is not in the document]
pepone: state fields are mutated.

examples/slice/GameOfTheGoose/Server/Game.cs
```diff
@@ -0,0 +1,132 @@
...
+
+using GameOfTheGoose;
+using ZeroC.Slice;
+
+namespace GameOfTheGooseServer;
+
+/// <summary>Pure game logic for the Game of the Goose. Not thread-safe — caller must synchronize.</summary>
+internal class Game
+{
+    private GameState _state;
```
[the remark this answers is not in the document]
pepone: state fields are mutated.

examples/slice/GameOfTheGoose/slice/Game.slice
```diff
@@ -0,0 +1,159 @@
...
+    GameAlreadyStarted,
+
+    /// The player attempted to make a move but the game is not in progress.
+    GameNotInProgress,
+
+    /// The player attempted to join a game that has already finished.
+    GameAlreadyFinished,
+
+    /// The player attempted to roll the dice when it is not their turn.
+    NotPlayerTurn
```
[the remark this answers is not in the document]
pepone: Removed all the superfluous commas

examples/slice/GameOfTheGoose/Client/Program.cs
```diff
@@ -0,0 +1,167 @@
...
+
+string? playerName;
+do
+{
+    Console.Write("Enter your name: ");
+    playerName = Console.ReadLine()?.Trim();
+}
+while (string.IsNullOrEmpty(playerName));
+
+var sink = new GameUpdateSinkService();
```
externl, line 21: I think the sink variable deserves a comment. As well as maybe one where we assign it as the dispatcher for the connection.

examples/slice/GameOfTheGoose/Client/Program.cs
```diff
@@ -0,0 +1,167 @@
...
+string sessionId = joined.SessionId;
+byte myPlayerIndex = joined.PlayerIndex;
+
+// Receive the snapshot and event stream from the server.
+var (snapshot, events) = await sink.Received;
+var state = new GameStateManager(snapshot.State);
+
+if (snapshot.State.Status is GameStatus.Started)
+{
+    Console.WriteLine(state.Turn == myPlayerIndex
```
externl, line 56: Can we call FormatEvent on this?

examples/slice/GameOfTheGoose/Server/Board.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Collections.Immutable;
+
+namespace GameOfTheGooseServer;
+
+/// <summary>Represents the game board for our simplified version of the Game of the Goose. The board has 40 spaces,
+/// and players start at space 0. The first player to reach space 39 wins.</summary>
+internal static class Board
+{
+    /// <summary>The number of spaces on the board. Players start at space 0 and the first to reach space 39
+    /// wins.</summary>
```
externl, line 12:
```suggestion
    /// <summary>The number of spaces on the board.</summary>
```

examples/slice/GameOfTheGoose/Server/Board.cs
```diff
@@ -0,0 +1,22 @@
...
+    public const byte TrackSize = 40;
+
+    /// <summary>Goose spaces that move you to the next goose space and grant you an extra turn. If you land on
+    /// space 5, you move to space 9 and get an extra turn. Landing on the last goose space moves you to the final
+    /// track space.</summary>
+    public static readonly ImmutableList<byte> GooseSpaces = [5, 9, 14, 18, 23, 27, 32];
+
+    /// <summary>Bad luck spaces that cause you to skip your next turn.</summary>
```
externl, line 20: I think we should use keep using third-person point of view. 

```suggestion
    /// <summary>Goose spaces that move the player to the next goose space and grant an extra turn. If the player lands on
    /// space 5, they move to space 9 and get an extra turn. Landing on the last goose space moves the player to the final
    /// track space.</summary>
    public static readonly ImmutableList<byte> GooseSpaces = [5, 9, 14, 18, 23, 27, 32];

  /// <summary>Bad luck spaces that cause the player to skip their next turn.</summary>
```

examples/slice/GameOfTheGoose/Server/Game.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            _state.Status = new GameStatus.Started();
+            events.Add(new GameEvent.GameStarted());
+        }
+
+        return playerIndex;
+    }
+
+    /// <summary>Rolls dice for the given player and applies board rules.</summary>
+    internal Result<byte, GameError> RollDice(byte playerIndex, List<GameEvent> events)
```
externl, line 47: I'm having a hard time finding the line between what logic has handled by  `Game` and what's the responsibility of the caller. 

For example `AddPlayer` doesn't check the player count is too high and relies on the caller to never mess this up. 

However, `RollDice` has various checks and returns a Result when arguably the caller should deal with this.

examples/slice/GameOfTheGoose/Server/GameService.cs
```diff
@@ -0,0 +1,170 @@
...
+        GameSnapshot snapshot,
+        GameUpdateSource source)
+    {
+        try
+        {
+            await sink.OnGameUpdateAsync(snapshot, source.ReadAllAsync());
+        }
+        catch (Exception ex)
+        {
+            Console.WriteLine($"[StreamUpdates] exception: {ex}");
```
externl, line 149: If a client is having issues don't we want to just tear down the whole game. Otherwise won't the clients start to hang waiting for their turn?

examples/slice/GameOfTheGoose/slice/Game.slice
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/// Represents the current state of the game
+struct GameState
+{
+    /// The list of players already in the game, in the order they joined
+    players: [cs::type("List<Player>")] Sequence<Player>
+
+    /// The current status of the game
+    status: GameStatus
```
externl, line 35: Can we ever return a `GameStatus::Finished` state? If you join a finished game?

examples/slice/GameOfTheGoose/Client/Program.cs
```diff
@@ -0,0 +1,167 @@
...
+        ? $"Joined! The game has started. {Ansi.Bold}{Ansi.Yellow}It's your turn{Ansi.Reset}"
+        : $"Joined! The game has started. It's {state.Players[state.Turn].Name}'s turn");
+}
+else
```
externl, line 60: Else can't this be game finished?

pepone: No, joining a finished game returns a JoinResult.Failure, handled above.

examples/slice/GameOfTheGoose/Server/Game.cs
pepone: I will prefer to keep/move the game rules here, and keep the services more focused on the IceRPC side, streaming events, handling the session.

examples/slice/GameOfTheGoose/Server/GameService.cs
pepone: We should probably add a disconnected event, and let rest of the player continues.

I initially wanted to show how you can implement session recovery, get a new snapshot of the state, and receive events. Where we show how session != connection model works here. But I didn't want to add much more code in this initial version.

examples/slice/GameOfTheGoose/slice/Game.slice
pepone: Need to check what I implemented, but the idea was to return an error like `GameAlreadyFinished`.

examples/slice/GameOfTheGoose/slice/Game.slice
```diff
@@ -0,0 +1,159 @@
...
+    /// The index of the player whose turn it is to play (0-based index into the players sequence)
+    turn: uint8
+}
+
+enum GameEvent
+{
+    /// A new player has joined the game
+    PlayerJoined(player: Player)
+
+    /// The dice have been rolled by the player whose turn it is, and the result of the roll is available.
```
bernardnormier, line 46: As I understand it, the result of the dice roll decides the next event(s) (player moved, trapped, etc.). 

So, you could just send the dice roll and let each client compute what comes next.

This suggests to me you should _not_ send the dice roll: keep this private to the server and only distribute  "visible" updates to the client.
But keeping the dice roll private to the server goes against the spirit of the game (each player wants to know what's happening). 

The main issue here is the Game of the Goose is (as far as I can tell) a deterministic game. Each dice roll is all you need to figure out mechanically a bunch of derived events.

examples/slice/GameOfTheGoose/Client/Program.cs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// Build the features with our session ID for all RollDice calls.
+IFeatureCollection sessionFeatures = new FeatureCollection().With<IRequestContextFeature>(
+    new RequestContextFeature { ["sessionId"] = sessionId });
+
+// Listen for game events in the background. Canceled on quit or Ctrl+C.
+using var cts = new CancellationTokenSource();
+_ = CancelKeyPressed.ContinueWith(_ => cts.Cancel(), TaskScheduler.Default);
+ulong expectedSeq = snapshot.LastSeq + 1;
```
bernardnormier, line 72: I don't think we should transmit a field to check our code is not buggy. That's an odd thing to do.

Unless I am missing something: if there is no bug in our code, can we ever hit this out-of-order update?

examples/slice/GameOfTheGoose/slice/Game.slice
```diff
@@ -0,0 +1,159 @@
...
+
+    /// The player attempted to roll the dice when it is not their turn.
+    NotPlayerTurn
+
+    /// The player attempted to rejoin a game with an invalid session ID.
+    InvalidSessionId
+}
+
+/// The result of a join operation.
+enum JoinResult
```
bernardnormier, line 142: We should use `Result<Success, Failure>`, not create our own enum.

examples/slice/GameOfTheGoose/slice/Game.slice
```diff
@@ -0,0 +1,159 @@
...
+    gameEvent: GameEvent
+}
+
+interface GameUpdateSink
+{
+    /// Receive the initial game state snapshot followed by a stream of incremental updates.
+    /// @param snapshot: The game state at the time the player joined.
+    /// @param event: A stream of game events that occurred after the snapshot. The first event
+    /// has a sequence number of snapshot.lastSeq + 1.
+    onGameUpdate(snapshot: GameSnapshot, event: stream GameUpdate)
```
bernardnormier, line 112: This is no a good operation name for a stream of updates. `onGameUpdate` suggests you get a call for each update.

examples/slice/GameOfTheGoose/slice/Game.slice
```diff
@@ -0,0 +1,159 @@
...
+struct GameUpdate
+{
+    /// The sequence number of this update, which is used to ensure that updates are applied in the correct order.
+    seq: uint64
+
+    /// The game event that occurred, which describes the change to the game state.
+    gameEvent: GameEvent
+}
+
+interface GameUpdateSink
```
bernardnormier, line 106: It's not clear to me why you picked this bidir design over a simpler call like:
```slice
struct PlayerReady {
    playerIndex: uint8
    session: string
    snapshot: GameSnapshot
}

join(playerName: string) -> 
    (result: Result<PlayerReady, GameError>, updates: stream GameUpdate)
```

Making calls "the other way around" is about acknowledgement, which becomes murkier with streams. Here, it would be that the server wants to make sure that once it's sent the GameFinished update, the client sends a "void" to acknowledge it received all updates? Doesn't sound obvious.

examples/slice/GameOfTheGoose/README.md
```diff
@@ -0,0 +1,74 @@
+# Game of the Goose
+
+This example implements a simplified [Game of the Goose][game-wiki] to demonstrate real-time state synchronization
+across processes using IceRPC streams.
+
+The server holds the authoritative game state. When a client joins, it receives a snapshot of the current state followed
+by a stream of incremental event updates. Each client applies these events locally to stay in sync with the server,
+which is a common pattern for games, chat rooms, and other interactive applications.
+
+Key concepts demonstrated:
```
bernardnormier, line 10: I would try to simplify and drop non-core features.

For example:
 - it's not clear that bidir is needed here (see other comments)
 - the session IDs aren't core to this example

examples/slice/GameOfTheGoose/slice/Game.slice
pepone: > The main issue here is the Game of the Goose is (as far as I can tell) a deterministic game. Each dice roll is all you need to figure out mechanically a bunch of derived events.

That is correct.

pepone:
Let's find a better example to showcase this

PR #4264, opened 2026-03-01
Update Interop examples to Ice 3.8

bernardnormier:
This PR upgrades the two Interop examples to use Ice 3.8. It also replaces a few references to Ice 3.7 by references to Ice 3.8.

### DIFF
--- a/examples/slice/InteropGreeter/README.md
+++ b/examples/slice/InteropGreeter/README.md
@@ -9,8 +9,8 @@ First, build the client and server applications for IceRPC with:
 dotnet build
 ```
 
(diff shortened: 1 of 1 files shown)


examples/slice/InteropGreeter/README.md
```diff
@@ -0,0 +1,31 @@
...
+mode.
+
+First, build the client and server applications for IceRPC with:
+
+``` shell
+dotnet build
+```
+
+Then, build the client and/or server for the Ice 3.8 `greeter` demo in the language of your choice, for example
+the [Ice C++ Greeter].
```
pepone: Do you want to show C++ to showcase a different language? Otherwise I will just use Ice for C# here.

PR #4267, opened 2026-03-02
Use .ice files and slice2cs instead of mode=Slice1

bernardnormier:
This is companion to https://github.com/zeroc-ice/ice/pull/5206

It makes minimal updates to icerpc-csharp to use .ice files instead of mode=Slice1 .slice files. A follow-up PR will remove all mode=Slice1 .slice files.

### DIFF
--- a/tests/IceRpc.Slice.Tests/StructTests.cs
+++ b/tests/IceRpc.Slice.Tests/StructTests.cs
@@ -27,6 +27,37 @@ public sealed class StructTests
         Assert.That(value, Is.EqualTo(expected));
     }
 
+    /*
+    TODO: reenable once the Slice1 and Slice2 tests are in separate files/assemblies.
+    [Test]
+    public void Decode_slice2_compact_struct_with_nullable_proxy(
+        [Values("icerpc://localhost/service", null)] string? serviceAddress)
+    {
+        var expected = new MyCompactStructWithNullableProxy(
+            10,
+            serviceAddress is null ? null :
+                new PingableProxy(InvalidInvoker.Instance, new Uri(serviceAddress)));
+        var buffer = new MemoryBufferWriter(new byte[256]);
+        var encoder = new SliceEncoder(buffer, SliceEncoding.Slice2);
+        var bitSequenceWriter = encoder.GetBitSequenceWriter(1);
+        encoder.EncodeInt32(expected.A);
+        if (expected.I is null)
+        {
+            bitSequenceWriter.Write(false);
+        }
+        else
+        {
+            bitSequenceWriter.Write(true);
+            encoder.EncodeServiceAddress(expected.I.Value.ServiceAddress);
+        }
+        var decoder = new SliceDecoder(buffer.WrittenMemory, SliceEncoding.Slice2);
+
+        var value = new MyCompactStructWithNullableProxy(ref decoder);
+
+        Assert.That(value, Is.EqualTo(expected));
+    }
+    */
+
     [Test]
     public void Decode_slice1_compact_struct_with_sequence_of_nullable_proxies()
     {
@@ -50,6 +81,33 @@ public sealed class StructTests
 
         Assert.That(value.I, Is.EqualTo(expected.I));
     }
+
+    /*
+    [Test]
+    public void Decode_slice2_compact_struct_with_sequence_of_nullable_proxies()
+    {
+        var expected = new MyCompactStructWithSequenceOfNullableProxies
+        {
+            I = new PingableProxy?[]
+            {
+                new PingableProxy(InvalidInvoker.Instance, new Uri("icerpc://localhost/service1")),
+                null,
+                new PingableProxy(InvalidInvoker.Instance, new Uri("icerpc://localhost/service2")),
+            }
+        };
+        var buffer = new MemoryBufferWriter(new byte[256]);
+        var encoder = new SliceEncoder(buffer, SliceEncoding.Slice2);
+        encoder.EncodeSequenceOfOptionals(
+            expected.I,
+            (ref SliceEncoder encoder, PingableProxy? value) =>
+                encoder.EncodeServiceAddress(value!.Value.ServiceAddress));
+        var decoder = new SliceDecoder(buffer.WrittenMemory, SliceEncoding.Slice2);
+
+        var value = new MyCompactStructWithSequenceOfNullableProxies(ref decoder);
+
+        Assert.That(value.I, Is.EqualTo(expected.I));
+    }
+    */
 
     [Test]
     public void Decode_slice1_compact_struct_with_dictionary_of_nullable_proxies()
@@ -76,6 +134,33 @@ public sealed class StructTests
         Assert.That(value.I, Is.EqualTo(expected.I));
     }
 
+    /*
+    [Test]
+    public void Decode_slice2_compact_struct_with_dictionary_of_nullable_proxies()
+    {
+        var expected = new MyCompactStructWithDictionaryOfNullableProxies
+        {
+            I = new Dictionary<int, PingableProxy?>
+            {
+                [1] = new PingableProxy(InvalidInvoker.Instance, new Uri("icerpc://localhost/service1")),
+                [2] = null,
+                [3] = new PingableProxy(InvalidInvoker.Instance, new Uri("icerpc://localhost/service2")),
+            }
+        };
+        var buffer = new MemoryBufferWriter(new byte[256]);
+        var encoder = new SliceEncoder(buffer, SliceEncoding.Slice2);
+        encoder.EncodeDictionaryWithOptionalValueType(
+            expected.I,
+            (ref SliceEncoder encoder, int value) => encoder.EncodeInt32(value),
+            (ref SliceEncoder encoder, PingableProxy? value) => encoder.EncodeServiceAddress(value!.Value.ServiceAddress));
+        var decoder = new SliceDecoder(buffer.WrittenMemory, SliceEncoding.Slice2);
+
+        var value = new MyCompactStructWithDictionaryOfNullableProxies(ref decoder);
+
+        Assert.That(value.I, Is.EqualTo(expected.I));
+    }
+    */
+
     [Test]
     public void Encode_slice1_compact_struct_with_nullable_proxy(
         [Values("icerpc://localhost/service", null)] string? serviceAddress)
@@ -93,6 +178,29 @@ public sealed class StructTests
         Assert.That(decoder.DecodeNullableProxy<AnotherPingableProxy>(), Is.EqualTo(expected.I));
 
     }
+
+    /*
+    [Test]
+    public void Encode_slice2_compact_struct_with_nullable_proxy(
+        [Values("icerpc://localhost/service", null)] string? serviceAddress)
+    {
+        var expected = new MyCompactStructWithNullableProxy(
+            10,
+            serviceAddress is null ? null : new PingableProxy(InvalidInvoker.Instance, new Uri(serviceAddress)));
+
+        var buffer = new MemoryBufferWriter(new byte[256]);
+        var encoder = new SliceEncoder(buffer, SliceEncoding.Slice2);
+
+        expected.Encode(ref encoder);
+
+        var decoder = new SliceDecoder(buffer.WrittenMemory, SliceEncoding.Slice2);
+        var bitSequenceReader = decoder.GetBitSequenceReader(1);
+        Assert.That(decoder.DecodeInt32(), Is.EqualTo(expected.A));
+        Assert.That(
+            bitSequenceReader.Read() ? decoder.DecodeProxy<PingableProxy>() : (PingableProxy?)null,
+            Is.EqualTo(expected.I));
+    }
+    */
 
     [Test]
     public void Encode_slice1_compact_struct_with_sequence_of_nullable_proxies()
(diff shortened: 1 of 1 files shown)


build/Common.props
```diff
@@ -15,6 +15,9 @@
     <StyleCopAnalyzersVersion>1.2.0-beta.556</StyleCopAnalyzersVersion>
     <DebugType>portable</DebugType>
     <DebugSymbols>true</DebugSymbols>
+
+     <!-- The version of ZeroC.Ice.Slice.Tools used to compile .ice files -->
+    <IceSliceToolsVersion>3.9.0-alpha0</IceSliceToolsVersion>
```
bernardnormier: We should use `3.9.0-nightly.*` once the updated ZeroC.Ice.Slice.Tools in available in the nightly builds.

examples/slice/InteropGreeter/slice/Greeter.ice
```diff
@@ -0,0 +1,13 @@
...
+
+module VisitorCenter
+{
+    /// Represents a simple greeter.
+    interface Greeter
+    {
+        /// Creates a personalized greeting.
+        /// @param name The name of the person to greet.
+        /// @return The greeting.
+        string greet(string name);
```
bernardnormier, line 11: The 3.8 demo Slice file without `cs:identifier`.

src/IceRpc.Locator/IceRpc.Locator.csproj
```diff
@@ -7,17 +7,24 @@
...
+    </SliceCompile>
+  </ItemDefinitionGroup>
   <ItemGroup>
     <ProjectReference Include="../IceRpc/IceRpc.csproj" ExactVersion="true" />
     <ProjectReference Include="../IceRpc.Slice/IceRpc.Slice.csproj" ExactVersion="true" />
-    <SliceFile
-      Include="../../slice/Ice/Locator.slice;../../slice/Ice/LocatorRegistry.slice;../../slice/Ice/Process.slice"
+
+    <SliceCompile
+      Include="../../slice/Ice/Locator.ice;../../slice/Ice/LocatorRegistry.ice;../../slice/Ice/Process.ice"
```
bernardnormier, line 20: This packaging most like breaks the the `include "xxx"` in these files.

src/IceRpc/Internal/IceIdentity.cs
```diff
@@ -2,17 +2,16 @@
 
 namespace IceRpc.Internal;
 
-internal readonly partial record struct Identity
```
bernardnormier, line 5: Should revert back to internal readonly once we add cs:internal and cs:readonly to Ice-Slice.

src/IceRpc/IceRpc.csproj
```diff
@@ -11,6 +11,9 @@
     <SliceFile>
       <Rpc>none</Rpc>
     </SliceFile>
+    <SliceCompile>
+      <Rpc>none</Rpc>
```
bernardnormier, line 15: This "none" option for SliceCompile is only need for this assembly.

Maybe we could find a better solution, like excluding all `*.IceRpc.cs` from the C# compilation?

tests/IceRpc.Locator.Tests/CacheLessLocationResolverTests.cs
```diff
@@ -17,7 +17,7 @@ public async Task Resolving_a_known_location_returns_a_proxy([Values] bool isAda
 
         (ServiceAddress? serviceAddress, bool fromCache) =
             await locationResolver.ResolveAsync(
-                new Location { IsAdapterId = isAdapterId, Value = "good" },
+                new Location { IsAdapterId = isAdapterId, Value = isAdapterId ? "good" : "/good" },
```
bernardnormier, line 20: "good" is not a valid path. Prior to this update, it didn't matter as we did not encode it; now we convert this path to an identity all the time, even when we don't encode the resulting identity.

tests/ZeroC.Slice.Tests/CustomTypeTests.cs
```diff
@@ -90,135 +90,6 @@ public void Encode_sequence_of_custom_types([Values] SliceEncoding encoding)
         Assert.That(expected.S, Is.EqualTo(value));
         Assert.That(decoder.Consumed, Is.EqualTo(buffer.WrittenMemory.Length));
     }
-
-    [Test]
-    public void Decode_slice1_sequence_of_optional_custom_types()
```
bernardnormier, line 95: No custom types with Ice-Slice.

tests/ZeroC.Slice.Tests/EnumTests.cs
```diff
@@ -193,38 +193,4 @@ public void Decode_slice1_enum([Values(MySlice1Enum.Enum1, MySlice1Enum.Enum2)]
         Assert.That(decoded, Is.EqualTo(expected));
         Assert.That(decoder.Remaining, Is.EqualTo(0));
     }
-
-    [Test]
-    public void Encode_slice1_unchecked_enum(
```
bernardnormier, line 198: All Ice-Slice enums are checked.

tests/IceRpc.Slice.Tests/ProxyTests.slice
```diff
@@ -8,15 +8,13 @@ interface MyDerivedInterface : MyBaseInterface {}
...
     receiveProxy() -> ReceiveProxyTestProxy
-    receiveObjectProxy() -> Ice::ObjectProxy
 }
 
 [cs::type("IceRpc.Slice.Tests.ReceiveProxyTestProxy")]
 custom ReceiveProxyTestProxy
 
 interface SendProxyTest {
     sendProxy(proxy: SendProxyTestProxy)
-    sendObjectProxy(proxy: Ice::ObjectProxy)
```
bernardnormier, line 19: Can't use types/proxies defined in .ice files in .slice files.

build/Common.props
pepone: No plan to merge `--rpc` into 3.8?

src/IceRpc.Locator/IceRpc.Locator.csproj
pepone: Do you mean it breaks the packaging of this Slice files?

src/IceRpc/IceRpc.csproj
pepone: I think this is conceptually clear. The SliceCompile task adds all generated items to the `Compile` ItemGroup, if we want to exclude C# generated items the tasks or targets would have to support this exclusions too.

build/Common.props
bernardnormier: We should merge it to 3.8 eventually.

src/IceRpc.Locator/IceRpc.Locator.csproj
bernardnormier: I mean the packaged Locator.ice won't be usable as it includes `"Identity.ice"` which is not in the same directory or in the include path.

bernardnormier: I think the fix here is to package all `Ice/*.ice` files "manually" in the same NuGet package, probably inside the Ice.Slice package.

build/Common.props
externl: Probably simplest to get it working with main for now?

src/IceRpc.Locator/IceRpc.Locator.csproj
bernardnormier: I removed the Pack attribute as it's actually not an attribute of SliceCompile. So this issue is moot - we currently don't package the .ice files.

tests/ZeroC.Slice.Tests/ClassTests.ice
```diff
@@ -0,0 +1,66 @@
...
+    // TODO: this metadata doesn't exist yet.
+    // ["cs:internal"]
+    class MyInternalBaseClass
+    {
+        string m1;
+    }
+
+    // TODO: this metadata doesn't exist yet.
+    // ["cs:internal"]
+    class MyInternalClass : MyInternalBaseClass
```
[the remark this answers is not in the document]
bernardnormier: See https://github.com/zeroc-ice/ice/issues/5225

tests/ZeroC.Slice.Tests/ClassTests.ice
```diff
@@ -0,0 +1,66 @@
...
+    class MyDerivedCompactClass(789) : MyCompactClass {}
+
+    class MyClassWithValues
+    {
+        int a;
+        optional(5) int b;
+        Value c;
+        optional(3) int d;
+        int e;
+    }
```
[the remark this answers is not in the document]
bernardnormier: I removed the non-nullable Value field of the previous definitions, as we cannot express it in Ice-Slice at present.

See https://github.com/zeroc-ice/ice/issues/5209

tests/IntegrationTests/ProtocolBridgingTests.ice
```diff
@@ -0,0 +1,32 @@
...
+
+    interface ProtocolBridgingTest
+    {
+        // Simple operations
+        int op(int x);
+        void opVoid();
+
+        // Oneway operation
+        // ["oneway"] TODO: not implemented yet
+        void opOneway(int x);
```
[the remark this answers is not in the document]
bernardnormier: See https://github.com/zeroc-ice/ice/issues/5224

tests/IceRpc.Slice.Tests/StructTests.cs
```diff
@@ -27,182 +27,83 @@ public void Decode_slice1_compact_struct_with_nullable_proxy(
...
+            (ref SliceEncoder encoder, AnotherPingableProxy? value) => encoder.EncodeNullableServiceAddress(value?.ServiceAddress));
         var decoder = new SliceDecoder(buffer.WrittenMemory, SliceEncoding.Slice1);
 
         var value = new MyCompactStructWithDictionaryOfNullableProxies(ref decoder);
 
         Assert.That(value.I, Is.EqualTo(expected.I));
     }
 
-    [Test]
-    public void Decode_slice2_compact_struct_with_dictionary_of_nullable_proxies()
```
pepone, line 133: No clear why did you remove this Slice2 tests.

tests/ZeroC.Slice.Tests.ReferencedAssemblies/C/C.csproj
```diff
@@ -1,22 +1,23 @@
 <Project Sdk="Microsoft.NET.Sdk">
```
pepone, line 1: We switch to ZeroC.Ice.Slice.Tools here because these tests are only relevant for class and exceptions, right?

tests/ZeroC.Slice.Tests/TaggedTests.Slice1.ice
```diff
@@ -0,0 +1,36 @@
...
+    {
+        optional(1) byte a;                 // Uses F1 tag format
+        optional(2) short b;                // Uses F2 tag format
+        optional(3) int c;                  // Uses F4 tag format
+        optional(4) long d;                 // Uses F8 tag format
+        optional(5) FixedSizeStruct e;      // Uses VSize tag format
+        optional(6) VarSizeStruct f;        // Use FSize tag format
+        optional(7) MyEnum g;               // Uses Size tag format
+        optional(8) ByteSeq h;              // Uses OptimizedVSize tag format
+        optional(9) IntSeq i;               // Uses VSize tag format
```
InsertCreativityHere, line 31: Good catch. This used to say `FSize` in the old test, which was incorrect.

tests/ZeroC.Slice.Tests.ReferencedAssemblies/C/C.csproj
bernardnormier: Yes, it's all about classes and exceptions.

tests/ZeroC.Slice.Tests/SlicingTests.slice
```diff
@@ -1,47 +0,0 @@
...
-
-module ZeroC::Slice::Tests
-
-class SlicingBaseClass {
-    m1: string
-}
-
-class SlicingDerivedClass : SlicingBaseClass {
-    m2: string
-    m3: SlicingBaseClass?
```
InsertCreativityHere, line 13: When we add support for `?` to `.ice` files, we should revisit this commit to tell which places class/proxy should be nullable or not. Since unfortunately, this information can't be captured in the new files.

tests/IceRpc.Slice.Tests/StructTests.cs
bernardnormier: It's a mistake. All these tests were using Slice1-defined types compatible with Slice2. The correct fix is to redefine these types in a Slice2 file. Fixing...

tests/ZeroC.Slice.Tests/ClassTests.ice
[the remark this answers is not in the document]
InsertCreativityHere: Should we leave a commented out field with a TODO? Otherwise we'll forget.
Once we implement zeroc-ice/ice#5209, it would be good to have this field in this test.

tests/IceRpc.Slice.Tests/StructTests.cs
bernardnormier: I restored and commented-out these slice2 tests. It will be much easier to restore them once slice1 and slice2 tests are no longer combined in the same file / assemblies.

PR #4271, opened 2026-03-04
Rename ZeroC.Slice -> ZeroC.Slice.Codec

### LINKED ISSUE #4255: Rename ZeroC.Slice to ZeroC.Slice.Codec
I propose to rename the ZeroC.Slice namespace and NuGet package ZeroC.Slice.Codec.

This name would be consistent with the `slice-codec` name we use in Rust.

ZeroC.Slice.Codec (currently simply ZeroC.Slice) is about encoding/decoding Slice types. While this is an important component of Slice, it's not the only component, and we'll soon add:
 - definitions produced by slicec (in ZeroC.Slice.Compiler)
 - a nicer representation of these definitions (in ZeroC.Slice.Symbols)
 - code-generator plug-ins (in ZeroC.Slice.CodeGen and IceRpc.Slice.CodeGen ?)

and more.

bernardnormier:
This PR renames the ZeroC.Slice namespace/NuGet package to ZeroC.Slice.Codec.

Fixes #4255.

### DIFF
--- a/src/ZeroC.Slice.Codec/README.md
+++ b/src/ZeroC.Slice.Codec/README.md
@@ -73,4 +73,4 @@ public partial record struct Person
 [package]: https://www.nuget.org/packages/ZeroC.Slice.Codec
 [pipelines]: https://learn.microsoft.com/en-us/dotnet/standard/io/pipelines
 [slicec-cs]: https://github.com/icerpc/icerpc-csharp/tree/main/tools/IceRpc.Slice.Tools
-[source]: https://github.com/icerpc/icerpc-csharp/tree/main/src/ZeroC.Slice
+[source]: https://github.com/icerpc/icerpc-csharp/tree/main/src/ZeroC.Slice.Codec


src/ZeroC.Slice.Codec/README.md
```diff
@@ -65,11 +66,11 @@ public partial record struct Person
 }
 ```
 
-[api]: https://docs.icerpc.dev/api/csharp/api/ZeroC.Slice.html
+[api]: https://docs.icerpc.dev/api/csharp/api/ZeroC.Slice.Codec.html
 [cs-buffers]: https://learn.microsoft.com/en-us/dotnet/standard/io/buffers
 [docs]: https://docs.icerpc.dev/slice2
 [idl]: https://en.wikipedia.org/wiki/Interface_description_language
-[package]: https://www.nuget.org/packages/ZeroC.Slice
+[package]: https://www.nuget.org/packages/ZeroC.Slice.Codec
```
externl, line 73: @pepone does NuGet support package renaming or do we just have a publish new package?

pepone: There is support for deprecating packages https://learn.microsoft.com/en-us/nuget/nuget-org/deprecate-packages

I wonder if we can create and last version of the old package that just has a transitive dependency on the new one, but not clear if this work. I didn't find explicit support for renaming.

PR #4272, opened 2026-03-04
Rework .IceRpc.cs file exclussion to not rely on --rpc none

pepone:
This PR is an initial step to get rid of `--rpc none`. It creates a target that can be imported to exclude `.IceRpc.cs` items. And updates projects that were using `<Rpc>none</Rpc>` to use it.


build/IceRpc.ExcludeSliceCompile.targets
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Import this file in projects that need to exclude .IceRpc.cs files generated by slice2cs. These files depend on
+     the IceRpc assembly and cannot be compiled in projects that don't reference it. -->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="_IceRpc_ExcludeSliceCompileGenerated" DependsOnTargets="SliceCompile" BeforeTargets="CoreCompile">
+    <ItemGroup>
+        <Compile Remove="@(Compile)" Condition="$([System.String]::Copy('%(Filename)').EndsWith('.IceRpc'))" />
```
bernardnormier: It's not end with `.IceRpc.cs`?

pepone: Filename medata doesn't include the extension

https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-well-known-item-metadata?view=visualstudio

tests/ZeroC.Slice.Tests.ReferencedAssemblies/B/B.csproj
```diff
@@ -2,11 +2,7 @@
   <PropertyGroup>
     <TargetFramework>net10.0</TargetFramework>
   </PropertyGroup>
-  <ItemDefinitionGroup>
-    <SliceCompile>
-      <Rpc>none</Rpc>
-    </SliceCompile>
-  </ItemDefinitionGroup>
+
```
InsertCreativityHere: We didn't leave this blank line in other files.

tests/ZeroC.Slice.Tests.ReferencedAssemblies/A/A.csproj
```diff
@@ -2,11 +2,7 @@
   <PropertyGroup>
     <TargetFramework>net10.0</TargetFramework>
   </PropertyGroup>
-  <ItemDefinitionGroup>
-    <SliceCompile>
-      <Rpc>none</Rpc>
-    </SliceCompile>
-  </ItemDefinitionGroup>
+
```
InsertCreativityHere: We should be consistent about whether or not there is a blank here,
although I suspect it's left my mistake.

PR #4274, opened 2026-03-05
Remove Socket.ConnectAsync cancellation workaround

### LINKED ISSUE #4259: Remove Socket.ConnectAsync workaround
https://github.com/icerpc/icerpc-csharp/blob/1701684c938363063c53c02b02ba3ac6e45d63ea/src/IceRpc/Transports/Tcp/Internal/TcpConnection.cs#L334-L336

Seems this was fixed in .NET 10 https://github.com/dotnet/runtime/pull/116943

We can probably remove this workaround in main, we should test.

pepone:
The workaround for dotnet/runtime#75889 is no longer needed since the issue was fixed in .NET 10 (dotnet/runtime#116943).

Fixes #4259


bernardnormier:
You should include Fixes ... in the comment.

PR #4275, opened 2026-03-05
Change enclosing namespace for Ice definitions

### LINKED ISSUE #4269: Enclosing namespace for mapped Ice (Glacier2, IceGrid) etc. modules
It's desirable to map all Ice top-level modules (Ice, Glacier2, IceGrid, IceStorm...) consistently in IceRPC.

Currently, we map the Ice module to IceRpc.Slice.Ice ... but this mapping is not great (why Slice in the name?).

## Proposal
Map the Ice top-level modules to `IceRpc.[ModuleName]`: IceRpc.Ice, IceRpc.Glacier2, IceRpc.IceGrid etc.

It's simple, consistent, and fortunately the number of top-level Ice modules is limited.

bernardnormier:
It's now simply IceRpc, and the Ice module maps to IceRpc.Ice (vs IceRpc.Slice.Ice previously).

Fixes #4269.

### DIFF
--- a/examples/slice/InteropIceGrid/Client/Program.cs
+++ b/examples/slice/InteropIceGrid/Client/Program.cs
@@ -1,8 +1,8 @@
 // Copyright (c) ZeroC, Inc.
 
 using IceRpc;
-using IceRpc.Retry;
 using IceRpc.Ice;
+using IceRpc.Retry;
 using Microsoft.Extensions.Logging;
 using VisitorCenter;
 


examples/slice/InteropIceGrid/Client/Program.cs
```diff
@@ -2,7 +2,7 @@
 
 using IceRpc;
 using IceRpc.Retry;
-using IceRpc.Slice.Ice;
+using IceRpc.Ice;
```
externl, line 4: Needs sorting.

PR #4276, opened 2026-03-05
Add slice.toml vendoring system to replace git subtree

pepone:
Add slice.toml to track the provenance of vendored Slice/Ice files and tools/sync-slice.py to synchronize them from upstream repositories. Add a CI workflow to verify vendored files match slice.toml on each PR.

Remove leftover git subtree metadata from the slice/ directory.

### DIFF
--- a/tools/sync-slice.py
+++ b/tools/sync-slice.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # Copyright (c) ZeroC, Inc.
 
-"""Synchronize vendored Slice/Ice files based on slice.toml.
+"""Synchronize vendored Slice files (*.ice, *.slice) based on slice.toml.
 
 Usage:
     python tools/sync-slice.py           # sync files from upstream


.github/workflows/slice-sync.yml
```diff
@@ -0,0 +1,19 @@
...
+  verify-slice:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+
+      - run: python tools/sync-slice.py --verify
```
externl: Would be nice to have `name:` on these steps.

tools/sync-slice.py
```diff
@@ -0,0 +1,225 @@
...
+    """Return sorted relative paths under *base_dir* matching any glob pattern."""
+    matches: set[str] = set()
+    for pattern in patterns:
+        for path in base_dir.glob(pattern):
+            if path.is_file():
+                matches.add(str(path.relative_to(base_dir)))
+    return sorted(matches)
+
+
+
```
externl:
```suggestion


```

.github/workflows/slice-sync.yml
```diff
@@ -0,0 +1,19 @@
+name: "Verify vendored Slice files"
```
bernardnormier: I would prefer a different name for this yml file: slice-sync.yml and sync-slice.py look strange.

Maybe: check-shared-slice.yml?

slice.toml
```diff
@@ -0,0 +1,17 @@
+# slice.toml — provenance for vendored IDL files
```
bernardnormier: I would call them Slice files

bernardnormier (approved):

.github/workflows/slice-sync.yml
externl: Maybe `vendored-slice.yml`

pepone: Fixed as suggested by @externl in Slack

slice.yml with a job verify-vendored-files

tools/sync-slice.py
```diff
@@ -0,0 +1,224 @@
+#!/usr/bin/env python3
+# Copyright (c) ZeroC, Inc.
+
+"""Synchronize vendored Slice/Ice files based on slice.toml.
```
externl:
```suggestion
"""Synchronize vendored Slice files (*.ice, *.slice) based on slice.toml.
```

slice.toml
```diff
@@ -0,0 +1,17 @@
...
+#
+# Each [source.<name>] section describes an upstream repository and which files
+# to vendor into this repo.  Run `python tools/sync-slice.py` to synchronize.
+
+[source.ice]
+repo    = "https://github.com/zeroc-ice/ice"
+ref     = "main"
+source  = "slice"
+dest    = "slice"
+include = ["Ice/Identity.ice", "Ice/Locator.ice", "Ice/LocatorRegistry.ice", "Ice/Process.ice"]
```
InsertCreativityHere, line 11: Not strictly against it, but do we really need an optional `source` field?
You can specify paths in include, so I feel like it would be simpler for us to not handle parsing/using this field, and just adding `slice/` to the 4 files we include
```suggestion
[source.ice]
repo    = "https://github.com/zeroc-ice/ice"
ref     = "main"
dest    = "slice"
include = ["slice/Ice/Identity.ice", "slice/Ice/Locator.ice", "slice/Ice/LocatorRegistry.ice", "slice/Ice/Process.ice"]
```
But again, if you like `source`, feel free to keep it. Not important either way.

InsertCreativityHere (approved):
Looks good to me.

slice.toml
InsertCreativityHere: Nevermind, I see the purpose of this with respect to the destination folder now.

PR #4280, opened 2026-03-06
Improve SocketError to IceRpcError mapping

### LINKED ISSUE #4031: Revisit client error when server aborts connection due to MaxConnections (ice protocol)
When an ice-protocol server rejects a new connection because MaxConnections is reached, it aborts the transport connection:

https://github.com/icerpc/icerpc-csharp/blob/c866c086c6e0b2064681351cf0c7c0a1b4f79d61/src/IceRpc/Server.cs#L908

This abort is in turn received by the client as an IceRpcError.ConnectionAborted (fine) or uncategorized error (IceRpcError.IceRpcError). See #2462.

This behavior is problematic because the ConnectionCache "move to the next server address" requires specific error codes, and IceRpcError.IceRpcError (= uncategorized error) is not among them. See #4027.

pepone:
- Add mappings for NetworkDown, HostDown, TimedOut, and DNS errors (HostNotFound, TryAgain, NoRecovery, NoData) to ServerUnreachable, and NetworkReset to ConnectionAborted, so that retry middleware can treat these failures appropriately and try alternative replicas.
- Handle SocketError.InvalidArgument when accessing endpoint properties after the peer disposes the connection during TLS handshake.
- Add detailed comments explaining each mapping group.

Fixes #4031


src/IceRpc/Transports/SocketExceptionExtensions.cs
```diff
@@ -18,17 +18,53 @@ public static IceRpcException ToIceRpcException(this SocketException exception,
...
+
+            // NetworkReset indicates that an established connection became invalid due to a network event
+            // (for example interface reset, Wi-Fi reconnect, VPN reconnect).
+            SocketError.NetworkReset => IceRpcError.ConnectionAborted,
+
+            // The server is reachable but no process is listening on the target port.
             SocketError.ConnectionRefused => IceRpcError.ConnectionRefused,
+
+            // These errors indicate the remote server cannot be reached. This includes routing failures,
+            // local network failures, timeouts, and host reachability failures.
```
bernardnormier: One aspect not clear to me and worth clarifying is whether or not these ServerUnreachable can only occur during connection establishment.

If no, then either:
 - the Retry interceptor should handle ServerUnreachable like ConnectionAborted, or
 - we should map these errors to ConnectionAborted

src/IceRpc/Transports/SocketExceptionExtensions.cs
```diff
@@ -18,17 +18,53 @@ public static IceRpcException ToIceRpcException(this SocketException exception,
...
+            // They are grouped as ServerUnreachable because from the RPC perspective the connection
+            // cannot be established or maintained.
+            SocketError.HostUnreachable => IceRpcError.ServerUnreachable,
+            SocketError.NetworkUnreachable => IceRpcError.ServerUnreachable,
+            SocketError.NetworkDown => IceRpcError.ServerUnreachable,
+            SocketError.HostDown => IceRpcError.ServerUnreachable,
+            SocketError.TimedOut => IceRpcError.ServerUnreachable,
+
+            // Name resolution failures are also mapped to ServerUnreachable so that callers and retry
+            // middleware can treat DNS failures and transport reachability failures uniformly.
```
bernardnormier: The Retry interceptor does not currently handle ServerUnreachable. ServerUnreachable is (so far) considered a connection establishment failure and is handled by the ConnectionCache.

src/IceRpc/Transports/SocketExceptionExtensions.cs
pepone: Updated comment

src/IceRpc/Transports/SocketExceptionExtensions.cs
pepone: updated the comment

bernardnormier:
We should add a test for #4031 (MaxConnections with the ice protocol) in a follow-up PR.

src/IceRpc/Transports/SocketExceptionExtensions.cs
```diff
@@ -18,17 +18,60 @@ public static IceRpcException ToIceRpcException(this SocketException exception,
...
+
+            // The server is reachable but no process is listening on the target port.
             SocketError.ConnectionRefused => IceRpcError.ConnectionRefused,
+
+            // These errors indicate the remote server cannot be reached. This includes routing failures,
+            // local network failures, timeouts, and host reachability failures.
+            //
+            // On async TCP, these errors occur almost exclusively during connection establishment.
+            // Post-connect, the TCP stack absorbs ICMP errors and retransmits; the kernel would
+            // typically return ConnectionReset or TimedOut (after minutes of retransmission) instead.
```
bernardnormier: This comment suggests TimedOut should be mapped to ConnectionAborted, not ServerUnreachable.

src/IceRpc/Transports/Tcp/Internal/TcpConnection.cs
```diff
@@ -353,6 +348,21 @@ await _sslStream.AuthenticateAsClientAsync(
...
+                localNetworkAddress: Socket.LocalEndPoint!,
+                remoteNetworkAddress: Socket.RemoteEndPoint!,
+                _sslStream?.RemoteCertificate);
+        }
+        catch (SocketException exception)
+        {
+            // This can happen if the peer closes the connection immediately after accepting it, which can
+            // cause the endpoint information to be unavailable. Any SocketException at this point means the
+            // connection is no longer usable.
+            throw new IceRpcException(IceRpcError.ConnectionAborted, exception);
```
bernardnormier, line 359: This is at the end of the connection establishment, so nobody is using this connection yet.

It seems `exception.ToIceRpcException()` would be better here:
 - a ServerUnreachable error code would allow the ConnectionCache to try the next server address
 - more importantly, it keeps a consistent mapping SocketErrorCode => IceRpcError. If you see an exception with a different mapping, you're going to wonder what's going on.

And you could move this return back to line 341.

src/IceRpc/Transports/SocketExceptionExtensions.cs
pepone: Reworked the comment

src/IceRpc/Transports/Tcp/Internal/TcpConnection.cs
pepone: Accessing the endpoint after the connection was closed throws `System.Net.Sockets.SocketException (22): Invalid argument`, this gets mapped to `IceRpc.Error`. Not clear how using `exception.ToIceRpcException()` would work here.

ConnectionCache also retries on ConnectionAborted doesn't it?

https://github.com/icerpc/icerpc-csharp/blob/b9d6c4d68c33aaf2811163f90e93e1531a6aa4d0/src/IceRpc/ConnectionCache.cs#L495-L510

If we move this to 341, and throw `IceRpcException`, this would be catch by the IOException catch block, which would wrap the `IceRpcException` in a new `IceRpcException` with `IceRpcError` code.

`IceRpcException` is an `IOEexception`

https://github.com/icerpc/icerpc-csharp/blob/b9d6c4d68c33aaf2811163f90e93e1531a6aa4d0/src/IceRpc/Transports/Tcp/Internal/IOExceptionExtensions.cs#L10-L13

bernardnormier: I meant move the `return new TransportConnectionInformation(...` statement to line 341 (where it used to be), not the entire try/catch block.

Then, based on this:
```
Accessing the endpoint after the connection was closed throws System.Net.Sockets.SocketException (22): Invalid argument, this gets mapped to IceRpc.Error. 
```

It sounds like the fix is to map 22 (Invalid Argument) to ServerUnreachable, with a good comment?

pepone: > It sounds like the fix is to map 22 (Invalid Argument) to ServerUnreachable, with a good comment?

I will prefer to not do that in SocketExcceptionExtensions, this error code can be used in other error conditions where this mapping would not make sense.

One option is to set a flag and explicitly handle this, something like:

```
catch (SocketException ex) when (connected) 
{
   ///  this can happen if the peer closed the connection shortly after it was connected.
   throw new IceRpcException(IceRpcError.ConnectionAborted, exception);
} 
catch (SocketException ...)
```

pepone: Fixed as suggested in the previous commnet.

tests/IceRpc.Conformance.Tests/Transports/DuplexConnectionConformanceTests.cs
```diff
@@ -113,7 +113,7 @@ public async Task Connect_succeeds_or_fails_with_connection_aborted_if_the_peer_
         {
             Assert.That(
                 exception?.IceRpcError,
-                Is.Null.Or.EqualTo(IceRpcError.ConnectionAborted).Or.EqualTo(IceRpcError.IceRpcError),
+                Is.Null.Or.EqualTo(IceRpcError.ConnectionAborted),
                 $"The test failed with an unexpected IceRpcError {exception}");
         }
```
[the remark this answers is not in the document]
bernardnormier: You should fix that before merging.

Issue #4270, opened 2026-03-04
Move Slice1 and Slice2 tests to separate test assemblies


bernardnormier:
We currently test Slice in ZeroC.Slice.Tests and Ice.Slice.Tests. 

We should split these tests in Slice1 (.ice) and Slice2 (.slice) test assemblies, and "duplicate" some of these tests now that we have separate code generators for .ice and .slice files.

PR #4282, opened 2026-03-06
Split ZeroC.Slice.Tests

bernardnormier:
This PR splits the Zeroc.Slice.Tests is several test assemblies:

- ZeroC.Tests.Common - common classes
- ZeroC.Slice.Codec.Tests - tests for the Codec assembly, doesn't rely on generated code
- ZeroC.Slice.IceGen.Tests - tests that rely on code generated from .ice files
- ZeroC.Slice.SlicecGen.Tests - tests that rely on code generated from .slice files

It's a partial fix for #4270.

### DIFF
--- a/slice/Compiler/DocComment.slice
+++ b/slice/Compiler/DocComment.slice
@@ -4,13 +4,11 @@
 module Compiler
 
 struct DocComment {
-    overview: Message
-    see: Sequence<EntityId>
+    overview: Sequence<MessageComponent>
+    seeTags: Sequence<EntityId>
 }
 
 enum MessageComponent {
     Text(v: string)
     Link(v: EntityId)
 }
-
-typealias Message = Sequence<MessageComponent>


tests/ZeroC.Slice.Codec.Tests/BitSequenceWriterTests.cs
```diff
@@ -4,8 +4,9 @@
 using System.IO.Pipelines;
 using ZeroC.Slice.Codec;
```
pepone: no longer need this using, same in other files  in the package.

tests/ZeroC.Slice.SlicecGen.Tests/TaggedTests.cs
```diff
@@ -0,0 +1,163 @@
+// Copyright (c) ZeroC, Inc.
+
```
pepone, line 2: I assume you just move these around without changing tests

tests/ZeroC.Slice.SlicecGen.Tests/WellKnownTypesTests.cs
```diff
@@ -2,34 +2,14 @@
```
pepone, line 2: We can rethink this test in a follow up PR, my understanding this is just testing a custom type. The code already tests the encoding/decoding of well known types  doesn't it?

externl:
Some slicec-cs tests need to be fixed as well.

tests/ZeroC.Slice.SlicecGen.Tests/TaggedTests.cs
bernardnormier: Yes, there was no change to any test.

tests/ZeroC.Slice.SlicecGen.Tests/WellKnownTypesTests.cs
bernardnormier: You're right, the name is not correct. I will rename it.

bernardnormier: Actually, it's testing well-known types in user-defined structs. Not custom types. This PR split the test file between ZeroC.Slice.Codec.Tests and ZeroC.Slice.SlicecGen.Tests.

PR #4285, opened 2026-03-09
Fix unknown reply status handling in ice protocol (#4281)

### LINKED ISSUE #4281: Add support for unknown Ice ReplyStatus
Ice 3.8 makes ReplyStatus extensible - you can define your own "system exception" with an error code (the reply status value) and a message. Future versions of Ice could also define additional ReplyStatus errors.

IceRPC's ice protocol doesn't support this, in particular this is not correct:
https://github.com/icerpc/icerpc-csharp/blob/37bc16cb09fae805523f5769ee37d37f3136ca22/src/IceRpc/Internal/IceProtocolConnection.cs#L730

ReplyStatus is now a "checked" exception, so we shouldn't use `AsReplyStatus` which throws for unknown enumerator values.

I think the encoding path is already correct and doesn't need to be updated.

pepone:
Decode unknown ReplyStatus values as their corresponding StatusCode instead of always mapping to InternalError. Custom status codes greater than Unauthorized now round-trip correctly through the ice protocol. Add tests for StatusCode round-trip encoding/decoding.

Fix #4281


externl:
```
IceRpc:Tests:IceProtocolConnectionTests:Dispatcher_failure(IceRpc:InlineDispatch
er,InternalError,"deadline message { Original StatusCode = DeadlineExceeded }")
       Assert.That(response.ErrorMessage, Is.EqualTo(expectedErrorMessage))     
       Expected string length 59 but was 16. Strings differ at index 16.        
       Expected: "deadline message { Original StatusCode = DeadlineExceeded }"  
       But was:  "deadline message"                                             
       ---------------------------^    
```

src/IceRpc/Internal/IceProtocolConnection.cs
```diff
@@ -903,9 +904,11 @@ private static void EncodeResponseHeader(
...
                     break;
                 default:
-                    encoder.EncodeReplyStatus(ReplyStatus.UnknownException);
-                    encoder.EncodeString(
-                        $"{response.ErrorMessage} {{ Original StatusCode = {response.StatusCode} }}");
+                    // Custom status codes greater than Unauthorized round-trip through the
+                    // ice protocol; others encode as UnknownException.
+                    encoder.EncodeReplyStatus(response.StatusCode > StatusCode.Unauthorized ?
+                        (ReplyStatus)response.StatusCode : ReplyStatus.UnknownException);
+                    encoder.EncodeString(response.ErrorMessage!);
```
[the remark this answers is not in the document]
pepone: Reverted this, to keep the original status code in the message.

tests/IceRpc.Tests/IceProtocolConnectionTests.cs
```diff
@@ -63,6 +63,29 @@ private static IEnumerable<TestCaseData> DispatchFailureSource
...
+            // InternalError encodes as ReplyStatus.UnknownException and decodes back to InternalError.
+            yield return new TestCaseData(StatusCode.InternalError, StatusCode.InternalError);
+
+            // Well-known status codes without a dedicated ReplyStatus encode as UnknownException,
+            // which decodes back to InternalError.
+            yield return new TestCaseData(StatusCode.ApplicationError, StatusCode.InternalError);
+            yield return new TestCaseData(StatusCode.DeadlineExceeded, StatusCode.InternalError);
+
+            // Custom status codes outside the defined ReplyStatus range round-trip correctly.
+            yield return new TestCaseData((StatusCode)42, (StatusCode)42);
```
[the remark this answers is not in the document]
pepone: Updated the comment to clarify this.

src/IceRpc/Internal/IceProtocolConnection.cs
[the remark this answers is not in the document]
pepone: Fixed to avoid truncating status codes.

src/IceRpc/Internal/IceProtocolConnection.cs
```diff
@@ -760,16 +760,17 @@ private static (StatusCode StatusCode, string? ErrorMessage, SequencePosition Co
...
                 ReplyStatus.ObjectNotExist => StatusCode.NotFound,
                 ReplyStatus.FacetNotExist => StatusCode.NotFound,
                 ReplyStatus.OperationNotExist => StatusCode.NotImplemented,
+                ReplyStatus.UnknownLocalException => StatusCode.InternalError,
+                ReplyStatus.UnknownUserException => StatusCode.InternalError,
+                ReplyStatus.UnknownException => StatusCode.InternalError,
                 ReplyStatus.InvalidData => StatusCode.InvalidData,
                 ReplyStatus.Unauthorized => StatusCode.Unauthorized,
-                _ => StatusCode.InternalError
+                _ => (StatusCode)replyStatus
```
bernardnormier: I don't find this correct at all. 

The ReplyStatus and StatusCode values are not synchronized. ReplyStatus 32 should not map to StatusCode 32 unless the IceRpc core knows them both. 

The only reasonable mapping when we receive an unknown ReplyStatus is to map it to StatusCode.InternalError (like we did previously).

src/IceRpc/Internal/IceProtocolConnection.cs
```diff
@@ -903,9 +904,21 @@ private static void EncodeResponseHeader(
                     encoder.EncodeString(response.ErrorMessage!);
                     break;
                 default:
-                    encoder.EncodeReplyStatus(ReplyStatus.UnknownException);
-                    encoder.EncodeString(
-                        $"{response.ErrorMessage} {{ Original StatusCode = {response.StatusCode} }}");
+                    // Custom status codes that fit in a single byte round-trip through the
```
bernardnormier: "round-tripping" should not work for unknown ReplyStatus values.

src/IceRpc/Internal/IceProtocolConnection.cs
pepone: How can an Ice server implemented with IceRPC set a custom reply status?

bernardnormier: It's not something that we support.

If we want to add such support, the application would need to perform the custom StatusCode -> ReplyStatus mapping through some callback.

tests/IceRpc.Tests/IceProtocolConnectionTests.cs
```diff
@@ -91,6 +144,38 @@ public async Task Dispatcher_failure(
         Assert.That(readResult.Buffer.IsEmpty, Is.True);
     }
 
+    /// <summary>Verifies that a StatusCode dispatched by the server is encoded as a ReplyStatus and decoded back to
+    /// the expected StatusCode by the client.</summary>
+    [Test, TestCaseSource(nameof(StatusCodeRoundTripSource))]
+    public async Task Response_reply_status_maps_to_expected_status_code(
```
bernardnormier: I would include "ice" in the test name, e.g. `Response_ice_reply_status_maps_to_expected_status_code`

tests/IceRpc.Tests/IceProtocolConnectionTests.cs
```diff
@@ -63,6 +63,59 @@ private static IEnumerable<TestCaseData> DispatchFailureSource
...
+            yield return new TestCaseData(
+                StatusCode.Unauthorized,
+                StatusCode.Unauthorized,
+                "The dispatch failed with status code Unauthorized.");
+            yield return new TestCaseData(
+                StatusCode.InternalError,
+                StatusCode.InternalError,
+                "The dispatch failed with status code InternalError.");
+
+            // Custom status codes that fit in a single byte round-trip correctly.
```
bernardnormier: You should fix the comment: it doesn't round-trip as "42".

tests/IceRpc.Tests/IceProtocolConnectionTests.cs
```diff
@@ -63,6 +63,59 @@ private static IEnumerable<TestCaseData> DispatchFailureSource
...
+            yield return new TestCaseData(
+                StatusCode.ApplicationError,
+                StatusCode.InternalError,
+                "The dispatch failed with status code ApplicationError. { Original StatusCode = ApplicationError }");
+            yield return new TestCaseData(
+                StatusCode.DeadlineExceeded,
+                StatusCode.InternalError,
+                "The dispatch failed with status code DeadlineExceeded. { Original StatusCode = DeadlineExceeded }");
+
+            // Custom status codes that don't fit in a single byte encode as UnknownException.
```
bernardnormier: You mean, InternalError? That's actually the same as 42, which fits in a byte.

PR #4288, opened 2026-03-10
Fully qualify type names in enum variant nested record classes

pepone:
## Summary
- Fix CS8910 errors in generated C# code for Slice enums with fields, where the nested record class name shadows the field type name (e.g., `record class Enum(Enum V)`)
- Force fully qualified type names with `global::` prefix for all user-defined types in enum variant primary constructor parameters
- Remove the now-unused `ContainerBuilder::add_fields` method

### DIFF
--- a/tools/slicec-cs/src/generators/enum_generator.rs
+++ b/tools/slicec-cs/src/generators/enum_generator.rs
@@ -136,7 +136,7 @@ fn enumerators_as_nested_records(enum_def: &Enum) -> CodeBlock {
         // We pass "" as the namespace to force fully qualified type names (global::...) for all
         // user-defined types. This avoids CS8910 errors where the nested record class name shadows
         // the field type name (e.g., `record class Enum(Enum V)` vs `record class Enum(global::...Enum V)`).
-        for field in &enumerator.fields() {
+        for field in enumerator.fields() {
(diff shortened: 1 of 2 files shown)


tools/slicec-cs/src/generators/enum_generator.rs
```diff
@@ -131,8 +131,19 @@ fn enumerators_as_nested_records(enum_def: &Enum) -> CodeBlock {
...
             .add_comments(enumerator.formatted_doc_comment_seealso())
             .add_obsolete_attribute(enumerator)
-            .add_base(enum_def.escape_identifier())
-            .add_fields(&enumerator.fields());
+            .add_base(enum_def.escape_identifier());
+
+        // We pass "" as the namespace to force fully qualified type names (global::...) for all
+        // user-defined types. This avoids CS8910 errors where the nested record class name shadows
+        // the field type name (e.g., `record class Enum(Enum V)` vs `record class Enum(global::...Enum V)`).
+        for field in &enumerator.fields() {
```
InsertCreativityHere: Very minor, but you don't need the borrow here:
```suggestion
        for field in enumerator.fields() {
```
This was just to conform to the type signature of `add_fields`.

InsertCreativityHere (approved):
Looks good to me!

PR #4293, opened 2026-03-11
Initial refactoring of service source generator

bernardnormier:
This PR combines the Slice service generator and the Ice service generator into a single service generator, IceRpc.ServiceGenerator.dll.

The controlling class attribute is now IceRpc.ServiceAttribute (in IceRpc.dll), and the source generator is packaged inside the main IceRpc NuGet package.

We want to merge the Protobuf service generator into this source generator in a follow-up PR.

Please note:

With this update, the IceRpc.Slice public API is primarily for use by the code generator / source generator, and not application code. So we have very few "using IceRpc.Slice;" left. We should consider further reducing this public API in a follow-up PR by moving ISliceFeature/SliceFeature to IceRpc.Features, and the ServiceProviderExtensions to IceRpc.Extensions.DependencyInjection.

The IceRpc.ServiceGenerator needs more refactoring to add Protobuf and also make Ice and Slice more separate.

### DIFF
--- a/src/IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.csproj
+++ b/src/IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.csproj
@@ -23,7 +23,5 @@
     <Content Include="../../LICENSE" Pack="true" PackagePath="/" />
     <Content Include="README.md" Pack="true" PackagePath="/" />
     <None Include="../../build/icerpc-icon.png" Pack="true" PackagePath="/" />
-    <None Include="IceRpc.ServiceGenerator.targets" Pack="true" PackagePath="build" />
-    <None Include="IceRpc.ServiceGenerator.targets" Pack="true" PackagePath="buildTransitive" />
   </ItemGroup>
 </Project>
--- a/src/IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.targets
+++ b/src/IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.targets
@@ -1,6 +1,0 @@
-<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <ItemGroup>
-    <Analyzer Include="$(MSBuildThisFileDirectory)../analyzers/dotnet/cs/netstandard2.0/IceRpc.ServiceGenerator.dll"
-              Condition="Exists('$(MSBuildThisFileDirectory)../analyzers/dotnet/cs/netstandard2.0/IceRpc.ServiceGenerator.dll')" />
-  </ItemGroup>
-</Project>


src/IceRpc.ServiceGenerator/Internal/ServiceMethod.cs
```diff
@@ -1,11 +1,14 @@
 // Copyright (c) ZeroC, Inc.
 
-namespace IceRpc.Slice.Generators.Internal;
+namespace IceRpc.ServiceGenerator.Internal;
 
 /// <summary>Represents an abstract method in a generated XxxService interface marked with the
 /// <c>IceRpc.Slice.SliceOperationAttribute</c> attribute.</summary>
```
pepone: Should we mention IceOperationAttribute too?

src/IceRpc.Ice/IceRpc.Ice.csproj
```diff
@@ -19,13 +19,8 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="../IceRpc/IceRpc.csproj" ExactVersion="true" />
+    <ProjectReference Include="../IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.csproj" ExactVersion="true" />
```
pepone: Why are we no longer using ` ReferenceOutputAssembly="false"`, shouldn't we also use `OutputItemType="Analyzer"` like the tests?

src/IceRpc.Slice/IceRpc.Slice.csproj
```diff
@@ -1,19 +1,15 @@
...
   <PropertyGroup>
-    <Description>IceRpc.Slice for C#.</Description>
+    <Description>IceRPC + Slice integration for C#.</Description>
     <AssemblyTitle>$(Description)</AssemblyTitle>
     <PackageTags>icerpc;slice;binary;serialization;format</PackageTags>
     <TargetFramework>net10.0</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="../IceRpc/IceRpc.csproj" ExactVersion="true" />
+    <ProjectReference Include="../IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.csproj" ExactVersion="true" />
```
pepone: ditto

pepone (approved):

src/IceRpc.ServiceGenerator/Internal/ServiceMethod.cs
bernardnormier: Fixed

src/IceRpc.Ice/IceRpc.Ice.csproj
bernardnormier: I fixed it as you suggested. And the current setup still pulls IceRpc.ServiceGenerator in projects with a reference to the IceRpc.Slice package.

bernardnormier: Actually, we can't use `ReferenceOutputAssembly="false"` as it prevents this dependency from flowing transitively to consumers of IceRpc.Slice/IceRpc.Ice. See updated project file.

src/IceRpc.ServiceGenerator/Internal/Parser.cs
```diff
@@ -173,8 +176,23 @@ private IReadOnlyList<ServiceMethod> GetServiceMethods(INamedTypeSymbol interfac
...
+                attribute = GetAttribute(method, _sliceOperationAttribute);
+                idl = Idl.Slice;
+            }
+            if (attribute is null && _iceOperationAttribute is not null)
             {
+                attribute = GetAttribute(method, _iceOperationAttribute);
+                idl = Idl.Ice;
+            }
+
+            if (attribute is null)
```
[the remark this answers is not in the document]
bernardnormier: These are generated methods, with a single attribute per method. We could emit a diagnostic if we encounter a method with more than one of these attributes.

PR #4294, opened 2026-03-12
Refactor IceRpc.ServiceGenerator

bernardnormier:
This PR refactors ServiceGenerator with a clean separation between the Ice and Slice support.

### DIFF
--- a/src/IceRpc.ServiceGenerator/Internal/SliceServiceMethod.cs
+++ b/src/IceRpc.ServiceGenerator/Internal/SliceServiceMethod.cs
@@ -18,7 +18,7 @@ internal class SliceServiceMethod : ServiceMethod
 
     private static readonly string[] _usingDirectives =
     [
-        "using IceRpc.Slice;",
+        "using IceRpc.Slice.Operations;",
         "using ZeroC.Slice.Codec;",
     ];
 
@@ -295,7 +295,7 @@ internal class SliceServiceMethodFactory : ServiceMethodFactory
     private readonly INamedTypeSymbol? _pipeReaderSymbol;
 
     internal SliceServiceMethodFactory(Compilation compilation)
-        : base(compilation.GetTypeByMetadataName("IceRpc.Slice.SliceOperationAttribute"))
+        : base(compilation.GetTypeByMetadataName("IceRpc.Slice.Operations.SliceOperationAttribute"))
     {
         _asyncEnumerableSymbol = compilation.GetTypeByMetadataName("System.Collections.Generic.IAsyncEnumerable`1");
         _pipeReaderSymbol = compilation.GetTypeByMetadataName("System.IO.Pipelines.PipeReader");
--- a/src/IceRpc.Slice/DependencyInjection/SliceServiceProviderExtensions.cs
+++ b/src/IceRpc.Slice/DependencyInjection/SliceServiceProviderExtensions.cs
@@ -1,6 +1,8 @@
 // Copyright (c) ZeroC, Inc.
 
-namespace IceRpc.Slice;
+using IceRpc.Slice;
+
+namespace IceRpc.Extensions.DependencyInjection;
 
 /// <summary>Provides extension methods for <see cref="IServiceProvider" /> to create Slice proxies.</summary>
 public static class SliceServiceProviderExtensions
--- a/src/IceRpc.Slice/Features/ISliceFeature.cs
+++ b/src/IceRpc.Slice/Features/ISliceFeature.cs
@@ -1,6 +1,8 @@
 // Copyright (c) ZeroC, Inc.
 
-namespace IceRpc.Slice;
+using IceRpc.Slice;
+
+namespace IceRpc.Features;
 
 /// <summary>Represents a feature used to customize the encoding and decoding of request and response payloads.
 /// </summary>
--- a/src/IceRpc.Slice/Features/SliceFeature.cs
+++ b/src/IceRpc.Slice/Features/SliceFeature.cs
@@ -1,6 +1,8 @@
 // Copyright (c) ZeroC, Inc.
 
-namespace IceRpc.Slice;
+using IceRpc.Slice;
+
+namespace IceRpc.Features;
 
 /// <summary>Default implementation of <see cref="ISliceFeature" />.</summary>
 public sealed class SliceFeature : ISliceFeature
--- a/src/IceRpc.Slice/Operations/AsyncEnumerableExtensions.cs
+++ b/src/IceRpc.Slice/Operations/AsyncEnumerableExtensions.cs
@@ -4,7 +4,7 @@ using System.Diagnostics;
 using System.IO.Pipelines;
 using ZeroC.Slice.Codec;
 
-namespace IceRpc.Slice;
+namespace IceRpc.Slice.Operations;
 
 /// <summary>Provides an extension method for <see cref="IAsyncEnumerable{T}" /> to encode elements into a <see
 /// cref="PipeReader"/>.</summary>
--- a/src/IceRpc.Slice/Operations/IncomingRequestExtensions.cs
+++ b/src/IceRpc.Slice/Operations/IncomingRequestExtensions.cs
@@ -1,11 +1,11 @@
 // Copyright (c) ZeroC, Inc.
 
 using IceRpc.Features;
-using IceRpc.Slice.Internal;
+using IceRpc.Slice.Operations.Internal;
 using System.IO.Pipelines;
 using ZeroC.Slice.Codec;
 
-namespace IceRpc.Slice;
+namespace IceRpc.Slice.Operations;
 
 /// <summary>Provides extension methods for <see cref="IncomingRequest" /> to decode its Slice-encoded payload.
 /// </summary>
--- a/src/IceRpc.Slice/Operations/IncomingResponseExtensions.cs
+++ b/src/IceRpc.Slice/Operations/IncomingResponseExtensions.cs
@@ -1,9 +1,10 @@
 // Copyright (c) ZeroC, Inc.
 
-using IceRpc.Slice.Internal;
+using IceRpc.Features;
+using IceRpc.Slice.Operations.Internal;
 using ZeroC.Slice.Codec;
 
-namespace IceRpc.Slice;
+namespace IceRpc.Slice.Operations;
 
 /// <summary>Provides extension methods for <see cref="IncomingResponse" /> to decode its Slice-encoded payload.
 /// </summary>
--- a/src/IceRpc.Slice/Operations/Internal/IncomingFrameExtensions.cs
+++ b/src/IceRpc.Slice/Operations/Internal/IncomingFrameExtensions.cs
@@ -1,10 +1,11 @@
 // Copyright (c) ZeroC, Inc.
 
+using IceRpc.Features;
 using IceRpc.Internal;
 using System.IO.Pipelines;
 using ZeroC.Slice.Codec;
 
-namespace IceRpc.Slice.Internal;
+namespace IceRpc.Slice.Operations.Internal;
 
 /// <summary>Provides extension methods for <see cref="IncomingFrame" /> to decode its payload when this payload is
 /// encoded with the Slice encoding.</summary>
--- a/src/IceRpc.Slice/Operations/PipeReaderExtensions.cs
+++ b/src/IceRpc.Slice/Operations/PipeReaderExtensions.cs
@@ -1,5 +1,6 @@
 // Copyright (c) ZeroC, Inc.
 
+using IceRpc.Features;
 using IceRpc.Internal;
 using System.Buffers;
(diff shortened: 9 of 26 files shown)


src/IceRpc.ServiceGenerator/Internal/Emitter.cs
```diff
@@ -102,136 +107,9 @@ private static CodeBlock GenerateDispatchBody(ServiceClass serviceClass)
         }
     }
 
-    private static CodeBlock GenerateDispatchCase(ServiceMethod serviceMethod) =>
+    private static CodeBlock GenerateDispatchCase(IServiceMethod serviceMethod) =>
         @$"case ""{serviceMethod.OperationName}"":
-{GenerateDispatchCaseBody(serviceMethod)}";
-
-    private static CodeBlock GenerateDispatchCaseBody(ServiceMethod serviceMethod)
```
bernardnormier, line 114: Moved to (Ice|Slice)ServiceMethod.GenerateDiscpatchCaseBody.

src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs
```diff
@@ -0,0 +1,245 @@
...
+    public string OperationName { get; }
+
+    public IEnumerable<string> UsingDirectives => _usingDirectives;
+
+    private static readonly string[] _usingDirectives =
+    [
+        "using IceRpc.Ice;",
+        "using ZeroC.Slice.Codec;",
+    ];
+
```
bernardnormier, line 24: No field for streams or compressReturn.

src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs
```diff
@@ -0,0 +1,245 @@
...
+                _returnCount = 1;
+            }
+        }
+    }
+}
+
+internal class IceServiceMethodFactory : ServiceMethodFactory
+{
+    internal IceServiceMethodFactory(Compilation compilation)
+        : base(compilation.GetTypeByMetadataName("IceRpc.Ice.IceOperationAttribute"))
```
bernardnormier, line 237: Fully encapsulated here.

src/IceRpc.ServiceGenerator/Internal/ServiceMethod.cs
```diff
@@ -0,0 +1,73 @@
...
+
+/// <summary>The common base implementation of <see cref="IServiceMethodFactory"/>.</summary>
+internal abstract class ServiceMethodFactory : IServiceMethodFactory
+{
+    private readonly INamedTypeSymbol? _operationAttribute;
+
+    public bool TryCreate(IMethodSymbol methodSymbol, out IServiceMethod? serviceMethod)
+    {
+        serviceMethod = null;
+        if (_operationAttribute is null)
```
bernardnormier, line 60: _operationAttribute corresponds to the operation-setting attribute:
IceRpc.Slice.SliceOperationAttribute for Slice
IceRpc.Slice.IceOperationAttribute for Ice
IceRpc.Protobuf.ProtobufServiceMethodAttribute for Protobuf (we should rename it)

src/IceRpc.ServiceGenerator/Internal/Parser.cs
```diff
@@ -119,27 +112,14 @@ static bool IsAllowedKind(SyntaxKind kind) =>
         return serviceDefinitions;
     }
 
-    private AttributeData? GetAttribute(ISymbol symbol, INamedTypeSymbol attributeSymbol)
```
bernardnormier, line 122: Moved to SymbolExtensions.

src/IceRpc.ServiceGenerator/Internal/Parser.cs
```diff
@@ -148,208 +128,31 @@ static bool IsAllowedKind(SyntaxKind kind) =>
         return null;
     }
 
-    private string GetFullName(ISymbol symbol)
```
bernardnormier, line 151: Moved to SymbolExtensions

src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs
```diff
@@ -0,0 +1,245 @@
...
+    private static readonly string[] _usingDirectives =
+    [
+        "using IceRpc.Ice;",
+        "using ZeroC.Slice.Codec;",
+    ];
+
+    /// <summary>The name of the C# method minus the Async suffix. For example: "FindObjectById".</summary>
+    private readonly string _dispatchMethodName;
+
+    /// <summary>A value indicating whether the non-stream portion of the return value is pre-encoded by the
```
bernardnormier: Bug: remove stream

src/IceRpc.ServiceGenerator/Internal/Parser.cs
```diff
@@ -64,20 +57,20 @@ internal IReadOnlyList<ServiceClass> GetServiceDefinitions(IEnumerable<ClassDecl
...
                 if (baseServiceClass is not null)
                 {
                     baseServiceMethods = GetServiceMethods(baseServiceClass.AllInterfaces);
                 }
 
-                IReadOnlyList<ServiceMethod> serviceMethods = GetServiceMethods(classSymbol.AllInterfaces)
+                IReadOnlyList<IServiceMethod> serviceMethods = GetServiceMethods(classSymbol.AllInterfaces)
                     .Except(baseServiceMethods)
                     .Distinct()
                     .ToList();
```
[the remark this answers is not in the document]
bernardnormier: This looks like a pre-existing bug we should fix.

We actually have a test for this case, and we should check its generated code.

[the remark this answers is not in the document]
bernardnormier: This bug was actually introduced by this PR when I converted ServiceMethod from record struct to an interface (later an abstract class).

Fixed.

src/IceRpc.ServiceGenerator/Internal/SymbolExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using Microsoft.CodeAnalysis;
+using System.Collections.Immutable;
+
+namespace IceRpc.ServiceGenerator.Internal;
+
+/// <summary>Extension methods for <see cref="ISymbol"/>.</summary>
+internal static class SymbolExtensions
+{
+    internal static AttributeData? GetAttribute(this ISymbol symbol, INamedTypeSymbol attributeSymbol)
```
pepone: I think this should be named `FindAttribute`.

src/IceRpc.ServiceGenerator/Internal/SliceServiceMethod.cs
```diff
@@ -0,0 +1,310 @@
...
+        : base(compilation.GetTypeByMetadataName("IceRpc.Slice.SliceOperationAttribute"))
+    {
+        _asyncEnumerableSymbol = compilation.GetTypeByMetadataName("System.Collections.Generic.IAsyncEnumerable`1");
+        _pipeReaderSymbol = compilation.GetTypeByMetadataName("System.IO.Pipelines.PipeReader");
+    }
+
+    private protected override ServiceMethod CreateServiceMethod(
+        IMethodSymbol methodSymbol,
+        AttributeData attribute) =>
+        new SliceServiceMethod(methodSymbol, attribute, _asyncEnumerableSymbol, _pipeReaderSymbol);
```
pepone:
```suggestion
    private protected override ServiceMethod CreateServiceMethod(IMethodSymbol methodSymbol, AttributeData attribute) =>
        new SliceServiceMethod(methodSymbol, attribute, _asyncEnumerableSymbol, _pipeReaderSymbol);
```

src/IceRpc.ServiceGenerator/Internal/Emitter.cs
```diff
@@ -12,11 +12,16 @@ internal string Emit(ServiceClass serviceClass, CancellationToken cancellationTo
         cancellationToken.ThrowIfCancellationRequested();
 
         CodeBlock codeBlock = Preamble();
-        foreach (Idl idl in serviceClass.ServiceMethods.Select(serviceMethod => serviceMethod.Idl).Distinct())
+
+        // The service methods for the same IDL all return the same IEnumerable<string>, so we typically have a single
+        // element after Distinct().
+        IEnumerable<IEnumerable<string>> usingGroups =
+            serviceClass.ServiceMethods.Select(serviceMethod => serviceMethod.UsingDirectives).Distinct();
```
pepone, line 19: Why do we need using directives for each service method? Can't we just have a set of using directives per service class?

I guess you can still have Ice and IceRpc methods in the same service?

src/IceRpc.ServiceGenerator/Internal/SliceServiceMethod.cs
```diff
@@ -0,0 +1,310 @@
...
+            .ArgumentsOnNewLine(true)
+            .AddArgument(
+                $"decodeArgs: global::{_fullInterfaceName}.Request.Decode{_dispatchMethodName}Async")
+            .AddArgument($"method: {method}");
+
+        // We don't use the generated Response.EncodeXxx method when _returnCount is 0. So we could not generate it.
+        if (_returnCount > 0)
+        {
+            if (_returnCount == 1)
+            {
```
pepone: We can remove this nested if

```
if (_refCount == 1)
{
....
}
else if (_refCount > 1)
{
...
}
// We don't use the generated ...
```

src/IceRpc.ServiceGenerator/Internal/Parser.cs
```diff
@@ -32,19 +24,20 @@ internal Parser(
...
-        _asyncEnumerableSymbol = _compilation.GetTypeByMetadataName("System.Collections.Generic.IAsyncEnumerable`1");
-        _iceOperationAttribute = _compilation.GetTypeByMetadataName(IceOperationAttribute);
-        _pipeReaderSymbol = _compilation.GetTypeByMetadataName("System.IO.Pipelines.PipeReader");
-        _serviceAttribute = _compilation.GetTypeByMetadataName(ServiceAttribute);
-        _sliceOperationAttribute = _compilation.GetTypeByMetadataName(SliceOperationAttribute);
+        _serviceMethodFactoryList = new IServiceMethodFactory[]
+        {
+            new SliceServiceMethodFactory(compilation),
+            new IceServiceMethodFactory(compilation),
+        };
```
pepone: use collection expression:

```
_serviceMethodFactoryList =
[
    new SliceServiceMethodFactory(compilation),
    new IceServiceMethodFactory(compilation),
];
```

src/IceRpc.ServiceGenerator/Internal/Parser.cs
```diff
@@ -32,19 +24,20 @@ internal Parser(
...
     }
 
     internal IReadOnlyList<ServiceClass> GetServiceDefinitions(IEnumerable<ClassDeclarationSyntax> classes)
     {
-        if ((_iceOperationAttribute is null && _sliceOperationAttribute is null) || _serviceAttribute is null)
+        if (_serviceAttribute is null)
         {
-            // nothing to do if these types aren't available
+            // nothing to do
```
pepone, line 40: Wondering if this should be an error or an assert. It seems unlikely we can hit with a `ServiceAttribute` defined in the same assembly.

src/IceRpc.ServiceGenerator/Internal/Emitter.cs
bernardnormier: Yes, a service class doesn't know which IDL(s) are used to define its operations. And I don't want to generate using directives for all known IDLs "just in case".

So my solution was:
 - each ServiceMethod provides the using directives it needs (based on its IDL)
 - the service class then gather all these using directives (as `IEnumerable<string>`) and eliminate duplicates before generating the code.

Note that at present, I don't think you can use Slice and Ice for the same service because some extension methods for proxies and services have the same names (like CheckNonIdempotent). We should fix this (and write a test) in a follow-up PR.

src/IceRpc.ServiceGenerator/Internal/Parser.cs
bernardnormier: I don't think it should be an assert or an error.

It could be a diagnostic: you add a reference to IceRpc.ServiceGenerator to a project that doesn't use IceRpc ... the source generator does nothing in this case, which is fine. Probably also means you added this reference by accident.

src/IceRpc.ServiceGenerator/Internal/SliceServiceMethod.cs
bernardnormier: Fixed

src/IceRpc.ServiceGenerator/Internal/Parser.cs
bernardnormier: Fixed as suggested.

externl (approved):
Looks good!

PR #4295, opened 2026-03-12
Update IceRpc.Slice structure

bernardnormier:
This PR updates the IceRpc.Slice structure and moves most classes to IceRpc.Slice.Operations.

"Operations" because it's support code for Slice Operations. For Ice-Slice, it will be Operations as well. For Protobuf, it will be Methods or RpcMethods.

The classes in IceRpc.Slice.Operations are primarily extension classes, used by code generated by either slicec-cs or IceRpc.ServiceGenerator.

As a result, these two code generators generate "using IceRpc.Slice.Operations;" in the generated file. Regular applications can use public symbols in IceRpc.Slice.Operations but are not expected to.

After this update, there are only a few classes remaining directly in IceRpc.Slice: ISliceProxy, SliceEncodeOptions, and the SliceEncoder/Decoder extensions for ServiceAddress.

We could remove the recently added Slice prefix from SliceProxy and SliceEncodeOptions since  "using IceRpc.Slice" + "using IceRpc.Ice" should be extremely uncommon.

### DIFF
--- a/tests/IceRpc.Slice.Tests/ServiceAddressTests.cs
+++ b/tests/IceRpc.Slice.Tests/ServiceAddressTests.cs
@@ -1,6 +1,5 @@
 // Copyright (c) ZeroC, Inc.
 
-using IceRpc.Slice.Operations;
 using NUnit.Framework;
 using ZeroC.Slice.Codec;
 using ZeroC.Tests.Common;


src/IceRpc.ServiceGenerator/Internal/Emitter.cs
```diff
@@ -15,6 +15,11 @@ internal string Emit(ServiceClass serviceClass, CancellationToken cancellationTo
         foreach (Idl idl in serviceClass.ServiceMethods.Select(serviceMethod => serviceMethod.Idl).Distinct())
         {
             codeBlock.WriteLine($"using IceRpc.{idl};");
+            // TODO: temporary
+            if (idl == Idl.Slice)
```
bernardnormier, line 19: Fixed in #4294.

src/IceRpc.Slice/Extensions/SliceOperationAttribute.cs
```diff
@@ -1,9 +1,8 @@
 // Copyright (c) ZeroC, Inc.
 
 using System.ComponentModel;
-using ZeroC.Slice.Codec;
 
-namespace IceRpc.Slice;
+namespace IceRpc.Slice.Extensions;
```
bernardnormier: The SliceOperationAttribute is not an extension but it's used exclusively by the code generators - that's how slicec-cs communicates info to IceRpc.ServiceGenerator.

pepone (approved):

PR #4297, opened 2026-03-13
Add ZeroC.Slice.Compiler and ZeroC.Slice.Symbols packages

pepone:
- ZeroC.Slice.Compiler contains the Slice compiler encode and decode generated code.
- ZeroC.Slice.Symbols contains the model to use in code generators.

### DIFF
--- a/src/ZeroC.Slice.Symbols/EnumWithFields.cs
+++ b/src/ZeroC.Slice.Symbols/EnumWithFields.cs
@@ -41,6 +41,11 @@ public record class EnumWithFields : Symbol
         public required EntityInfo EntityInfo { get; init; }
 
         /// <summary>
+        /// Gets the discriminant value used for encoding/decoding this enumerator.
+        /// </summary>
+        public required int Discriminant { get; init; }
+
+        /// <summary>
         /// Gets the list of fields for this enumerator.
         /// </summary>
         public required ImmutableList<Field> Fields { get; init; }
--- a/src/ZeroC.Slice.Symbols/SymbolConverter.cs
+++ b/src/ZeroC.Slice.Symbols/SymbolConverter.cs
@@ -117,7 +117,12 @@ public sealed class SymbolConverter
             return cached;
         }
 
-        (Compiler.SliceFile file, Compiler.Symbol symbol) = _named[typeId];
+        if (!_named.TryGetValue(typeId, out var entry))
+        {
+            throw new InvalidOperationException($"Failed to resolve named type with ID '{typeId}'.");
+        }
+
+        (Compiler.SliceFile file, Compiler.Symbol symbol) = entry;
         Module module = ConvertModule(file.ModuleDeclaration);
         Symbol converted = ConvertSymbol(symbol, file, module);
         _cache[typeId] = converted;
@@ -190,6 +195,7 @@ public sealed class SymbolConverter
                 Enumerators = raw.Enumerators.Select(e => new EnumWithFields.Enumerator
                 {
                     EntityInfo = ConvertEntityInfo(e.EntityInfo, module),
+                    Discriminant = (int)e.Value.AbsoluteValue,
                     Fields = e.Fields.Select(f => ConvertField(f, file, module)).ToImmutableList(),
                 }).ToImmutableList(),
             };
--- a/src/ZeroC.Slice.Symbols/Attribute.cs
+++ b/src/ZeroC.Slice.Symbols/Attribute.cs
@@ -25,7 +25,7 @@ public record class Attribute
     public const string CsReadonly = "cs::readonly";
 
     /// <summary>
-    /// The attributes directive, e.g. "cs::readonly" for [cs::readonly] attribute.
+    /// The attribute's directive, e.g. "cs::readonly" for [cs::readonly] attribute.
     /// </summary>
     public required string Directive { get; init; }
 
--- a/src/ZeroC.Slice.Symbols/TypeRef.cs
+++ b/src/ZeroC.Slice.Symbols/TypeRef.cs
@@ -31,7 +31,7 @@ public record class TypeRef
     {
         Builtin b => b.Kind != BuiltinKind.String,
         Struct => true,
-        EnumWithUnderlying e => !e.IsUnchecked,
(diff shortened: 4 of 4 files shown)


src/ZeroC.Slice.Symbols/ZeroC.Slice.Symbols.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../ZeroC.Slice.Codec/ZeroC.Slice.Codec.csproj" />
+    <ProjectReference Include="../ZeroC.Slice.Compiler/ZeroC.Slice.Compiler.csproj" />
```
bernardnormier, line 9: Should we include the ZeroC.Slice.Compiler "checked-in generated code" directly in the ZeroC.Slice.Symbols assembly?

The advantage of putting ZeroC.Slice.Compiler in its own assembly and NuGet package is not immediately clear to me.

src/ZeroC.Slice.Symbols/SymbolConverter.cs
```diff
@@ -0,0 +1,266 @@
...
+                {
+                    string key = string.IsNullOrEmpty(moduleScope) ? id : $"{moduleScope}::{id}";
+                    _named.TryAdd(key, (file, symbol));
+                }
+            }
+        }
+    }
+
+    /// <summary>Converts source files into rich symbol types with all TypeRefs resolved.</summary>
+    public ImmutableList<SliceFile> ConvertFiles(IEnumerable<Compiler.SliceFile> sourceFiles) =>
```
bernardnormier, line 57: I don't follow this API. Presumably the sourceFiles must be a subset of allFiles passed to the constructor?

src/ZeroC.Slice.Symbols/Field.cs
```diff
@@ -0,0 +1,36 @@
...
+/// <summary>
+/// Represents a field in a Slice definition.
+/// Slice fields are used in various contexts, such as structs, enumerations, operation parameters, and return values.
+/// </summary>
+public record class Field
+{
+    /// <summary>
+    /// Gets the field's entity information.
+    /// </summary>
+    public required EntityInfo EntityInfo { get; init; }
```
bernardnormier, line 14: I think the Symbols API could/should diverge more from the Compiler API, with interfaces, inheritance, references ... all things that can't be expressed in the Compiler API (for good reasons).

For example, we could have an  Entity base class and an `IType` interface (represents a Slice type), and Field would be a more natural:
```cs
public record class Field : Entity
{
    public IType Type { get; }
    public int? Tag { get; }
    ...
}
```

The exact API depends on what's most convenient for the code generator.

src/ZeroC.Slice.Symbols/EnumWithFields.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        /// Gets the enumerator's entity information.
+        /// </summary>
+        public required EntityInfo EntityInfo { get; init; }
+
+        /// <summary>
+        /// Gets the list of fields for this enumerator.
+        /// </summary>
+        public required ImmutableList<Field> Fields { get; init; }
+    }
```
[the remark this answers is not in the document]
pepone: Added missing discriminant

src/ZeroC.Slice.Symbols/SymbolConverter.cs
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    private Interface ConvertInterface(Compiler.Interface raw, Compiler.SliceFile file, Module module) =>
+        new()
+        {
+            EntityInfo = ConvertEntityInfo(raw.EntityInfo, module),
+            Bases = raw.Bases
+            .Select(baseId => ResolveNamedType(baseId))
+            .OfType<Interface>()
+            .ToImmutableList(),
```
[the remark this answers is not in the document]
pepone: Bases are warrantee to be interfaces.

src/ZeroC.Slice.Compiler/ZeroC.Slice.Compiler.csproj
```diff
@@ -0,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
```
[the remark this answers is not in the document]
pepone: These are internal packages for now.

src/ZeroC.Slice.Symbols/ZeroC.Slice.Symbols.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
```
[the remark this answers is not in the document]
pepone: These are internal packages.

src/ZeroC.Slice.Symbols/SymbolConverter.cs
pepone: In the draft generator I'm using for testing. I use it like.

```
// Convert decoded types into rich symbols with resolved references.
var converter = new ZeroC.Slice.Symbols.SymbolConverter(sourceFiles.Concat(referenceFiles));
ImmutableList<ZeroC.Slice.Symbols.SliceFile> symbolFiles = converter.ConvertFiles(sourceFiles);
```

bernardnormier: Should we simply mimic the code generator API, with a static method like:

```cs
ImmutableList<ZeroC.Slice.Symbols.SliceFile> symbolFiles = SymbolConverter.ConvertFiles(sourceFiles, referenceFiles);
```

PR #4299, opened 2026-03-13
Initial code generator to replace slicec-cs

pepone:


bernardnormier:
Can I use the code in this PR to generate code from a Slice file? If yes, how do I do that?

pepone:
You can do something like:

```
../slicec/target/debug/slicec slice/Compiler/*.slice | dotnet run --project src/ZeroC.Slice.Generator/ZeroC.Slice.Generator.csproj
```

- You need to use slicec from Austin PR
- It currently prints the code to stdout.

src/ZeroC.Slice.Generator/StringExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+internal static class StringExtensions
+{
+    /// <summary>Converts a Slice identifier to PascalCase.</summary>
+    internal static string ToPascalCase(this string identifier)
+    {
+        var sb = new StringBuilder(identifier.Length);
+        bool newWord = true;
+        foreach (char c in identifier)
+        {
+            if (char.IsLetterOrDigit(c))
```
bernardnormier: I think we should also support snake case conversion. See https://github.com/zeroc-ice/ice/blob/cecd85079bd112c3dd9c3e8285554ce41d6df5d5/cpp/src/Slice/SliceUtil.cpp#L251.

src/ZeroC.Slice.Symbols/Symbol.cs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) ZeroC, Inc.
+
+namespace ZeroC.Slice.Symbols;
+
+/// <summary>
+/// Base for all types that can be referenced by a <see cref="TypeRef"/>.
```
bernardnormier: I think it would be useful to distinguish between Type (with an IType C# interface?) and Interface. 

A field can only reference a Type, not an Interface. In other words, a TypeRef always can only hold a Type.
And I believe only Interfaces reference Interfaces (for base Interfaces).

src/ZeroC.Slice.Symbols/Attribute.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace ZeroC.Slice.Symbols;
+
+/// <summary>
+/// Represents an attribute applied to a Slice declaration.
+/// </summary>
+public record class Attribute
+{
+    /// <summary>The directive for the cs::attribute attribute.</summary>
+    public const string CsAttribute = "cs::attribute";
```
bernardnormier: Should we move these const to ZeroC.Slice.Generator?

ZeroC.Slice.Symbols seems to be (and should be) language agnostic with respect to attributes. It's the Generator that checks and consumes these cs attributes.

src/ZeroC.Slice.Symbols/Compiler/CodeGenerator.cs
```diff
@@ -0,0 +1,417 @@
...
+#pragma warning disable CS1573 // Parameter has no matching param tag in the XML comment
+#pragma warning disable CS0612 // Type or member is obsolete
+#pragma warning disable CS0618 // Type or member is obsolete
+#pragma warning disable CS0619 // Type or member is obsolete
+
+using ZeroC.Slice.Codec;
+
+[assembly:Slice("CodeGenerator.slice")]
+
+namespace ZeroC.Slice.Compiler;
```
bernardnormier: Since we control how/where we map module Compiler in C#, what about ZeroC.Slice.Symbols.Internal.Compiler?

src/ZeroC.Slice.Generator/CsAttributeExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using ZeroC.CodeBuilder;
+using ZeroC.Slice.Symbols;
+
+using Attribute = ZeroC.Slice.Symbols.Attribute;
+
+namespace ZeroC.Slice.Generator;
+
+/// <summary>Extension methods for writing cs::attribute attributes to code builders.</summary>
+internal static class CsAttributeExtensions
```
bernardnormier: An XxxExtensions class should hold only extensions for Xxx.

src/ZeroC.Slice.Symbols/EntityInfo.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Collections.Immutable;
+
+namespace ZeroC.Slice.Symbols;
+
+/// <summary>
+/// Represents the information for an entity defined in Slice.
+/// </summary>
+public record class EntityInfo
```
bernardnormier: Notwithstanding the "prefer composition over inheritance" advice from the Design Patterns book, I feel a base Entity record class would be more natural in C#.

src/ZeroC.Slice.Symbols/Attribute.cs
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Collections.Immutable;
+
+namespace ZeroC.Slice.Symbols;
+
+/// <summary>
```
bernardnormier: Nitpick: I prefer the more compact summary style we use throughout the IceRpc code base, even though it's not the Microsoft default style. 

```cs
/// <summary>....</summary>
```

src/ZeroC.Slice.Symbols/TypeRef.cs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Collections.Immutable;
+
+namespace ZeroC.Slice.Symbols;
+
+/// <summary>
+/// Represents a reference to a Slice type.
+/// </summary>
+public record class TypeRef
```
bernardnormier: Should this be a struct?

I think we should use classes (possibly record classes) for "shared" objects, like say n Struct that can be referenced by multiple Fields through a TypeRef, or more generally entities that can be referenced by doc-comments.

On the other hand, a TypeRef can't be referenced so a struct would make more sense.

src/ZeroC.Slice.Symbols/Struct.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Collections.Immutable;
+
+namespace ZeroC.Slice.Symbols;
+
+/// <summary>
+/// Represents a struct defined in Slice.
+/// </summary>
+public record class Struct : Symbol
```
bernardnormier: I don't see the logic of using `record` classes over classes for many types.

record classes are meaningful when you're comparing two instances for equality based on the values of its fields. I don't think this should ever happen for Symbols. Reference equality is sufficient.

src/ZeroC.Slice.Generator/Generator.cs
```diff
@@ -0,0 +1,383 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Collections.Immutable;
+using ZeroC.CodeBuilder;
+using ZeroC.Slice.Symbols;
+using Attribute = ZeroC.Slice.Symbols.Attribute;
+
+namespace ZeroC.Slice.Generator;
+
+/// <summary>Abstract base class for code generators. Owns the type registry (symbol-to-namespace mapping) and
```
bernardnormier: I don't like this huge helper base class.

It's also not clear to me what this type registry is. I can't spot it; and if we have a type registry, it would be a code smell.

src/ZeroC.Slice.Symbols/EnumWithFields.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Collections.Immutable;
+
+namespace ZeroC.Slice.Symbols;
+
+/// <summary>
+/// Represents a Slice enumeration with fields, mapped to a Dunet discriminated union.
```
bernardnormier: We should not describe the C# mapping in the Symbols doc-comments.

src/ZeroC.Slice.Symbols/Field.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <summary>
+    /// Gets a value indicating whether this field is tagged. When <see langword="true"/>, the <see cref="Tag"/>
+    /// property has a value; otherwise, it is <see langword="null"/>.
+    /// </summary>
+    public bool IsTagged => Tag.HasValue;
+
+    /// <summary>
+    /// Gets a value indicating whether this field should have the 'required' keyword (non-optional reference type).
+    /// </summary>
+    public bool IsRequired => !Type.IsOptional && !Type.IsValueType;
```
bernardnormier: Since IsRequired is C#-mapping specific, this should be an extension (in Generator).

src/ZeroC.Slice.Generator/StringExtensions.cs
pepone: Updated to reuse the C++ implementation in slice2cs.

src/ZeroC.Slice.Symbols/EnumWithUnderlying.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        /// <summary>
+        /// Gets the enumerator's entity information.
+        /// </summary>
+        public required EntityInfo EntityInfo { get; init; }
+
+        /// <summary>
+        /// Gets the absolute value of this enumerator.
+        /// </summary>
+        public required ulong AbsoluteValue { get; init; }
```
bernardnormier: My idea for the C# code was to use a generic and avoid this awkward AbsoluteValue, as in:

`Enumerator<T>` and `Enum<T>` where T is the corresponding C# underlying numeric type (I avoid "mapped" here as ZeroC.Slice.Symbols should be language-mapping agnostic).

PR #4301, opened 2026-03-13
Move Protobuf source generator into IceRpc.ServiceGenerator

bernardnormier:
Protobuf now uses `[Service]` like Ice and Slice.

### DIFF
--- a/src/IceRpc.ServiceGenerator/Internal/ProtobufServiceMethod.cs
+++ b/src/IceRpc.ServiceGenerator/Internal/ProtobufServiceMethod.cs
@@ -7,7 +7,7 @@ using ZeroC.CodeBuilder;
 
 namespace IceRpc.ServiceGenerator.Internal;
 
-/// <summary>Implements <see cref="ServiceMethod" /> for the Slice IDL.</summary>
+/// <summary>Implements <see cref="ServiceMethod" /> for the Protobuf IDL.</summary>
 internal class ProtobufServiceMethod : ServiceMethod
 {
     /// <inheritdoc />


src/IceRpc.ServiceGenerator/Internal/ProtobufServiceMethod.cs
```diff
@@ -0,0 +1,105 @@
...
+    private readonly string _interfaceName;
+
+    // The kind of the RPC method: "Unary", "ClientStreaming", "ServerStreaming", or "BidiStreaming".
+    private readonly string _methodKind;
+
+    // The name of the mapped C# method on the Service interface. For example: "GreetAsync".
+    private readonly string _methodName;
+
+    /// <inheritdoc />
+    internal override CodeBlock GenerateDispatchCaseBody() =>
```
bernardnormier, line 35: As you can see, the Protobuf-specific code generation is very simple.

src/IceRpc.ServiceGenerator/Internal/ProtobufServiceMethod.cs
```diff
@@ -0,0 +1,105 @@
...
+    $@"return request.Dispatch{_methodKind}Async(
+        global::{_inputTypeName}.Parser,
+        (global::{_interfaceName})this,
+        static (service, input, features, cancellationToken) =>
+            service.{_methodName}(input, features, cancellationToken),
+        cancellationToken);";
+
+    internal ProtobufServiceMethod(
```
bernardnormier, line 42: The Protobuf-specific parsing, previously in Parser.cs.

tests/IceRpc.Protobuf.Tests/ServiceTests.cs
```diff
@@ -82,7 +82,7 @@ public async Task Proto_less_service_returns_not_implemented()
         Assert.That(response.StatusCode, Is.EqualTo(StatusCode.NotImplemented));
     }
 
-    [ProtobufService]
+    [Service]
     internal partial class MultipleServices : IMyFirstWidgetService, IMySecondWidgetService
```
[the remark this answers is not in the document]
bernardnormier: You're mistaken, the test itself is in an IceRpc subnamespace.

src/IceRpc.Protobuf/IceRpc.Protobuf.csproj
```diff
@@ -1,28 +1,27 @@
...
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="../../src/IceRpc/IceRpc.csproj" />
     <PackageReference Include="Google.Protobuf" Version="$(NuGetProtobufVersion)" />
-    <!-- Includes the analyzer in the package without consuming it. -->
-    <ProjectReference
-      Include="../IceRpc.Protobuf.Generators/IceRpc.Protobuf.Generators.csproj"
-      ReferenceOutputAssembly="false"
-    />
+    <ProjectReference Include="../IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.csproj"
```
pepone, line 11: One thing not clear about this is how this works if the same generator gets included multiple times, because we package it with multiple nuget packages.

tests/IceRpc.Protobuf.Tests/IceRpc.Protobuf.Tests.csproj
```diff
@@ -22,7 +22,7 @@
 
   <ItemGroup>
     <ProjectReference Include="../../src/IceRpc.Protobuf/IceRpc.Protobuf.csproj" />
-    <ProjectReference Include="../../src/IceRpc.Protobuf.Generators/IceRpc.Protobuf.Generators.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
+    <ProjectReference Include="../../src/IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```
bernardnormier, line 25: Claude explanation for why this necessary:

NuGet scenario: When IceRpc.Protobuf is packed, the ProjectReference to IceRpc.ServiceGenerator (with PrivateAssets="compile;runtime") becomes a NuGet dependency in the .nuspec. When a consumer installs the IceRpc.Protobuf NuGet package, NuGet also installs the IceRpc.ServiceGenerator NuGet package as a transitive dependency. Because IceRpc.ServiceGenerator has BuildOutputTargetFolder=analyzers/dotnet/cs, NuGet places its DLL in the analyzer folder and Roslyn loads it as a source generator.

ProjectReference scenario: When the test project references IceRpc.Protobuf via ProjectReference, MSBuild resolves transitive project dependencies — but OutputItemType="Analyzer" metadata on a ProjectReference does not propagate transitively. MSBuild only adds it to the Analyzer item group for the immediately referencing project (IceRpc.Protobuf itself), not for downstream consumers (the test project). This is a known MSBuild limitation — there's no equivalent of NuGet's transitive analyzer dependency flow for ProjectReference chains.

src/IceRpc.Protobuf/README.md
```diff
@@ -3,10 +3,8 @@
 The IceRPC framework allows you to make RPCs with the serialization format and [IDL] of your choice. It provides full
 support for both Slice and Protobuf.
 
-The IceRpc.Protobuf NuGet package is part of the IceRPC + Protobuf integration, and includes two .NET assemblies:
-
-- IceRpc.Protobuf.dll is used by the code generated by [`protoc-gen-icerpc-csharp`][protobuf-tools]
-- IceRpc.Protobuf.Generators.dll provides a [source generator] that you enable with the `[ProtobufService]` attribute.
+IceRpc.Protobuf provides integration code between the IceRPC core APIs and code generated by IceRPC's protoc plug-in for
```
bernardnormier, line 6: We need to rework these READMEs.

The "integration code" here is actually for both code generators: the protoc plug-in and the service generator.

src/IceRpc.Protobuf/IceRpc.Protobuf.csproj
bernardnormier: It's not included, just referenced. IceRpc.ServiceGenerator is a separate NuGet package.

PR #4303, opened 2026-03-13
Update IceRpc.Protobuf structure

bernardnormier:
This is the Protobuf counterpart of #4295 and #4300.


src/IceRpc.Protobuf/RpcMethods/RpcMethodAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+/// can deduct most information from the signature of the method decorated with this attribute, such as the method
+/// kind (unary, client streaming, server streaming, or bidi streaming).</remarks>
+[AttributeUsage(AttributeTargets.Method, Inherited = false)]
+[EditorBrowsable(EditorBrowsableState.Never)]
+public sealed class RpcMethodAttribute : Attribute
+{
+    /// <summary>Gets the method name. It corresponds to the name of the rpc method in the Protobuf file, with the
+    /// same spelling and the same case.</summary>
+    /// <value>The method name.</value>
+    public string Name { get; }
```
bernardnormier, line 21: These properties are unused, and provided only to make the attribute class look normal.

The source generator only examines the construction of the attribute.

src/IceRpc.ProtocGen/Program.cs
```diff
@@ -46,6 +46,7 @@
 #pragma warning disable CS0619 // Type or member is obsolete
 
 using IceRpc.Protobuf;
+using IceRpc.Protobuf.RpcMethods;
```
bernardnormier, line 49: After some back and forth, I decided to keep the `using IceRpc.Protobuf` in the code generated by the compiler. These generated files are RPC-specific and there is no downside in avoiding some qualification in this generated code (like we did previously).

I'll update Ice and Slice to do the same, in separate PRs.

src/IceRpc.Protobuf/RpcMethods/RpcMethodAttribute.cs
pepone: I would not add properties we don't need just to look "normal".

src/IceRpc.ProtocGen/Program.cs
pepone: The downside is a bigger chance to collide with user generated type isn't it, or do we fully qualify the others?

src/IceRpc.Protobuf/RpcMethods/RpcMethodAttribute.cs
bernardnormier: Not sure if we should change that. This is true for all attributes used by the Service generator.

src/IceRpc.ProtocGen/Program.cs
bernardnormier: We used to generate "using IceRpc.Slice", and "using IceRpc.Protobuf", so it's not worse than before.

And now, there are very few classes left in these namespaces, so even fewer changes of collisions.

Issue #4284, opened 2026-03-07
Fully separate Ice and Slice codec and IceRpc integrations


bernardnormier:
Currently, the Ice and Slice IDLs (in .ice and .slice files), are implemented by the same assemblies, namely:
 - ZeroC.Slice.Codec for encoding/decoding. Ice corresponds to SliceEncoding.Slice1 while Slice corresponds to SliceEncoding.Slice2
 - IceRpc.Slice for the IceRpc integration

Using the same assemblies provide some opportunities for code reuse, such as:
 - the encoding/decoding of integers and strings
 - the logic for encoding/decoding sequences and dictionaries
 - the helper code for proxies and dispatch methods (with some differences)

However, this sharing also increases complexity. It's all the more apparent now that we use different IDLs and compilers for Ice and Slice.

## Proposal

- Move Ice and Slice support to separate assemblies
- Drop the terms Slice1 and Slice2
- Use "Ice" when talking about the Ice IDL saved in .ice files, and "Slice" when talking about the new Slice IDL introduced with IceRPC.

The resulting assemblies and namespaces are:

| Assembly | Description |
|-----------|-------------|
|ZeroC.Slice.Codec | provides SliceEncoder and SliceDecoder |
| ZeroC.Ice.Codec | provides IceEncoder, IceDecoder, IceException, IceClass, … |
| IceRpc.Slice | provides SliceService and ISliceProxy |
| IceRpc.Slice.Generators | a source generator triggered by `[SliceService]` |
| IceRpc.Ice | provides IceService, IIceProxy in addition to IIceObject, Identity |
| IceRpc.Ice.Generators | a source generator triggered by `[IceService]` |

This proposal has a few downsides:
- Changes the IceRPC Ice APIs (but we're already doing some changes for the 0.6.0 release)
- Some code duplication
- `[SliceService]` and `[IceService]` are incompatible, which means the same class cannot implement both Ice and Slice interfaces. However, we can't identify a clear use-case for such a service class.

PR #4305, opened 2026-03-14
Rename/duplicate projects

bernardnormier:
This is the last preliminary renaming for #4284:
 - replace CodeGen by Generator in various projects
 - rename ProtocGen to Protobuf.Generator

and finally duplicate/split ZeroC.Slice.Codec.Tests into two projects:
 - ZeroC.Slice.Codec.Tests is Slice2-only (also includes tests for ZeroC.Tests.Common classes)
 - IceRpc.Ice.Codec.Tests is Slice1-only

### DIFF
--- a/src/IceRpc.Protobuf.BuildTelemetry/protoc-gen-icerpc-build-telemetry.bat
+++ b/src/IceRpc.Protobuf.BuildTelemetry/protoc-gen-icerpc-build-telemetry.bat
@@ -1,3 +1,3 @@
 @echo OFF
 
-call dotnet %~dp0\IceRpc.Protobuf.BuildTelemetry.dll -- %*
+call dotnet "%~dp0\IceRpc.Protobuf.BuildTelemetry.dll" -- %*


IceRpc.slnx
```diff
@@ -40,23 +40,24 @@
...
     <Project Path="tests/IceRpc.Slice.Tests/IceRpc.Slice.Tests.csproj" />
     <Project Path="tests/IceRpc.Telemetry.Tests/IceRpc.Telemetry.Tests.csproj" />
     <Project Path="tests/IceRpc.Tests/IceRpc.Tests.csproj" />
     <Project Path="tests/IceRpc.Tests.Common/IceRpc.Tests.Common.csproj" />
     <Project Path="tests/IntegrationTests/IntegrationTests.csproj" />
+    <Project Path="tests/IceRpc.Ice.Codec.Tests/IceRpc.Ice.Codec.Tests.csproj" />
     <Project Path="tests/ZeroC.Slice.Codec.Tests/ZeroC.Slice.Codec.Tests.csproj" />
-    <Project Path="tests/ZeroC.Ice.CodeGen.Tests/ZeroC.Ice.CodeGen.Tests.csproj" />
-    <Project Path="tests/ZeroC.Slice.CodeGen.Tests/ZeroC.Slice.CodeGen.Tests.csproj" />
+    <Project Path="tests/IceRpc.Ice.Generator.None.Tests/IceRpc.Ice.Generator.None.Tests.csproj" />
```
bernardnormier, line 52: That's a potentially controversial renaming: I named the test project for the IceRpc.Ice code-gen tests that don't depend on the IceRpc.Ice assembly "IceRpc.Ice.Generator.None.Tests". None is a reference to the IceRpc.ExcludeSliceCompile.targets import, previous known as `<Rpc>none</Rpc>` in the .csproj to suppress the .IceRpc.cs files.

tests/IceRpc.Ice.Generator.None.Tests.ReferencedAssemblies/A/A.ice
```diff
@@ -2,7 +2,7 @@
 
 #pragma once
 
-module Ice::CodeGen::Tests::ReferencedAssemblies
+module Ice::Generator::None::Tests::ReferencedAssemblies
```
pepone, line 5: Why None?

IceRpc.slnx
pepone: My understading is that `IceRpc.Ice.Generator` tests the slice2cs generated code, and we were going to remove this `none` option from slice2cs, so it doesn't make sense to keep it here.

bernardnormier: In other words, you want to a single test assembly for all the tests with .ice-generated code, where this assembly depends on the IceRpc.Ice assembly.

With my PR, the IceRpc.Ice.Generator.None.Tests assembly doesn't depend on IceRpc.Ice assembly. For example, it includes plain "class" testing, without a dependency on the generated code for interfaces (in .IceRpc.cs files).

Just confirming: you'd rather merge these test projects?

pepone: Yes, I think the merge is fine here given that we don't have this separation in slice2cs.

PR #4309, opened 2026-03-16
Upgrade GitHub Actions to latest versions

externl:


.github/workflows/build-packages.yml
```diff
@@ -118,7 +118,7 @@ jobs:
 
       - name: Sign .NET Assemblies and slicec-cs.exe compiler
         if: inputs.authenticode_sign
-        uses: azure/trusted-signing-action@v0
+        uses: azure/artifact-signing-action@v1
```
externl: @pepone I assume there isn't a reason we need to stay on `v0`?

PR #4310, opened 2026-03-16
Move Slice1 support to IceRpc.Ice.Codec with IceEncoder/IceDecoder etc.

bernardnormier:
This PR moves the Ice encoding to a new IceRpc.Ice.Codec namespace inside the main IceRpc assembly.

ZeroC.Slice.Codec is largely unchanged; I'll remove Slice1 support in a follow-up PR.

Note: this PR depends on an updated slice2cs  compiler (see PR in ice repo).

### DIFF
--- a/tests/IceRpc.Ice.Generator.None.Tests/EnumTests.cs
+++ b/tests/IceRpc.Ice.Generator.None.Tests/EnumTests.cs
@@ -2,7 +2,6 @@
 
 using IceRpc.Ice.Codec;
 using NUnit.Framework;
-
 using ZeroC.Tests.Common;
 
 namespace IceRpc.Ice.Generator.None.Tests;


src/IceRpc.Ice/Operations/Internal/IncomingFrameExtensions.cs
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) ZeroC, Inc.
 
 using IceRpc.Features;
+using IceRpc.Ice.Codec;
```
bernardnormier, line 4: The namespace is what matters in these files, since IceRpc has a transitive dependency on ZeroC.Slice.Codec.

src/IceRpc/Ice/Codec/ClassFormat.cs
```diff
@@ -0,0 +1,18 @@
...
+/// <summary>Describes the possible formats when encoding classes.</summary>
+public enum ClassFormat
+{
+    /// <summary>The Compact format assumes the sender and receiver have the same Ice definitions for classes. If an
+    /// application receives a derived class it does not know, it is not capable of decoding it into a known base
+    /// class because there is not enough information in the encoded payload. The Compact format is the default.
+    /// </summary>
+    Compact,
+
+    /// <summary>The Sliced format allows the receiver to slice off unknown slices. If an application receives a
```
bernardnormier, line 14: A benefit of the "Ice codec" and "IceEncoder" terminology is "slice" is no longer ambiguous.

src/IceRpc/Ice/Codec/IceEncoder.cs
```diff
@@ -0,0 +1,322 @@
...
+        _classContext = new ClassContext(classFormat);
+    }
+
+    // Encode methods for basic types
+
+    /// <summary>Encodes a bool into an Ice bool.</summary>
+    /// <param name="v">The boolean to encode.</param>
+    public void EncodeBool(bool v) => EncodeUInt8(v ? (byte)1 : (byte)0);
+
+    /// <summary>Encodes a float into an Ice float32.</summary>
```
bernardnormier, line 60: For now, I've kept the Slice terminology. It would be desirable to switch back to the Ice terminology (int, short, float, double, etc. in a follow-up PR).

src/IceRpc/Ice/Codec/IceEncoder.cs
```diff
@@ -0,0 +1,322 @@
...
+    // Other methods
+
+    /// <summary>Encodes a non-null tagged value. The number of bytes needed to encode the value is known before
+    /// encoding the value. This method always use the VSize tag format.</summary>
+    /// <typeparam name="T">The type of the value being encoded.</typeparam>
+    /// <param name="tag">The tag.</param>
+    /// <param name="size">The number of bytes needed to encode the value.</param>
+    /// <param name="v">The value to encode.</param>
+    /// <param name="encodeAction">The delegate that encodes the value after the tag header.</param>
+    public void EncodeTagged<T>(int tag, int size, T v, EncodeAction<T> encodeAction) where T : notnull
```
bernardnormier, line 180: As an exception, I would however keep the tagged terminology over the Ice optional terminology.

src/IceRpc/Ice/Codec/IceEncoderExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+            default:
+                encoder.EncodeSequence(
+                    v,
+                    (ref IceEncoder encoder, T element) => encoder.EncodeFixedSizeNumeric(element));
+                break;
+        }
+    }
+
+    /// <summary>Encodes a sequence.</summary>
+    /// <typeparam name="T">The type of the sequence elements. It is non-nullable except for nullable class.</typeparam>
```
bernardnormier: `+` nullable proxies.

src/IceRpc/Internal/IceProtocol.cs
```diff
@@ -11,7 +11,7 @@ internal sealed class IceProtocol : Protocol
     internal static IceProtocol Instance { get; } = new();
 
     /// <summary>Checks if this absolute path is well-formed.</summary>
-    /// <remarks>This check is more lenient than the check performed when encoding a service address with Slice1 because
+    /// <remarks>This check is more lenient than the check performed when encoding a service address with Ice because
     /// we want the default path (`/`) to be a valid path for all protocols. Sending a request to a null/empty identity
     /// is in itself ok and will most likely result in a dispatch exception with a
```
bernardnormier: We should clarify this comment.

src/IceRpc/AssemblyInfo.cs
```diff
@@ -11,6 +11,7 @@
 // Make internals visible to the tests assembly, to allow writing unit tests for the internal classes
 [assembly: InternalsVisibleTo("IceRpc.Tests")]
 [assembly: InternalsVisibleTo("IceRpc.Ice.Tests")]
+[assembly: InternalsVisibleTo("IceRpc.Ice.Generator.None.Tests")]
```
bernardnormier, line 14: It's because the IceRpc.Ice.Codec is now in this assembly.

tests/IceRpc.Ice.Generator.None.Tests/SequenceDecodingTests.ice
```diff
@@ -1,11 +0,0 @@
-// Copyright (c) ZeroC, Inc.
-
-module IceRpc::Ice::Generator::None::Tests
```
bernardnormier, line 3: Moved to the Slice tests.

src/IceRpc/Ice/Codec/IceEncoder.cs
```diff
@@ -0,0 +1,322 @@
...
+    public void EncodeInt64(long v) => EncodeFixedSizeNumeric(v);
+
+    /// <summary>Encodes a size on variable number of bytes.</summary>
+    /// <param name="value">The size to encode.</param>
+    public void EncodeSize(int value)
+    {
+        if (value < 0)
+        {
+            throw new ArgumentException(
+                $"The {nameof(value)} argument must be greater than 0.",
```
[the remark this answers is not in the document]
bernardnormier: Fixed.

src/IceRpc/Ice/Codec/IceDecoder.Class.cs
```diff
@@ -0,0 +1,581 @@
...
+            // We need to decrement index since position 0 in the indirection table corresponds to index 1.
+            index--;
+            if (index < _classContext.Current.IndirectionTable?.Length)
+            {
+                return _classContext.Current.IndirectionTable[index];
+            }
+            else
+            {
+                throw new InvalidDataException("The index is too big for the indirection table.");
+            }
```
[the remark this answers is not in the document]
bernardnormier: Claude's answer:

> Copilot is wrong. The code is valid C#.

The ?. in IndirectionTable?.Length produces an int?, not a bool?. When you write index < nullableInt, C# uses lifted comparison operators — the < operator is automatically "lifted" to accept nullable operands and returns a plain bool, not bool?. If the right-hand side is null, the comparison simply evaluates to false.

This is defined in the C# language spec (§12.4.8): for relational operators like <, the lifted form returns false when either operand is null. So index < null → false, which is exactly the desired behavior here — if there's no indirection table, the if body is skipped.


I'll add a comment.

src/IceRpc/Ice/Codec/SliceInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+
+/// <summary>Encapsulates the details of a class slice (as in slice of cake) that an <see cref="IActivator" /> could
+/// not decode.</summary>
+public sealed class SliceInfo
+{
+    /// <summary>Gets the Ice type ID or compact ID for this slice.</summary>
+    public string TypeId { get; }
+
+    /// <summary>Gets the encoded bytes for this slice, including the leading size integer.</summary>
+    public ReadOnlyMemory<byte> Bytes { get; }
```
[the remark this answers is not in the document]
bernardnormier: The comment is correct: the bytes include the slice size encoded on 4 bytes (the first 4 bytes). In particular, this Bytes buffer size is "slice size" and includes these 4 bytes.

See https://docs.zeroc.com/ice/3.8/cpp/simple-example-of-class-encoding#SimpleExampleofClassEncoding-ClassEncodingintheSlicedFormat for an example.

src/IceRpc/Ice/Codec/IceEncoderExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+        EncodeAction<TKey> keyEncodeAction,
+        EncodeAction<TValue> valueEncodeAction)
+        where TKey : notnull
+    {
+        encoder.EncodeSize(v.Count());
+        foreach ((TKey key, TValue value) in v)
+        {
+            keyEncodeAction(ref encoder, key);
+            valueEncodeAction(ref encoder, value);
+        }
```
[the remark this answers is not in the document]
bernardnormier: After discussing this issue with Claude, the best solution here is to make sure `v` is countable, with:

```cs
if (!v.TryGetNonEnumeratedCount(out int count))
{
    throw new ArgumentException("The value must be a countable collection.", nameof(v));
}
encoder.EncodeSize(count);
```

For a follow-up PR. This also affects SliceDecoder.

src/IceRpc/Ice/Codec/IceEncoderExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// <param name="encoder">The Ice encoder.</param>
+    /// <param name="v">The sequence to encode.</param>
+    /// <param name="encodeAction">The encode action for an element.</param>
+    public static void EncodeSequence<T>(this ref IceEncoder encoder, IEnumerable<T> v, EncodeAction<T> encodeAction)
+    {
+        encoder.EncodeSize(v.Count()); // potentially slow Linq Count()
+        foreach (T item in v)
+        {
+            encodeAction(ref encoder, item);
+        }
```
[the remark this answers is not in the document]
bernardnormier: See above.

src/IceRpc/Ice/Codec/IceEncoderExtensions.cs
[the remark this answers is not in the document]
pepone: Another option is to reserve 4/8 bytes and patch the size for the case that it is not available as a non enumerated count, but maybe is not worth the trouble.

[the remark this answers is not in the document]
bernardnormier: I think non-countable enumerable should be uncommon. If it turns out it's not, we can implement your solution later.

src/IceRpc/Ice/Codec/IceEncoder.cs
bernardnormier: See #4313

src/IceRpc.Ice/Operations/IncomingRequestExtensions.cs
externl:
```suggestion
/// <param name="inExceptionSpecification">A function that returns <see langword="true" /> when the provided Ice
/// exception conforms to the exception specification; otherwise, <see langword="false" />.</param>
```

src/IceRpc/Ice/Codec/IceEncoderExtensions.cs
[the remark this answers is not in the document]
bernardnormier: See #4312

src/IceRpc.Ice/Operations/IncomingRequestExtensions.cs
externl:
```suggestion
/// <param name="inExceptionSpecification">A function that returns <see langword="true" /> when the provided Ice
/// exception conforms to the exception specification; otherwise, <see langword="false" />.</param>
```

externl:
```suggestion
/// <param name="inExceptionSpecification">A function that returns <see langword="true" /> when the provided Ice
/// exception conforms to the exception specification; otherwise, <see langword="false" />.</param>
```

externl, line 1: Best to double check this file for "Slice exception"

src/IceRpc.Ice/Operations/IncomingResponseExtensions.cs
```diff
@@ -117,23 +117,22 @@ private static async ValueTask<SliceException> DecodeSliceExceptionAsync(
...
 
-        SliceException exception = DecodeBuffer(readResult.Buffer);
+        IceException exception = DecodeBuffer(readResult.Buffer);
         response.Payload.AdvanceTo(readResult.Buffer.End);
         return exception;
 
-        SliceException DecodeBuffer(ReadOnlySequence<byte> buffer)
+        IceException DecodeBuffer(ReadOnlySequence<byte> buffer)
         {
             // A Slice exception never sets Message, even when received over icerpc.
```
externl:
```suggestion
            // An Ice exception never sets Message, even when received over icerpc.
```

src/IceRpc.Ice/Operations/IncomingResponseExtensions.cs
externl, line 1: Ditto.

src/IceRpc.Ice/Operations/IceProxyExtensions.cs
```diff
@@ -48,7 +48,7 @@ public static class IceProxyExtensions
     /// <typeparam name="TProxy">The type of the proxy struct.</typeparam>
     /// <typeparam name="T">The response type.</typeparam>
     /// <param name="proxy">A proxy to the remote service.</param>
-    /// <param name="operation">The name of the operation, as specified in Slice.</param>
+    /// <param name="operation">The name of the operation, as specified in Ice Slice.</param>
```
InsertCreativityHere, line 51: Do we really want to make use of the term "Ice Slice".
It's meaning is clear to me, but may be very confusing to users given our new terminology.

tests/IceRpc.Ice.Generator.Tests/ClassTests.cs
```diff
@@ -1,8 +1,9 @@
 // Copyright (c) ZeroC, Inc.
 
+using IceRpc.Ice.Codec;
 using NUnit.Framework;
 using System.IO.Pipelines;
-using ZeroC.Slice.Codec;
+
```
InsertCreativityHere: Extra space?

```suggestion

```

tests/IceRpc.Ice.Generator.Tests/InterfaceTests.cs
```diff
@@ -22,7 +22,7 @@ public sealed class InterfaceTests
     [TestCase(typeof(IMyOtherWidgetService), "::IceRpc::Ice::Generator::Tests::myOtherWidget")]
     public void Get_slice_type_id(Type type, string? expected)
```
InsertCreativityHere: Should this be:
```suggestion
    public void Get_ice_type_id(Type type, string? expected)
```
now?

tests/IceRpc.Ice.Generator.Tests/ProxyTests.cs
```diff
@@ -1,9 +1,10 @@
 // Copyright (c) ZeroC, Inc.
 
+using IceRpc.Ice.Codec;
 using IceRpc.Ice.Operations;
 using IceRpc.Tests.Common;
 using NUnit.Framework;
-using ZeroC.Slice.Codec;
+
```
InsertCreativityHere: Is this extra space intentional?

```suggestion
```

tests/IceRpc.Ice.Generator.Tests/StructTests.cs
```diff
@@ -1,8 +1,9 @@
 // Copyright (c) ZeroC, Inc.
 
+using IceRpc.Ice.Codec;
 using IceRpc.Ice.Operations;
 using NUnit.Framework;
-using ZeroC.Slice.Codec;
+
```
InsertCreativityHere: Same:
```suggestion
```

tests/IceRpc.Ice.Generator.None.Tests/TypeIdAttributeTests.cs
```diff
@@ -14,7 +14,7 @@ public sealed class TypeIdAttributeTests
     [TestCase(typeof(MyOtherClass), "::IceRpc::Ice::Generator::None::Tests::TypeIdAttributeTestNamespace::myOtherClass")]
     public void Get_slice_type_id(Type type, string? expected)
```
InsertCreativityHere: Same:
```suggestion
    public void Get_ice_type_id(Type type, string? expected)
```

tests/IceRpc.Ice.Generator.None.Tests/ExceptionTests.cs
```diff
@@ -1,7 +1,8 @@
 // Copyright (c) ZeroC, Inc.
 
+using IceRpc.Ice.Codec;
 using NUnit.Framework;
-using ZeroC.Slice.Codec;
+
```
InsertCreativityHere: Some files have this space, some do not. We should be consistent.
And I would lean towards to space, like how we had originally.
```suggestion
```

src/IceRpc/Internal/IceDefinitions.cs
```diff
@@ -17,7 +17,7 @@ internal static class IceDefinitions
     internal static readonly byte[] Magic = "IceP"u8.ToArray(); // 'I', 'c', 'e', 'P'
 
     // 4-bytes after magic that provide the protocol version (always 1.0 for an ice frame) and the encoding of the
-    // frame header (always set to 1.0 with an ice frame, even though we use Slice1).
+    // frame header (always set to 1.0 with an ice frame, even though the IceEncoder/Decoder are really 1.1).
```
InsertCreativityHere: I just find the current wording to be kind of odd.
```suggestion
    // frame header (always set to 1.0 with an ice frame, even though the IceEncoder/Decoder really use 1.1).
```

PR #4314, opened 2026-03-17
Initial C# code generator

pepone:
- ZeroC.Slice.Generator generates C# code for Slice types
- IceRpc.Slice.Generator generates C# code for proxy/dispatch
- ZeroC.Slice.Symbols C# model build from compiler definitions to use in C# code genertors.

The generator can be run with:

```
 ../slicec/target/debug/slicec tests/IceRpc.Slice.CodeGen.Tests/*.slice -R slice | dotnet run --project src/ZeroC.Slice.Generator/ZeroC.Slice.Generator.csproj
```

or

```
 ../slicec/target/debug/slicec tests/IceRpc.Slice.CodeGen.Tests/*.slice -R slice | dotnet run --project src/IceRpc.Slice.Generator/IceRpc.Slice.Generator.csproj
```

- DocComment generation is not yet done, in part because slicec is not yet encoding comments.
- ZeroC.Slice.Generator/GeneratorDriver.cs contains the boilerplate code to decode and encode the generator request and convert from the compiler symbols to the symbols used by the generators.

There are separate generators for enums, enum struct, struct. Removed the base Generator class that did too many things from previous PR, and split this into extension for IType, Field, ...

### DIFF
--- a/tests/IceRpc.Ice.Generator.Tests/ProxyTests.cs
+++ b/tests/IceRpc.Ice.Generator.Tests/ProxyTests.cs
@@ -11,6 +11,40 @@ namespace IceRpc.Ice.Generator.Tests;
 [Parallelizable(scope: ParallelScope.All)]
 public partial class ProxyTests
 {
+    [Test]
+    public void Create_proxy_with_null_protocol_fails()
+    {
+        // Arrange
+        var serviceAddress = new ServiceAddress(protocol: null) { Path = "/foo" };
+
+        // Act & Assert
+        Assert.That(() => new PingableProxy(InvalidInvoker.Instance, serviceAddress),
+            Throws.TypeOf<ArgumentException>());
+    }
+
+    [Test]
+    public void Initialize_proxy_with_null_protocol_fails()
+    {
+        // Arrange
+        var serviceAddress = new ServiceAddress(protocol: null) { Path = "/foo" };
+
+        // Act & Assert
+        Assert.That(
+            () => new PingableProxy{ Invoker = InvalidInvoker.Instance, ServiceAddress = serviceAddress },
+            Throws.TypeOf<ArgumentException>());
+    }
+
+    [Test]
+    public void Proxy_has_default_service_path_with_ice_protocol()
+    {
+        // Arrange
+        var proxy = new PingableProxy{ Invoker = InvalidInvoker.Instance };
+
+        // Assert
+        Assert.That(proxy.ServiceAddress.Path, Is.EqualTo(PingableProxy.DefaultServicePath));
+        Assert.That(proxy.ServiceAddress.Protocol, Is.EqualTo(Protocol.Ice));
+    }
+
     /// <summary>Verifies that calling DecodeProxy correctly decodes a proxy.</summary>
     /// <param name="value">The service address of the proxy to encode.</param>
     /// <param name="expected">The expected URI string of the service address.</param>
--- a/tests/IntegrationTests/CustomTransportTests.cs
+++ b/tests/IntegrationTests/CustomTransportTests.cs
@@ -93,7 +93,7 @@ public partial class CustomTransportTests
             },
             multiplexedClientTransport: new CustomClientTransport());
 
-        var proxy = new PingableProxy(connection);
+        var proxy = new PingableProxy(connection, new Uri($"icerpc:{PingableProxy.DefaultServicePath}"));
         await proxy.PingAsync();
     }
 
@@ -127,7 +127,7 @@ public partial class CustomTransportTests
                 },
                 multiplexedClientTransport: new CustomClientTransport());
 
-            var proxy = new PingableProxy(connection1);
+            var proxy = new PingableProxy(connection1, new Uri($"icerpc:{PingableProxy.DefaultServicePath}"));
             await proxy.PingAsync();
         }
     }


.vscode/cspell.json
```diff
@@ -3,6 +3,7 @@
   "language": "en",
   "words": [
     "aarch",
+    "aeiou",
```
bernardnormier: For some words like aeiou, it's better to disable cspell with 

```cs
// cspell:disable-line
```

src/ZeroC.Slice.Symbols/SymbolConverter.cs
```diff
@@ -0,0 +1,458 @@
...
+
+        foreach (Compiler.SliceFile file in allFiles)
+        {
+            string moduleScope = file.ModuleDeclaration.Identifier;
+            foreach (Compiler.Symbol symbol in file.Contents)
+            {
+                if (GetNamedIdentifier(symbol) is string id)
+                {
+                    _named.TryAdd($"{moduleScope}::{id}", (file, symbol));
+                }
```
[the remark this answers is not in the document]
pepone: module is always not empty, we should remove the other check.

src/ZeroC.Slice.Symbols/SymbolConverter.cs
```diff
@@ -0,0 +1,458 @@
...
+                Module = module,
+                IsCompact = raw.IsCompact,
+                IsUnchecked = raw.IsUnchecked,
+                Enumerators = raw.Enumerators.Select(e => new EnumWithFields.Enumerator
+                {
+                    Identifier = e.EntityInfo.Identifier,
+                    Attributes = ConvertAttributes(e.EntityInfo.Attributes),
+                    Module = module,
+                    Discriminant = (int)e.Value.AbsoluteValue,
+                    Fields = e.Fields.Select(f => ConvertField(f, file, module)).ToImmutableList(),
```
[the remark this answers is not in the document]
pepone: Cast to int is fine here

> The range of these discriminants is 0 to 2,147,483,647 (int32 max).

See https://docs.icerpc.dev/slice2/language-guide/enum-types

src/ZeroC.Slice.Symbols/SliceFile.cs
```diff
@@ -0,0 +1,21 @@
...
+    public required string Path { get; init; }
+
+    /// <summary>Gets the module defined in the Slice file.</summary>
+    public required Module Module { get; init; }
+
+    /// <summary>Gets the list of file-level attributes defined in the Slice file.</summary>
+    public required ImmutableList<Attribute> Attributes { get; init; }
+
+    /// <summary>Gets the list of symbols defined in the Slice file.</summary>
+    public required ImmutableList<ISymbol> Contents { get; init; }
```
bernardnormier, line 20: As I understand it, an ISymbol is either a type or an interface. So if we split this Contents in Types and Interfaces, we can eliminate ISymbol?

src/ZeroC.Slice.Symbols/ISymbol.cs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) ZeroC, Inc.
+
+namespace ZeroC.Slice.Symbols;
+
+/// <summary>Base for all symbols that can appear in a <see cref="SliceFile"/>.</summary>
+#pragma warning disable CA1040 // Avoid empty interfaces
+public interface ISymbol;
```
bernardnormier, line 7: I would prefer to avoid ISymbol as it's also Microsoft.CodeAnalysis.ISymbol.

src/ZeroC.Slice.Symbols/TypeRef.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) ZeroC, Inc.
+
+using System.Collections.Immutable;
+
+namespace ZeroC.Slice.Symbols;
+
+/// <summary>Represents a reference to a Slice type.</summary>
+public class TypeRef
```
bernardnormier: Can we use a readonly record struct instead, just like Attribute?

bernardnormier (commented):
Looks good. Will re-review later.

src/IceRpc.Slice.Generator/ProxyGenerator.cs
```diff
@@ -0,0 +1,628 @@
...
+            "encodeOptions",
+            "null",
+            "The encode options, used to customize the encoding of request payloads.");
+        uriCtor.AddSetsRequiredMembersAttribute();
+        uriCtor.AddBaseParameters(["invoker", "new IceRpc.ServiceAddress(serviceAddressUri)", "encodeOptions"]);
+
+        var defaultCtor = new FunctionBuilder("public", "", proxyName, FunctionType.BlockBody);
+        defaultCtor.AddComment(
+            "summary",
+            @"Constructs a proxy with an icerpc service address with path <see cref=""DefaultServicePath"" />.");
```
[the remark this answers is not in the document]
pepone: The empty body is correct, this is as per slicec-cs impl:
```
    /// <summary>Constructs a proxy with an icerpc service address with path <see cref="DefaultServicePath" />.</summary>
    public MoreDerivedProxy()
    {
    }
```

src/IceRpc.Slice.Generator/IceRpc.Slice.Generator.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <!--Because an application's API isn't typically referenced from outside the assembly, types can be made internal -->
+    <NoWarn>CA1515</NoWarn>
```
bernardnormier, line 5: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1515

Meaning we have public types that can/should be internal. Why disable this warning?

src/IceRpc.Slice.Generator/OperationHelpers.cs
```diff
@@ -0,0 +1,111 @@
...
+
+using System.Collections.Immutable;
+using ZeroC.CodeBuilder;
+using ZeroC.Slice.Symbols;
+
+using static ZeroC.Slice.Generator.OperationExtensions;
+
+namespace ZeroC.Slice.Generator;
+
+/// <summary>Helper methods for generating operation encode/decode code blocks.</summary>
```
bernardnormier: A more typical C#-way to write helpers is to add extension method, such as OperationExtensions.

src/IceRpc.Slice.Generator/ProxyGenerator.cs
```diff
@@ -0,0 +1,628 @@
...
+            {
+                request.AddBlock(BuildEncodeStreamMethod(op, streamParam, currentNamespace));
+            }
+        }
+
+        return request.Build();
+    }
+
+    /// <summary>Builds the EncodeStreamOf{Op} method for a streamed field (parameter or return value).</summary>
+    internal static CodeBlock BuildEncodeStreamMethod(Operation op, Field streamParam, string currentNamespace)
```
bernardnormier: I am surprised you don't get a build warning for defining an internal method after a private one.

src/IceRpc.Slice.Generator/ProxyGenerator.cs
```diff
@@ -0,0 +1,628 @@
...
+    {
+        string opName = op.Name;
+        string streamType = OperationExtensions.GetStreamTypeString(streamParam, currentNamespace);
+
+        var fn = new FunctionBuilder(
```
bernardnormier: I would prefer "builder" or "functionBuilder" over "fn"

src/IceRpc.Slice.Generator/ProxyGenerator.cs
```diff
@@ -0,0 +1,628 @@
...
+
+                fn.SetBody(body);
+                response.AddBlock(fn.Build());
+            }
+        }
+
+        return response.Build();
+    }
+
+    private static CodeBlock BuildProxyOperationImpl(Operation op, string currentNamespace)
```
bernardnormier: We use "Core" instead of "Impl".

src/IceRpc.Slice.Generator/IceRpc.Slice.Generator.csproj
pepone: I got this when including the code builder

```
eroC.Slice.Generator net10.0 failed with 11 error(s) (0,2s)
    CSC : error CA1515: Because an application's API isn't typically referenced from outside the assembly, types can be made internal (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1515)
    /Users/jose/Documents/IceRPC/icerpc-csharp/src/ZeroC.CodeBuilder/CodeBlock.cs(8,21): error CA1515: Because an application's API isn't typically referenced from outside the assembly, types can be made internal (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1515)
    /Users/jose/Documents/IceRPC/icerpc-csharp/src/ZeroC.CodeBuilder/CommentTag.cs(8,21): error CA1515: Because an application's API isn't typically referenced from outside the assembly, types can be made internal (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1515)
```

src/IceRpc.Slice.Generator/ProxyGenerator.cs
pepone: renamed

src/IceRpc.Slice.Generator/ProxyGenerator.cs
pepone: moved to op extensions

PR #4318, opened 2026-03-18
Replace calls to LINQ Count()

### LINKED ISSUE #4312: Update Sequence and Dictionary encoding code
After discussing this issue with Claude, the best solution here is to make sure `v` is countable, with:

```cs
if (!v.TryGetNonEnumeratedCount(out int count))
{
    throw new ArgumentException("The value must be a countable collection.", nameof(v));
}
encoder.EncodeSize(count);
```

For a follow-up PR. This also affects SliceEncoder.

_Originally posted by @bernardnormier in https://github.com/icerpc/icerpc-csharp/pull/4310#discussion_r2943621201_

bernardnormier:
This PR replaces calls to `.Count()` on `IEnumerable<T>` in the source code.

The concern is `Count()` followed by iteration is not necessarily correct - the two can be out of sync. Naturally that's not a concern for "nornal" collections, that are all countable.

The initial fix was to throw an ArgumentException for non-countable collections. But we actually use a non-countable collection in the Telemetry interceptor (Baggage), which shows such collections aren't that unusual. So I changed the fallback to instead materialize the collection with `ToArray()`.

Fixes #4312

### DIFF
--- a/src/IceRpc/Ice/Codec/Internal/Activator.cs
+++ b/src/IceRpc/Ice/Codec/Internal/Activator.cs
@@ -20,18 +20,23 @@ internal class Activator : IActivator
     /// <summary>Merge activators into a single activator; duplicate entries are ignored.</summary>
     internal static Activator Merge(IEnumerable<Activator> activators)
     {
-        var dict = new Dictionary<string, Type>();
-        int activatorCount = 0;
-        foreach (Activator activator in activators)
+        if (activators.Count() == 1)
         {
-            activatorCount++;
-            foreach ((string typeId, Type factory) in activator._dict)
-            {
-                dict[typeId] = factory;
-            }
+            return activators.First();
         }
+        else
+        {
+            var dict = new Dictionary<string, Type>();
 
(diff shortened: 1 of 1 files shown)


src/IceRpc/Ice/Codec/Internal/Activator.cs
```diff
@@ -20,23 +20,18 @@ internal class Activator : IActivator
     /// <summary>Merge activators into a single activator; duplicate entries are ignored.</summary>
     internal static Activator Merge(IEnumerable<Activator> activators)
     {
-        if (activators.Count() == 1)
+        var dict = new Dictionary<string, Type>();
```
pepone: If the point of not calling Count is to avoid enumerating more than once, why is this any better, We still iterate and later call first, which might require a second pass if the enumerable doesn't provide `[]`.

bernardnormier: I just reverted the changes to Activator. You're right, it's confusing and not really better.

PR #4321, opened 2026-03-19
Remove Slice build telemetry

bernardnormier:
This PR removes Slice build telemetry. We should reintroduce it later as a slicec plug-in.

### DIFF
--- a/slice.toml
+++ b/slice.toml
@@ -5,13 +5,20 @@
 
 [source.ice]
 repo    = "https://github.com/zeroc-ice/ice"
-rev     = "4e7c327a184ca5330854d16b88976b46ddb24f30"
+rev     = "32d11c5b7d6cdc58dcf7b7cfd85ea655a5ad5913"
 source  = "slice"
 dest    = "slice"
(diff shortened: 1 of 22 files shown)


src/IceRpc.Protobuf.BuildTelemetry/slice/BuildTelemetry.slice
InsertCreativityHere, line 1: Should we maybe move this file into `icerpc-slice` eventually?

InsertCreativityHere: Also should we remove the `SliceTelemetryData` definitions from this file?

bernardnormier: Yes, we should move the BuildTelemetry Slice file somewhere else. And we should update the Slice BuildTelemetry data in a follow-up PR. 

Previously, we shared the Slice file between Protobuf and Slice build telemetry, and this PR just moved this shared file to Protobuf.

PR #4322, opened 2026-03-19
Synchronize Slice compiler Definitions with 'slicec'

InsertCreativityHere:
https://github.com/icerpc/slicec/pull/737 moved the Slice compiler definitions from `icerpc-slice` to `slicec`.
So, now we need to pull them in from `slicec`.

### DIFF
--- a/slice.toml
+++ b/slice.toml
@@ -5,14 +5,14 @@
 
 [source.ice]
 repo    = "https://github.com/zeroc-ice/ice"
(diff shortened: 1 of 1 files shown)


InsertCreativityHere:
> We need to wait for the PR that removes the files from icerpc-slice, and synchronise that too.

Yep, I just wanted to make CI worked how I expected it to while waiting!
I updated all the commit hashes to their latests and synchronized everything.

PR #4329, opened 2026-03-20
Doc comment support

pepone:


src/ZeroC.Slice.Generator/GeneratorDriver.cs
```diff
@@ -85,35 +85,35 @@ internal static async Task RunAsync(
                     }
                 }
 
-                if (codeBlocks.Count > 0)
-                {
-                    string fileName = Path.GetFileName(file.Path);
+                string fileName = Path.GetFileName(file.Path);
```
pepone, line 88: This is unrelated to doc comments, we want to still emit a file with the preamble even if the generator doesn't generate any code blocks.

src/ZeroC.Slice.Symbols/Compiler/SyntaxElements.cs
```diff
@@ -126,24 +126,27 @@ public EntityInfo(
     [global::System.Diagnostics.CodeAnalysis.SetsRequiredMembers]
     public EntityInfo(ref SliceDecoder decoder)
     {
+        var bitSequenceReader = decoder.GetBitSequenceReader(1);
```
pepone, line 129: Regenerated the Compiler definitions with latest code.

src/IceRpc.Slice.Generator/DispatchGenerator.cs
```diff
@@ -303,7 +308,9 @@ private static CodeBlock BuildServiceOperationDeclaration(Operation op, string c
         {
             attrParts.Add("CompressReturn = true");
         }
-        operationBuilder.AddAttribute($"SliceOperation({string.Join(", ", attrParts)})");
+        operationBuilder
+            .AddDocCommentSeeAlso(op.Comment, currentNamespace)
```
bernardnormier, line 312: This generates a see also when op.Comment has a see also, right?

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -0,0 +1,84 @@
...
+using ZeroC.Slice.Symbols;
+
+namespace ZeroC.Slice.Generator;
+
+/// <summary>Formats Slice doc comments into C# XML doc comment strings.</summary>
+internal static class DocCommentFormatter
+{
+    /// <summary>Formats the overview of a <see cref="Comment"/> as XML doc comment summary text.
+    /// Returns null if the comment has no overview.</summary>
+    internal static string? FormatSummary(Comment? comment, string currentNamespace)
```
bernardnormier: I find this method name pretty confusing: FormatSummary doesn't include the `<summary>` while FormatSeeAlso includes `<seealso>`.

src/ZeroC.Slice.Symbols/Comment.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    /// <summary>Gets the overview message components.</summary>
+    public required ImmutableList<CommentMessageComponent> Overview { get; init; }
+
+    /// <summary>Gets the @see tag references.</summary>
+    public required ImmutableList<CommentLink> SeeTags { get; init; }
+}
+
+/// <summary>A component of a doc comment overview message.</summary>
+public abstract record class CommentMessageComponent
```
bernardnormier: Should we just use plain inheritance instead of this Dunet-light record classes?

src/IceRpc.Slice.Generator/DispatchGenerator.cs
pepone: Yes

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
pepone: I agree, I was already thinking about renaming this to make it clear.

I think `FormatOverview` is a better name because it formats the `Comment.Overview` and is used to build both the`<summary` and `<param` tags contents.

Also renaming `FormatSeeAlso` to `FormatSeeAlsoTags` to make clear in builds the seealso tags.

src/ZeroC.Slice.Symbols/Comment.cs
pepone: Reworked to use plain inheritance

PR #4330, opened 2026-03-20
Add more Ice tests

bernardnormier:
This PR duplicates a number of Slice tests into Ice tests.

It also moves some Slice.Generator tests that don't depend on generated code into non-Generator Slice tests.

Finally, it removes a few payloadContinuation from IceRpc.Ice.Operations: the Ice operations don't support streaming and as a result never use payload continuations.


tests/IceRpc.Ice.Codec.Tests/StringDecodingTests.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) ZeroC, Inc.
 
 using NUnit.Framework;
+using System.Buffers;
 using System.IO.Pipelines;
```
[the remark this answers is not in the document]
bernardnormier: It's required.

tests/IceRpc.Ice.Codec.Tests/StringDecodingTests.cs
```diff
@@ -65,4 +66,45 @@ public void Decode_multi_segment_string(string value)
...
+    }
+
+    [Test]
+    public void Decode_multi_segment_non_utf8_string_fails()
+    {
+        Assert.That(() =>
+        {
+            // Arrange
+            // A custom memory pool with a tiny max buffer size
+            using var customPool = new TestMemoryPool(7);
```
pepone, line 88: What is the point of the memory pool in this test?

tests/IceRpc.Ice.Generator.None.Tests/StructTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        expected.Encode(ref encoder);
+
+        var decoder = new IceDecoder(buffer.WrittenMemory);
+        Assert.That(decoder.DecodeInt(), Is.EqualTo(expected.I));
+        Assert.That(decoder.DecodeInt(), Is.EqualTo(expected.J));
+        Assert.That(decoder.Consumed, Is.EqualTo(buffer.WrittenMemory.Length));
+    }
+
+    [Test]
+    public void Cs_attribute_on_field()
```
pepone: I think it should be `CS_` because is two letters

tests/IceRpc.Ice.Generator.None.Tests/StructTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        var memberInfos = typeof(MyStructWithFieldAttributes).GetMember("I");
+        var attributes = memberInfos[0].GetCustomAttributes(typeof(System.ComponentModel.DescriptionAttribute), false);
+        var description = ((System.ComponentModel.DescriptionAttribute)attributes[0]).Description;
+
+        // Assert
+        Assert.That(description, Is.EqualTo("An integer"));
+    }
+
+    [Test]
+    public void Clone_cs_readonly_struct()
```
pepone, line 56: The clone is provided by record struct isn't it? Why should we test it?

tests/IceRpc.Ice.Generator.Tests/ServiceTests.cs
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) ZeroC, Inc.
+
+using IceRpc.Features;
+using IceRpc.Tests.Common;
+using NUnit.Framework;
+
+namespace IceRpc.Ice.Generator.Tests;
+
+[Parallelizable(scope: ParallelScope.All)]
+public partial class ServiceTests
```
pepone: Why is this class partial?

tests/ZeroC.Slice.Codec.Tests/SequenceDecodingTests.cs
```diff
@@ -0,0 +1,159 @@
...
+    {
+        A = 1,
+        B = 2,
+        C = 3,
+        D = 4,
+    };
+
+    // This test is unique to the Slice encoding where enums can use a fixed number of bytes.
+    [Test]
+    public void Decode_sequence_with_enum_range_action()
```
pepone: Why `_range_action`?

pepone (approved):

tests/IceRpc.Ice.Codec.Tests/StringDecodingTests.cs
```diff
@@ -65,4 +66,45 @@ public void Decode_multi_segment_string(string value)
...
 
+    [Test]
+    public void Decode_single_segment_non_utf8_string_fails()
+    {
+        Assert.That(() =>
+        {
+            var encodedString = new byte[] { 0x02, 0xFD, 0xFF }; // Ice size 2 + invalid UTF-8 bytes
+            var sut = new IceDecoder(encodedString);
+
+            sut.DecodeString();
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested.

tests/IceRpc.Ice.Generator.None.Tests/StructTests.cs
bernardnormier: Fixed.

tests/IceRpc.Ice.Generator.Tests/ServiceTests.cs
bernardnormier: Fixed. Also Slice equivalent.

tests/ZeroC.Slice.Codec.Tests/SequenceDecodingTests.cs
bernardnormier: Renamed `_with_element_action` and added equivalent test for IceRpc.Ice.Codec.

tests/IceRpc.Ice.Generator.None.Tests/StructTests.cs
bernardnormier: You're right, and we have the same test for Slice. We should replace it by a "readonly" test ... for a follow-up PR.

tests/IceRpc.Ice.Codec.Tests/StringDecodingTests.cs
bernardnormier: It's really a white box test, where we understand that one difficulty in the decoding implementation is the crossing of underlying memory segment boundaries. The default segments are 4K, so not easy to test. With this custom pool, we can actually test segment boundaries.

*However*, this white box description corresponds to an early implementation of DecodeString where we were dealing with these memory segments in our implementation. We now use the Encoding.GetString extension that implements all this logic. So we're really testing Encoding.GetString, and we should not. 

Let's clean this up in a follow-up PR.

tests/IceRpc.Ice.Generator.None.Tests/SequenceDecodingTests.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    [Test]
+    public void Decode_bool_sequence_field()
+    {
+        // Arrange
+        bool[] expected = [false, true, false];
+        var buffer = new MemoryBufferWriter(new byte[256]);
+        var encoder = new IceEncoder(buffer);
+        encoder.EncodeSize(3);
+        encoder.WriteByteSpan(new byte[] { 0x00, 0x01, 0x00 });
```
pepone, line 21: We can use EncodeSequence here, as we unit test it separately.

tests/IceRpc.Ice.Generator.Tests/TypeNameQualificationTests.cs
```diff
@@ -0,0 +1,36 @@
...
+using IceRpc.Features;
+using IceRpc.Tests.Common;
+using NUnit.Framework;
+
+namespace IceRpc.Ice.Generator.Tests;
+
+[Parallelizable(scope: ParallelScope.All)]
+public partial class TypeNameQualificationTests
+{
+    /// <summary>Verifies that when a type is defined in multiple modules, the generated code doesn't mix up the
```
pepone: Is the terminology correct here " when a type is defined in multiple modules", maybe is clear to use the "type name" for example they are different types with the same unqualified name. Is not the same type defined in multiple module.

tests/ZeroC.Slice.Codec.Tests/DictionaryDecodingTests.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    [Test]
+    public void Decode_dictionary()
+    {
+        // Arrange
+        var buffer = new MemoryBufferWriter(new byte[1024 * 256]);
+        var encoder = new SliceEncoder(buffer);
+        var expected = Enumerable.Range(0, 1024).ToDictionary(key => key, value => $"value-{value}");
+
+        // A dictionary is encoded like a sequence.
```
pepone:
```suggestion
        // A dictionary is encoded like a sequence of key value pairs.
```

tests/ZeroC.Slice.Codec.Tests/SequenceDecodingTests.cs
```diff
@@ -0,0 +1,158 @@
...
+namespace ZeroC.Slice.Codec.Tests;
+
+[Parallelizable(scope: ParallelScope.All)]
+public class SequenceDecodingTests
+{
+    /// <summary>Tests <see cref="SliceDecoderExtensions.DecodeSequence{T}(ref SliceDecoder, Action{T}?)" /> with a
+    /// fixed-size numeric value type.</summary>
+    [Test]
+    public void Decode_fixed_sized_numeric_sequence()
+    {
```
pepone, line 16: Add `// Arrange // Act // Assert`

tests/ZeroC.Slice.Codec.Tests/SequenceDecodingTests.cs
```diff
@@ -0,0 +1,158 @@
...
+        Assert.That(result, Is.EqualTo(expected));
+        Assert.That(sut.Consumed, Is.EqualTo(buffer.WrittenMemory.Length));
+    }
+
+    /// <summary>Tests <see cref="SliceDecoderExtensions.DecodeSequence{T}(ref SliceDecoder, Action{T}?)" /> with a
+    /// string sequence.</summary>
+    [Test]
+    public void Decode_string_sequence()
+    {
+        string[] expected = Enumerable.Range(0, 256).Select(i => $"string-{i}").ToArray();
```
pepone, line 42: Here too

tests/ZeroC.Slice.Codec.Tests/SequenceDecodingTests.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        // Act
+        int?[] decoded = sut.DecodeSequenceOfOptionals<int?>((ref SliceDecoder decoder) => decoder.DecodeInt32());
+
+        // Assert
+        Assert.That(decoded, Is.EqualTo(expected));
+    }
+
+    [Test]
+    public void Decode_sequence_with_bit_sequence_exceeds_default_max_collection_allocation()
```
pepone, line 88: We can improve this test to set the limit and test with several sizes, for a follow up PR.

pepone (approved):

tests/IceRpc.Ice.Generator.Tests/DictionaryMappingTests.cs
```diff
@@ -0,0 +1,346 @@
...
+        {
+            Payload = requestPayload
+        };
+        Dictionary<MyStruct, MyStruct> decodedValue =
+            await IDictionaryMappingOperationsService.Request.DecodeSendDictionaryOfMyStructAsync(
+                request,
+                default);
+        Assert.That(decodedValue, Is.EqualTo(value));
+    }
+
```
externl, line 228: Should/do we have test for dictionary of classes (as the values)?

externl (approved):
Looks good to me!

tests/IceRpc.Ice.Generator.Tests/TypeNameQualificationTests.cs
bernardnormier: Fixed. Also added missing Ice test.

tests/IceRpc.Ice.Generator.None.Tests/SequenceDecodingTests.cs
bernardnormier: It's really for consistency with the following test, that encodes invalid bool values.

tests/ZeroC.Slice.Codec.Tests/DictionaryDecodingTests.cs
bernardnormier: Fixed

tests/ZeroC.Slice.Codec.Tests/SequenceDecodingTests.cs
bernardnormier: Fixed

tests/ZeroC.Slice.Codec.Tests/SequenceDecodingTests.cs
bernardnormier: Fixed

tests/IceRpc.Ice.Generator.Tests/DictionaryMappingTests.cs
bernardnormier: We certainly could. However, I don't think the code generator has a special path for dictionary where the value is a class.

PR #4341, opened 2026-03-24
Replace EncodedReturn attribute argument

### LINKED ISSUE #4339: Remove IceOperationAttribute.EncodedReturn property
This property is not necessary: IceRpc.ServiceGenerator can figure out if the Ice operation has the "marshaled-result" attribute from the signature of the dispatch method.

bernardnormier:
Fixes #4339.


src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs
```diff
@@ -169,6 +161,9 @@ internal IceServiceMethod(IMethodSymbol method, AttributeData attribute)
                         _idempotent = idempotent;
                     }
                     break;
+                // TODO: issue a diagnostic if we encounter an unknown named argument
```
externl, line 164: Always best to open an issue unless you plan to do immediately.

externl (approved):

PR #4343, opened 2026-03-24
Update Slice enum terminology

pepone:
Update C# generators to use the new enumeration terminology

### DIFF
--- a/src/ZeroC.Slice.Symbols/VariantEnum.cs
+++ b/src/ZeroC.Slice.Symbols/VariantEnum.cs
@@ -7,8 +7,8 @@ namespace ZeroC.Slice.Symbols;
 /// <summary>Represents a Slice variant enumeration.</summary>
 public class VariantEnum : Entity, ISymbol, IType
 {
-    /// <summary>Gets a value indicating whether this variant enumeration is a compact enumeration. Compact variant
-    /// enumerations do not accept tagged fields.</summary>
+    /// <summary>Gets a value indicating whether this variant enumeration is a compact enumeration. Variants in compact
+    /// variant enumerations don't accept tagged fields.</summary>
     public required bool IsCompact { get; init; }
 
     /// <summary>Gets a value indicating whether this enumeration is unchecked.</summary>


src/ZeroC.Slice.Generator/CSAttributeValidator.cs
```diff
@@ -72,9 +72,9 @@ private static void ValidateSymbol(ISymbol symbol, List<Compiler.Diagnostic> dia
                 }
                 break;
 
-            case EnumWithUnderlying e:
+            case BasicEnum e:
                 ValidateAttributes(e.Attributes, Target.Enum, diagnostics);
-                ValidateEnumWithUnderlyingEnumerators(e, diagnostics);
+                ValidateBasicEnumEnumerators(e, diagnostics);
```
InsertCreativityHere: I'd just name this `ValidateEnumerators`, since `Enumerators` are always in a `BasicEnum`.
`VariantEnum`s have `Variants` instead of `Enumerators`, so no ambiguity.

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
```diff
@@ -101,46 +101,46 @@ private static void ValidateSymbol(ISymbol symbol, List<Compiler.Diagnostic> dia
...
-    private static void ValidateEnumWithUnderlyingEnumerators(
-        EnumWithUnderlying e,
+    private static void ValidateBasicEnumEnumerators(
+        BasicEnum e,
         List<Compiler.Diagnostic> diagnostics)
     {
-        // TODO will be better to refactor EnumWithUnderlying to avoid the need for this switch,
-        // We can have Enumerators and TypedEnumerators and only used the typed one to access the value.
+        // TODO will be better to refactor BasicEnum to avoid the need for this switch,
+        // We can have Variants and TypedEnumerators and only used the typed one to access the value.
```
InsertCreativityHere: Is `Variants` correct here?

src/ZeroC.Slice.Symbols/Compiler/SyntaxElements.cs
```diff
@@ -465,7 +465,7 @@ public readonly void Encode(ref SliceEncoder encoder)
     }
 }
 
-/// <remarks>The Slice compiler generated this record struct from the Slice struct <c>Compiler::Enumerator</c>.</remarks>
+/// <remarks>The Slice compiler generated this record struct from the Slice struct <c>Compiler::Variant</c>.</remarks>
```
InsertCreativityHere: This code should only be generated right?
It's incorrect to edit it directly.

src/ZeroC.Slice.Symbols/VariantEnum.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System.Collections.Immutable;
+
+namespace ZeroC.Slice.Symbols;
+
+/// <summary>Represents a Slice variant enumeration.</summary>
+public class VariantEnum : Entity, ISymbol, IType
+{
+    /// <summary>Gets a value indicating whether this variant enumeration is a compact enumeration. Compact variant
+    /// enumerations cannot include unknown variants.</summary>
```
InsertCreativityHere: I don't think this comment is accurate. The only effect of `IsCompact` is on whether tags are accepted, nothing to do with unknown variants.

InsertCreativityHere (approved):
Looks good to me!

src/ZeroC.Slice.Symbols/VariantEnum.cs
bernardnormier: We don't support compact for unchecked enum because unchecked enum are already as compact as possible.

bernardnormier (approved):

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
```diff
@@ -101,46 +101,46 @@ private static void ValidateSymbol(ISymbol symbol, List<Compiler.Diagnostic> dia
...
         switch (e)
         {
-            case EnumWithUnderlying<sbyte> enumWithUnderlying:
+            case BasicEnum<sbyte> enumWithUnderlying:
```
externl: `enumWithUnderlying` I would just name these all `enum` or `basicEnum`.

src/ZeroC.Slice.Symbols/Compiler/SyntaxElements.cs
externl: I believe so.

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
pepone: Makes sense, fixed as suggested!

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
pepone: No, I removed this TODO for now I will open a separate issue to address what this was about.

src/ZeroC.Slice.Symbols/Compiler/SyntaxElements.cs
pepone: Yes reverted the changes here.

src/ZeroC.Slice.Symbols/VariantEnum.cs
pepone: Updated to:

Gets a value indicating whether this variant enumeration is a compact enumeration. Variants in compact variant enumerations don't accept tagged fields.

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
pepone: Fixed as suggested

Issue #4317, opened 2026-03-18
Move ALPN to the protocol layer


pepone:
https://github.com/icerpc/icerpc-csharp/blob/0ea64b253e00be3fba1bcb9b6f845836718858c6/src/IceRpc/Transports/Tcp/TcpClientTransport.cs#L69-L76

The code above sets ALPN to the protocol name. Setting the ALPN value should be a protocol concern not a transport concern. 

QUIC transport has similar code:

https://github.com/icerpc/icerpc-csharp/blob/0ea64b253e00be3fba1bcb9b6f845836718858c6/src/IceRpc/Transports/Quic/QuicClientTransport.cs#L58-L62

PR #4346, opened 2026-03-24
Move ALPN to the protocol layer (#4317)

pepone:
Add ApplicationProtocol property to DuplexConnectionOptions and MultiplexedConnectionOptions so the protocol layer can communicate the ALPN value to the transport layer without touching SSL authentication options directly.

The protocol layer (ClientProtocolConnectionFactory and Server) sets ApplicationProtocol on the transport connection options. The transport layer uses this value instead of serverAddress.Protocol.Name when configuring SSL ApplicationProtocols.

QUIC transports now validate that ApplicationProtocol is set, since ALPN is mandatory for QUIC.

### DIFF
--- a/src/IceRpc/ClientProtocolConnectionFactory.cs
+++ b/src/IceRpc/ClientProtocolConnectionFactory.cs
@@ -43,6 +43,7 @@ public sealed class ClientProtocolConnectionFactory : IClientProtocolConnectionF
         _duplexClientTransport = duplexClientTransport ?? IDuplexClientTransport.Default;
         _duplexConnectionOptions = new DuplexConnectionOptions
         {
+            ApplicationProtocol = Protocol.Ice.Name,
             Pool = connectionOptions.Pool,
             MinSegmentSize = connectionOptions.MinSegmentSize,
         };
@@ -53,6 +54,7 @@ public sealed class ClientProtocolConnectionFactory : IClientProtocolConnectionF
         // which is accepted locally is the control stream created by the peer.
         _multiplexedConnectionOptions = new MultiplexedConnectionOptions
         {
+            ApplicationProtocol = Protocol.IceRpc.Name,
             HandshakeTimeout = connectTimeout,
 
             MaxBidirectionalStreams = connectionOptions.Dispatcher is null ? 0 :
--- a/src/IceRpc/Server.cs
+++ b/src/IceRpc/Server.cs
@@ -101,6 +101,7 @@ public sealed class Server : IAsyncDisposable
                     _serverAddress,
                     new DuplexConnectionOptions
                     {
+                        ApplicationProtocol = _serverAddress.Protocol.Name,
                         MinSegmentSize = options.ConnectionOptions.MinSegmentSize,
                         Pool = options.ConnectionOptions.Pool,
                     },
@@ -116,6 +117,7 @@ public sealed class Server : IAsyncDisposable
                     _serverAddress,
                     new MultiplexedConnectionOptions
                     {
+                        ApplicationProtocol = _serverAddress.Protocol.Name,
                         HandshakeTimeout = options.ConnectTimeout,
                         MaxBidirectionalStreams = options.ConnectionOptions.MaxIceRpcBidirectionalStreams,
                         // Add an additional stream for the icerpc protocol control stream.
--- a/src/IceRpc/Transports/DuplexConnectionOptions.cs
+++ b/src/IceRpc/Transports/DuplexConnectionOptions.cs
@@ -7,6 +7,11 @@ namespace IceRpc.Transports;
 /// <summary>A property bag used to configure a <see cref="IDuplexConnection" />.</summary>
 public record class DuplexConnectionOptions
 {
+    /// <summary>Gets or sets the application protocol for ALPN (Application-Layer Protocol Negotiation).</summary>
+    /// <value>The application protocol name, or <see langword="null" /> if no application protocol is configured.
+    /// Defaults to <see langword="null" />.</value>
+    public string? ApplicationProtocol { get; set; }
+
     /// <summary>Gets or sets the minimum size of the segment requested from the <see cref="Pool" />.</summary>
     /// <value>The minimum size in bytes of the segment requested from the <see cref="Pool" />. It cannot be less than
     /// <c>1</c> KB. Defaults to <c>4</c> KB.</value>
--- a/src/IceRpc/Transports/Quic/Internal/QuicMultiplexedListener.cs
+++ b/src/IceRpc/Transports/Quic/Internal/QuicMultiplexedListener.cs
@@ -55,9 +55,9 @@ internal class QuicMultiplexedListener : IListener<IMultiplexedConnection>
         _options = options;
 
         authenticationOptions = authenticationOptions.Clone();
-        authenticationOptions.ApplicationProtocols ??= new List<SslApplicationProtocol> // Mandatory with Quic
+        authenticationOptions.ApplicationProtocols ??= new List<SslApplicationProtocol>
         {
-            new SslApplicationProtocol(serverAddress.Protocol.Name)
+            new SslApplicationProtocol(options.ApplicationProtocol!)
         };
 
         _quicServerOptions = new QuicServerConnectionOptions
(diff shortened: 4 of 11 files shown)


src/IceRpc/Transports/Tcp/TcpClientTransport.cs
```diff
@@ -59,19 +59,14 @@ public IDuplexConnection CreateConnection(
...
-            // specify any application protocol. This way, a proxy server listening on a port shared by multiple
-            // application protocols can use this ALPN protocol ID to forward all ice/icerpc traffic to an ice/icerpc
-            // back-end server.
-            // We do this only when the port is not the default port for ice or icerpc; when we use the IANA-registered
-            // default port, the server can and should use this port number to identify the application protocol when no
-            // ALPN protocol ID is provided.
+            // Set ApplicationProtocols when the application does not specify any application protocol and the
+            // connection options provide one.
             if (authenticationOptions.ApplicationProtocols is null &&
-                serverAddress.Port != serverAddress.Protocol.DefaultPort)
```
pepone, line 70: I removed this default port check it doesn't make sense to embed this protocol detail in the transport.

src/IceRpc/Transports/Quic/QuicClientTransport.cs
```diff
@@ -48,6 +48,13 @@ public IMultiplexedConnection CreateConnection(
                 nameof(serverAddress));
         }
 
+        if (options.ApplicationProtocol is null)
+        {
+            throw new ArgumentException(
+                "The QUIC client transport requires the multiplexed connection options to have ApplicationProtocol set.",
```
bernardnormier, line 54: I would simplify/clarify the message, like:

ApplicationProtocol must be set when establishing a connection with QUIC.

bernardnormier: Also, is this check at the right spot?

If you use QuicClientTransport on its own, you could specify this ApplicationProtocol via the clientAuthenticationOptions.

src/IceRpc/Transports/Quic/QuicServerTransport.cs
```diff
@@ -54,6 +54,13 @@ public IListener<IMultiplexedConnection> Listen(
                 "The QUIC server transport requires the SSL server authentication options to be set.");
         }
 
+        if (options.ApplicationProtocol is null)
```
bernardnormier, line 57: Same here. If you use QuicServerTransport on its own, you could provide this application protocol using serverAuthenticationOptions.

src/IceRpc/Transports/Quic/Internal/QuicMultiplexedListener.cs
```diff
@@ -81,7 +81,7 @@ internal QuicMultiplexedListener(
                 {
                     ListenEndPoint = new IPEndPoint(ipAddress, serverAddress.Port),
                     ListenBacklog = quicTransportOptions.ListenBacklog,
-                    ApplicationProtocols = authenticationOptions.ApplicationProtocols,
+                    ApplicationProtocols = authenticationOptions.ApplicationProtocols!,
```
bernardnormier, line 84: I don't understand why we need this extra "!". It's clear from line 58 above that authenticationOptions.ApplicationProtocols is not null.

src/IceRpc/Server.cs
```diff
@@ -101,6 +101,7 @@ public Server(
                     _serverAddress,
```
bernardnormier, line 101: I assume to goal is to eventually not transmit _serverAddress to the transport, right?

src/IceRpc/ClientProtocolConnectionFactory.cs
```diff
@@ -43,6 +43,7 @@ public ClientProtocolConnectionFactory(
         _duplexClientTransport = duplexClientTransport ?? IDuplexClientTransport.Default;
         _duplexConnectionOptions = new DuplexConnectionOptions
         {
+            ApplicationProtocol = Protocol.Ice.Name,
```
bernardnormier, line 46: We could set ApplicationProtocol to null in CreateConnection below, when the serverAddress port matches the default port, but I guess it's just simpler to always set this protocol all the time, like for QUIC.

Do we set ALPN all the time with Ice?

bernardnormier (commented):
Looks good.

pepone:
> Have you considered the alternative where the caller (protocol) sets ALPN in the authentication options?

Yes, this was my initial approach. I switched because when the protocol layer sets ALPN doesn't know if the transport would use it. 

For example we allow create secure client connections without specify the authentication options, then if this is connection uses a secure transport it creates the authentication options. What will we do the protocol layer here if there are no authentication options, always creating authentication options would be wrong because we use this in Tcp transport to decide whether to create a secure or non secure connection. I move this to the transport connection options and let the transport decide if it must be used.

This made me think if we should add a property in the transport interface `IsSecure` or `UsesApln` but it wouldn't work. For tcp the same transport can create both secure and non secure connections and this depends on whether the connection creation is given the SSL authentication options or not.  I think it will be clear to force either tcp or ssl when the transport is created. Kind of what we already do for Ice protocol in Tcp transport.

src/IceRpc/Transports/Quic/QuicClientTransport.cs
```diff
@@ -57,8 +64,7 @@ public IMultiplexedConnection CreateConnection(
         clientAuthenticationOptions.TargetHost ??= serverAddress.Host;
         clientAuthenticationOptions.ApplicationProtocols ??=
         [
-            // Mandatory with Quic
-            new(serverAddress.Protocol.Name)
+            new(options.ApplicationProtocol!)
```
externl, line 67: Is this n necessary since we checked for null above?

src/IceRpc/Transports/DuplexConnectionOptions.cs
```diff
@@ -7,6 +7,11 @@ namespace IceRpc.Transports;
 /// <summary>A property bag used to configure a <see cref="IDuplexConnection" />.</summary>
 public record class DuplexConnectionOptions
 {
+    /// <summary>Gets or sets the application protocol for ALPN (Application-Layer Protocol Negotiation).</summary>
+    /// <value>The application protocol name, or <see langword="null" /> if no application protocol is configured.
+    /// Defaults to <see langword="null" />.</value>
+    public string? ApplicationProtocol { get; set; }
```
externl, line 13: ALPN is a an extension of TLS. I don't think we should expose this at his level. To me this is akin to having other TLS settings here.

bernardnormier:
> > Have you considered the alternative where the caller (protocol) sets ALPN in the authentication options?
> 
> Yes, this was my initial approach. I switched because when the protocol layer sets ALPN doesn't know if the transport would use it.
[...]

Let's focus on the client-side. I think this addition to the transport interfaces would solve this issue:

```cs

public interface IMultiplexedClientTransport
{
    // When true, clientAuthenticationOptions must be non-null, and the caller
    // will create a empty clientAuthenticationOptions with the correct ALPN unless the user provided it already
    // When false, a null clientAuthenticationOptions is accepted. A non-null clientAuthenticationOptions results in a 
    // secure connection or is rejected if the transport is non secure-capable.
    // TODO: maybe IsSecure should be "Secure Only" - it doesn't mean "Secure Capable".
    // Like Name, IsSecure doesn't change during the lifetime of a transport instance.
    bool IsSecure { get; } 
}

public interface IDuplexClientTransport
{
    // Same as above. transport name is the server address transport name, which can be null
    // It can be "tcp" or "ssl" for the TcpClienTransport.
    bool IsSecure(string? transportName);
}
```

The logic in ClientConnectionFactory would be something like:
 - is the clientAuthenticationOptions null? (else proceed)
    - if the transport is secure, create an empty clientAuthenticationOptions with the correct ALPN.

src/IceRpc/Server.cs
pepone: Yes, the first part is to move this protocol transport dependency out.

src/IceRpc/ClientProtocolConnectionFactory.cs
pepone: Ice/SSL doesn't set ALPN in any of the implementations.

pepone:
I will create a new PR with a different approach, and full scope of removing the protocol dependencies on transports, not just ALPN.

PR #4351, opened 2026-03-25
Remove useTagEndMarker parameter

bernardnormier:
This PR simplifies the Ice decoding of tagged fields by removing the useTageEndMarker parameter. It was not needed since we can compute it.

Note: this PR depends on the corresponding fix in slice2cs --icerpc.

### DIFF
--- a/src/IceRpc/Ice/Codec/IceDecoder.cs
+++ b/src/IceRpc/Ice/Codec/IceDecoder.cs
@@ -278,14 +278,12 @@ public ref partial struct IceDecoder
     /// <param name="tag">The tag.</param>
     /// <param name="tagFormat">The expected tag format of this tag when found in the underlying buffer.</param>
     /// <param name="decodeFunc">A decode function that decodes the value of this tag.</param>
-    /// <param name="useTagEndMarker">When <see langword="true" />, a tag end marker marks the end of the tagged fields.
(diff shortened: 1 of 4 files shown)


InsertCreativityHere (approved):
Looks good to me!

src/IceRpc/Ice/Codec/IceDecoder.cs
```diff
@@ -333,21 +331,11 @@ public void Skip(int count)
...
-                nameof(useTagEndMarker));
-        }
-        else if (useTagEndMarker && _classContext.Current.InstanceType == InstanceType.None)
-        {
-            throw new ArgumentException(
-                $"The {nameof(useTagEndMarker)} argument must be false when decoding parameters.",
-                nameof(useTagEndMarker));
-        }
+        // True when decoding a class or exception, false when decoding parameters. Keep in mind we never decode a
+        // class while decoding a tagged parameter.
```
pepone, line 337: There is no ambiguity because we dropped support for tagged class and exceptions right?

pepone (approved):

src/IceRpc/Ice/Codec/IceDecoder.cs
bernardnormier: I believe so, yes.

PR #4353, opened 2026-03-26
Make transports protocol-independent

pepone:
Replace ServerAddress with a new TransportAddress type in all transport
interfaces, breaking the circular dependency between the protocol and
transport layers.

Key changes:

- Add TransportAddress (readonly record struct with Host, Port, TrasnportName, Params)
  used by all transport CreateConnection/Listen methods and IListener
- The protocol layer sets ALPN directly on the SSL authentication options and validates that
  user-provided ALPN matches the expected protocol
- Add IsSslRequired(string? transportName) method to transport interfaces;
  TCP returns true for "ssl", QUIC always true, others false
- Transports validate TransportAddress.TransportName and reject unknown Params
- TCP transport accepts both "tcp" and "ssl" transport names
- For Ice protocol, the protocol layer strips "t"/"z" params before passing to the transport
- QUIC transport requires SSL auth options with ApplicationProtocols pre-set by the caller
- Add Protocol.AlpnProtocol internal property

### DIFF
--- a/src/IceRpc/Transports/Quic/Internal/QuicMultiplexedListener.cs
+++ b/src/IceRpc/Transports/Quic/Internal/QuicMultiplexedListener.cs
@@ -52,21 +52,10 @@ internal class QuicMultiplexedListener : IListener<IMultiplexedConnection>
                 nameof(transportAddress));
         }
 
-        if (serverAuthenticationOptions is null)
-        {
-            throw new ArgumentNullException(
-                nameof(serverAuthenticationOptions),
-                "The QUIC server transport requires the SSL server authentication options to be set.");
-        }
         serverAuthenticationOptions = serverAuthenticationOptions.Clone();
 
-        if (serverAuthenticationOptions.ApplicationProtocols
-            is not List<SslApplicationProtocol> applicationProtocols || applicationProtocols.Count == 0)
-        {
-            throw new ArgumentException(
-                "The QUIC server transport requires ApplicationProtocols to be set in the SSL server authentication options.",
-                nameof(serverAuthenticationOptions));
-        }
+        // Always set by the caller QuicServerTransport.
+        Debug.Assert(serverAuthenticationOptions.ApplicationProtocols is not null);
 
         _options = options;
 
--- a/src/IceRpc/Transports/Tcp/TcpClientTransport.cs
+++ b/src/IceRpc/Transports/Tcp/TcpClientTransport.cs
@@ -40,8 +40,7 @@ public class TcpClientTransport : IDuplexClientTransport
         // "ssl" is only accepted for the Ice protocol, identified by the ALPN.
         if (transportAddress.TransportName == "ssl")
         {
-            if (clientAuthenticationOptions?.ApplicationProtocols
-                is not List<SslApplicationProtocol> alpnProtocols ||
+            if (clientAuthenticationOptions?.ApplicationProtocols is not List<SslApplicationProtocol> alpnProtocols ||
                 alpnProtocols.Count != 1 ||
                 alpnProtocols[0] != new SslApplicationProtocol("ice"))
             {
--- a/src/IceRpc/Transports/Tcp/TcpServerTransport.cs
+++ b/src/IceRpc/Transports/Tcp/TcpServerTransport.cs
@@ -32,7 +32,13 @@ public class TcpServerTransport : IDuplexServerTransport
         // "ssl" is only accepted for the Ice protocol, identified by the ALPN.
         if (transportAddress.TransportName == "ssl")
         {
-            if (serverAuthenticationOptions?.ApplicationProtocols
+            if (serverAuthenticationOptions is null)
+            {
+                throw new ArgumentNullException(
+                    nameof(serverAuthenticationOptions),
+                    "The SSL server transport requires the SSL server authentication options to be set.");
+            }
+            else if (serverAuthenticationOptions?.ApplicationProtocols
                 is not List<SslApplicationProtocol> alpnProtocols ||
                 alpnProtocols.Count != 1 ||
                 alpnProtocols[0] != new SslApplicationProtocol("ice"))


src/IceRpc/Transports/TransportAddress.cs
```diff
@@ -0,0 +1,31 @@
...
+public readonly record struct TransportAddress
+{
+    /// <summary>Gets or initializes the host.</summary>
+    /// <value>The host name or IP address. Defaults to <c>::0</c>.</value>
+    public string Host { get; init; } = "::0";
+
+    /// <summary>Gets or initializes the transport name.</summary>
+    /// <value>The transport name (e.g., "tcp", "ssl", "quic"), or <see langword="null" /> if unspecified. Defaults to
+    /// <see langword="null" />.</value>
+    public string? Name { get; init; }
```
bernardnormier: I would rename this property TransportName, and reorder these properties in a more URI-like order, like

Host, Port, TransportName, Params

src/IceRpc/Transports/TransportAddress.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Immutable;
+
+namespace IceRpc.Transports;
+
+/// <summary>Represents a transport address with the host and port to connect to or listen on.</summary>
+public readonly record struct TransportAddress
+{
+    /// <summary>Gets or initializes the host.</summary>
+    /// <value>The host name or IP address. Defaults to <c>::0</c>.</value>
+    public string Host { get; init; } = "::0";
```
bernardnormier: Not sure about these various defaults. It's for tests that create a TransportAddress? Is `::0` and Port = 0 really meaningful/common for tests?

src/IceRpc/Transports/IMultiplexedClientTransport.cs
```diff
@@ -35,17 +35,24 @@ public static IMultiplexedClientTransport Default
     [SupportedOSPlatform("windows")]
     private static readonly QuicClientTransport _quicClientTransport = new();
 
-    /// <summary>Gets the transport's name.</summary>
+    /// <summary>Gets a value indicating whether this transport requires SSL.</summary>
+    /// <value><see langword="true" /> if this transport requires SSL; otherwise, <see langword="false" />. Defaults to
+    /// <see langword="false" />.</value>
+    bool IsSslRequired(string? transportName) => false;
```
bernardnormier: This should be a clean interface with no default method impl.

src/IceRpc/Transports/TransportAddress.cs
pepone: Removed

src/IceRpc/Transports/TransportAddress.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>Gets or initializes the transport name.</summary>
+    /// <value>The transport name (e.g., "tcp", "ssl", "quic"), or <see langword="null" /> if unspecified.</value>
+    public string? TransportName { get; init; }
+
+    /// <summary>Gets or initializes the transport-specific parameters.</summary>
+    /// <value>The transport parameters. Defaults to an empty dictionary.</value>
+    public ImmutableDictionary<string, string> Params { get; init; } = [];
+
+    /// <summary>Constructs a transport address.</summary>
+    public TransportAddress()
```
bernardnormier, line 27: Are you sure you need this constructor?

src/IceRpc/Transports/IDuplexClientTransport.cs
```diff
@@ -12,18 +12,25 @@ public interface IDuplexClientTransport
...
+    /// <value>The transport accepts transport addresses that use this name as the
+    /// <see cref="TransportAddress.TransportName"/>. Some transports may accept additional names beyond this default.
+    /// </value>
+    string DefaultName { get; }
 
-    /// <summary>Creates a new transport connection to the specified server address.</summary>
-    /// <param name="serverAddress">The server address of the connection.</param>
+    /// <summary>Gets a value indicating whether this transport requires SSL.</summary>
+    /// <value><see langword="true" /> if this transport requires SSL; otherwise, <see langword="false" />. Defaults to
+    /// <see langword="false" />.</value>
```
bernardnormier: There should be no default. 
I would also add that `null` is equivalent to DefaultName.

src/IceRpc/Transports/IDuplexServerTransport.cs
```diff
@@ -12,18 +12,25 @@ public interface IDuplexServerTransport
...
+    /// <see cref="TransportAddress.TransportName"/>. Some transports may accept additional names beyond this default.
+    /// </value>
+    string DefaultName { get; }
 
-    /// <summary>Starts listening on a server address.</summary>
-    /// <param name="serverAddress">The server address of the listener.</param>
+    /// <summary>Gets a value indicating whether this transport requires SSL.</summary>
+    /// <value><see langword="true" /> if this transport requires SSL; otherwise, <see langword="false" />. Defaults to
+    /// <see langword="false" />.</value>
+    bool IsSslRequired(string? transportName);
```
bernardnormier: I don't think we need or should add this IsSslRequired on the server-side.

It doesn't make sense for the caller to create an empty auth options, so we don't need it.

src/IceRpc/Transports/IMultiplexedClientTransport.cs
```diff
@@ -35,17 +35,25 @@ public static IMultiplexedClientTransport Default
...
+    /// <value>The transport accepts transport addresses that use this name as the
+    /// <see cref="TransportAddress.TransportName"/>. Some transports may accept additional names beyond this default.
+    /// </value>
+    string DefaultName { get; }
 
-    /// <summary>Creates a new transport connection to the specified server address.</summary>
-    /// <param name="serverAddress">The server address of the connection.</param>
+    /// <summary>Gets a value indicating whether this transport requires SSL.</summary>
+    /// <value><see langword="true" /> if this transport requires SSL; otherwise, <see langword="false" />. Defaults to
+    /// <see langword="false" />.</value>
```
bernardnormier: See Duplex comments.

src/IceRpc/Transports/IMultiplexedServerTransport.cs
```diff
@@ -35,18 +35,25 @@ public static IMultiplexedServerTransport Default
...
-    string Name { get; }
+    /// <summary>Gets the default transport name.</summary>
+    /// <value>The transport accepts transport addresses that use this name as the
+    /// <see cref="TransportAddress.TransportName"/>. Some transports may accept additional names beyond this default.
+    /// </value>
+    string DefaultName { get; }
 
-    /// <summary>Starts listening on a server address.</summary>
-    /// <param name="serverAddress">The server address of the listener.</param>
+    /// <summary>Gets a value indicating whether this transport requires SSL.</summary>
```
bernardnormier: See Duplex comments.

src/IceRpc/Transports/Quic/QuicServerTransport.cs
```diff
@@ -54,11 +62,13 @@ public IListener<IMultiplexedConnection> Listen(
                 "The QUIC server transport requires the SSL server authentication options to be set.");
         }
 
-        if (serverAddress.Transport is null)
+        if (serverAuthenticationOptions.ApplicationProtocols is null or { Count: 0 })
         {
-            serverAddress = serverAddress with { Transport = Name };
+            throw new ArgumentException(
+                "The QUIC server transport requires ApplicationProtocols to be set on the SSL server authentication options.",
```
bernardnormier: Should it be "set _in_ the SSL server authentication options"?

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
```diff
@@ -28,87 +28,35 @@ public TcpClientTransport()
...
+                nameof(transportAddress));
         }
 
-        SslClientAuthenticationOptions? authenticationOptions = clientAuthenticationOptions?.Clone() ??
-            (serverAddress.Transport == SslName ? new SslClientAuthenticationOptions() : null);
-
-        if (authenticationOptions is not null)
+        SslClientAuthenticationOptions? authenticationOptions = clientAuthenticationOptions;
+        if (authenticationOptions is not null && authenticationOptions.TargetHost is null)
         {
```
bernardnormier: I would keep the ssl is only for Ice check, where we detect ice from the ALPN name.

src/IceRpc/Transports/Tcp/TcpServerTransport.cs
```diff
@@ -28,37 +28,30 @@ public TcpServerTransport()
...
-                nameof(serverAddress));
+                $"The TCP server transport does not support transport '{name}'.");
         }
 
-        if (serverAddress.Transport is null)
-        {
-            serverAddress = serverAddress with { Transport = Name };
-        }
-        else if (serverAddress.Transport == SslName && serverAuthenticationOptions is null)
+        if (transportAddress.TransportName == "ssl" && serverAuthenticationOptions is null)
```
bernardnormier: Same in this code: "ssl" is only for Ice, where we detect Ice from the ALPN.

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
```diff
@@ -28,87 +28,35 @@ public TcpClientTransport()
...
-            serverAddress = serverAddress with { Transport = Name };
+                "The transport address contains parameters that are not valid for the TCP client transport.",
+                nameof(transportAddress));
         }
 
-        SslClientAuthenticationOptions? authenticationOptions = clientAuthenticationOptions?.Clone() ??
-            (serverAddress.Transport == SslName ? new SslClientAuthenticationOptions() : null);
-
-        if (authenticationOptions is not null)
+        SslClientAuthenticationOptions? authenticationOptions = clientAuthenticationOptions;
```
externl: So we're no longer cloning all the time?

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
```diff
@@ -28,87 +28,35 @@ public TcpClientTransport()
...
-            if (authenticationOptions.ApplicationProtocols is null &&
-                serverAddress.Port != serverAddress.Protocol.DefaultPort)
-            {
-                authenticationOptions.ApplicationProtocols = new List<SslApplicationProtocol>
-                {
-                    new SslApplicationProtocol(serverAddress.Protocol.Name)
-                };
-            }
+            authenticationOptions = authenticationOptions.Clone();
+            authenticationOptions?.TargetHost ??= transportAddress.Host;
```
externl: What does it mean (if anything) if `authenticationOptions?.TargetHost != transportAddress.Host`;

src/IceRpc/ClientProtocolConnectionFactory.cs
```diff
@@ -77,24 +79,55 @@ public ClientProtocolConnectionFactory(
...
-                        _multiplexedConnectionOptions,
-                        _clientAuthenticationOptions),
-                    transportConnectionInformation: null,
-                    _connectionOptions,
-                    taskExceptionObserver: _logger == NullLogger.Instance ? null :
-                        new LogTaskExceptionObserver(_logger));
+        var transportAddress = serverAddress.ToTransportAddress();
+
+        // Validate user-provided ALPN.
+        if (_clientAuthenticationOptions?.ApplicationProtocols is { } applicationProtocols)
```
bernardnormier: I didn't know this `{ }`, and I don't find it easy to read. Can you switch to the traditional "is TypeName xxx" syntax?

src/IceRpc/ClientProtocolConnectionFactory.cs
```diff
@@ -77,24 +79,55 @@ public ClientProtocolConnectionFactory(
...
+        {
+            if (applicationProtocols.Count != 1 || applicationProtocols[0] != serverAddress.Protocol.AlpnProtocol)
+            {
+                throw new ArgumentException(
+                    $"The ApplicationProtocols must contain exactly '{serverAddress.Protocol.Name}'.");
```
bernardnormier: must be null or contain ...

src/IceRpc/ClientProtocolConnectionFactory.cs
```diff
@@ -77,24 +79,55 @@ public ClientProtocolConnectionFactory(
...
+            }
+        }
+
+        IProtocolConnection connection;
+        if (serverAddress.Protocol == Protocol.Ice)
+        {
+            SslClientAuthenticationOptions? authOptions = _clientAuthenticationOptions?.Clone();
```
bernardnormier: I would use authenticationOptions, like everywhere else.

src/IceRpc/Transports/IDuplexClientTransport.cs
```diff
@@ -12,18 +12,25 @@ public interface IDuplexClientTransport
...
+    bool IsSslRequired(string? transportName);
```
externl, line 27: I don't think this should default to anything? And if it had to it should probably be true. 

Should we also clarify what an implementation is supposed to do if it receives a transport name it doesn't know about/support. Returning a value seems wrong.

src/IceRpc/Transports/IDuplexServerTransport.cs
```diff
@@ -12,18 +12,25 @@ public interface IDuplexServerTransport
...
+    /// <value>The transport accepts transport addresses that use this name as the
+    /// <see cref="TransportAddress.TransportName"/>. Some transports may accept additional names beyond this default.
+    /// </value>
+    string DefaultName { get; }
 
-    /// <summary>Starts listening on a server address.</summary>
-    /// <param name="serverAddress">The server address of the listener.</param>
+    /// <summary>Gets a value indicating whether this transport requires SSL.</summary>
+    /// <value><see langword="true" /> if this transport requires SSL; otherwise, <see langword="false" />. Defaults to
+    /// <see langword="false" />.</value>
```
externl: Ditto.

src/IceRpc/Transports/IMultiplexedServerTransport.cs
```diff
@@ -35,18 +35,25 @@ public static IMultiplexedServerTransport Default
...
+    /// <value><see langword="true" /> if this transport requires SSL; otherwise, <see langword="false" />. Defaults to
+    /// <see langword="false" />.</value>
+    bool IsSslRequired(string? transportName);
```
externl: Ditto

src/IceRpc/ClientProtocolConnectionFactory.cs
```diff
@@ -77,24 +79,55 @@ public ClientProtocolConnectionFactory(
...
+            if (authOptions is null && _duplexClientTransport.IsSslRequired(serverAddress.Transport))
+            {
+                authOptions = new SslClientAuthenticationOptions();
+            }
+            authOptions?.ApplicationProtocols ??= [Protocol.Ice.AlpnProtocol];
+
+            // Strip Ice-specific params ("t", "z") before passing to transport.
```
bernardnormier: I would add something like: these params have no effect with IceRPC transports.

src/IceRpc/ServerAddress.cs
```diff
@@ -212,6 +213,16 @@ internal ServerAddress(
         _params = serverAddressParams;
         OriginalUri = null;
     }
+
+    /// <summary>Converts a <see cref="ServerAddress" /> to a <see cref="TransportAddress" />.</summary>
+    internal TransportAddress ToTransportAddress() =>
```
externl: Should this be public in case a user needs to covert from on to another.

src/IceRpc/ClientProtocolConnectionFactory.cs
```diff
@@ -77,24 +79,55 @@ public ClientProtocolConnectionFactory(
...
-                        serverAddress,
-                        _multiplexedConnectionOptions,
-                        _clientAuthenticationOptions),
-                    transportConnectionInformation: null,
-                    _connectionOptions,
-                    taskExceptionObserver: _logger == NullLogger.Instance ? null :
-                        new LogTaskExceptionObserver(_logger));
+        var transportAddress = serverAddress.ToTransportAddress();
+
+        // Validate user-provided ALPN.
```
bernardnormier: I think it would be cleaner to perform some of this cloning and validation in the constructor.

We probably need two separate _clientAuthenticationOptions fields, the _iceClientAuthenticationOptions (with the Ice ALPN) and the _iceRpcClientAuthenticationOptions (IceRPC ALPN).

And maybe the simplest would be to reject any user-provide ALPN, i.e. the user-provided ApplicationProtocols must be null.

src/IceRpc/ServerAddress.cs
```diff
@@ -212,6 +213,16 @@ internal ServerAddress(
         _params = serverAddressParams;
         OriginalUri = null;
     }
+
+    /// <summary>Converts a <see cref="ServerAddress" /> to a <see cref="TransportAddress" />.</summary>
```
bernardnormier: I don't understand why this is not an extension method.

src/IceRpc/Transports/IDuplexClientTransport.cs
pepone: Need t to fix the comment, the default were already removed!

src/IceRpc/Server.cs
```diff
@@ -88,32 +88,77 @@ public Server(
...
                     transportListener,
+                    _serverAddress,
                     options.ConnectionOptions);
             }
             else
             {
+                SslServerAuthenticationOptions? authOptions = options.ServerAuthenticationOptions;
+                if (authOptions is null)
+                {
+                    if (multiplexedServerTransport.IsSslRequired(_serverAddress.Transport))
```
bernardnormier: I don't think we need that. This can fail when we call Listen.

src/IceRpc/ServerAddress.cs
pepone: Right now is only use by the protocol connection factory, and in the server implementation. I don't expect users having to deal with this conversions.

src/IceRpc/ServerAddress.cs
bernardnormier: Or nothing since it's trivial and used like once.

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
pepone: We should, I just fixed it.

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
pepone: This is not new behavior, same history on main:

```
        if (authenticationOptions is not null)
        {
            // We are establishing a secure TLS connection. It can rely on system certificates.

            authenticationOptions.TargetHost ??= serverAddress.Host;
```

The `authenticationOptions.TargetHost` is just for SSL name verification, we default to the server address host. I think you might want to use an IP address for `serverAddress.Host` and keep a DNS name on `authenticationOptions.TargetHost` probably not very common but still possible to use different names.

src/IceRpc/Transports/IDuplexClientTransport.cs
externl: What about how to deal with an unknown name?

pepone:
> From a user point of view its really not clear how ServiceAddress and TransportAddress differ.

You will only deal with TransportAddress if you want to implement a transport. None of IceRPC examples will use TransportAddress.

Then for server vs service address my take is:
- A server address is the addressing information to reach a given IceRPC or Ice server
- Service address is the addressing information to reach a given Service (or Object using Ice terminology).

externl:
> > From a user point of view its really not clear how ServiceAddress and TransportAddress differ.
> 
> You will only deal with TransportAddress if you want to implement a transport. None of IceRPC examples will use TransportAddress.
> 
> Then for server vs service address my take is:
> 
> * A server address is the addressing information to reach a given IceRPC or Ice server
> * Service address is the addressing information to reach a given Service (or Object using Ice terminology).

Sorry I meant to write `ServerAddress` vs `TransportAddress`.

tests/IceRpc.Tests.Common/ServiceCollectionExtensions.cs
```diff
@@ -71,43 +71,74 @@ public static IServiceCollection AddSlicTransport(this IServiceCollection servic
...
+        var alpn = new SslApplicationProtocol(serverAddress.Protocol.Name);
+
+        return services
             .AddSingleton(provider =>
-                provider.GetRequiredService<IMultiplexedServerTransport>().Listen(
-                    new ServerAddress(serverAddressUri ?? new Uri("icerpc://colochost")),
-                    provider.GetService<IOptions<MultiplexedConnectionOptions>>()?.Value ?? new(),
-                    serverAuthenticationOptions: provider.GetService<SslServerAuthenticationOptions>()))
+            {
+                // Ensure ALPN is set on SSL options before any transport uses them.
```
bernardnormier, line 122: Should we do the same for duplex transport above?

src/IceRpc/Server.cs
pepone: So for the server we create auth options with ALPN if the users didn't provide them. And then we let listen fail? Which allows to remove IsSslRequired from the server side.

src/IceRpc/Transports/TransportAddress.cs
pepone: Removed

src/IceRpc/Transports/IDuplexServerTransport.cs
pepone: Removed

src/IceRpc/Server.cs
bernardnormier: On the server-side, the application always provide the auth options. IceRpc never creates empty server auth options.

src/IceRpc/Transports/Quic/QuicClientTransport.cs
```diff
@@ -15,7 +15,10 @@ namespace IceRpc.Transports.Quic;
 public class QuicClientTransport : IMultiplexedClientTransport
 {
     /// <inheritdoc/>
-    public string Name => "quic";
+    public string DefaultName => "quic";
+
+    /// <inheritdoc/>
+    public bool IsSslRequired(string? transportName) => true;
```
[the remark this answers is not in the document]
bernardnormier: I think we should indeed perform the transport name validation in `IsSslRequired`.

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
```diff
@@ -28,87 +33,43 @@ public TcpClientTransport()
 
     /// <inheritdoc/>
     public IDuplexConnection CreateConnection(
```
externl, line 35: Don't we also need a check in this function (`CreateConnection`) that if `TransportName` == `ssl` that `clientAuthenticationOptions != null`?

src/IceRpc/Transports/Quic/Internal/QuicMultiplexedListener.cs
```diff
@@ -40,25 +40,35 @@ internal class QuicMultiplexedListener : IListener<IMultiplexedConnection>
...
+        {
+            throw new ArgumentNullException(
+                nameof(serverAuthenticationOptions),
+                "The QUIC server transport requires the SSL server authentication options to be set.");
+        }
+        serverAuthenticationOptions = serverAuthenticationOptions.Clone();
 
-        authenticationOptions = authenticationOptions.Clone();
-        authenticationOptions.ApplicationProtocols ??= new List<SslApplicationProtocol> // Mandatory with Quic
+        if (serverAuthenticationOptions.ApplicationProtocols
```
bernardnormier: Is this really necessary? Presumably this is enforced by ListenAsync below.

src/IceRpc/ClientProtocolConnectionFactory.cs
```diff
@@ -77,24 +96,56 @@ public ClientProtocolConnectionFactory(
...
+                Host = serverAddress.Host,
+                Port = serverAddress.Port,
+                TransportName = serverAddress.Transport,
+                Params = serverAddress.Params
+            };
+
+        IProtocolConnection connection;
+        if (serverAddress.Protocol == Protocol.Ice)
+        {
+            SslClientAuthenticationOptions? authenticationOptions = _iceClientAuthenticationOptions?.Clone();
```
bernardnormier, line 110: Do transports like Quic and Tcp actually modify the authentication options they are given?

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
```diff
@@ -28,87 +33,43 @@ public TcpClientTransport()
...
         SslClientAuthenticationOptions? clientAuthenticationOptions)
     {
-        if (serverAddress.Transport is string transport && !IsValidTransportName(transport, serverAddress.Protocol))
+        // "ssl" is only accepted for the Ice protocol, identified by the ALPN.
+        if (transportAddress.TransportName == "ssl")
         {
-            throw new NotSupportedException(
-                $"The Tcp client transport does not support server addresses with transport '{transport}'.");
+            if (clientAuthenticationOptions?.ApplicationProtocols
+                is not List<SslApplicationProtocol> alpnProtocols ||
```
bernardnormier: This doesn't fit on a single line?

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
```diff
@@ -28,87 +33,43 @@ public TcpClientTransport()
...
+                alpnProtocols.Count != 1 ||
+                alpnProtocols[0] != new SslApplicationProtocol("ice"))
+            {
+                throw new NotSupportedException(
```
bernardnormier, line 47: For me, the main "pro" of this check is that "tcp" and "ssl" are really the same transport. In particular tcp + auth options is completely identical to ssl + auth options.

If we tolerate the two names for all protocols, in particular icerpc, we would give the incorrect impression that Slic-over-tcp and Slic-over-ssl are valid and different, when they are really the same.

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
```diff
@@ -28,87 +33,43 @@ public TcpClientTransport()
...
-        if (serverAddress.Transport is null)
+        if (transportAddress.Params.Count > 0)
         {
-            serverAddress = serverAddress with { Transport = Name };
+            throw new ArgumentException(
+                "The transport address contains parameters that are not valid for the TCP client transport.",
+                nameof(transportAddress));
         }
 
         SslClientAuthenticationOptions? authenticationOptions = clientAuthenticationOptions?.Clone() ??
```
bernardnormier, line 63: We should review all these Clones.

src/IceRpc/Transports/Tcp/TcpServerTransport.cs
```diff
@@ -28,37 +25,34 @@ public TcpServerTransport()
...
         DuplexConnectionOptions options,
         SslServerAuthenticationOptions? serverAuthenticationOptions)
     {
-        if (serverAddress.Transport is string transport && !IsValidTransportName(transport, serverAddress.Protocol))
+        // "ssl" is only accepted for the Ice protocol, identified by the ALPN.
+        if (transportAddress.TransportName == "ssl")
         {
-            throw new NotSupportedException(
-                $"The Tcp server transport does not support server addresses with transport '{transport}'.");
+            if (serverAuthenticationOptions?.ApplicationProtocols
```
bernardnormier: A null serverAuthenticationOptions should be rejected here, like in the previous code.

src/IceRpc/Transports/Tcp/TcpServerTransport.cs
```diff
@@ -28,37 +25,34 @@ public TcpServerTransport()
 
     /// <inheritdoc/>
     public IListener<IDuplexConnection> Listen(
-        ServerAddress serverAddress,
+        TransportAddress transportAddress,
         DuplexConnectionOptions options,
         SslServerAuthenticationOptions? serverAuthenticationOptions)
     {
-        if (serverAddress.Transport is string transport && !IsValidTransportName(transport, serverAddress.Protocol))
+        // "ssl" is only accepted for the Ice protocol, identified by the ALPN.
```
externl, line 32: I would prefer to check the supported names. Then next check the configuration. Not mix them together.

src/IceRpc/Transports/Quic/Internal/QuicMultiplexedListener.cs
pepone: I will check

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
bernardnormier: Can wait until a follow-up PR.

src/IceRpc/ClientProtocolConnectionFactory.cs
bernardnormier: Related to the Clone review, for a follow-up PR.

src/IceRpc/Transports/Tcp/TcpServerTransport.cs
pepone: Here the supported names depend on the configuration.

Issue #4350, opened 2026-03-25
Change default mapping for Slice to internal, add cs:public attribute


bernardnormier:
The Slice code generator currently maps everything to public, e.g. a Slice struct is mapped by default to a public C# struct. If you'd rather get an internal C# struct, you add the `[cs:internal]` to your Slice struct. 

This default mapping is a legacy of seeing Slice-generated APIs as public APIs, when they are usually (and should be) internal APIs. 

A good example of the recommended domain-separation approach is the ZeroC.Slice.Symbols assembly, where the generated code is all internal and the corresponding public APIs are hand-written.

Mapping to public remains useful when you package up your generated code in a shared assembly, and reference this assembly in other assemblies. But it's not the most common and should not be the default.

## Proposal

Change the default mapping to internal. Replace `cs:internal` by `cs:public` to allow you to mark the generated C# code as public instead of internal.

## Ice Slice to C# mapping

We currently also map Ice Slice to public C# constructs by default. I propose to keep this mapping unchanged (like in Ice); mapping Ice classes and exceptions to internal classes isn't great as they need to be activated at runtime. Also, C# exceptions are supposed to be public.

externl:
Sounds good to me!

PR #4354, opened 2026-03-26
Replace cs::internal by cs::public

bernardnormier:
This PR replaces cs::internal by cs::public (with the correct semantics changes).

Implements #4350.

### DIFF
--- a/slice/IceRpc/CompressionFormat.slice
+++ b/slice/IceRpc/CompressionFormat.slice
@@ -3,6 +3,7 @@
 module IceRpc
 
 /// The compression format of a payload.
+[cs::public]
 unchecked enum CompressionFormat : uint8 {
     /// Reserved value which should not be encoded.
     NotCompressed = 0
--- a/slice/IceRpc/Internal/IceRpcDefinitions.slice
+++ b/slice/IceRpc/Internal/IceRpcDefinitions.slice
@@ -5,7 +5,6 @@ module IceRpc::Internal
 // These definitions help with the encoding of icerpc frames.
 
 /// Each icerpc control frame has a type identified by this enumeration.
-[cs::internal]
 enum IceRpcControlFrameType : uint8 {
     /// The Settings frame is sent by each peer on connection establishment to exchange settings.
     Settings = 0
@@ -19,7 +18,6 @@ enum IceRpcControlFrameType : uint8 {
 /// - a request header size (varuint62)
 /// - a request header (below)
 /// - a request payload
-[cs::internal]
 [cs::readonly]
 compact struct IceRpcRequestHeader {
     path: string
@@ -31,7 +29,6 @@ compact struct IceRpcRequestHeader {
 /// - a response header size (varuint62)
 /// - a response header (below)
 /// - a response payload
-[cs::internal]
 [cs::readonly]
 compact struct IceRpcResponseHeader {
     statusCode: StatusCode
@@ -40,14 +37,12 @@ compact struct IceRpcResponseHeader {
 }
 
 /// The settings frame is sent by each peer on connection establishment to exchange settings.
-[cs::internal]
 [cs::readonly]
 compact struct IceRpcSettings {
     value: Dictionary<IceRpcSettingKey, varuint62>
 }
 
 /// The keys for Settings entries.
-[cs::internal]
 unchecked enum IceRpcSettingKey : varuint62 {
     MaxHeaderSize = 0
 }
@@ -55,7 +50,6 @@ unchecked enum IceRpcSettingKey : varuint62 {
 /// The GoAway frame is sent on connection shutdown to notify the peer that it shouldn't send additional requests and to
 /// provide two stream IDs. Requests carried by streams with IDs greater or equal to these stream IDs were not accepted
 /// or otherwise processed, and as a result can be safely retried.
-[cs::internal]
 [cs::readonly]
 compact struct IceRpcGoAway {
     bidirectionalStreamId: varuint62
--- a/src/IceRpc.Templates/Templates/IceRpc-Protobuf-DI-Server/ServerHostedService.cs
+++ b/src/IceRpc.Templates/Templates/IceRpc-Protobuf-DI-Server/ServerHostedService.cs
@@ -4,7 +4,11 @@ using Microsoft.Extensions.Hosting;
 namespace IceRpc_Protobuf_DI_Server;
 
 /// <summary>The server hosted service is ran and managed by the .NET Generic Host.</summary>
-public class ServerHostedService : IHostedService
+[System.Diagnostics.CodeAnalysis.SuppressMessage(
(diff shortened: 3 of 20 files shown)


slice/IceRpc/Internal/IceRpcDefinitions.slice
```diff
@@ -5,7 +5,6 @@ module IceRpc::Internal
 // These definitions help with the encoding of icerpc frames.
 
 /// Each icerpc control frame has a type identified by this enumeration.
-[cs::internal]
```
bernardnormier, line 8: Since I updated many vendored files, I'll need to update/sync their source as well (first?).

tools/slicec-cs/src/generators/struct_generator.rs
```diff
@@ -41,38 +41,41 @@ pub fn generate_struct(struct_def: &Struct) -> CodeBlock {
...
-        struct_def.access_modifier(),
-        "",
-        &escaped_identifier,
-        FunctionType::BlockBody,
-    );
-
-    if has_required_field {
-        main_constructor.add_sets_required_members_attribute();
-    }
+    // We don't need to "override" the implicit public parameterless constructor.
```
bernardnormier, line 44: And we wanted to override it, it would need to public no matter what.

tools/slicec-cs/src/generators/struct_generator.rs
```diff
@@ -41,38 +41,41 @@ pub fn generate_struct(struct_def: &Struct) -> CodeBlock {
...
+    if !fields.is_empty() {
+        let mut main_constructor = FunctionBuilder::new(
```
bernardnormier, line 46: Only a formatting change inside this block.

tests/ZeroC.Slice.Generator.Tests/TaggedTests.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace ZeroC.Slice.Generator.Tests;
 
-public class TaggedTests
+internal class TaggedTests
```
bernardnormier, line 9: Switch to internal because some parameters (of the public test methods) are now internal.

This works fine with NUnit.

tests/ZeroC.Slice.Generator.Tests/StructTests.slice
```diff
@@ -20,6 +20,7 @@ compact struct MyCompactStruct {
     j: int32
 }
 
+[cs::public]
```
bernardnormier, line 23: We test it using reflection and how test code apparently requires a public struct or at least public fields.

slice/IceRpc/CompressionFormat.slice
```diff
@@ -3,6 +3,7 @@
 module IceRpc
 
 /// The compression format of a payload.
+[cs::public]
```
bernardnormier, line 6: All these public Slice definitions are meant to be public in C# to, and exported by the IceRpc assembly.

Note that custom types don't need to be `[cs::public]`: the corresponding code is provided, not generated.

bernardnormier (commented):
Although not obvious on this PR, we generate too much "public" when mapping interfaces. slice2cs --icerpc is better in that respect; we should improve this mapping once slicec-cs is replaced by a C# code generator.

src/IceRpc.Templates/Templates/IceRpc-Slice-DI-Client/ClientHostedService.cs
```diff
@@ -4,7 +4,11 @@
 namespace IceRpc_Slice_DI_Client;
 
 /// <summary>The hosted client service is ran and managed by the .NET Generic Host.</summary>
-public class ClientHostedService : BackgroundService
+[System.Diagnostics.CodeAnalysis.SuppressMessage(
```
bernardnormier, line 7: This is now necessary because the IGreeter constructor parameter is now internal (see below).

This internal and diagnostics suppression matches what we already did in the examples.

src/IceRpc.Templates/Templates/IceRpc-Protobuf-DI-Server/ServerHostedService.cs
```diff
@@ -4,7 +4,11 @@
 namespace IceRpc_Protobuf_DI_Server;
 
 /// <summary>The server hosted service is ran and managed by the .NET Generic Host.</summary>
-public class ServerHostedService : IHostedService
+[System.Diagnostics.CodeAnalysis.SuppressMessage(
+    "Performance",
+    "CA1812:Avoid uninstantiated internal classes",
+    Justification = "This class is instantiated dynamically by the dependency injection container.")]
+internal class ServerHostedService : IHostedService
```
bernardnormier, line 11: Only for consistency with the Protobuf examples and the Slice templates.

slice/IceRpc/Internal/IceRpcDefinitions.slice
InsertCreativityHere: It shouldn't make a big difference. But if you merge this without making the change in `icerpc-slice`,
you'll have to open another PR for `icerpc-csharp` to update the commit-id in the `slice.toml`.

tests/ZeroC.Slice.Generator.Tests/StructTests.slice
InsertCreativityHere: Probably worth adding a comment saying this.
So we don't wonder why this random test is `public` 3 years from now.

tools/slicec-cs/src/generators/struct_generator.rs
```diff
@@ -41,38 +41,41 @@ pub fn generate_struct(struct_def: &Struct) -> CodeBlock {
...
+                field.formatted_doc_comment_summary(),
+            );
         }
-        code
-    });
-    builder.add_block(main_constructor.build());
+        main_constructor.set_body({
+            let mut code = CodeBlock::default();
+            for field in &fields {
+                writeln!(code, "this.{} = {};", field.field_name(), field.parameter_name(),);
```
InsertCreativityHere: You didn't write this code, but it's clearly bogus and should be fixed.
Sometimes `rustfmt` gets confused when moving things around and inserts these spurious commas:

```suggestion
                writeln!(code, "this.{} = {};", field.field_name(), field.parameter_name());
```

InsertCreativityHere (approved):
Looks good to me!

pepone (approved):
Looks good we have to update the new `ZeroC.Slice.Generator` too, until we switch from slicec-cs to slicec will be good to minimize the generated code changes.

externl (approved):

tests/ZeroC.Slice.Generator.Tests/StructTests.slice
bernardnormier: You just remove cs::public and you'll figure out why :).

PR #4355, opened 2026-03-27
Remove check for QUIC in Default property

bernardnormier:
This check was incorrect since it affects applications that don't use QUIC or multiplexed transports in general, such as the Ice interop tests.

Note that the QUIC transport implementation already performs this check  in Listen and ConnectAsync.

### DIFF
--- a/src/IceRpc/Transports/IMultiplexedClientTransport.cs
+++ b/src/IceRpc/Transports/IMultiplexedClientTransport.cs
@@ -1,7 +1,6 @@
 // Copyright (c) ZeroC, Inc.
 
 using IceRpc.Transports.Quic;
-using System.Net.Quic;
 using System.Net.Security;
 using System.Runtime.Versioning;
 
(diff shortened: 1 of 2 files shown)


externl (approved):

src/IceRpc/Transports/IMultiplexedClientTransport.cs
```diff
@@ -18,15 +17,10 @@ public static IMultiplexedClientTransport Default
         {
             if (OperatingSystem.IsLinux() || OperatingSystem.IsMacOS() || OperatingSystem.IsWindows())
```
pepone, line 18: Why do we need to check the OS here, if we include all .NET supported OS?

src/IceRpc/Transports/IMultiplexedClientTransport.cs
```diff
@@ -18,15 +17,10 @@ public static IMultiplexedClientTransport Default
...
-                {
-                    return _quicClientTransport;
-                }
-                throw new NotSupportedException(
-                    "The default QUIC client transport is not available on this system. Please review the Platform Dependencies for QUIC in the .NET documentation.");
+                return _quicClientTransport;
             }
             throw new PlatformNotSupportedException(
-                "The default QUIC client transport is not supported on this platform.");
+                "The default multiplexed client transport, QUIC, is only available on Linux, macOS, and Windows.");
```
pepone, line 23: Should this be "Quic" which is the standard syntax in .NET? Also consistent with "Tcp" in the tcp transport.

src/IceRpc/Transports/IMultiplexedClientTransport.cs
bernardnormier: It's because QUIC is only available on the 3 platforms:

```cs
[SupportedOSPlatform("linux")]
[SupportedOSPlatform("macos")]
[SupportedOSPlatform("windows")]
```

I wish we could remove all that, but we can't because of this constraint in System.Net.Quic.

src/IceRpc/Transports/IMultiplexedClientTransport.cs
bernardnormier: We should generally use the correct capitalization in error message, e.g QUIC, TCP, UDP. 

Quic and Tcp are for C# identifiers, not messages.

pepone (approved):

PR #4356, opened 2026-03-28
Add Generator to handle encoding and decoding of Slice compiler data

pepone:
This PR sync Slice compiler definitions with icerpc/slicec.

The PR also adds a base Generator in ZeroC.Slice.Symbols, which deals with the conversion from Slice encoded definitions to the C# symbols model, and with encoding generator responses. This allow to keep all Slice compiler definitions internal in ZerocC.Slice.Compiler.

### DIFF
--- a/src/ZeroC.Slice.Generator/GeneratorDriver.cs
+++ b/src/ZeroC.Slice.Generator/GeneratorDriver.cs
@@ -8,8 +8,7 @@ using ZeroC.Slice.Symbols;
 
 namespace ZeroC.Slice.Generator;
 
-/// <summary>Shared pipeline for Slice code generators. Handles the stdin/stdout protocol, attribute validation,
-/// preamble generation, and response encoding.</summary>
+/// <summary>Shared pipeline for Slice code generators.</summary>
 internal static class GeneratorDriver
 {
     /// <summary>Runs the generator pipeline: reads stdin, decodes symbols, generates code, writes stdout.</summary>


src/ZeroC.Slice.Symbols/SymbolConverter.cs
```diff
@@ -9,7 +9,7 @@ public sealed class SymbolConverter
 {
     /// <summary>Converts source files into rich symbol types with all TypeRefs resolved. Reference files are used for
     /// type resolution but are not included in the output.</summary>
-    public static ImmutableList<SliceFile> ConvertFiles(
+    internal static ImmutableList<SliceFile> ConvertFiles(
```
bernardnormier, line 12: Should the entire class become internal?

src/ZeroC.Slice.Generator/GeneratorDriver.cs
```diff
@@ -25,109 +21,85 @@ internal static async Task RunAsync(
...
-            ?? assembly.GetName().Version?.ToString()
-            ?? "unknown";
-
-        // Generate code for each source file, skipping generation if there are validation errors.
-        var generatedFiles = new List<Compiler.GeneratedFile>();
-        if (!diagnostics.Any(d => d.Level == Compiler.DiagnosticLevel.Error))
+        using Stream stdout = Console.OpenStandardOutput();
+        var writer = PipeWriter.Create(stdout);
+
+        await Symbols.Generator.RunAsync(reader, writer, symbolFiles =>
```
bernardnormier: A static method would be more elegant than this big lambda expression.

src/ZeroC.Slice.Symbols/Generator.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// symbols, invokes <paramref name="transform"/> with the converted source files, and writes the Slice-encoded
+    /// response to <paramref name="output"/>.</summary>
+    /// <param name="input">The pipe reader to read the Slice-encoded request from.</param>
+    /// <param name="output">The pipe writer to write the Slice-encoded response to.</param>
+    /// <param name="transform">A function that receives the converted source files and returns the generator
+    /// response.</param>
+    public static async Task RunAsync(
+        PipeReader input,
+        PipeWriter output,
+        Func<ImmutableList<SliceFile>, Task<GeneratorResponse>> transform)
```
bernardnormier: Maybe we should start with a sync Func (returns GeneratorResponse) and only add async (returns Task) when we have a use-case for async generation.

src/ZeroC.Slice.Symbols/SymbolConverter.cs
```diff
@@ -243,16 +244,19 @@ static IEnumerable<string> InterfaceDeps(Compiler.Interface i)
                 }
             }
 
-            static IEnumerable<string> EnumDeps(Compiler.Enum e) =>
-                e.Enumerators.SelectMany(en => en.Fields.Select(f => f.DataType.TypeId));
+            static IEnumerable<string> BasicEnumDeps(Compiler.BasicEnum e) => [];
```
bernardnormier, line 247: This whole code with Deps needs more comments and probably better names than Deps. For another PR.

src/ZeroC.Slice.Generator/GeneratorDriver.cs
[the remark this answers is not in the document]
pepone: This is fine as is, because both have `ZeroC.Slice` prefix.

src/ZeroC.Slice.Symbols/Generator.cs
bernardnormier: One use-case could be build telemetry, where transform will establish a connection and send a oneway request.

Issue #4307, opened 2026-03-14
Quic Stream_read_returns_canceled_read_result_on_cancel_pending_read hang on macos


bernardnormier:
This test was hanging/crashed here (on main)
https://github.com/icerpc/icerpc-csharp/actions/runs/23097413673/job/67092220240

```
  Passed IceRpc:Tests:Transports:QuicStreamConformanceTests:Stream_read_returns_canceled_read_result_after_cancel_pending_read [259 ms]
Blame: Dumping 18861 - dotnet
[createdump] Gathering state for process 18861 
[createdump] Writing full dump to file /var/folders/7k/p5jlfs4x2vbgq5sbjd_s8vyr0000gn/T/8a620ce6-378c-434e-aba9-e0bee41a12b0/dotnet_18861_20260314T221331_hangdump.dmp
[createdump] Written 6055383416 bytes (369591 pages) to core file
[createdump] Target process is alive
[createdump] Dump successfully written in 42710ms
The active test run was aborted. Reason: Test host process crashed
Data collector 'Blame' message: The specified inactivity time of 60 seconds has elapsed. Collecting hang dumps from testhost and its child processes.
Results File: /Users/runner/work/icerpc-csharp/icerpc-csharp/TestResults/_sjc22-bm206-2316f2c5-cece-4401-8bfb-a1b371d7d9bb-7A5B1BE3AEDB_2026-03-14_22_12_06[2].trx

Test Run Aborted.
Total tests: Unknown
     Passed: 108
 Total time: 2.1654 Minutes

The active Test Run was aborted because the host process exited unexpectedly. Please inspect the call stack above, if available, to get more information about where the exception originated from.
The test running when the crash occurred: 
IceRpc.Tests.Transports.QuicStreamConformanceTests.Stream_read_returns_canceled_read_result_on_cancel_pending_read

This test may, or may not be the source of the crash.
     1>Project "/Users/runner/work/icerpc-csharp/icerpc-csharp/IceRpc.slnx" (1) is building "/Users/runner/work/icerpc-csharp/icerpc-csharp/tests/IceRpc.Quic.Tests/IceRpc.Quic.Tests.csproj" (33) on node 2 (VSTest target(s)).
    33>_VSTestConsole:
         MSB4181: The "VSTestTask" task returned false but did not log an error.
[...]

PR #4360, opened 2026-03-30
Improve Stream_read_returns_canceled_read_result_on_cancel_pending_read

bernardnormier:
This PR improves Stream_read_returns_canceled_read_result_on_cancel_pending_read, primarily by adding a CancelAfter attribute.

See #4307

### DIFF
--- a/tests/IceRpc.Conformance.Tests/Transports/MultiplexedStreamConformanceTests.cs
+++ b/tests/IceRpc.Conformance.Tests/Transports/MultiplexedStreamConformanceTests.cs
@@ -564,43 +564,41 @@ public abstract class MultiplexedStreamConformanceTests
     }
 
     [Test]
-    public async Task Stream_read_returns_canceled_read_result_on_cancel_pending_read()
+    [CancelAfter(2000)] // Cancel after 2000 milliseconds
+    public async Task Stream_read_returns_canceled_read_result_on_cancel_pending_read(
+        CancellationToken cancellationToken)
     {
         // Arrange
         await using ServiceProvider provider = CreateServiceCollection().BuildServiceProvider(validateScopes: true);
         var clientServerConnection = provider.GetRequiredService<ClientServerMultiplexedConnection>();
-        await clientServerConnection.AcceptAndConnectAsync();
-        using var sut = await clientServerConnection.CreateAndAcceptStreamAsync();
+        await clientServerConnection.AcceptAndConnectAsync(cancellationToken);
+        using var sut = await clientServerConnection.CreateAndAcceptStreamAsync(cancellationToken: cancellationToken);
 
         // Act
-        ValueTask<ReadResult> readTask = sut.Local.Input.ReadAsync();
+        ValueTask<ReadResult> readTask = sut.Local.Input.ReadAsync(cancellationToken);
         sut.Local.Input.CancelPendingRead();
 
         // Assert
-        ReadResult readResult1 = await readTask;
+        ReadResult readResult1 = default;
+        Assert.That(async () => readResult1 = await readTask, Throws.Nothing);
+        Assert.That(readResult1.IsCanceled, Is.True);
+        Assert.That(readResult1.IsCompleted, Is.False);
         sut.Local.Input.AdvanceTo(readResult1.Buffer.Start);
 
-        Assert.That(async () => await sut.Remote.Output.WriteAsync(_oneBytePayload), Throws.Nothing);
+        Assert.That(async () => await sut.Remote.Output.WriteAsync(_oneBytePayload, cancellationToken), Throws.Nothing);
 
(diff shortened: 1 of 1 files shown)


tests/IceRpc.Conformance.Tests/Transports/MultiplexedStreamConformanceTests.cs
```diff
@@ -564,44 +564,42 @@ public async Task Stream_read_returns_canceled_read_result_after_cancel_pending_
     }
 
     [Test]
-    public async Task Stream_read_returns_canceled_read_result_on_cancel_pending_read()
+    [CancelAfter(2000)] // Cancel after 2000 milliseconds
```
bernardnormier, line 567: Better cancel test run than hang forever.

tests/IceRpc.Conformance.Tests/Transports/MultiplexedStreamConformanceTests.cs
```diff
@@ -564,44 +564,42 @@ public async Task Stream_read_returns_canceled_read_result_after_cancel_pending_
...
         // Act
-        ValueTask<ReadResult> readTask = sut.Local.Input.ReadAsync();
+        ValueTask<ReadResult> readTask = sut.Local.Input.ReadAsync(cancellationToken);
         sut.Local.Input.CancelPendingRead();
 
         // Assert
-        ReadResult readResult1 = await readTask;
+        ReadResult readResult1 = default;
+        Assert.That(async () => readResult1 = await readTask, Throws.Nothing);
+        Assert.That(readResult1.IsCanceled, Is.True);
```
bernardnormier, line 584: Moved readResult1 asserts here, in case it has unexpected values that make the next ReadAsync hang with QUIC.

tests/IceRpc.Conformance.Tests/Transports/MultiplexedStreamConformanceTests.cs
```diff
@@ -564,44 +564,41 @@ public async Task Stream_read_returns_canceled_read_result_after_cancel_pending_
...
 
         // Act
-        ValueTask<ReadResult> readTask = sut.Local.Input.ReadAsync();
+        ValueTask<ReadResult> readTask = sut.Local.Input.ReadAsync(cancellationToken);
         sut.Local.Input.CancelPendingRead();
 
         // Assert
-        ReadResult readResult1 = await readTask;
+        ReadResult readResult1 = default;
+        Assert.That(async () => readResult1 = await readTask, Throws.Nothing);
```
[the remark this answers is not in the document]
bernardnormier: Claude's response:

Neither comment is a legitimate concern. Both are based on the same incorrect premise.

Comment 1 claims Assert.That(async () => readResult1 = await readTask, Throws.Nothing) won't compile because the assignment expression produces a ReadResult value, making the lambda Func<Task<ReadResult>> instead of Func<Task>.

Comment 2 claims Assert.That(async () => await sut.Remote.Output.WriteAsync(...), Throws.Nothing) won't compile because WriteAsync returns ValueTask<FlushResult>.

Both are wrong. In C#, an expression-bodied async lambda like async () => someExpression can target Func<Task> (NUnit's AsyncTestDelegate). The expression result is simply discarded. This is no different from writing await someValueTask; as a statement — the return value doesn't force the lambda's return type.

Comment 2 is especially easy to dismiss: the old code already had the exact same pattern without the cancellationToken parameter, and it compiled fine:
```
// old code (line 588 before your PR)
Assert.That(async () => await sut.Remote.Output.WriteAsync(_oneBytePayload), Throws.Nothing);
```

Your code compiles and runs — as you noted — because there is no issue.

externl (approved):

tests/IceRpc.Conformance.Tests/Transports/MultiplexedStreamConformanceTests.cs
pepone: Not clear this is the correct fix, we already set the blame hang timeout for this.

https://github.com/bernardnormier/icerpc-csharp/blob/6ec345edefe78f5cbfce76fc9ecad4bb2ca3075e/.github/actions/test/action.yaml#L19

Didn't the hang test uploaded the dumps? With this new cancel token if the test is canceled we have no idea why it was hanging.

bernardnormier: I assume we'll get a stack trace that pinpoints where the hang occurred.

pepone: This can give a stack of what task we were waiting when the token gets canceled. But the blame hang timeout dumps all threads.

bernardnormier:
This does not help with #4307

PR #4361, opened 2026-03-31
Review SSL authentication options cloning

### LINKED ISSUE #4357: Review authentication options clones
https://github.com/icerpc/icerpc-csharp/pull/4353#discussion_r3002792624

We need to clarify what the cloning policy should be here.

I think we should always clone options pass to Server and ClientConnection, ConnectionCache. This way if the application modifies them it doesn't has unexpected side effects.

The transports can either do the same, or only clone on write. For the transport clone on write will be my preference.

### LINKED ISSUE #4358: clientAuthenticationOptions and ssl transport
Don't we also need a check in this function (`CreateConnection`) that if `TransportName` == `ssl` that `clientAuthenticationOptions != null`?

_Originally posted by @externl in https://github.com/icerpc/icerpc-csharp/pull/4353#discussion_r3002733481_

I forget to fix that one before merging the PR.

pepone:
Fix #4357
Fix #4358

### DIFF
--- a/src/IceRpc/ClientProtocolConnectionFactory.cs
+++ b/src/IceRpc/ClientProtocolConnectionFactory.cs
@@ -78,10 +78,10 @@ public sealed class ClientProtocolConnectionFactory : IClientProtocolConnectionF
         // Clone and set ALPN for each protocol.
         if (clientAuthenticationOptions is not null)
         {
-            _iceClientAuthenticationOptions = clientAuthenticationOptions.Clone();
+            _iceClientAuthenticationOptions = clientAuthenticationOptions.ShallowClone();
             _iceClientAuthenticationOptions.ApplicationProtocols = [Protocol.Ice.AlpnProtocol];
 
-            _iceRpcClientAuthenticationOptions = clientAuthenticationOptions.Clone();
+            _iceRpcClientAuthenticationOptions = clientAuthenticationOptions.ShallowClone();
             _iceRpcClientAuthenticationOptions.ApplicationProtocols = [Protocol.IceRpc.AlpnProtocol];
         }
 
--- a/src/IceRpc/Server.cs
+++ b/src/IceRpc/Server.cs
@@ -114,7 +114,7 @@ public sealed class Server : IAsyncDisposable
             SslServerAuthenticationOptions? serverAuthenticationOptions = options.ServerAuthenticationOptions;
             if (serverAuthenticationOptions is not null)
             {
-                serverAuthenticationOptions = serverAuthenticationOptions.Clone();
+                serverAuthenticationOptions = serverAuthenticationOptions.ShallowClone();
                 serverAuthenticationOptions.ApplicationProtocols = [_serverAddress.Protocol.AlpnProtocol];
             }
 
--- a/src/IceRpc/Transports/Quic/QuicClientTransport.cs
+++ b/src/IceRpc/Transports/Quic/QuicClientTransport.cs
@@ -72,7 +72,7 @@ public class QuicClientTransport : IMultiplexedClientTransport
                 nameof(clientAuthenticationOptions));
         }
 
-        clientAuthenticationOptions = clientAuthenticationOptions.Clone();
+        clientAuthenticationOptions = clientAuthenticationOptions.ShallowClone();
         clientAuthenticationOptions.TargetHost ??= transportAddress.Host;
 
         EndPoint endPoint = IPAddress.TryParse(transportAddress.Host, out IPAddress? ipAddress) ?
--- a/src/IceRpc/Transports/SslAuthenticationOptionsExtensions.cs
+++ b/src/IceRpc/Transports/SslAuthenticationOptionsExtensions.cs
@@ -11,7 +11,7 @@ public static class SslAuthenticationOptionsExtensions
     /// <summary>Makes a shallow copy of an SSL client authentication options.</summary>
     /// <param name="value">The options to copy.</param>
     /// <returns>The shallow copy.</returns>
-    public static SslClientAuthenticationOptions Clone(this SslClientAuthenticationOptions value) =>
+    public static SslClientAuthenticationOptions ShallowClone(this SslClientAuthenticationOptions value) =>
         new()
         {
             AllowRenegotiation = value.AllowRenegotiation,
@@ -36,7 +36,7 @@ public static class SslAuthenticationOptionsExtensions
     /// <summary>Makes a shallow copy of an SSL server authentication options.</summary>
     /// <param name="value">The options to copy.</param>
     /// <returns>The shallow copy.</returns>
-    public static SslServerAuthenticationOptions Clone(this SslServerAuthenticationOptions value) =>
+    public static SslServerAuthenticationOptions ShallowClone(this SslServerAuthenticationOptions value) =>
         new()
         {
             AllowRenegotiation = value.AllowRenegotiation,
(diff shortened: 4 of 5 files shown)


src/IceRpc/Transports/Quic/QuicClientTransport.cs
```diff
@@ -73,15 +72,20 @@ public IMultiplexedConnection CreateConnection(
                 nameof(clientAuthenticationOptions));
         }
 
-        clientAuthenticationOptions.TargetHost ??= transportAddress.Host;
+        SslClientAuthenticationOptions? authenticationOptions = clientAuthenticationOptions;
+        if (authenticationOptions.TargetHost is null)
```
pepone: This is the only clone remaining. Here the authentication options can be used for multiple hosts, and makes sense to clone and set when not null. We do the same in the TCP client transport.

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
```diff
@@ -40,12 +40,17 @@ public IDuplexConnection CreateConnection(
         // "ssl" is only accepted for the Ice protocol, identified by the ALPN.
         if (transportAddress.TransportName == "ssl")
         {
-            if (clientAuthenticationOptions?.ApplicationProtocols is not List<SslApplicationProtocol> alpnProtocols ||
+            if (clientAuthenticationOptions is null)
```
pepone, line 43: This is the fix for #4358

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
```diff
@@ -40,12 +40,17 @@ public IDuplexConnection CreateConnection(
...
+                    nameof(clientAuthenticationOptions),
+                    "The SSL client transport requires the SSL client authentication options to be set.");
+            }
+            else if (clientAuthenticationOptions.ApplicationProtocols is not List<SslApplicationProtocol> alpnProtocols ||
                 alpnProtocols.Count != 1 ||
                 alpnProtocols[0] != new SslApplicationProtocol("ice"))
             {
-                throw new NotSupportedException(
-                    "The 'ssl' transport name is only supported with the Ice protocol.");
+                throw new NotSupportedException("The 'ssl' transport name is only supported with the Ice protocol.");
```
bernardnormier: I think Ice should be lowercase here. It's the ice and icerpc protocols.

src/IceRpc/Server.cs
```diff
@@ -111,8 +111,7 @@ public Server(
         {
             IConnectorListener listener;
 
-            SslServerAuthenticationOptions? serverAuthenticationOptions = options.ServerAuthenticationOptions?.Clone();
-            serverAuthenticationOptions?.ApplicationProtocols = [_serverAddress.Protocol.AlpnProtocol];
+            options.ServerAuthenticationOptions?.ApplicationProtocols = [_serverAddress.Protocol.AlpnProtocol];
```
externl: Seems like we should do a clone of this if we're going to mutate it.

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
```diff
@@ -60,9 +65,12 @@ public IDuplexConnection CreateConnection(
                 nameof(transportAddress));
         }
 
-        SslClientAuthenticationOptions? authenticationOptions = clientAuthenticationOptions?.Clone() ??
-            (transportAddress.TransportName == "ssl" ? new SslClientAuthenticationOptions() : null);
-        authenticationOptions?.TargetHost ??= transportAddress.Host;
+        SslClientAuthenticationOptions? authenticationOptions = clientAuthenticationOptions;
```
bernardnormier: I don't like this update. I'd rather clone all the time like we did before, not in some corner case. And I find 
`SslClientAuthenticationOptions? authenticationOptions = clientAuthenticationOptions;` error prone: it's easy to mix them up.

src/IceRpc/Transports/Quic/QuicClientTransport.cs
```diff
@@ -73,15 +72,20 @@ public IMultiplexedConnection CreateConnection(
                 nameof(clientAuthenticationOptions));
         }
 
-        clientAuthenticationOptions.TargetHost ??= transportAddress.Host;
+        SslClientAuthenticationOptions? authenticationOptions = clientAuthenticationOptions;
```
bernardnormier: Same as for the TCP code: I'd rather Clone all the time.

src/IceRpc/ClientProtocolConnectionFactory.cs
```diff
@@ -107,7 +107,7 @@ public IProtocolConnection CreateConnection(ServerAddress serverAddress)
         IProtocolConnection connection;
         if (serverAddress.Protocol == Protocol.Ice)
         {
-            SslClientAuthenticationOptions? authenticationOptions = _iceClientAuthenticationOptions?.Clone();
+            SslClientAuthenticationOptions? authenticationOptions = _iceClientAuthenticationOptions;
```
bernardnormier: For the variable name, I'd prefer clientAuthenticationOptions.

src/IceRpc/ClientProtocolConnectionFactory.cs
```diff
@@ -130,7 +130,7 @@ public IProtocolConnection CreateConnection(ServerAddress serverAddress)
         }
         else
         {
-            SslClientAuthenticationOptions? authenticationOptions = _iceRpcClientAuthenticationOptions?.Clone();
+            SslClientAuthenticationOptions? authenticationOptions = _iceRpcClientAuthenticationOptions;
```
bernardnormier: Ditto

src/IceRpc/Server.cs
bernardnormier: I'd rather keep the old code, and allow the user to reuse these ServerAuthenticationOptions for multiple servers.

src/IceRpc/Transports/Tcp/TcpClientTransport.cs
externl: I'd prefer to not make overly defensive clones all over the place personally.

externl: Claude's assessment of dealing with these options objects, referencing how they're used in the .NET Core and Kestrel. 


> All three follow the same rule: the options object is treated as a long-lived cached value; clone it only at the point you need to diverge from the cached version. Nobody clones defensively on receipt "just in case."

pepone: I updated https://github.com/icerpc/icerpc-csharp/issues/4357 with the description of what I think is best.

pepone: I keep the simpler approach of cloning all the time in the client transport.

src/IceRpc/Transports/SslAuthenticationOptionsExtensions.cs
bernardnormier: I would rename these methods ShallowClone, like in the dotnet runtime.

src/IceRpc/Server.cs
```diff
@@ -111,8 +111,12 @@ public Server(
         {
             IConnectorListener listener;
 
-            SslServerAuthenticationOptions? serverAuthenticationOptions = options.ServerAuthenticationOptions?.Clone();
-            serverAuthenticationOptions?.ApplicationProtocols = [_serverAddress.Protocol.AlpnProtocol];
+            SslServerAuthenticationOptions? serverAuthenticationOptions = options.ServerAuthenticationOptions;
```
bernardnormier, line 114: It's identical to the old code except with 3 more lines.

bernardnormier (approved):
Looks good!

src/IceRpc/Server.cs
pepone: The previous  version using `?` checks twice for null.

PR #4362, opened 2026-03-31
Review cs::attribute validation rules for enums

### LINKED ISSUE #4347: Review CSAttributeValidator for Slice enums
Need to review rules here. I think we might want different rules for BasicEnum, VariantEnum, Enumerator, and Variant. For example `cs::attribute` is allowed in `BasicEnum` C# mapped enums because you cannot do that in a partial. At first sight looks like VariantEnum should be more like struct.

pepone:
Fix #4347

### DIFF
--- a/src/ZeroC.Slice.Generator/CSAttributeValidator.cs
+++ b/src/ZeroC.Slice.Generator/CSAttributeValidator.cs
@@ -214,6 +214,8 @@ internal static class CsAttributeValidator
         switch (attr.Directive)
         {
             case CSAttributes.CSAttribute:
+                // The cs::attribute attribute is only allow for C# construction were adding attributes using a partial
+                // declaration is not possible,such as on enumerators and fields.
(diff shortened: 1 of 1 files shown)


src/ZeroC.Slice.Generator/CSAttributeValidator.cs
```diff
@@ -213,7 +215,7 @@ private static void ValidateCSAttribute(
         {
             case CSAttributes.CSAttribute:
                 RequireArgs(attr, 1, diagnostics);
-                if (target is not (Target.Enum or Target.Enumerator or Target.Field or Target.FieldInStruct))
+                if (target is not (Target.BasicEnum or Target.Enumerator or Target.Field or Target.FieldInStruct))
```
bernardnormier, line 220: This code needs comments explaining why `cs::attribute` is applicable to some constructs and not others.

PR #4365, opened 2026-03-31
Icerpc slice tools

pepone:
This PR updates IceRpc.Slice.Tools to use slicec and the new C# based generators instead of slicec-cs.

It depends on https://github.com/icerpc/slicec/pull/752

It downloads latest slicec nightly and caches it with the local build, so it will not work until slicec PR is merged and new nightly builds are uploaded.

There is a few additional fixes for generated code, to match more closely the current slicec-cs generated code.


build/IceRpc.Version.props
```diff
@@ -2,6 +2,9 @@
   <PropertyGroup>
     <!-- The version of all assembly files built in this repository -->
     <Version Condition="'$(Version)' == ''">0.6.0-preview.1</Version>
+    <!-- The slicec compiler version -->
+    <SlicecMMVersion>0.4</SlicecMMVersion>
```
bernardnormier, line 6: I assume this is temporary and will be eventually a simpler `SlicecVersion`?

src/IceRpc.Slice.Generator/slicec-icerpc-csharp-generator.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
bernardnormier, line 1: I would use simpler names for these scripts: slicec doesn't have the same requirements as protoc.

For example: slicec-csharp-generator (for the "base" csharp generator), slicec-icerpc-csharp-generator

src/IceRpc.Slice.Generator/slicec-icerpc-csharp-generator.sh
```diff
@@ -0,0 +1,10 @@
...
+
+set -ue
+
+# Everything in this script is relative to the directory containing this script.
+# This works most of the time but is not perfect. It will fail if the script is sourced or if the script is
+# executed from a symlink.
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+dotnet IceRpc.Slice.Generator.dll -- "$@"
```
bernardnormier: My understanding is `IceRpc.Slice.Generator.dll` rejects all args, and you read this script, you get the impression `IceRpc.Slice.Generator.dll` accepts some args. 
So maybe add something like:

```
# We forward the arguments to IceRpc.Slice.Generator, even though IceRpc.Slice.Generator
# currently rejects all arguments.
```

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.csproj
```diff
@@ -14,17 +15,17 @@
...
-    <RustConfiguration Condition="'$(Configuration)' == 'Debug'">debug</RustConfiguration>
-    <SliceCompilerExe Condition="'$(OS)' == 'Windows_NT'">slicec-cs.exe</SliceCompilerExe>
-    <SliceCompilerExe Condition="'$(OS)' != 'Windows_NT'">slicec-cs</SliceCompilerExe>
-    <!-- Internal properties used to compute the slicec-cs compiler location for the current platform. -->
+    <!-- Internal properties used to compute the slicec compiler location for the current platform. -->
     <IceRPC_OSName Condition="$([MSBuild]::IsOSPlatform('Linux'))">linux</IceRPC_OSName>
     <IceRPC_OSName Condition="$([MSBuild]::IsOSPlatform('Windows'))">windows</IceRPC_OSName>
     <IceRPC_OSName Condition="$([MSBuild]::IsOSPlatform('OSX'))">macos</IceRPC_OSName>
     <IceRPC_OSArch>$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString().ToLower())</IceRPC_OSArch>
+    <SliceCompilerExe Condition="'$(OS)' == 'Windows_NT'">slicec.exe</SliceCompilerExe>
```
bernardnormier: We can't use the IceRPC_OSName set a couple lines above?

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.props
```diff
@@ -3,7 +3,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <EnableDefaultSliceFileItems Condition="'$(EnableDefaultSliceFileItems)' == ''">true</EnableDefaultSliceFileItems>
-    <!-- Internal properties used to compute the slicec-cs compiler location for the current platform. -->
+    <!-- Internal properties used to compute the slicec compiler location for the current platform. -->
```
bernardnormier, line 6: I thought internal properties start with an `_`?

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets
```diff
@@ -55,11 +58,13 @@
   <Target Name="SliceC" BeforeTargets="CoreCompile" Condition="@(SliceFile) != ''">
     <MakeDir Directories="%(SliceFile.OutputDir)" />
     <!-- Compile the Slice files -->
-    <SliceCCSharpTask
+    <SliceCTask
       WorkingDirectory="$(MSBuildProjectDirectory)"
       OutputDir="%(SliceFile.OutputDir)"
       References="@(SliceDirectory->'%(FullPath)')"
       ToolsPath="$(SliceCompilerPath)"
+      ZeroCGenPath="$(SliceZeroCGenPath)"
```
bernardnormier: We should be consistent with the capitalized spelling of slicec: either Slicec or SliceC.

src/IceRpc.Slice.Tools/SliceCTask.cs
```diff
@@ -9,68 +9,76 @@
...
     public string[] References { get; set; } = [];
 
-    /// <summary>
-    /// The RPC provider to generate code for, corresponds to the <c>--rpc</c> slicec-cs compiler option.
-    /// </summary>
+    /// <summary>Gets or sets a value indicating whether to run the IceRpc.Slice.Generator plugin in addition to
+    /// ZeroC.Slice.Generator.</summary>
     [Required]
-    public string Rpc { get; set; } = "";
+    public bool Rpc { get; set; }
```
bernardnormier: Is it possible to move this logic -- which generator(s) to invoke -- into the caller?

This way, SliceCTask would only accept a list of generators and would get this Rpc option.

src/IceRpc.Slice.Tools/SliceCTask.cs
```diff
@@ -9,68 +9,76 @@
...
+    /// <summary>The directory containing the <c>slicec-gen-zeroc-slice-csharp</c> generator scripts.</summary>
+    [Required]
+    public string ZeroCGenPath { get; set; } = "";
+
     /// <inheritdoc/>
     protected override string ToolName =>
-        RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "slicec-cs.exe" : "slicec-cs";
+        RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "slicec.exe" : "slicec";
 
     private readonly JsonSerializerOptions _jsonSerializeOptions = new() { PropertyNameCaseInsensitive = true };
```
bernardnormier, line 50: What is this about?

src/IceRpc.Slice.Tools/SliceFile.FileProperties.xaml
```diff
@@ -93,10 +93,7 @@
     </StringProperty.DataSource>
   </StringProperty>
 
-  <EnumProperty Name="Rpc" DisplayName="RPC Provider" Description="Generate code for the specified RPC framework.">
-    <EnumValue Name="none" DisplayName="None" />
-    <EnumValue Name="icerpc" DisplayName="IceRPC" />
-  </EnumProperty>
+  <BoolProperty Name="Rpc" DisplayName="RPC Provider" Description="Generate IceRPC service and proxy code." />
```
bernardnormier: The display name is confusing here. You can't target various RPC frameworks: it's IceRPC or nothing.

src/IceRpc.Slice.Tools/SliceFile.ProjectProperties.xaml
```diff
@@ -36,10 +36,7 @@
     </StringProperty.DataSource>
   </StringProperty>
 
-  <EnumProperty Name="Rpc" DisplayName="RPC Provider" Description="Generate code for the specified RPC framework.">
-    <EnumValue Name="none" DisplayName="None" />
-    <EnumValue Name="icerpc" DisplayName="IceRPC" />
-  </EnumProperty>
+  <BoolProperty Name="Rpc" DisplayName="RPC Provider" Description="Generate IceRPC service and proxy code." />
```
bernardnormier: Same here. I would also rename the property IceRPC or IceRpc.

src/ZeroC.Slice.Generator/CSAttribute.cs
```diff
@@ -14,8 +14,8 @@ internal static class CSAttributes
     /// <summary>The directive for the <c>cs::identifier</c> attribute.</summary>
     internal const string CSIdentifier = "cs::identifier";
 
-    /// <summary>The directive for the <c>cs::internal</c> attribute.</summary>
-    internal const string CSInternal = "cs::internal";
+    /// <summary>The directive for the <c>cs::public</c> attribute.</summary>
```
bernardnormier: Resort alphabetically.

src/ZeroC.Slice.Generator/VariantEnumGenerator.cs
```diff
@@ -139,7 +137,9 @@ private static CodeBlock GenerateVariantRecord(
             ? $"{variantName}({BuildParameterList(variant.Fields, "")})"
             : variantName;
 
-        return new ContainerBuilder($"{accessModifier} partial record class", nameWithParams)
+        // Nested variant records are always public to match dunet's generated partials.
+        // Their effective visibility is limited by the parent type's access modifier.
+        return new ContainerBuilder("public partial record class", nameWithParams)
```
bernardnormier: Why not simply omit the access modifier?

That's what we do for partial classes generated by IceRpc.ServiceGenerator.

build/IceRpc.Version.props
pepone: This is used for the download URL, currently is not the same as the version:

```
<SlicecBaseUrl>https://zeroc.com/download/icerpc/slicec/nightly/$(SlicecMMVersion)</SlicecBaseUrl>
```

We need to rework this to support both nightly and release versions.

src/IceRpc.Slice.Generator/slicec-icerpc-csharp-generator.sh
pepone: added comments to the scripts

src/IceRpc.Slice.Tools/SliceCTask.cs
pepone: This is to decode the JSON diagnostics emit by slicec, which use lowercase names for the properties.

src/IceRpc.Slice.Tools/SliceFile.FileProperties.xaml
pepone: renamed IceRPC

src/IceRpc.Slice.Generator/OperationExtensions.cs
```diff
@@ -229,4 +276,59 @@ private static string BuildReturnType(string taskType, Operation op, string curr
...
+                if (value is not null)
+                {
+                    {{encodeExpr}};
+                }
+            }
+            """;
+    }
+
+    /// <summary>Returns a decode lambda for an optional stream element with inline bool markers.</summary>
+    internal static string GetStreamDecodeLambda(Field streamField, string currentNamespace)
```
externl: Shouldn't this be something more like `GetStreamOfOptionalDecodeLambda` The current name gives the impression you can use it for all streams.

src/IceRpc.Slice.Generator/OperationExtensions.cs
```diff
@@ -229,4 +276,59 @@ private static string BuildReturnType(string taskType, Operation op, string curr
...
+    {
+        IType elemType = streamField.DataType.Type;
+        string csType = elemType.ToTypeString(currentNamespace);
+        string decodeExpr = elemType.DecodeExpression(currentNamespace);
+        return $"(ref SliceDecoder decoder) => decoder.DecodeBool() ? ({csType}?){decodeExpr} : null";
```
externl, line 329: Why not call `GetDecodeLambda`. I think it would  be good to re-use the helper code where we can and there seems to be a special case for non-dictionary/sequence types. Is this somehow not applicable to streams of options of these types?

```
// For non-dict/non-seq optional types, add the nullable cast.
        if (isOptional && type is not DictionaryType and not SequenceType)
        {
            string csType = type.ToTypeString(currentNamespace);
            return $"(ref SliceDecoder decoder) => ({csType}?){decodeExpr}";
        }

        return $"(ref SliceDecoder decoder) => {decodeExpr}";
````

File: src/IceRpc.Slice.Generator/OperationExtensions.cs [hunk 15]
```diff
@@ -229,4 +276,59 @@ private static string BuildReturnType(string taskType, Operation op, string curr
...
+                encoder.EncodeBool(value is not null);
+                if (value is not null)
+                {
+                    {{encodeExpr}};
+                }
+            }
+            """;
+    }
+
+    /// <summary>Returns a decode lambda for an optional stream element with inline bool markers.</summary>
```
externl: What are a "inline bool markers"? Isn't this decoding bit-sequence of one element?

src/IceRpc.Slice.Generator/slicec-icerpc-csharp-generator.bat
```diff
@@ -0,0 +1,4 @@
+@echo OFF
+
+REM We forward the arguments to IceRpc.Slice.Generator, even though it currently ignores all arguments.
+call dotnet "%~dp0\IceRpc.Slice.Generator.dll" -- %*
```
externl: I think we should actually update it to error if it receives any arguments.

src/IceRpc.Slice.Generator/slicec-icerpc-csharp-generator.sh
externl: Agree. We should reject all arguments, not ignore.

src/IceRpc.Slice.Tools/SliceCTask.cs
```diff
@@ -9,68 +9,58 @@
...
+/// <summary>A MSBuild task to compile Slice files using the <c>slicec</c> compiler with configured generator
+/// plugins.</summary>
+public class SliceCTask : ToolTask
 {
-    /// <summary>Additional options to pass to the <c>slicec-cs</c> compiler.</summary>
+    /// <summary>Additional options to pass to the <c>slicec</c> compiler.</summary>
     public string[] AdditionalOptions { get; set; } = [];
 
+    /// <summary>The code-generator plugins to run after parsing and validation. Each entry is the full path to a
+    /// generator script.</summary>
```
externl, line 20: What about its options? Do we pass those where too, or will be have a separate property?

src/IceRpc.Slice.Tools/SliceCTask.cs
externl: The slices compiler can be configured to return its output: diagnostics, etc. in machine readable format, JSON.

src/IceRpc.Slice.Generator/slicec-icerpc-csharp-generator.bat
pepone: How do you see the rejection, just print an error to stderr and exit with an error code, or should we emit an Slice encoded diagnostic.

src/IceRpc.Slice.Tools/SliceCTask.cs
pepone: I guess we can include them here too if needed, this is passed as the `--generator` argument to slicec. Not sure it is worht mentioning right now as we don't support any options.

src/ZeroC.CodeBuilder/IAttributeBuilder.cs
```diff
@@ -28,7 +28,16 @@ public static T AddNeverEditorBrowsableAttribute<T>(this T builder) where T : IA
...
+    public static T AddObsoleteAttribute<T>(
+        this T builder,
+        bool condition = true,
+        string? message = null) where T : IAttributeBuilder<T> =>
+        condition
+            ? builder.AddAttribute(
+                message is not null
+                    ? $"global::System.Obsolete(\"{message}\")"
+                    : "global::System.Obsolete")
+            : builder;
```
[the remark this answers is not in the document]
pepone: It is up to the Slice author to not include invalid C# attributes.

src/IceRpc.Slice.Generator/OperationExtensions.cs
pepone: Fixed the doc comments.

src/IceRpc.Slice.Generator/OperationExtensions.cs
pepone: Here for stream of optional we handle the bit-sequence too.

src/IceRpc.Slice.Tools/SliceCTask.cs
pepone: Updated to take a list of generators.

src/IceRpc.Slice.Generator/slicec-icerpc-csharp-generator.bat
externl: I would do an error code and stderr message.

src/IceRpc.Slice.Generator/DispatchGenerator.cs
```diff
@@ -72,7 +76,7 @@ private static CodeBlock BuildServiceRequestClass(Interface interfaceDef, string
             if (streamParam is null)
             {
                 // Non-streaming: expression body
-                var decodeBuilder = new FunctionBuilder(
+                FunctionBuilder decodeBuilder = new FunctionBuilder(
```
externl, line 79: Is `var` not our preferred choice here as we're calling the constructor directly?

src/IceRpc.Slice.Generator/OperationExtensions.cs
externl: I mean calling `GetDecodeLambda` inside the new extension function.

BUILDING.md
```diff
@@ -24,31 +24,25 @@ build.cmd -help
     - [Command line](#command-line)
     - [Visual Studio Code](#visual-studio-code)
   - [Running the tests](#running-the-tests)
-    - [C# tests](#c-tests)
-    - [Rust tests](#rust-tests)
   - [Creating and publishing NuGet packages](#creating-and-publishing-nuget-packages)
-    - [Slice tools](#slice-tools)
```
InsertCreativityHere, line 30: The `Slice tools` section was kept in this document (line 117), so we should probably keep it in the table of contents too right?

src/IceRpc.Slice.Generator/DispatchGenerator.cs
pepone: we don't assign the result of `new ` directly, we use the builder pattern and chain methods to it.

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets
```diff
@@ -54,13 +57,21 @@
   </ItemGroup>
   <Target Name="SliceC" BeforeTargets="CoreCompile" Condition="@(SliceFile) != ''">
     <MakeDir Directories="%(SliceFile.OutputDir)" />
+    <!-- Build the generator list: always include the base C# generator, conditionally include the IceRPC generator. -->
+    <ItemGroup>
+      <_SliceGenerators Remove="@(_SliceGenerators)" />
```
InsertCreativityHere: I'm not super familiar with all this, but what's the point to clearing `_SliceGenerators`?
What do we expect may of been in this list before we add our 2 generators?

src/IceRpc.Slice.Tools/SliceCTask.cs
```diff
@@ -9,68 +9,58 @@
 
 namespace IceRpc.Slice.Tools;
 
-/// <summary>A MSBuild task to compile Slice files to C# using the IceRPC <c>slicec-cs</c> compiler.</summary>
-public class SliceCCSharpTask : ToolTask
+/// <summary>A MSBuild task to compile Slice files using the <c>slicec</c> compiler with configured generator
+/// plugins.</summary>
```
InsertCreativityHere: This should be "an". Even though 'MSBuild' starts with a consonant, you using the letter as a initial, not like you would normally pronounce it. And the spelling of the letter 'M' itself is "em" (as opposed to "mm").
```suggestion
/// <summary>An MSBuild task to compile Slice files using the <c>slicec</c> compiler with configured generator
/// plugins.</summary>
```

src/IceRpc.Slice.Generator/OperationExtensions.cs
pepone: We want that the lambda we generate here includes the bit-sequence decoding, which GetDecodeLambda doesn't include. Then because this is only for optionals we always include the cast, which the special part in GetDecodeLambda to deal with nested sequences types.

src/ZeroC.Slice.Generator/AttributeExtensions.cs
```diff
@@ -12,5 +12,18 @@ internal static class AttributeExtensions
         /// <summary>Returns all cs::attribute attributes from the list.</summary>
         internal IEnumerable<Attribute> CSAttributes() =>
             attributes.Where(a => a.Directive == Generator.CSAttributes.CSAttribute);
+
+        /// <summary>Gets whether the entity has the <c>deprecated</c> attribute.</summary>
+        internal bool IsDeprecated => attributes.Any(a => a.Directive == "deprecated");
+
+        /// <summary>Gets the deprecation message, or null if not deprecated or no message provided.</summary>
```
InsertCreativityHere:
```suggestion
        /// <summary>Gets the deprecation message, or null if not deprecated or no was message provided.</summary>
```

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
InsertCreativityHere: Pretty minor, but it's nice keeping these cases blocks in the same (alphabetical) order as in `CSAttribute.cs`.
i.e., moving the `CSPublic` block here between `CSNamespace` and `CSReadonly` in this file.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -45,6 +45,8 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
 
     private static string FormatInlineLink(CommentLink link, string currentNamespace) => link switch
     {
+        // Type aliases don't generate C# types, so output the identifier as plain text.
+        ResolvedCommentLink { Entity: TypeAlias } r => XmlEscape(r.Entity.Identifier),
```
InsertCreativityHere: Should we put the identifier in code-formatting to show that it's an identifier and not just normal text?
```suggestion
        ResolvedCommentLink { Entity: TypeAlias } r => $"<c>{XmlEscape(r.Entity.Identifier)}</c>",
```
Just actually not sure what the convention would be in C# here.

src/ZeroC.Slice.Generator/EntityExtensions.cs
```diff
@@ -22,26 +22,24 @@ internal static class EntityExtensions
...
         internal string Namespace => entity.Module.Namespace;
 
         /// <summary>Gets the camelCase parameter name (checks cs::identifier attribute).</summary>
-        internal string ParameterName
+        internal string ParameterName => entity.Attributes.FindAttribute(CSAttributes.CSIdentifier) is Attribute attr ?
+            attr.Args[0] : entity.Identifier.ToCamelCase();
+
+        /// <summary>Gets the fully-qualified name of the encoder or decoder extensions class, using
+        /// <c>global::</c> prefix when the entity is in a different namespace.</summary>
+        internal string ScopedExtensionsClass(string currentNamespace, bool decoder)
```
InsertCreativityHere: This function feels very similar to `EntityToTypeString` in `ITypeExtensions`. And this function is only called from with `ITypeExtensions.cs`.

Might be simpler to move this function in there and pull out the common namespace qualification logic.
Also might not be though, since these are both very small functions! :) Just a thought I noticed.

src/ZeroC.Slice.Generator/GeneratorDriver.cs
```diff
@@ -45,6 +49,29 @@ GeneratorResponse RunCore(ImmutableList<SliceFile> symbolFiles)
                 ?? assembly.GetName().Version?.ToString()
                 ?? "unknown";
 
+            // Check for duplicate file names — two files with the same name would produce the same
+            // output file, with the second silently overwriting the first.
+            var seenFileNames = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            foreach (SliceFile file in symbolFiles)
```
InsertCreativityHere, line 55: This check looks good, but I'm wondering if we should put this responsibility on `slicec`, to check and report if a file will be written multiple times during code-generation.

This way plugins don't need to worry about it, and `slicec` is the only thing that can catch cross-plugin overwriting. For example if I have these Slice files: `Definitions.IceRpc.slice`, `Definitions.slice`
The zeroc-generator and the icerpc-generator would both write `Definitions.IceRpc.cs`, but since the plugins are separate, they could never catch this. Only `slicec` could.

pepone: Not sure, what about languages like java/matlab/python where the name a number of generated files is totally independent of the Slice file names.

InsertCreativityHere: So, this C# code is relying on the fact that the generated names are 1-to-1 with the Slice file names, but I don't think that matters in general.

`slicec` can keep a `generated_file_paths: HashSet<String>` as it's actually writing the files to disk.
And if I tries to write to a file that it's already written it can report it.

Since none of the code-generators actually write to disk, only `slicec` does. And each generator tells `slicec` the paths it wants to write, so `slicec` has all the information it needs to detect these kinds of overwrites.

pepone: Doing the check base on the returned generated files, will work and is independent of how Slice files are mapped to file in the target language.

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets
pepone: I reworked this a bit, there was a bigger issue here. This didn't work well when you have SliceFile items with different IceRpc values in the project. We need to use  MSBuild batching.

src/ZeroC.Slice.Generator/EntityExtensions.cs
pepone: This is defined here because it is an extension method for Entity.

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
pepone: Fixed

src/ZeroC.CodeBuilder/FunctionBuilder.cs
```diff
@@ -36,7 +37,7 @@ public FunctionBuilder AddAttribute(string attribute)
         return this;
     }
 
-    /// <summary>Adds arguments to pass to the base constructor.</summary>
+    /// <summary>Adds arguments to pass to the base constructor (<c>: base(...)</c>).</summary>
     /// <param name="arguments">The arguments to pass to base.</param>
     /// <returns>This builder instance for method chaining.</returns>
     public FunctionBuilder AddBaseParameters(IEnumerable<string> arguments)
```
InsertCreativityHere, line 43: Unless I'm missing something, I only see one place where we call `AddThisParameter`, and none for `AddBaseParameters`. So, should probably delete `AddBaseParameters`, and replace `_baseArguments`+`_useThisConstructorChain` with just `_thisArguments`.

src/ZeroC.Slice.Generator/FieldExtensions.cs
```diff
@@ -89,18 +109,35 @@ internal static int GetBitSequenceSize(this ImmutableList<Field> fields) =>
...
         }
         else if (field.DataTypeIsOptional)
         {
             string decodeExpr = field.DecodeField(currentNamespace);
+            if (field.DataType.IsValueType)
+            {
+                string csType = useIncomingType
+                    ? field.DataType.IncomingParameterTypeString(false, currentNamespace)
+                    : field.DataType.FieldTypeString(false, currentNamespace);
+                return $"bitSequenceReader.Read() ? ({csType}?){decodeExpr} : null";
```
InsertCreativityHere, line 139: Would you be opposed to doing this instead, to reduce a little duplication, and reduce the number of returns:
```suggestion
                decodeExpr = decodeExpr.insert(0, "({csType}?)");
```
Maybe the re-assignment is more messy than the savings here though in your eyes.

src/ZeroC.Slice.Generator/Program.cs
```diff
@@ -11,5 +17,7 @@ await GeneratorDriver.RunAsync(
         VariantEnum e => VariantEnumGenerator.Generate(e),
         _ => null,
     },
-    mapOutputPath: path => Path.ChangeExtension(path, ".cs"),
+    mapOutputPath: path => Path.ChangeExtension(Path.GetFileName(path), ".cs"),
```
InsertCreativityHere, line 20: Why do we need to get the file name here? Is it to avoid changing directory's "extensions"?

src/ZeroC.Slice.Generator/VariantEnumGenerator.cs
```diff
@@ -25,12 +25,11 @@ internal static CodeBlock Generate(VariantEnum enumDef)
 
     private static CodeBlock GenerateUnknownRecord(
         VariantEnum enumDef,
-        string parentIdentifier,
-        string accessModifier)
+        string parentIdentifier)
     {
         string enumName = enumDef.Name;
         return new ContainerBuilder(
-                $"{accessModifier} partial record class",
```
InsertCreativityHere, line 33: My understanding is that nested classes in C# are always private by default.
Does Dunet do some magic to make these accessible from the outside?
Otherwise if they're always private, won't this cause issues for users trying to access them?

BUILDING.md
```diff
@@ -60,13 +55,13 @@ install ReportGenerator as follows:
 
 ```mermaid
 flowchart LR
-    compiler(slicec-cs) --> icerpc(ZeroC.Slice.Codec<br>IceRpc.*)
+    compiler(slicec + generators) --> icerpc(ZeroC.Slice.Codec<br>IceRpc.*)
```
bernardnormier, line 58: We should update all this in a follow-up PR. Maybe we don't need this build script anymore.

src/IceRpc.Slice.Tools/README.md
```diff
@@ -82,24 +88,24 @@ types such as `WellKnownTypes::Duration`, `WellKnownTypes::TimeStamp` and `WellK
...
-| Rpc               | `icerpc`  | Specifies the RPC framework to generate code for. Possible values: `icerpc`, `none`. This metadata corresponds to the `--rpc` option of `slicec-cs` |
+| Name              | Default   | Description                                                                                                                                                                                       |
+|-------------------|-----------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| AdditionalOptions |           | Specifies additional options to pass to the Slice compiler.                                                                                                                                       |
+| IceRpc            | `true`    | When `true`, the IceRpc.Slice.Generator runs in addition to the ZeroC.Slice.Generator, generating IceRPC-specific code (proxies, dispatchers). When `false`, only the ZeroC.Slice.Generator runs. |
```
bernardnormier: IceRpc.Slice.Generator doesn't generate dispatchers. It generates "skeletons", but so far we've never used this term with new Slice. We could use "service interfaces".

src/IceRpc.Slice.Tools/SliceCTask.cs
```diff
@@ -9,68 +9,58 @@
 
 namespace IceRpc.Slice.Tools;
 
-/// <summary>A MSBuild task to compile Slice files to C# using the IceRPC <c>slicec-cs</c> compiler.</summary>
-public class SliceCCSharpTask : ToolTask
+/// <summary>An MSBuild task to compile Slice files using the <c>slicec</c> compiler with configured generator
+/// plugins.</summary>
+public class SliceCTask : ToolTask
```
bernardnormier, line 14: Are we standardizing on Slicec or SliceC? We have ProtocTask, which seems inconsistent with SliceCTask.

I believe both Joe and Austin prefer `Slicec`.

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets
```diff
@@ -52,53 +55,44 @@
...
     <ItemGroup>
-      <SliceFile>
-        <!--
-          Add GeneratedPath and IceRpcGeneratedPath metadata with the normalized path, this is required for
-          excludes below to work with different path separators.
-        -->
-        <GeneratedPath>$([MSBuild]::NormalizePath('%(OutputDir)/%(Filename).cs'))</GeneratedPath>
-        <IceRpcGeneratedPath>$([MSBuild]::NormalizePath('%(OutputDir)/%(Filename).IceRpc.cs'))</IceRpcGeneratedPath>
-      </SliceFile>
-      <!--
```
bernardnormier, line 81: We should keep this comment. It's not clear to me why we need Exclude in this Compile item - I would expect _SliceFile.GeneratedPath to contain exactly the generated files and nothing else.

src/IceRpc.Slice.Tools/SliceCTask.cs
InsertCreativityHere: That is my preference, but more than anything I just want us to pick a consistent option.

src/ZeroC.CodeBuilder/FunctionBuilder.cs
pepone: Added #4376 to fix in a separate PR.

src/IceRpc.Slice.Tools/SliceCTask.cs
pepone: Added #4377 to fix in a separate PR.

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets
pepone: Added comment back.

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets
pepone: See #4377

src/ZeroC.Slice.Generator/Program.cs
pepone: The input is a full-path to where the Slice file is located. We want to return a relative path. So if the input is `C:\icerpc-chsharp\slice\Identity.slice`, the mapped path is just `Identity.cs` and not `C:\icerpc-chsharp\Identity.cs`

src/ZeroC.Slice.Generator/VariantEnumGenerator.cs
pepone: My understanding is the `partial` will take whatever access modifier was declared in the other partial. For example whatever Dunet picks.

src/ZeroC.Slice.Generator/FieldExtensions.cs
pepone: I find the current code easier to read.

PR #4374, opened 2026-04-06
Update DI templates

bernardnormier:
This PR resynchronizes all the DI templates with the generic host examples.


src/IceRpc.Templates/Templates/IceRpc-Protobuf-Client/Program.Authentication.cs
```diff
@@ -1,5 +1,3 @@
-// Copyright (c) ZeroC, Inc.
-
```
bernardnormier, line 2: We don't want these copyright notices in the templates "generated code".

externl (approved):

pepone (approved):

PR #4392, opened 2026-04-10
 Mix Ice, Slice, and Protobuf services

pepone:
Fix #4386

### DIFF
--- a/src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs
+++ b/src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs
@@ -58,7 +58,7 @@ internal class IceServiceMethod : ServiceMethod
         var codeBlock = new CodeBlock();
         if (!_idempotent)
         {
-            codeBlock.WriteLine("request.CheckNonIdempotent();");
+            codeBlock.WriteLine("IceRpc.Ice.Operations.IncomingRequestExtensions.CheckNonIdempotent(request);");
         }
 
         string thisInterface = $"((global::{_fullInterfaceName})this)";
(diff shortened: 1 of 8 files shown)


src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs
```diff
@@ -58,7 +58,7 @@ internal override CodeBlock GenerateDispatchCaseBody()
         var codeBlock = new CodeBlock();
         if (!_idempotent)
         {
-            codeBlock.WriteLine("request.CheckNonIdempotent();");
+            codeBlock.WriteLine("IceRpc.Ice.Operations.IncomingRequestExtensions.CheckNonIdempotent(request);");
```
bernardnormier, line 61: Maybe move CheckNonIdempotent to the IceRpc assembly/namespace?

bernardnormier (approved):
For a follow-up PR: I think we should add "Ice" and "Slice" to various "overloaded" methods in Ice and Slice's IncomingRequestExtensions class.

E.g DispatchIceOperationAsync and DispatchSliceOperationAsync, instead of DispatchOperationAsync for both.

src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs
bernardnormier: Or just inline it in the generated code. CheckNonIdempotent is pretty trivial:

```csharp
if (request.Fields.ContainsKey(RequestFieldKey.Idempotent))
{
        $"Invocation mode mismatch for operation '{request.Operation}': received idempotent field for an operation not marked as idempotent.");
}
```

PR #4393, opened 2026-04-10
Review access modifier generation in IceRpc.Slice.Generator

pepone:
The IceRpc.Slice.Generator generates many public types that can be internal by default, and only use `public` when strictly required, or `cs::pubic` is set.

### DIFF
--- a/src/IceRpc.Slice.Generator/DispatchGenerator.cs
+++ b/src/IceRpc.Slice.Generator/DispatchGenerator.cs
@@ -43,23 +43,27 @@ internal static class DispatchGenerator
 
         if (interfaceDef.Operations.Count > 0)
         {
-            builder.AddBlock(BuildServiceRequestClass(interfaceDef, scopedId, currentNamespace));
-            builder.AddBlock(BuildServiceResponseClass(interfaceDef, scopedId, currentNamespace));
+            builder.AddBlock(BuildServiceRequestClass(interfaceDef, scopedId, accessModifier, currentNamespace));
+            builder.AddBlock(BuildServiceResponseClass(interfaceDef, scopedId, accessModifier, currentNamespace));
 
             foreach (Operation op in interfaceDef.Operations)
             {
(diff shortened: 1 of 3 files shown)


src/IceRpc.Slice.Generator/DispatchGenerator.cs
```diff
@@ -43,23 +43,27 @@ Your service implementation must implement this interface.
...
         {
-            builder.AddBlock(BuildServiceRequestClass(interfaceDef, scopedId, currentNamespace));
-            builder.AddBlock(BuildServiceResponseClass(interfaceDef, scopedId, currentNamespace));
+            builder.AddBlock(BuildServiceRequestClass(interfaceDef, scopedId, accessModifier, currentNamespace));
+            builder.AddBlock(BuildServiceResponseClass(interfaceDef, scopedId, accessModifier, currentNamespace));
 
             foreach (Operation op in interfaceDef.Operations)
             {
-                builder.AddBlock(BuildServiceOperationDeclaration(op, currentNamespace));
+                builder.AddBlock(BuildServiceOperationDeclaration(op, accessModifier, currentNamespace));
```
bernardnormier: It's an abstract method in an interface. It should not have any access modifier.

I am surprised "internal" actually works here.

src/IceRpc.Slice.Generator/ProxyGenerator.cs
```diff
@@ -636,6 +644,7 @@ private static CodeBlock BuildProxyConstructors(string proxyName)
             .AddThisParameters(["invoker", "new IceRpc.ServiceAddress(serviceAddressUri)", "encodeOptions"])
             .Build();
 
+        // The parameterless struct constructor must be public (CS8958).
```
bernardnormier, line 647: We should call it parameterlessCtor, not defaultCtor.

bernardnormier (approved):
I think some proxy members like FromPath and DefaultPath are missing from this PR.

PR #4401, opened 2026-04-13
Inline service generator idempotent check

bernardnormier:
This PR removes CheckNonIdempotent and inlines the check.

See #4392.

### DIFF
--- a/src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs
+++ b/src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs
@@ -58,7 +58,12 @@ internal class IceServiceMethod : ServiceMethod
         var codeBlock = new CodeBlock();
         if (!_idempotent)
         {
-            codeBlock.WriteLine("IceRpc.Ice.Operations.IncomingRequestExtensions.CheckNonIdempotent(request);");
+            codeBlock.AddBlock(@"if (request.Fields.ContainsKey(IceRpc.RequestFieldKey.Idempotent))
+{
+    throw new IceRpc.DispatchException(
+        IceRpc.StatusCode.InvalidData,
+        $""Invocation mode mismatch for operation '{request.Operation}': received idempotent field for an operation not marked as idempotent."");
+}");
         }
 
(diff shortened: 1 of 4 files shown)


src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs
```diff
@@ -58,7 +58,12 @@ internal override CodeBlock GenerateDispatchCaseBody()
         var codeBlock = new CodeBlock();
         if (!_idempotent)
         {
-            codeBlock.WriteLine("IceRpc.Ice.Operations.IncomingRequestExtensions.CheckNonIdempotent(request);");
+            codeBlock.AddBlock(@"if (request.Fields.ContainsKey(IceRpc.RequestFieldKey.Idempotent))
+{
+    throw new IceRpc.DispatchException(
```
bernardnormier, line 63: We currently don't have any test for this mismatch.

src/IceRpc.ServiceGenerator/Internal/SliceServiceMethod.cs
```diff
@@ -62,8 +62,14 @@ internal override CodeBlock GenerateDispatchCaseBody()
         var codeBlock = new CodeBlock();
         if (!_idempotent)
         {
-            codeBlock.WriteLine("IceRpc.Slice.Operations.IncomingRequestExtensions.CheckNonIdempotent(request);");
+            codeBlock.AddBlock(@"if (request.Fields.ContainsKey(IceRpc.RequestFieldKey.Idempotent))
+{
+    throw new IceRpc.DispatchException(
```
pepone, line 67: You changed InvalidDataException -> DispatchException(InvalidData) because this always happens in the dispatch, and InvalidDataException is more for decoding errors where we don't know the decoding context?

pepone (approved):

externl (approved):

src/IceRpc.ServiceGenerator/Internal/SliceServiceMethod.cs
bernardnormier: That's correct.

PR #4402, opened 2026-04-14
Add IceRpc.BuildTelemetry generator

pepone:
This is a initial version of the IceRpc.Slice.BuildTelemetry generator.

For a follow up PR:

- We want to upgrade the data we collect.
- We want to pass arguments to the generator for additional dev diagnostics, and use them in CI.


src/IceRpc.Slice.BuildTelemetry/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+            new SslClientAuthenticationOptions(),
+            multiplexedClientTransport: QuicConnection.IsSupported ?
+                new QuicClientTransport() : new SlicClientTransport(new TcpClientTransport()));
+
+        try
+        {
+            var reporter = new ReporterProxy(connection);
+            reporter.UploadAsync(new BuildTelemetry.Slice(data), cancellationToken: cts.Token)
+                .ConfigureAwait(false).GetAwaiter().GetResult();
+            connection.ShutdownAsync(cts.Token).ConfigureAwait(false).GetAwaiter().GetResult();
```
bernardnormier: I don't like this GetAwaiter().GetResult(). It would be much nicer to use an async function.

src/IceRpc.Slice.BuildTelemetry/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+
+using Stream stdin = Console.OpenStandardInput();
+var reader = PipeReader.Create(stdin);
+
+using Stream stdout = Console.OpenStandardOutput();
+var writer = PipeWriter.Create(stdout);
+
+await Generator.RunAsync(reader, writer, BuildResponse).ConfigureAwait(false);
+
+static GeneratorResponse BuildResponse(
```
bernardnormier: RunAsync should accept an async callback, and BuildResponse should be async.

src/IceRpc.Slice.BuildTelemetry/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+        byte[] newHash = SHA256.HashData(content);
+        hashBytes = hashBytes.Length == 0 ? newHash : SHA256.HashData(newHash.Concat(hashBytes).ToArray());
+    }
+
+    // Determine the IceRPC version using the assembly version.
+    var assembly = Assembly.GetExecutingAssembly();
+    string toolVersion = assembly!.GetName().Version!.ToString();
+
+    string compilationHash = hashBytes.Length == 0 ?
+        string.Empty : Convert.ToHexString(hashBytes).ToLowerInvariant();
```
bernardnormier: For me, when symbolFiles is empty, we're not compiling anything and there is no point in "reporting build telemetry".

So no files => no report. That's what we do in the Protobuf build telemetry.

src/IceRpc.Slice.BuildTelemetry/slice/BuildTelemetry.slice
```diff
@@ -0,0 +1,91 @@
...
+    Slice(data: SliceTelemetryData)
+}
+
+/// Represents telemetry data collected from an IceRPC + Slice build.
+struct SliceTelemetryData {
+    /// The architecture of the client machine (e.g., x86_64, arm64).
+    architecture: string
+
+    /// A hash representing the the Slice files being compiled.
+    compilationHash: string
```
pepone, line 42: Wondering if we should provide the hash with the encoded compilation symbols to avoid compute this in multiple build-telemetry generators.

bernardnormier: When would you use multiple build telemetry generators together?

pepone: What I mean is we have to implement the hash computation in all build telemetry generators. For example generators for other languages.

bernardnormier: We only need one build telemetry generator per IDL. So most likely two: one for Slice and one for Protobuf.

I expect we'll write them in Rust once we have IceRPC for Rust.

src/IceRpc.Slice.BuildTelemetry/slice/BuildTelemetry.slice
```diff
@@ -0,0 +1,91 @@
...
+
+    /// The operating system running the client (e.g., Linux, Windows, macOS).
+    operatingSystem: string
+
+    /// The target language of the telemetry data.
+    targetLanguage: TargetLanguage
+
+    /// The version of the telemetry reporting tool used in this build.
+    toolVersion: string
+}
```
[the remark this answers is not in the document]
pepone: We will move into a separate repository an common location in a follo-up PR.

src/IceRpc.Slice.BuildTelemetry/Program.cs
```diff
@@ -0,0 +1,146 @@
...
+
+    var sliceTelemetryData = new SliceTelemetryData(
+        RuntimeInformation.ProcessArchitecture.ToString(),
+        compilationHash,
+        containsSlice1: false,
+        containsSlice2: true,
+        IsCi(),
+        RuntimeInformation.OSDescription,
+        referenceFileCount: 0,
```
externl, line 80: I guess we'll lose the reference file count? Not sure if we really care.

externl (approved):
Looks good. We need to have another pass over what we collect.

src/IceRpc.Slice.BuildTelemetry/IceRpc.Slice.BuildTelemetry.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="../IceRpc.Slice.Tools/IceRpc.Slice.Tools.props" />
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net10.0</TargetFramework>
+    <RollForward>major</RollForward>
```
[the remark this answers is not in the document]
bernardnormier: We already set IceRpcBuildTelemetry to false for all projects in the src directory.

src/ZeroC.Slice.Generator/GeneratorDriver.cs
```diff
@@ -29,7 +29,7 @@ internal static async Task RunAsync(
 
         await Symbols.Generator.RunAsync(reader, writer, RunCore).ConfigureAwait(false);
 
-        GeneratorResponse RunCore(ImmutableList<SliceFile> symbolFiles, Dictionary<string, string> options)
+        Task<GeneratorResponse> RunCore(ImmutableList<SliceFile> symbolFiles, Dictionary<string, string> options)
```
bernardnormier: Rename to `TransformAsync`?

src/IceRpc.Slice.BuildTelemetry/Program.cs
bernardnormier: Let's leave is as 0 until we refactor the build telemetry data.

bernardnormier (approved):

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.props
```diff
@@ -2,6 +2,8 @@
 <!-- Copyright (c) ZeroC, Inc. -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
+    <!-- Report build telemetry by default. The default is overridden to 'false' in the source-build branch below. -->
```
InsertCreativityHere: > The default is overridden to 'false' in the source-build branch below.

Where is this? I don't see any overriding to false here.

src/IceRpc.Slice.BuildTelemetry/slice/BuildTelemetry.slice
```diff
@@ -0,0 +1,91 @@
...
+
+/// Represents telemetry data from an IceRPC + Protobuf build.
+struct ProtobufTelemetryData {
+    /// The architecture of the client machine (e.g., x86_64, arm64).
+    architecture: string
+
+    /// A hash representing the the Protobuf files being compiled.
```
InsertCreativityHere: This still has the leftover `the the` in it:
```suggestion
    /// A hash representing the Protobuf files being compiled.
```

src/IceRpc.Slice.BuildTelemetry/slice/BuildTelemetry.slice
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) ZeroC, Inc.
```
InsertCreativityHere, line 1: Should we move this file somewhere it can be shared between the two build-telemetries to avoid them falling out of sync? Or is this painful to do with C# project's setup?

src/IceRpc.Slice.BuildTelemetry/IceRpc.Slice.BuildTelemetry.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <AssemblyName>IceRpc.Slice.BuildTelemetry</AssemblyName>
+    <!-- We don't generate a NuGet package for IceRpc.Slice.BuildTelemetry. The generator is included with
```
InsertCreativityHere: We say "plug-in" in the `IceRpc.Protobuf.BuildTelemetry.csproj` file, and I think "plug-in" is more true-to-intent than "generator" for the telemtry, even if it's technically using the same API as the generators.
```suggestion
    <!-- We don't generate a NuGet package for IceRpc.Slice.BuildTelemetry. The plug-in is included with
```

InsertCreativityHere (approved):
Looks good to me.

At some point, we should discuss what information the telemetry should be collecting which is useful to us.

src/IceRpc.Slice.BuildTelemetry/slice/BuildTelemetry.slice
pepone: I want to move this file to slicec repository, and vendor it to shared location like we do for other Slice files. For a follow-up PR.

PR #4406, opened 2026-04-14
Update IncreaseCollectionAllocation to avoid int overflows

bernardnormier:
This PR also adds tests to verify the ZeroC.Slice.Codec, IceRpc.Ice.Codec, and the icerpc header decoding code handles properly input that could trigger an int overflow.


src/ZeroC.Slice.Codec/SliceDecoder.cs
```diff
@@ -419,18 +419,20 @@ public BitSequenceReader GetBitSequenceReader(int bitSequenceSize)
     }
 
     /// <summary>Increases the number of bytes in the decoder's collection allocation.</summary>
-    /// <param name="byteCount">The number of bytes to add.</param>
+    /// <param name="byteCount">The number of bytes to add. Must be greater than 0.</param>
     /// <exception cref="InvalidDataException">Thrown when the total number of bytes exceeds the max collection
     /// allocation.</exception>
     /// <seealso cref="SliceDecoder(ReadOnlySequence{byte}, object?, int)" />
-    public void IncreaseCollectionAllocation(int byteCount)
+    public void IncreaseCollectionAllocation(long byteCount)
```
bernardnormier: Since this is a public API, we cannot apply the exact same fix to 0.5.x.

tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs
```diff
@@ -446,6 +446,96 @@ public async Task Invalid_request_refused(byte[] invalidRequestBytes, bool succe
...
+        ReadOnlyMemory<byte> responseBytes = responseWriter.WrittenMemory;
+
+        bool invalidRead = false;
+        await using ServiceProvider provider = new ServiceCollection()
+            .AddProtocolTest(Protocol.IceRpc)
+            .AddTestMultiplexedTransportDecorator(
+                clientOperationsOptions: new MultiplexedTransportOperationsOptions()
+                {
+                    StreamInputDecorator = decoratee =>
+                        invalidRead ? PipeReader.Create(new ReadOnlySequence<byte>(responseBytes)) : decoratee
```
bernardnormier, line 520: This technique (replace the transport bytes to provide a bogus response) was already used in another test.

src/ZeroC.Slice.Codec/SliceDecoderExtensions.cs
```diff
@@ -218,8 +218,8 @@ public static TSequence DecodeSequence<TSequence, TElement>(
         }
         else
         {
+            decoder.IncreaseCollectionAllocation((long)count * Unsafe.SizeOf<T?>());
```
bernardnormier: We now perform the collection allocation check before getting the bitSequence. Also note `Unsafe.SizeOf<T?>`.

src/IceRpc/Ice/Codec/IceDecoder.cs
```diff
@@ -302,18 +303,20 @@ public void CopyTo(Span<byte> destination)
...
-        if (_currentCollectionAllocation > _maxCollectionAllocation)
+        Debug.Assert(byteCount > 0);
+        long newAllocation = _currentCollectionAllocation + byteCount;
+        if (newAllocation > _maxCollectionAllocation)
         {
             throw new InvalidDataException(
                 $"The decoding exceeds the max collection allocation of '{_maxCollectionAllocation}'.");
         }
+        _currentCollectionAllocation = (int)newAllocation;
     }
```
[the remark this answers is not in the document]
bernardnormier: Fixed

src/ZeroC.Slice.Codec/SliceDecoder.cs
```diff
@@ -419,18 +419,20 @@ public BitSequenceReader GetBitSequenceReader(int bitSequenceSize)
...
-        _currentCollectionAllocation += byteCount;
-        if (_currentCollectionAllocation > _maxCollectionAllocation)
+        Debug.Assert(byteCount > 0);
+        long newAllocation = _currentCollectionAllocation + byteCount;
+        if (newAllocation > _maxCollectionAllocation)
         {
             throw new InvalidDataException(
                 $"The decoding exceeds the max collection allocation of '{_maxCollectionAllocation}'.");
         }
+        _currentCollectionAllocation = (int)newAllocation;
```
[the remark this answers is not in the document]
bernardnormier: Fixed

src/ZeroC.Slice.Codec/SliceDecoder.cs
```diff
@@ -419,18 +419,27 @@ public BitSequenceReader GetBitSequenceReader(int bitSequenceSize)
...
+    public void IncreaseCollectionAllocation(long byteCount)
     {
-        _currentCollectionAllocation += byteCount;
-        if (_currentCollectionAllocation > _maxCollectionAllocation)
+        if (byteCount <= 0)
+        {
+            throw new ArgumentException(
+                $"The {nameof(byteCount)} argument must be greater than 0.",
+                nameof(byteCount));
+        }
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested. We now tolerate 0.

tests/ZeroC.Slice.Codec.Tests/SequenceDecodingTests.cs
```diff
@@ -219,4 +219,47 @@ public void Decode_sequence_within_max_collection_allocation(int count)
...
             Throws.Nothing);
     }
+
+    /// <summary>Verifies that a crafted count near int.MaxValue / elementSize that would overflow int arithmetic
+    /// is correctly rejected by the allocation check.</summary>
+    [Test]
+    public void Decode_sequence_with_overflowing_allocation_cost_is_rejected()
+    {
+        // Arrange
+        // count * sizeof(int) would overflow in unchecked int arithmetic, wrapping to a small positive number.
```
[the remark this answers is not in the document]
bernardnormier: Fixed

tests/IceRpc.Ice.Codec.Tests/SequenceDecodingTests.cs
```diff
@@ -128,4 +128,26 @@ public void Decode_sequence_within_max_collection_allocation(int count)
...
             Throws.Nothing);
     }
+
+    /// <summary>Verifies that a crafted count near int.MaxValue / elementSize that would overflow int arithmetic
+    /// is correctly rejected by the allocation check.</summary>
+    [Test]
+    public void Decode_sequence_with_overflowing_allocation_cost_is_rejected()
+    {
+        // Arrange
+        // count * sizeof(int) would overflow in unchecked int arithmetic, wrapping to a small positive number.
```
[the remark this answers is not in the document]
bernardnormier: Fixed

tests/ZeroC.Slice.Codec.Tests/SequenceDecodingTests.cs
```diff
@@ -219,4 +219,47 @@ public void Decode_sequence_within_max_collection_allocation(int count)
...
+        encoder.EncodeSize(count);
+
+        // Act/Assert
+        Assert.That(
+            () =>
+            {
+                var sut = new SliceDecoder(buffer.WrittenMemory, maxCollectionAllocation: 1024);
+                _ = sut.DecodeSequence((ref SliceDecoder decoder) => decoder.DecodeInt32());
+            },
+            Throws.InstanceOf<InvalidDataException>().With.Message.Contains("max collection allocation"));
```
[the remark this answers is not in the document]
bernardnormier: Claude:  I think Copilot's concern is overly cautious here. Let me explain why.

For the massive allocation to actually happen, multiple protections would need to regress simultaneously: [...]
The OOM scenario requires all three protections to regress at once, which isn't a realistic concern. The tests are fine as-is.

tests/ZeroC.Slice.Codec.Tests/SequenceDecodingTests.cs
```diff
@@ -219,4 +219,47 @@ public void Decode_sequence_within_max_collection_allocation(int count)
...
+        var buffer = new MemoryBufferWriter(new byte[16]);
+        var encoder = new SliceEncoder(buffer);
+        encoder.EncodeSize(count);
+
+        // Act/Assert
+        Assert.That(
+            () =>
+            {
+                var sut = new SliceDecoder(buffer.WrittenMemory, maxCollectionAllocation: 1024);
+                _ = sut.DecodeSequenceOfOptionals<long?>((ref SliceDecoder decoder) => decoder.DecodeInt64());
```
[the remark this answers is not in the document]
bernardnormier: See above

src/ZeroC.Slice.Codec/SliceDecoder.cs
```diff
@@ -419,18 +419,26 @@ public BitSequenceReader GetBitSequenceReader(int bitSequenceSize)
...
+                $"The {nameof(byteCount)} argument must be greater than or equal to 0.",
+                nameof(byteCount));
+        }
+        long newAllocation = _currentCollectionAllocation + byteCount;
+        if (newAllocation > _maxCollectionAllocation)
         {
             throw new InvalidDataException(
                 $"The decoding exceeds the max collection allocation of '{_maxCollectionAllocation}'.");
         }
+        _currentCollectionAllocation = (int)newAllocation;
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested.

src/IceRpc/Ice/Codec/IceDecoder.cs
```diff
@@ -302,18 +302,26 @@ public void CopyTo(Span<byte> destination)
...
+                $"The {nameof(byteCount)} argument must be greater than or equal to 0.",
+                nameof(byteCount));
+        }
+        long newAllocation = _currentCollectionAllocation + byteCount;
+        if (newAllocation > _maxCollectionAllocation)
         {
             throw new InvalidDataException(
                 $"The decoding exceeds the max collection allocation of '{_maxCollectionAllocation}'.");
         }
+        _currentCollectionAllocation = (int)newAllocation;
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested.

externl (approved):
Looks good to me!

src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs
```diff
@@ -218,7 +218,7 @@ private static ServiceAddress DecodeServiceAddressCore(this ref IceDecoder decod
             {
                 // An Ice-encoded server address consumes at least 8 bytes (2 bytes for the server address type and 6
                 // bytes for the encapsulation header). SizeOf ServerAddress is large but less than 8 * 8.
-                decoder.IncreaseCollectionAllocation(count * Unsafe.SizeOf<ServerAddress>());
+                decoder.IncreaseCollectionAllocation((long)count * Unsafe.SizeOf<ServerAddress>());
```
pepone: I don't like this very much because the purposes of the cast is not obvious. I wonder if we should also do a initial check of "count" and reject it when is greater than the allocation limit and avoid the allocation.

Maybe we can move the cast to IncreaseCollectionAllocation to make this clear:

```
public void IncreaseCollectionAllocation(int count, int elementSize)
{
    if (count < 0 || elementSize < 0)
    {
        throw new ArgumentOutOfRangeException(...);
    }
    long byteCount = (long)count * elementSize;   // overflow-safe by construction
    int remaining = _maxCollectionAllocation - _currentCollectionAllocation;
    if (byteCount > remaining)
    {
        throw new InvalidDataException(...);
    }
    _currentCollectionAllocation += (int)byteCount;
}
```

pepone:
We should also fix the constructor to check for overflows, probably in a separate PR:

```
_maxCollectionAllocation = maxCollectionAllocation == -1 ? 8 * (int)buffer.Length
```

src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs
bernardnormier: > initial check of "count" 

I certainly prefer having a single check.

bernardnormier: Updated as suggested.

bernardnormier:
The Slice Codec's IncreaseCollectionAllocation is public because the decoding of a custom type can and should it whenever it allocates a collection.

Since there is no custom type with the Ice IDL, I made the Ice Codec's IncreaseCollectionAllocation internal and replaced the argument validation by Debug.Assert.

src/ZeroC.Slice.Codec/SliceDecoder.cs
```diff
@@ -419,18 +419,35 @@ public BitSequenceReader GetBitSequenceReader(int bitSequenceSize)
...
+    public void IncreaseCollectionAllocation(int count, int elementSize)
     {
-        _currentCollectionAllocation += byteCount;
-        if (_currentCollectionAllocation > _maxCollectionAllocation)
+        if (count < 0)
+        {
+            throw new ArgumentOutOfRangeException(
+                nameof(count),
+                $"The {nameof(count)} argument must be greater than or equal to 0.");
+        }
```
pepone: Should we exit earlier for count == 0?

src/IceRpc/Ice/Codec/IceDecoder.cs
```diff
@@ -372,6 +358,27 @@ public void SkipSize()
...
+    /// <param name="elementSize">The size of each element in bytes. Must be greater than 0.</param>
+    /// <exception cref="InvalidDataException">Thrown when the total number of bytes exceeds the max collection
+    /// allocation.</exception>
+    /// <seealso cref="IceDecoder(ReadOnlySequence{byte}, object?, int, IActivator?, int)" />
+    internal void IncreaseCollectionAllocation(int count, int elementSize)
+    {
+        Debug.Assert(count >= 0, $"{nameof(count)} must be greater than or equal to 0.");
+        Debug.Assert(elementSize > 0, $"{nameof(elementSize)} must be greater than 0.");
+
+        long byteCount = (long)count * elementSize;
```
pepone, line 371: Can we add a comment here: `// Widen to long so count * elementSize cannot overflow.`

tests/IceRpc.Ice.Codec.Tests/DictionaryDecodingTests.cs
```diff
@@ -93,4 +93,30 @@ public void Decode_dictionary_within_max_collection_allocation(int count)
...
+        int count = (int.MaxValue / entrySize) + 1;
+        var buffer = new MemoryBufferWriter(new byte[16]);
+        var encoder = new IceEncoder(buffer);
+        encoder.EncodeSize(count);
+
+        // Act/Assert
+        Assert.That(
+            () =>
+            {
+                var sut = new IceDecoder(buffer.WrittenMemory, maxCollectionAllocation: 1024);
```
pepone, line 114: I think we should set the higher value for maxCollectionAllocation, to ensure we are in the overflow case.

tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs
```diff
@@ -446,6 +446,96 @@ public async Task Invalid_request_refused(byte[] invalidRequestBytes, bool succe
...
+                })
+            .BuildServiceProvider(validateScopes: true);
+
+        var sut = provider.GetRequiredService<ClientServerProtocolConnection>();
+        await sut.ConnectAsync();
+
+        using var request = new OutgoingRequest(new ServiceAddress(Protocol.IceRpc));
+        invalidRead = true; // Use the crafted response bytes for the incoming response.
```
pepone, line 528: do we need this after connect async to allow connect to succeed?

tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs
```diff
@@ -446,6 +446,96 @@ public async Task Invalid_request_refused(byte[] invalidRequestBytes, bool succe
             Throws.Nothing);
     }
 
+    /// <summary>Verifies that a request with a field count large enough that count * 2 would overflow int arithmetic
+    /// is correctly rejected.</summary>
+    [Test]
+    public async Task Request_with_overflowing_field_count_is_rejected()
+    {
```
pepone, line 453: The test looks good, but there is lot of setup, I wonder if we should add an internal DecodeHeader helpers to make this easy to unit test.

src/ZeroC.Slice.Codec/SliceDecoder.cs
bernardnormier: No, this is a rare situation. No need to optimize for it.

src/IceRpc/Ice/Codec/IceDecoder.cs
bernardnormier: Done.

tests/IceRpc.Ice.Codec.Tests/DictionaryDecodingTests.cs
bernardnormier: We are in the overflow case since we verify the error message:
```cs
Throws.InstanceOf<InvalidDataException>().With.Message.Contains("max collection allocation"));
```

tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs
bernardnormier: I believe so, yes. See the Invocation_exception_handling_on_protocol_error test.

bernardnormier:
I removed the "count >= 0" check and ArgumentException thrown by IncreaseCollectionAllocation:

- the count >= 0 was about int overflow, that the new API avoids; I also want to use a negative count in a follow-up PR for the string decoding
- the elementSize > 0 remains as a Debug.Assert.

tests/IceRpc.Ice.Codec.Tests/DictionaryDecodingTests.cs
pepone: My understanding is that overflow case now always produce a large positive number. The message just checks we exceeded this small 1024 allocation limit, not necessarily an overflow case whereas a large limit means we hit the overflow.

bernardnormier: I am not following.

pepone: Here the exception means the allocation was requesting more than 1024 bytes, my understanding is that with the updated code the overflow results in a long value > than int32 max value.

pepone (approved):

externl (approved):
Looks good to me!

PR #4407, opened 2026-04-15
Detect short writes on TCP sockets

bernardnormier:
This code checks for short writes when writing to TCP sockets. Short writes should never occur.


src/IceRpc/Transports/Tcp/Internal/TcpConnection.cs
```diff
@@ -198,12 +198,21 @@ async ValueTask PerformWriteAsync()
...
                     if (buffer.IsSingleSegment)
                     {
-                        _ = await Socket.SendAsync(buffer.First, SocketFlags.None, cancellationToken)
-                            .ConfigureAwait(false);
+                        int bytesSent = await Socket.SendAsync(
+                            buffer.First,
+                            SocketFlags.None,
+                            cancellationToken).ConfigureAwait(false);
+
+                        if (bytesSent != buffer.First.Length)
```
pepone, line 206: I guess is fine to keep this check, but my understanding is that this can only happen if there is a bug in .NET. The documentation doesn't mention this but from what I remember of looking at the code it either writes everything or throws.

pepone (approved):

src/IceRpc/Transports/Tcp/Internal/TcpConnection.cs
pepone: A side note, seems kestrel ASP.NET, doesn't care about this and relies on the undocumented behavior.

Maybe we should throw https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.unreachableexception?view=net-10.0

externl (approved):

src/IceRpc/Transports/Tcp/Internal/TcpConnection.cs
bernardnormier: I would find UnreachableException confusing here.

I think we should throw `IceRpcException(IceRpcError.IceRpcError)` instead of InvalidOperationException. It's not really an invalid operation, but an unexpected behavior.

Issue #4504, opened 2026-04-15
Slice Tools JSon decoding error from MSBuild taks


pepone:
https://github.com/icerpc/icerpc-csharp/actions/runs/24456625271/job/71458986567

I think this is probably related to slicec and the genertors in this repository getting out of sync, but the error should be clear here.

```
  IceRpc.Slice.Tools -> /home/runner/work/icerpc-csharp/icerpc-csharp/src/IceRpc.Slice.Tools/bin/Debug/net10.0/IceRpc.Slice.Tools.dll
/home/runner/work/icerpc-csharp/icerpc-csharp/src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets(75,5): error MSB4018: The "SlicecTask" task failed unexpectedly. [/home/runner/work/icerpc-csharp/icerpc-csharp/src/IceRpc/IceRpc.csproj]
/home/runner/work/icerpc-csharp/icerpc-csharp/src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets(75,5): error MSB4018: System.Text.Json.JsonException->Microsoft.Build.Framework.BuildException.GenericBuildTransferredException: 'U' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 0. [/home/runner/work/icerpc-csharp/icerpc-csharp/src/IceRpc/IceRpc.csproj]
/home/runner/work/icerpc-csharp/icerpc-csharp/src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets(75,5): error MSB4018:  ---> System.Text.Json.JsonReaderException->Microsoft.Build.Framework.BuildException.GenericBuildTransferredException: 'U' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0. [/home/runner/work/icerpc-csharp/icerpc-csharp/src/IceRpc/IceRpc.csproj]
/home/runner/work/icerpc-csharp/icerpc-csharp/src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets(75,5): error MSB4018:    at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes) [/home/runner/work/icerpc-csharp/icerpc-csharp/src/IceRpc/IceRpc.csproj]
[...]

PR #4505, opened 2026-04-15
Sync slice compiler definitions with slicec repository

pepone:
The build is failing after picking latest nightly slicec binaries.

This PR syncs the definition and generated code with slicec repository. To do so you need to run the generator with and older version of slicec, for example

```
../slicec/target/debug/slicec \
    --generator src/ZeroC.Slice.Generator/bin/Debug/net10.0/ZeroC.Slice.Generator \
    slice/Compiler/*.slice --output-dir src/ZeroC.Slice.Symbols/Compiler
```

Where `slicec` is a build that matches the current generator definitions.

See #4504


src/ZeroC.Slice.Symbols/Compiler/DocComment.cs
```diff
@@ -120,7 +120,7 @@ internal static MessageComponent DecodeMessageComponent(this ref SliceDecoder de
             MessageComponent.Text.Discriminant => DecodeText(ref decoder),
             MessageComponent.Link.Discriminant => DecodeLink(ref decoder),
             int value => throw new global::System.IO.InvalidDataException(
-                $"Received invalid discriminant value '{value}' for MessageComponent.")
+                $"Received invalid discriminant value '{value}' for Compiler::MessageComponent.")
```
InsertCreativityHere, line 123: This is something we changed in the C# generator? To emit the whole scope?

Nothing wrong with it! Just making sure I didn't accidentally change something in `slicec` unintentionally.

InsertCreativityHere (approved):

src/ZeroC.Slice.Symbols/Compiler/DocComment.cs
pepone: yes, it was probably not regenerated when we changed that.

externl (approved):

bernardnormier (approved):

PR #4511, opened 2026-04-15
Rework IncreaseCollectionAllocation in string decoding

### LINKED ISSUE #4493: [Audit-Medium] String decoding enforces the allocation limit only aft…

### Medium: String decoding enforces the allocation limit only after allocating the string — **CONFIRMED**

Affected code:

- [src/ZeroC.Slice.Codec/SliceDecoder.cs:159-209](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/ZeroC.Slice.Codec/SliceDecoder.cs#L159) — `DecodeString`
- Both the fast path ([line 173](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/ZeroC.Slice.Codec/SliceDecoder.cs#L173)) and the slow multi-segment path ([line 192](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/ZeroC.Slice.Codec/SliceDecoder.cs#L192)) invoke `_utf8.GetString(...)` before the allocation counter is touched
- The counter is updated at [line 207](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/ZeroC.Slice.Codec/SliceDecoder.cs#L207): `IncreaseCollectionAllocation(result.Length * Unsafe.SizeOf<char>());`

Verification:

Confirmed. Additionally, the computation `result.Length * 2` is unchecked, so for strings longer than `int.MaxValue / 2` characters the guard double-fails: the string is already allocated, and the accounting wraps.

The maximum practical string size is bounded by `_reader.Remaining`, and the `size > bytes.Length` check protects against reading past the buffer, so the worst case is "one oversize string allocation followed by throw". Impact is therefore limited when the caller bounds buffer size at the transport layer, but the guard is not doing what its documentation implies.

Status: **Valid**. Severity: Medium.

Recommendation:

- Compute `size * 2` (estimated byte cost) using checked arithmetic *before* calling `_utf8.GetString`, and check against `_maxCollectionAllocation`.
- At minimum, make the post-allocation accounting `checked` so overflow surfaces as an error rather than silently wrapping.

---

**Source report:** [src-ZeroC.Slice.Codec-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-ZeroC.Slice.Codec-audit-2026-04-14.md) (finding `String decoding enforces the allocation limit only after allocating the string — **CONFIRMED**`)

bernardnormier:
Fixes #4493.

### DIFF
--- a/src/IceRpc/Ice/Codec/IceDecoder.cs
+++ b/src/IceRpc/Ice/Codec/IceDecoder.cs
@@ -212,6 +212,10 @@ public ref partial struct IceDecoder
         }
         else
         {
+            // In the worse case scenario, each byte becomes a new character. We'll adjust this allocation increase
+            // after decoding the string.
+            IncreaseCollectionAllocation(size, Unsafe.SizeOf<char>());
+
             string result;
             if (_reader.UnreadSpan.Length >= size)
             {
@@ -249,9 +253,9 @@ public ref partial struct IceDecoder
 
             _reader.Advance(size);
 
-            // We can only compute the new allocation _after_ decoding the string. For dictionaries and sequences,
-            // we perform this check before the allocation.
-            IncreaseCollectionAllocation(result.Length, Unsafe.SizeOf<char>());
+            // Make the adjustment. Note that `result.Length - size` can be negative. The overall increase is
+            // result.Length * SizeOf<char>().
+            IncreaseCollectionAllocation(result.Length - size, Unsafe.SizeOf<char>());
             return result;
         }
     }
--- a/src/IceRpc/Ice/Codec/IceDecoder.cs
+++ b/src/IceRpc/Ice/Codec/IceDecoder.cs
@@ -212,7 +212,7 @@ public ref partial struct IceDecoder
         }
         else
         {
-            // In the worse case scenario, each byte becomes a new character. We'll adjust this allocation increase
+            // In the worst-case scenario, each byte becomes a new character. We'll adjust this allocation increase
             // after decoding the string.
             IncreaseCollectionAllocation(size, Unsafe.SizeOf<char>());
 
--- a/src/IceRpc/Ice/Codec/IceDecoder.cs
+++ b/src/IceRpc/Ice/Codec/IceDecoder.cs
@@ -369,6 +369,7 @@ public ref partial struct IceDecoder
     /// <seealso cref="IceDecoder(ReadOnlySequence{byte}, object?, int, IActivator?, int)" />
     internal void IncreaseCollectionAllocation(int count, int elementSize)
     {
+        Debug.Assert(count >= 0, $"{nameof(count)} must be greater than or equal to 0.");
         Debug.Assert(elementSize > 0, $"{nameof(elementSize)} must be greater than 0.");
 
         long byteCount = (long)count * elementSize;
--- a/src/IceRpc/Ice/Codec/IceDecoder.cs
+++ b/src/IceRpc/Ice/Codec/IceDecoder.cs
@@ -253,9 +253,8 @@ public ref partial struct IceDecoder
 
             _reader.Advance(size);
 
-            // Make the adjustment. Note that `result.Length - size` can be negative. The overall increase is
-            // result.Length * SizeOf<char>().
-            IncreaseCollectionAllocation(result.Length - size, Unsafe.SizeOf<char>());
+            // Make the adjustment. The overall increase in allocation is result.Length * SizeOf<char>().
+            DecreaseCollectionAllocation(size - result.Length, Unsafe.SizeOf<char>());
             return result;
         }
     }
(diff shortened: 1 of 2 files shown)


src/IceRpc/Ice/Codec/IceDecoder.cs
```diff
@@ -212,6 +212,10 @@ public string DecodeString()
         }
         else
         {
+            // In the worse case scenario, each byte becomes a new character. We'll adjust this allocation increase
```
InsertCreativityHere:
```suggestion
            // In the worst-case scenario, each byte becomes a new character. We'll adjust this allocation increase
```

src/ZeroC.Slice.Codec/SliceDecoder.cs
```diff
@@ -165,6 +165,10 @@ public string DecodeString()
         }
         else
         {
+            // In the worse case scenario, each byte becomes a new character. We'll adjust this allocation increase
```
InsertCreativityHere:
```suggestion
            // In the worst-case scenario, each byte becomes a new character. We'll adjust this allocation increase
```

InsertCreativityHere (approved):
Looks good. Neat solution!

src/ZeroC.Slice.Codec/SliceDecoder.cs
```diff
@@ -202,9 +206,9 @@ public string DecodeString()
 
             _reader.Advance(size);
 
-            // We can only compute the new allocation _after_ decoding the string. For dictionaries and sequences,
-            // we perform this check before the allocation.
-            IncreaseCollectionAllocation(result.Length, Unsafe.SizeOf<char>());
+            // Make the adjustment. Note that `result.Length - size` can be negative. The overall increase is
+            // result.Length * SizeOf<char>().
+            IncreaseCollectionAllocation(result.Length - size, Unsafe.SizeOf<char>());
```
pepone: I find misleading to call IncreaseCollectionAllocation with negative values to decrease the value. Maybe this should be named AdjustCollectionAllocation?

Or maybe is simpler to keep >= 0 in IncreaseCollectionAllocation and do the adjustment inline here given this is only for strings.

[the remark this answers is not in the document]
bernardnormier: Reworked code along those lines.

bernardnormier: Fixed as suggested.

pepone (approved):

src/ZeroC.Slice.Codec/SliceDecoder.cs
```diff
@@ -424,6 +427,7 @@ public BitSequenceReader GetBitSequenceReader(int bitSequenceSize)
     /// <seealso cref="SliceDecoder(ReadOnlySequence{byte}, object?, int)" />
     public void IncreaseCollectionAllocation(int count, int elementSize)
     {
+        Debug.Assert(count >= 0, $"{nameof(count)} must be greater than or equal to 0.");
```
InsertCreativityHere, line 430: Should we also add this to the `IncreaseCollectionAllocation` in `IceDecoder`?

InsertCreativityHere (approved):
Still looks good.
I didn't have any issue with it being a single function instead of `Increase/Decrease` though.

externl (approved):

src/ZeroC.Slice.Codec/SliceDecoder.cs
```diff
@@ -165,6 +165,10 @@ public string DecodeString()
         }
         else
         {
+            // In the worst-case scenario, each byte becomes a new character. We'll adjust this allocation increase
+            // after decoding the string.
+            IncreaseCollectionAllocation(size, Unsafe.SizeOf<char>());
+
```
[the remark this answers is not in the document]
bernardnormier: That's true, but this allocation limit is not expected to be exact. It's to prevent malicious clients: whenever we decode a string, the allocation limit should have enough headroom to accept byteCount chars.

src/IceRpc/Ice/Codec/IceDecoder.cs
```diff
@@ -249,9 +253,8 @@ public string DecodeString()
 
             _reader.Advance(size);
 
-            // We can only compute the new allocation _after_ decoding the string. For dictionaries and sequences,
-            // we perform this check before the allocation.
-            IncreaseCollectionAllocation(result.Length, Unsafe.SizeOf<char>());
+            // Make the adjustment. The overall increase in allocation is result.Length * SizeOf<char>().
+            DecreaseCollectionAllocation(size - result.Length, Unsafe.SizeOf<char>());
             return result;
```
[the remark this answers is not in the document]
bernardnormier: That's over the top.

src/ZeroC.Slice.Codec/SliceDecoder.cs
```diff
@@ -424,6 +427,7 @@ public BitSequenceReader GetBitSequenceReader(int bitSequenceSize)
     /// <seealso cref="SliceDecoder(ReadOnlySequence{byte}, object?, int)" />
     public void IncreaseCollectionAllocation(int count, int elementSize)
     {
+        Debug.Assert(count >= 0, $"{nameof(count)} must be greater than or equal to 0.");
         Debug.Assert(elementSize > 0, $"{nameof(elementSize)} must be greater than 0.");
```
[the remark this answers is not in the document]
bernardnormier: The Debug.Assert are to catch egregious logic error bugs in the caller. We don't need to convert them to runtime checks with exceptions.

src/ZeroC.Slice.Codec/SliceDecoder.cs
bernardnormier: Yes, fixed.

PR #4514, opened 2026-04-15
Reject empty buffer in ColocConnection.ReadAsync

### LINKED ISSUE #4481: [Audit-Low] ColocConnection.ReadAsync silently accepts a zero-length…

### Low: `ColocConnection.ReadAsync` silently accepts a zero-length buffer, diverging from the TCP transport's contract

Affected code:

- [src/IceRpc.Transports.Coloc/Internal/ColocConnection.cs:31-101](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Transports.Coloc/Internal/ColocConnection.cs#L31) — `ReadAsync` has no length check on `buffer`
- Compare: [src/IceRpc/Transports/Tcp/Internal/TcpConnection.cs:61-66](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc/Transports/Tcp/Internal/TcpConnection.cs#L61) — `TcpConnection.ReadAsync` throws `ArgumentException` when `buffer.Length == 0`

Description:

`TcpConnection.ReadAsync` explicitly rejects empty buffers. `ColocConnection.ReadAsync` does not. If a caller passes `Memory<byte>.Empty`, coloc falls through to `_reader.ReadAsync` (which blocks until data arrives), then copies zero bytes via `CopySegmentToMemory` (destination length zero), then calls `AdvanceTo(GetPosition(0))` and returns 0.

Net behavior: a coloc caller passing an empty buffer *blocks* waiting for peer data, reads nothing, and returns 0 — potentially misleading it about the connection's state. A TCP caller would immediately get `ArgumentException` for the same mistake.

Two coloc-specific wrinkles make this worse:

1. If the peer writes data exactly when the empty read is pending, coloc returns 0 but *leaves the data buffered* because `GetPosition(0)` advances past zero bytes. The next real read gets the buffered data. That's actually correct semantically — but the blocking wait was unnecessary.
2. If the peer shuts down during the empty read, coloc returns 0 (end of stream) which is indistinguishable from "I asked for zero bytes and got zero".

Impact:

- Contract drift between transports.
- Misleading behavior for a caller bug (empty buffer) that's pure programmer error.

Recommendation:

- Add `if (buffer.Length == 0) throw new ArgumentException("...", nameof(buffer));` at the top of `ReadAsync`, matching `TcpConnection`.

Severity: Low.

---

**Source report:** [src-IceRpc.Transports.Coloc-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.Transports.Coloc-audit-2026-04-14.md) (finding ``ColocConnection.ReadAsync` silently accepts a zero-length buffer, diverging from the TCP transport's contract`)

pepone:
Fixes #4481.

### DIFF
--- a/tests/IceRpc.Conformance.Tests/Transports/DuplexConnectionConformanceTests.cs
+++ b/tests/IceRpc.Conformance.Tests/Transports/DuplexConnectionConformanceTests.cs
@@ -211,7 +211,7 @@ public abstract class DuplexConnectionConformanceTests
         }
     }
 
-    /// <summary>Verifies that a read with an empty buffer throws <see cref="ArgumentException" />.</summary>
+    /// <summary>Verifies that a read into an empty buffer throws <see cref="ArgumentException" />.</summary>
     [Test]
     public async Task Read_with_empty_buffer_fails()
     {
--- a/tests/IceRpc.Conformance.Tests/Transports/DuplexConnectionConformanceTests.cs
+++ b/tests/IceRpc.Conformance.Tests/Transports/DuplexConnectionConformanceTests.cs
@@ -211,6 +211,19 @@ public abstract class DuplexConnectionConformanceTests
         }
     }
 
+    /// <summary>Verifies that a read with an empty buffer throws <see cref="ArgumentException" />.</summary>
+    [Test]
+    public async Task Read_with_empty_buffer_fails()
+    {
+        await using ServiceProvider provider = CreateServiceCollection().BuildServiceProvider(validateScopes: true);
+        var sut = provider.GetRequiredService<ClientServerDuplexConnection>();
+        await sut.AcceptAndConnectAsync();
+
+        Assert.That(
+            async () => await sut.Client.ReadAsync(Memory<byte>.Empty, default),
+            Throws.TypeOf<ArgumentException>());
+    }
+
     [Test]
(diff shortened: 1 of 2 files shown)


tests/IceRpc.Conformance.Tests/Transports/DuplexConnectionConformanceTests.cs
```diff
@@ -211,6 +211,19 @@ static async Task ReadAsync(IDuplexConnection connection, int size)
         }
     }
 
+    /// <summary>Verifies that a read with an empty buffer throws <see cref="ArgumentException" />.</summary>
```
bernardnormier: a read into an empty buffer?

bernardnormier (approved):

tests/IceRpc.Conformance.Tests/Transports/DuplexConnectionConformanceTests.cs
pepone:
```suggestion
    /// <summary>Verifies that a read into an empty buffer throws <see cref="ArgumentException" />.</summary>
```

PR #4522, opened 2026-04-16
Remove ColocListener from shared dictionary on dispose

### LINKED ISSUE #4484: [Audit-Medium] ColocListener.DisposeAsync doesn't remove itself from…

### Medium: `ColocListener.DisposeAsync` doesn't remove itself from the shared `_listeners` dictionary

Affected code:

- [src/IceRpc.Transports.Coloc/Internal/ColocListener.cs:66-90](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Transports.Coloc/Internal/ColocListener.cs#L66) — `DisposeAsync` cancels the accept loop, completes the channel, drains pending items, and disposes the CTS — but never touches `_listeners`
- [src/IceRpc.Transports.Coloc/Internal/ColocServerTransport.cs:40-49](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Transports.Coloc/Internal/ColocServerTransport.cs#L40) — `Listen` calls `_listeners.TryAdd((host, port), listener)` and throws `AddressInUse` on collision
- [src/IceRpc.Transports.Coloc/ColocTransport.cs:32-36](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Transports.Coloc/ColocTransport.cs#L32) — the dictionary is shared between the client and server transports of a single `ColocTransport` instance, i.e. lives as long as the `ColocTransport` object

Description:

When an application calls `DisposeAsync` on a `ColocListener`, the listener stops accepting connections and completes its channel, but the dictionary entry under its `(Host, Port)` key persists. Two real consequences:

1. **Address "stuck in use" for the life of the `ColocTransport`.** A subsequent `Listen` at the same address fails with `IceRpcError.AddressInUse`, because `TryAdd` sees the stale entry. This forces applications to tear down and recreate the entire `ColocTransport` (and all its existing connections) to reuse an address, rather than just re-listening.

2. **Unbounded memory growth on transports with lots of listen/dispose churn.** A test harness or a server that rotates through many ephemeral coloc addresses accumulates one stale `ColocListener` per cycle. Each retains its `PipeOptions`, `Channel`, and `CancellationTokenSource`.

The fix is straightforward: `DisposeAsync` should call `_listeners.TryRemove` on its own `(Host, Port)` key. That requires the listener to know both the dictionary and its key — either passed in at construction or made available via a small callback.

Clients racing a connect against a listener mid-dispose already handle the "listener exists but channel completed" case correctly (`TryWrite` returns false → `ConnectionRefused`), so removing the entry during dispose doesn't break that path.

Impact:

- Permanent address reservation per process lifetime (within a single `ColocTransport`).
- Memory growth proportional to listen/dispose cycles.

Recommendation:

- Pass the `_listeners` dictionary (and the listener's own key) to `ColocListener`, and call `_listeners.TryRemove(key, out _)` at the top of `DisposeAsync`.
- Add a regression test that listens, disposes, and re-listens at the same address.

Severity: Medium.

---

**Source report:** [src-IceRpc.Transports.Coloc-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.Transports.Coloc-audit-2026-04-14.md) (finding ``ColocListener.DisposeAsync` doesn't remove itself from the shared `_listeners` dictionary`)

pepone:
Fixes #4484.

### DIFF
--- a/src/IceRpc.Transports.Coloc/Internal/ColocListener.cs
+++ b/src/IceRpc.Transports.Coloc/Internal/ColocListener.cs
@@ -1,6 +1,5 @@
 // Copyright (c) ZeroC, Inc.
 
-using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO.Pipelines;
@@ -15,7 +14,7 @@ internal class ColocListener : IListener<IDuplexConnection>
     public TransportAddress TransportAddress { get; }
 
     private readonly CancellationTokenSource _disposeCts = new();
-    private readonly ConcurrentDictionary<(string Host, ushort Port), ColocListener> _listeners;
+    private readonly Action _onDispose;
     private readonly EndPoint _networkAddress;
     private readonly PipeOptions _pipeOptions;
 
@@ -73,10 +72,8 @@ internal class ColocListener : IListener<IDuplexConnection>
             return default;
         }
 
-        // Remove this listener from the shared dictionary so the address can be reused.
-        _listeners.TryRemove(new KeyValuePair<(string, ushort), ColocListener>(
-            (TransportAddress.Host, TransportAddress.Port),
-            this));
+        // Notify the owner (e.g. the server transport) so it can release its reference to this listener.
+        _onDispose();
 
         // Cancel pending AcceptAsync.
         _disposeCts.Cancel();
@@ -98,13 +95,13 @@ internal class ColocListener : IListener<IDuplexConnection>
 
     internal ColocListener(
         TransportAddress transportAddress,
-        ConcurrentDictionary<(string Host, ushort Port), ColocListener> listeners,
+        Action onDispose,
         ColocTransportOptions colocTransportOptions,
         DuplexConnectionOptions duplexConnectionOptions)
     {
         TransportAddress = transportAddress;
 
-        _listeners = listeners;
+        _onDispose = onDispose;
         _networkAddress = new ColocEndPoint(transportAddress);
         _pipeOptions = new PipeOptions(
             pool: duplexConnectionOptions.Pool,
--- a/src/IceRpc.Transports.Coloc/Internal/ColocListener.cs
+++ b/src/IceRpc.Transports.Coloc/Internal/ColocListener.cs
@@ -14,7 +14,7 @@ internal class ColocListener : IListener<IDuplexConnection>
     public TransportAddress TransportAddress { get; }
 
     private readonly CancellationTokenSource _disposeCts = new();
-    private readonly Action _onDispose;
+    private readonly Action<ColocListener> _onDispose;
     private readonly EndPoint _networkAddress;
     private readonly PipeOptions _pipeOptions;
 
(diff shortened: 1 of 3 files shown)


src/IceRpc.Transports.Coloc/Internal/ColocListener.cs
```diff
@@ -71,6 +73,11 @@ public ValueTask DisposeAsync()
             return default;
         }
 
+        // Remove this listener from the shared dictionary so the address can be reused.
+        _listeners.TryRemove(new KeyValuePair<(string, ushort), ColocListener>(
+            (TransportAddress.Host, TransportAddress.Port),
+            this));
```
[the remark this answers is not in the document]
pepone: It does https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.tryremove?view=net-10.0#system-collections-concurrent-concurrentdictionary-2-tryremove(system-collections-generic-keyvaluepair((-0-1)))

src/IceRpc.Transports.Coloc/Internal/ColocListener.cs
```diff
@@ -91,11 +98,13 @@ public ValueTask DisposeAsync()
 
     internal ColocListener(
         TransportAddress transportAddress,
+        ConcurrentDictionary<(string Host, ushort Port), ColocListener> listeners,
```
externl: Seems wrong to me for the `ColocListener` to know about all other coloc listeners. What if instead we passed a callback which call instead. Then the `ColocServerTransport` can be responsible for the cleanup.

pepone: Fixed as suggested

src/IceRpc.Transports.Coloc/Internal/ColocServerTransport.cs
```diff
@@ -37,12 +37,16 @@ public IListener<IDuplexConnection> Listen(
                 nameof(transportAddress));
         }
 
-        var listener = new ColocListener(
+        var key = (transportAddress.Host, transportAddress.Port);
+        ColocListener? listener = null;
+        listener = new ColocListener(
             transportAddress,
+            onDispose: () => _listeners.TryRemove(
+                new KeyValuePair<(string, ushort), ColocListener>(key, listener!)),
```
bernardnormier: I don't find this code elegant (with `listener!`). There must be a cleaner solution.

pepone: Reworked the callback to take the listener as a parameter

externl (approved):

src/IceRpc.Transports.Coloc/Internal/ColocServerTransport.cs
```diff
@@ -37,12 +37,18 @@ public IListener<IDuplexConnection> Listen(
                 nameof(transportAddress));
         }
 
+        var key = (transportAddress.Host, transportAddress.Port);
+
+        void OnDispose(ColocListener disposedListener) =>
```
bernardnormier: We don't insert local functions in the middle of the code. It should be at the end of the method.

bernardnormier (approved):

PR #4530, opened 2026-04-21
Add PauseWriterThreshold to bound per-stream Slic outbound buffering

pepone:
## Summary

- Add a configurable `PauseWriterThreshold` option (default 64 KB) to `SlicTransportOptions` that limits how much outbound stream payload data Slic buffers locally per stream, independent of the peer's advertised flow control window.
- Introduce per-stream local credit tracking with completion callbacks on the shared `SlicDuplexConnectionWriter`. Local credit is only released after bytes are written to the underlying `IDuplexConnection`, not when they move between Slic-internal buffers.
- Add ping flood protection: the connection is closed if more than 16 Pong replies are pending in the shared writer, preventing a malicious peer from causing unbounded control-frame buffering.

Related: icerpc/icerpc-csharp-audit#15

### DIFF
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -14,7 +14,7 @@ namespace IceRpc.Transports.Slic.Internal;
 
 /// <summary>The Slic connection implements an <see cref="IMultiplexedConnection" /> on top of a <see
 /// cref="IDuplexConnection" />.</summary>
-internal class SlicConnection : IMultiplexedConnection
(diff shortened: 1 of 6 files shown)


bernardnormier (commented):
It would be preferable to fix the Ping issue in a separate PR.

Also, the motivation for the new per-stream local credit tracking is not clear. Is there a corresponding issue that explains why we need it?

pepone:
> It would be preferable to fix the Ping issue in a separate PR.

I will split this in two PRs, event the completion callback is required in both.

> Also, the motivation for the new per-stream local credit tracking is not clear. Is there a corresponding issue that explains why we need it?

see https://github.com/icerpc/icerpc-csharp-audit/issues/15

pepone:
Replaced by https://github.com/icerpc/icerpc-csharp/pull/4532

PR #4532, opened 2026-04-21
Add PauseWriterThreshold to bound per-stream Slic outbound buffering

pepone:
- Add a configurable PauseWriterThreshold option (default: 64 KB) to SlicTransportOptions. This option limits how much outbound payload data Slic buffers locally per stream, independently of the peer’s advertised flow-control window.
- Introduce per-stream local credit tracking with completion callbacks on the shared SlicDuplexConnectionWriter. Local credit is released only after bytes are written to the underlying IDuplexConnection, not when they move between Slic-internal buffers.

Replaces https://github.com/icerpc/icerpc-csharp/pull/4530. The ping-flood protection from that PR will be handled separately.

This change introduces a per-stream cap on locally buffered data, independent of the flow-control window.

The peer’s window size represents how much data the peer is willing to buffer, but it does not constrain how much data the sender may accumulate locally before it is written to the connection. Without an additional limit, a sender can buffer large amounts of data per stream, leading to excessive memory usage.

PauseWriterThreshold addresses this by bounding local buffering per stream.

See also: icerpc/icerpc-csharp-audit#15

### DIFF
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -38,6 +38,9 @@ internal class SlicConnection : IMultiplexedConnection
     /// <summary>Gets the initial stream window size.</summary>
     internal int InitialStreamWindowSize { get; }
 
+    /// <summary>Gets the pause writer threshold for per-stream local buffering.</summary>
+    internal int PauseWriterThreshold { get; }
+
     /// <summary>Gets the window update threshold. When the window size is increased and this threshold reached, a <see
     /// cref="FrameType.StreamWindowUpdate" /> frame is sent.</summary>
     internal int StreamWindowUpdateThreshold => InitialStreamWindowSize / StreamWindowUpdateRatio;
@@ -537,6 +540,7 @@ internal class SlicConnection : IMultiplexedConnection
         _maxUnidirectionalStreams = options.MaxUnidirectionalStreams;
 
         InitialStreamWindowSize = slicOptions.InitialStreamWindowSize;
+        PauseWriterThreshold = slicOptions.PauseWriterThreshold;
         _localIdleTimeout = slicOptions.IdleTimeout;
         _maxStreamFrameSize = slicOptions.MaxStreamFrameSize;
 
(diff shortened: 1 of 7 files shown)


src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -754,13 +758,19 @@ internal async ValueTask<FlushResult> WriteStreamDataFrameAsync(
...
                 {
-                    sendCredit = await stream.AcquireSendCreditAsync(writeCts.Token).ConfigureAwait(false);
-                    Debug.Assert(sendCredit > 0);
+                    // Wait for local buffering credit. This blocks if this stream already has PauseWriterThreshold
+                    // bytes in-flight across Slic-owned buffers.
+                    int localCredit = await stream.AcquireLocalCreditAsync(writeCts.Token).ConfigureAwait(false);
+                    Debug.Assert(localCredit > 0);
+
+                    // Wait for peer-granted send credit. This blocks if the peer's flow-control window is exhausted.
+                    int peerCredit = await stream.AcquireSendCreditAsync(writeCts.Token).ConfigureAwait(false);
```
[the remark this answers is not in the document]
pepone: I think this is fine as is, because the stream is torn down after the cancelation and the credit is local to it.

externl (approved):
Looks good to me!

src/IceRpc/Transports/Slic/Internal/SlicStream.cs
```diff
@@ -103,6 +103,9 @@ internal SlicStream(SlicConnection connection, bool isBidirectional, bool isRemo
     /// cref="FrameType.StreamLast"/> frame to ensure enough send credit is available. If no send credit is available,
```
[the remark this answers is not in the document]
bernardnormier: You should add a doc-comment for each method.

src/IceRpc/Transports/Slic/Internal/SlicPipeWriter.cs
```diff
@@ -229,4 +237,40 @@ internal void ReceivedWindowUpdateFrame(int size)
...
+    }
+
+    /// <summary>Decrements local credit by the actual frame payload size.</summary>
+    /// <param name="size">The payload size of the frame.</param>
+    /// <remarks>Must be called while the local credit semaphore is held.</remarks>
+    internal void ConsumedLocalCredit(int size)
+    {
+        Debug.Assert(_localCreditSemaphore.CurrentCount == 0);
+
+        int newLocalCredit = Interlocked.Add(ref _localCredit, -size);
```
bernardnormier, line 257: My understanding is if you use Interlocked, you can but should not use volatile.

src/IceRpc/Transports/Slic/Internal/SlicPipeWriter.cs
```diff
@@ -153,9 +156,14 @@ public override async ValueTask<FlushResult> WriteAsync(
         }
     }
 
+    /// <inheritdoc/>
+    void SlicDuplexConnectionWriter.ICompletionCallback.WriteCompleted(int creditBytes) =>
```
bernardnormier, line 160: We rarely if ever use this syntax for the implementation of interface methods.

bernardnormier (commented):
We need a design note for this local credit mechanism. Why do we need it? How does it work?

src/IceRpc/Transports/Slic/Internal/SlicPipeWriter.cs
pepone: This follow the existing design for `_peerWindowSize`, but it is probably fine to drop it as you say.

src/IceRpc/Transports/Slic/Internal/SlicPipeWriter.cs
pepone: This is required because is an internal interface in SlicDuplexConnectionWriter

pepone:
> We need a design note for this local credit mechanism. Why do we need it? How does it work?

Where should we put this?

src/IceRpc/Transports/Slic/Internal/SlicStream.cs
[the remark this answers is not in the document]
pepone: Fixed

src/IceRpc/Transports/Slic/Internal/SlicPipeWriter.cs
pepone: removed volatile for both.

bernardnormier:
> > We need a design note for this local credit mechanism. Why do we need it? How does it work?
> 
> Where should we put this?

Assuming it's reasonably short, as a comment on the issue.

pepone:
> Assuming it's reasonably short, as a comment on the issue.

Commented on the issue.

PR #4533, opened 2026-04-21
Fix idle timeout handling for large peer values

### LINKED ISSUE #4409: [Audit-Medium] IdleTimeout peer parameter not range-checked for large…

### Medium: `IdleTimeout` peer parameter not range-checked for large values

Affected code:

- [src/IceRpc/Transports/Slic/Internal/SlicConnection.cs:925-933](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs#L925)

Description:

Only `TimeSpan.Zero` is rejected. A peer can advertise an `IdleTimeout` of `int.MaxValue` milliseconds (~24 days), effectively disabling our liveness detection on this connection while the peer holds resources on our side. `_peerIdleTimeout` then loses its protective function in the `idleTimeout = min(local, peer)` calculation when the local timeout is infinite (the documented default).

Impact: Weakens liveness detection against a malicious or defective peer.

Recommendation: cap the accepted peer idle timeout at the local configured value or at a fixed policy ceiling (e.g. 10 minutes).

---

**Source report:** [src-IceRpc-Transports-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc-Transports-audit-2026-04-14.md) (finding ``IdleTimeout` peer parameter not range-checked for large values`)

pepone:
## Summary

- Fix the `min(local, peer)` idle timeout comparison to correctly treat `Timeout.InfiniteTimeSpan` as greater than any positive timeout. Previously, `InfiniteTimeSpan` (-1 ms) would win the `<` comparison, causing the effective timeout to be infinite when only the local side was configured as infinite and the peer advertised a valid timeout.
- Validate `IdleTimeout` setter to reject negative values other than `Timeout.InfiniteTimeSpan`.

Fixes #4409

### DIFF
--- a/src/IceRpc/Transports/Slic/SlicTransportOptions.cs
+++ b/src/IceRpc/Transports/Slic/SlicTransportOptions.cs
@@ -7,16 +7,23 @@ namespace IceRpc.Transports.Slic;
 public sealed record class SlicTransportOptions
 {
     /// <summary>Gets or sets the idle timeout. This timeout is used to monitor the transport connection health. If no
-    /// data is received within the idle timeout period, the transport connection is aborted.
-    /// </summary>
-    /// <value>The idle timeout. Defaults to <c>30</c> s.</value>
+    /// data is received within the idle timeout period, the transport connection is aborted. Use
+    /// <see cref="Timeout.InfiniteTimeSpan" /> to disable idle timeout monitoring.</summary>
+    /// <value>The idle timeout. It must be positive or <see cref="Timeout.InfiniteTimeSpan" />.
+    /// Defaults to <c>30</c> s.</value>
     public TimeSpan IdleTimeout
     {
         get => _idleTimeout;
-        set => _idleTimeout = value != TimeSpan.Zero ? value :
-            throw new ArgumentException(
-                $"The value '0' is not a valid for {nameof(IdleTimeout)} property.",
-                nameof(value));
+        set
+        {
+            if (value != Timeout.InfiniteTimeSpan && value <= TimeSpan.Zero)
+            {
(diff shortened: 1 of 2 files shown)


externl (approved):

src/IceRpc/Transports/Slic/SlicTransportOptions.cs
```diff
@@ -7,16 +7,23 @@ namespace IceRpc.Transports.Slic;
 public sealed record class SlicTransportOptions
 {
     /// <summary>Gets or sets the idle timeout. This timeout is used to monitor the transport connection health. If no
-    /// data is received within the idle timeout period, the transport connection is aborted.
-    /// </summary>
-    /// <value>The idle timeout. Defaults to <c>30</c> s.</value>
+    /// data is received within the idle timeout period, the transport connection is aborted. Use
+    /// <see cref="Timeout.InfiniteTimeSpan" /> to disable idle timeout monitoring.</summary>
```
bernardnormier: Is this comment correct? If you pick InfiniteTimeSpan, my understanding is you use the peer's idle timeout. This doesn't disable the idle timeout.

bernardnormier (approved):

src/IceRpc/Transports/Slic/SlicTransportOptions.cs
pepone: You are right, the timeout is negotiated, I will update the comment.

PR #4534, opened 2026-04-21
Bound baggage entry count in TelemetryMiddleware

### LINKED ISSUE #4477: [Audit-Low] baggage decoding is unbounded and amplifies on request fo…

### Low: baggage decoding is unbounded and amplifies on request forwarding

Affected code:

- [src/IceRpc.Telemetry/TelemetryMiddleware.cs:82-93](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Telemetry/TelemetryMiddleware.cs#L82) — `DecodeSequence` over baggage entries + `Reverse()` + loop of `AddBaggage`
- [src/IceRpc.Telemetry/TelemetryMiddleware.cs:58](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Telemetry/TelemetryMiddleware.cs#L58) — decoder constructed with default `maxCollectionAllocation = 8 * buffer.Length`
- [src/IceRpc.Telemetry/TelemetryInterceptor.cs:102-108](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Telemetry/TelemetryInterceptor.cs#L102) — outgoing path re-encodes `activity.Baggage` on every request

Description:

The middleware places no explicit upper bound on the number of baggage entries. The default `maxCollectionAllocation` of `8 * buffer.Length` bounds *byte* allocation but not *entry count*: a peer can pack the `TraceContext` field with thousands of empty-key / empty-value entries (2 bytes each), fitting ~500 entries per KiB. Each entry becomes:

- A `(string Key, string Value)` tuple allocated during `DecodeSequence`.
- A materialized position in the `Reverse()` pass.
- An `Activity.Baggage` linked-list node.

More concerning is the amplification pattern: if the service forwards the request through an outgoing pipeline that includes `TelemetryInterceptor`, the 500+ baggage entries are re-enumerated and re-encoded on every hop. Baggage propagation is an OpenTelemetry feature — clipping it isn't universally safe — but "accept unbounded size" is the wrong default for a middleware that can be exposed to untrusted peers.

Impact:

- Memory pressure on busy servers under a crafted `TraceContext` field.
- Extra CPU and wire bandwidth on every forwarded hop.
- No peer-exploitable DoS beyond the cap on the field size itself, but a measurable amplification when the service fan-outs.

Recommendation:

- Add an explicit cap on baggage entry count (OpenTelemetry spec recommends 180 entries as a soft limit). Translate overage into a discard, a clipping, or an `InvalidDataException` with a clear message.
- Consider caching the encoded form of `activity.Baggage` on the interceptor side so the serialization cost isn't paid per-hop.

Severity: Low.

---

**Source report:** [src-IceRpc.Telemetry-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.Telemetry-audit-2026-04-14.md) (finding `baggage decoding is unbounded and amplifies on request forwarding`)

pepone:
Fixes #4477.

### DIFF
--- a/src/IceRpc.Telemetry/TelemetryMiddleware.cs
+++ b/src/IceRpc.Telemetry/TelemetryMiddleware.cs
@@ -81,22 +81,17 @@ public class TelemetryMiddleware : IDispatcher
         // metadata is less damaging than failing the RPC, and silent clipping matches the behavior of
         // OpenTelemetry .NET's and Python's BaggagePropagator on incoming headers. Only the first
         // MaxBaggageEntries entries are read from the buffer; the remainder is left unconsumed.
+        //
+        // Activity.Baggage's enumeration order is undocumented, so duplicate-key resolution across the
+        // wire is inherently undefined on both ends — we don't attempt to preserve it.
         int count = decoder.DecodeSize();
         int kept = Math.Min(count, TelemetryInterceptor.MaxBaggageEntries);
 
-        var baggage = new (string Key, string Value)[kept];
         for (int i = 0; i < kept; i++)
         {
             string key = decoder.DecodeString();
             string value = decoder.DecodeString();
-            baggage[i] = (key, value);
-        }
-
-        // Restore in reverse order to keep the order in which the peer adds baggage entries,
-        // this is important when there are duplicate keys.
-        for (int i = kept - 1; i >= 0; i--)
-        {
-            activity.AddBaggage(baggage[i].Key, baggage[i].Value);
+            activity.AddBaggage(key, value);
         }
     }
 }
--- a/src/IceRpc.Telemetry/TelemetryMiddleware.cs
+++ b/src/IceRpc.Telemetry/TelemetryMiddleware.cs
@@ -76,19 +76,16 @@ public class TelemetryMiddleware : IDispatcher
         // Read TraceState encoded as a string
         activity.TraceStateString = decoder.DecodeString();
 
-        // Decode the baggage sequence with an explicit entry-count cap. Well-behaved peers clip their
-        // outgoing baggage at TelemetryInterceptor.MaxBaggageEntries; exceeding the cap here indicates
-        // a malformed or malicious sender. Check the count before decoding the entries so we don't
-        // allocate or deserialize anything for a bogus sequence.
+        // Decode the baggage sequence, silently clipping to MaxBaggageEntries. OpenTelemetry SDKs
+        // follow a strict no-throw policy for observability operations: losing a piece of contextual
+        // metadata is less damaging than failing the RPC, and silent clipping matches the behavior of
+        // OpenTelemetry .NET's and Python's BaggagePropagator on incoming headers. Only the first
+        // MaxBaggageEntries entries are read from the buffer; the remainder is left unconsumed.
         int count = decoder.DecodeSize();
-        if (count > TelemetryInterceptor.MaxBaggageEntries)
-        {
-            throw new InvalidDataException(
-                $"The baggage sequence has {count} entries, which exceeds the maximum of {TelemetryInterceptor.MaxBaggageEntries}.");
-        }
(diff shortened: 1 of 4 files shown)


src/IceRpc.Telemetry/TelemetryMiddleware.cs
```diff
@@ -76,19 +76,30 @@ internal static void RestoreActivityContext(ReadOnlySequence<byte> buffer, Activ
...
-                string key = decoder.DecodeString();
-                string value = decoder.DecodeString();
-                return (key, value);
-            });
+        // Decode the baggage sequence with an explicit entry-count cap. Well-behaved peers clip their
+        // outgoing baggage at TelemetryInterceptor.MaxBaggageEntries; exceeding the cap here indicates
+        // a malformed or malicious sender. Check the count before decoding the entries so we don't
+        // allocate or deserialize anything for a bogus sequence.
+        int count = decoder.DecodeSize();
+        if (count > TelemetryInterceptor.MaxBaggageEntries)
```
externl: What if both sides have different configuration? Seems aggressive to throw an exception here.

pepone: Yes, I reworked this to cap things in both side at 64 entries, which is the limit in the Spec, .NET uses 180 for its cap which will be problematic once we start adding more mappings which have implementations that adhere to the spec.

We should also not fail the request if there is some telemetry extra data, just ignoring seems better.

externl (approved):

src/IceRpc.Telemetry/TelemetryMiddleware.cs
```diff
@@ -76,18 +76,21 @@ internal static void RestoreActivityContext(ReadOnlySequence<byte> buffer, Activ
...
+        // Decode the baggage sequence, silently clipping to MaxBaggageEntries. OpenTelemetry SDKs
+        // follow a strict no-throw policy for observability operations: losing a piece of contextual
+        // metadata is less damaging than failing the RPC, and silent clipping matches the behavior of
+        // OpenTelemetry .NET's and Python's BaggagePropagator on incoming headers. Only the first
+        // MaxBaggageEntries entries are read from the buffer; the remainder is left unconsumed.
+        //
+        // Activity.Baggage's enumeration order is undocumented, so duplicate-key resolution across the
+        // wire is inherently undefined on both ends — we don't attempt to preserve it.
+        int count = decoder.DecodeSize();
+        int kept = Math.Min(count, TelemetryInterceptor.MaxBaggageEntries);
```
InsertCreativityHere, line 88: If `count > MaxBaggageEntries`, won't this leave a bunch of entries sitting in the buffer, undecoded, screwing up future reads? Or is `buffer` only for decoding the baggage and nothing else will care if we stop reading half-way through the buffer?

InsertCreativityHere (approved):

src/IceRpc.Telemetry/TelemetryMiddleware.cs
pepone: The activity context is encoded in the `RequestFieldKey.TraceContext` request field. Then this baggage is a dictionary at the end of this field, there is nothing else to decode from the field.

InsertCreativityHere: Okay, so only real risk would be if we added another field after `Baggage` then in the future.
Seems not worth worrying about to me.

pepone: Here the trace context follows the W3C standard, is unlikely this will change as the standard is used to allow to pass the context across heterogeneous systems.

PR #4537, opened 2026-04-22
Make ColocListener.DisposeAsync safe under concurrent calls

### LINKED ISSUE #4483: [Audit-Medium] ColocListener.DisposeAsync is not safe under concurren…

### Medium: `ColocListener.DisposeAsync` is not safe under concurrent calls

Affected code:

- [src/IceRpc.Transports.Coloc/Internal/ColocListener.cs:66-90](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Transports.Coloc/Internal/ColocListener.cs#L66) — `DisposeAsync`

Description:

`DisposeAsync` uses `_disposeCts.IsCancellationRequested` as its idempotency guard at [line 68](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Transports.Coloc/Internal/ColocListener.cs#L68). The check is racy: two concurrent `DisposeAsync` callers can both observe `false`, both proceed to `_disposeCts.Cancel()` (idempotent, fine), both call `_channel.Writer.Complete()` (idempotent after the first), both drain the channel (benign — second finds the channel empty), and both reach `_disposeCts.Dispose()` at [line 87](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Transports.Coloc/Internal/ColocListener.cs#L87). `CancellationTokenSource.Dispose` is not documented as safely callable more than once; the second call can produce `ObjectDisposedException` or undefined behavior depending on internal state.

The `DisposeAsync` contract in .NET allows idempotent multiple calls. In tests or shutdown paths where multiple owners independently dispose a shared listener, the race is real.

Impact:

- Concurrent dispose can throw `ObjectDisposedException` from the second caller.
- With unlucky timing, a CTS internally mid-dispose can race with an in-flight `CreateLinkedTokenSource(_disposeCts.Token, …)` inside `AcceptAsync`, producing further exceptions.

Recommendation:

- Guard `DisposeAsync` with `Interlocked.CompareExchange` on an `int` flag so only one caller proceeds.
- Or use `lock (_mutex)` around the state transitions.
- Don't call `_disposeCts.Dispose()` synchronously if `AcceptAsync` is still live — wait for in-flight accept tasks to settle, then dispose. Alternatively, skip disposing `_disposeCts` entirely (CTS without a timer is GC-collectable).

Severity: Medium.

---

**Source report:** [src-IceRpc.Transports.Coloc-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.Transports.Coloc-audit-2026-04-14.md) (finding ``ColocListener.DisposeAsync` is not safe under concurrent calls`)

pepone:
## Summary

Use the `_disposeTask` + `lock` pattern so concurrent callers await the same task, which completes only after all cleanup is done.

Fixes #4483

### DIFF
--- a/src/IceRpc.Transports.Coloc/Internal/ColocListener.cs
+++ b/src/IceRpc.Transports.Coloc/Internal/ColocListener.cs
@@ -13,8 +13,12 @@ internal class ColocListener : IListener<IDuplexConnection>
 {
     public TransportAddress TransportAddress { get; }
 
+    [SuppressMessage(
+        "Usage",
+        "CA2213:Disposable fields should be disposed",
+        Justification = "Disposing this CTS races with AcceptAsync creating a linked token source from its Token; a CTS with no timer is safely reclaimed by the GC.")]
     private readonly CancellationTokenSource _disposeCts = new();
-    private Task? _disposeTask;
+    private bool _disposed;
     private readonly Action<ColocListener> _onDispose;
     private readonly Lock _mutex = new();
     private readonly EndPoint _networkAddress;
@@ -33,7 +37,7 @@ internal class ColocListener : IListener<IDuplexConnection>
         CancellationTokenSource cts;
         lock (_mutex)
         {
-            ObjectDisposedException.ThrowIf(_disposeTask is not null, this);
+            ObjectDisposedException.ThrowIf(_disposed, this);
             cts = CancellationTokenSource.CreateLinkedTokenSource(_disposeCts.Token, cancellationToken);
         }
         using var _ = cts;
@@ -74,12 +78,12 @@ internal class ColocListener : IListener<IDuplexConnection>
     {
         lock (_mutex)
         {
-            _disposeTask ??= PerformDisposeAsync();
-        }
-        return new(_disposeTask);
+            if (_disposed)
+            {
+                return default;
+            }
+            _disposed = true;
 
-        Task PerformDisposeAsync()
-        {
             // Notify the owner (e.g. the server transport) so it can release its reference to this listener.
             _onDispose(this);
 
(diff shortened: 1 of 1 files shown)


externl (approved):

src/IceRpc.Transports.Coloc/Internal/ColocListener.cs
```diff
@@ -66,31 +72,34 @@ internal class ColocListener : IListener<IDuplexConnection>
...
         {
-            // Dispose already called.
-            return default;
+            _disposeTask ??= PerformDisposeAsync();
         }
+        return new(_disposeTask);
 
-        // Notify the owner (e.g. the server transport) so it can release its reference to this listener.
-        _onDispose(this);
+        Task PerformDisposeAsync()
```
bernardnormier: I don't think this pattern is desirable for a synchronous Dispose. There is no real "dispose task" since dispose is synchronous.

bernardnormier (commented):
I think we should use a flag as suggested by the Audit.

Also, we should remove the  `_disposeCts.Dispose();` (+ comment) as suggested by the Audit.

pepone:
>  Guard DisposeAsync with Interlocked.CompareExchange on an int flag so only one caller proceeds.

The problem I see with just using a flag is that DisposeAsync can return before object is disposed. For example concurrent dispose calls can not warrantee the object is disposed after return and reuse the port.

Updated the code to hold the mutex for the whole DisposeAsync body.

externl (approved):

bernardnormier (approved):
A volatile bool _disposed would not be sufficient here?

pepone:
> A volatile bool _disposed would not be sufficient here?

The mutex ensures that a second call to Dispose doesn't return before the dispose finished

Issue #4410, opened 2026-04-15
[Audit-Medium] No upper bound on peer-advertised Slic connection para…


pepone:
> **AI-generated audit finding** — this issue was opened from an automated security/correctness audit. It **has not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

### Medium: No upper bound on peer-advertised Slic connection parameters

Affected code:

- [src/IceRpc/Transports/Slic/Internal/SlicConnection.cs:899-993](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs#L899) — `DecodeParameters`

Description:

`MaxStreamFrameSize` and `InitialStreamWindowSize` are validated with a lower bound of 1024 bytes but have no upper bound (only the `int` range imposed by `DecodeParamValue`). `PeerMaxStreamFrameSize` and `PeerInitialStreamWindowSize` feed send-sizing math in `WriteStreamDataFrameAsync` ([SlicConnection.cs:762](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs#L762)) and flow-control accounting in `SlicPipeWriter`. A peer declaring `2 GiB` values does not cause immediate local allocation, but it disables the receive-window protection on the *sending* side and forces `int` arithmetic close to overflow boundaries.

Impact: Low to Medium — no direct memory blow-up on the receiver, but the invariant "peer-advertised sizes are sane" is load-bearing elsewhere in the implementation.

Recommendation: enforce a sensible upper bound (e.g. `MaxStreamFrameSize <= 16 MiB`, `InitialStreamWindowSize <= 16 MiB`) at decode time and reject as `InvalidDataException`.

---

**Source report:** [src-IceRpc-Transports-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc-Transports-audit-2026-04-14.md) (finding `No upper bound on peer-advertised Slic connection parameters`)

**Severity (auditor-assigned):** Medium

pepone:
Fixed in #4541

PR #4541, opened 2026-04-22
Bound Slic stream frame size and fix peer window overflow check

pepone:
## Summary

Addresses #4410.

- Cap `MaxStreamFrameSize` at 16,777,215 (HTTP/2 `SETTINGS_MAX_FRAME_SIZE` ceiling) in the local setter and in peer Initialize decode. A peer advertising a very large value would otherwise disable head-of-line protection for multiplexed streams on a TCP-based transport.
- Reject inbound stream frames whose size exceeds the locally advertised `MaxStreamFrameSize` in `ReadStreamDataFrameAsync`.
- Fix the cumulative peer-window overflow check in `SlicPipeWriter.ReceivedWindowUpdateFrame`: the previous `> MaxWindowSize` check was unreachable because `Interlocked.Add` wraps on overflow. The new `<= 0` check catches the wrap (valid because `_peerWindowSize >= 0` and `size > 0` by invariant).

## Test plan

- [x] New tests cover the local setter bound, peer-advertised bound, window-update overflow, and inbound-frame-size rejection
- [x] `dotnet test tests/IceRpc.Tests/IceRpc.Tests.csproj --filter "FullyQualifiedName~Slic"` - 188 passed


externl (approved):
Looks good!

src/IceRpc/Transports/Slic/SlicTransportOptions.cs
```diff
@@ -47,21 +47,33 @@ public int InitialStreamWindowSize
...
+            value;
     }
 
+    // Upper bound on MaxStreamFrameSize (local and peer-advertised). Matches HTTP/2's SETTINGS_MAX_FRAME_SIZE
+    // ceiling. Larger frames worsen head-of-line blocking across multiplexed streams without improving
+    // throughput at any realistic link speed.
+    internal const int MaxFrameSize = 16_777_215;
+
     // We use the HTTP/2 maximum window size (2GB).
     internal const int MaxWindowSize = int.MaxValue;
```
bernardnormier, line 69: You should review all usages of MaxWindowSize. For example, the `value > MaxWindowSize ? ...` above is bogus since value is an int.

It's likely we should simply remove this constant.

src/IceRpc/Transports/Slic/SlicTransportOptions.cs
```diff
@@ -47,21 +47,33 @@ public int InitialStreamWindowSize
...
-                nameof(value));
+                $"The {nameof(MaxStreamFrameSize)} value cannot be larger than {MaxFrameSize}.",
+                nameof(value)) :
+            value;
     }
 
+    // Upper bound on MaxStreamFrameSize (local and peer-advertised). Matches HTTP/2's SETTINGS_MAX_FRAME_SIZE
+    // ceiling. Larger frames worsen head-of-line blocking across multiplexed streams without improving
+    // throughput at any realistic link speed.
+    internal const int MaxFrameSize = 16_777_215;
```
bernardnormier: `MaxStreamFrameSizeCeiling` would be clearer. 

With the current names, it seems MaxStreamFrameSize and MaxFrameSize are different but related settings.

bernardnormier (approved):

src/IceRpc/Transports/Slic/SlicTransportOptions.cs
pepone: Removed the dead checks, but kept the constant randomly use int.MaxValue would be unclear and the window size is sent as a `varuint62` over the wire.

PR #4542, opened 2026-04-23
Document the no-copy contract on RequestContextFeature

### LINKED ISSUE #4454: [Audit-Low] outbound request context is encoded from a live mutable d…

### Low: outbound request context is encoded from a live mutable dictionary at send time — **CONFIRMED**

Affected code:

- [src/IceRpc.RequestContext/RequestContextInterceptor.cs:24](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.RequestContext/RequestContextInterceptor.cs#L24) — captures `IRequestContextFeature.Value` into `context`
- [src/IceRpc.RequestContext/RequestContextInterceptor.cs:35-58](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.RequestContext/RequestContextInterceptor.cs#L35) — stores closures that dereference `context` at encode time (deferred)
- [src/IceRpc/OutgoingFieldValue.cs](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc/OutgoingFieldValue.cs) — field value is an `Action<IBufferWriter<byte>>`, invoked later by the protocol connection during frame serialization

Verification:

Confirmed. The closure captures the `context` reference, not a snapshot. The protocol connection invokes the closure asynchronously during frame serialization (well after `InvokeAsync` returned to the caller). If the caller mutates or reuses the underlying dictionary before the frame is encoded onto the wire, the transmitted context can:

- Contain entries the caller did not intend to send,
- Miss entries the caller did intend to send, or
- Throw `InvalidOperationException` ("Collection was modified") during enumeration, which then surfaces as a send failure.

The risk is especially real when the same dictionary instance is used as the backing store for multiple invocations (a common pattern with a singleton context feature).

Impact:

- Cross-request contamination when a shared dictionary is reused.
- Nondeterministic metadata on concurrent or delayed sends.
- Send-time exceptions that surface well after the caller observed a "successful" `InvokeAsync` setup.

Recommendation:

- Snapshot the dictionary at interception time: `var snapshot = context.ToDictionary(kv => kv.Key, kv => kv.Value);` (or `ImmutableDictionary.CreateRange(context)`) and capture the snapshot in the closure.
- `OutgoingFieldValue`'s API could be tightened upstream to accept a `ReadOnlySequence<byte>` or a pre-encoded payload, eliminating deferred-encode hazards entirely, but that's out of scope for this package.
- Add a regression test that mutates the context dictionary between `InvokeAsync` return and the underlying send.

---

**Source report:** [src-IceRpc.RequestContext-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.RequestContext-audit-2026-04-14.md) (finding `outbound request context is encoded from a live mutable dictionary at send time — **CONFIRMED**`)

pepone:
The outgoing request context is encoded lazily, when the request is sent. If the caller continues mutating the dictionary it handed to `RequestContextFeature` after install, those mutations can race with the encoding and change what gets written on the wire (#4454).

Original approach was to snapshot the dictionary in the constructor. Per review (the snapshot would also drop a custom key comparer if one was supplied, and forces an allocation on every construction), pivot to **document the no-copy contract** instead and put the snapshot responsibility on the caller:

- Tighten the `RequestContextFeature(IDictionary)` doc comment to state that the feature stores the reference (no copy), and the caller must not mutate the dictionary after handing it over.
- Add `<remarks>` and a copy-on-write `<example>` on `IRequestContextFeature` showing how an interceptor that needs to add or remove entries should build a fresh dictionary and install a new feature, rather than mutating `Value` in place.

Fixes #4454


src/IceRpc/Features/RequestContextFeature.cs
```diff
@@ -24,9 +24,11 @@ public string this[string key]
...
     public RequestContextFeature() => Value = new Dictionary<string, string>();
 
-    /// <summary>Constructs a request context feature with the specified dictionary.</summary>
-    /// <param name="dictionary">The dictionary that the new feature will hold.</param>
-    public RequestContextFeature(IDictionary<string, string> dictionary) => Value = dictionary;
+    /// <summary>Constructs a request context feature with a copy of the specified dictionary.</summary>
+    /// <param name="dictionary">The source dictionary. Its entries are copied into the new feature so that later
+    /// mutations of <paramref name="dictionary"/> do not affect this feature.</param>
+    public RequestContextFeature(IDictionary<string, string> dictionary) =>
+        Value = new Dictionary<string, string>(dictionary);
```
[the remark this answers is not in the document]
pepone: While this sound possible, the decoder always use the default comparer for decoding the request context and will reject such dictionary as mallformed.

src/IceRpc/Features/RequestContextFeature.cs
```diff
@@ -24,9 +24,11 @@ public string this[string key]
     /// <summary>Constructs an empty writeable request context feature.</summary>
     public RequestContextFeature() => Value = new Dictionary<string, string>();
 
-    /// <summary>Constructs a request context feature with the specified dictionary.</summary>
-    /// <param name="dictionary">The dictionary that the new feature will hold.</param>
-    public RequestContextFeature(IDictionary<string, string> dictionary) => Value = dictionary;
+    /// <summary>Constructs a request context feature with a copy of the specified dictionary.</summary>
+    /// <param name="dictionary">The source dictionary. Its entries are copied into the new feature so that later
+    /// mutations of <paramref name="dictionary"/> do not affect this feature.</param>
+    public RequestContextFeature(IDictionary<string, string> dictionary) =>
```
pepone: I'm not 100% sure of adding this defensive copy, maybe the better fix is to clearly document this ..., if you are going to modify just pass a copy.

InsertCreativityHere: Yes, I put off reviewing this because I also wasn't sure about making this copy 100% of the time...
I think it'd be better to put this responsibility on the user, instead of making a copy.

pepone: Reworked this PR it just update the docs, and includes an example for the correct mutation pattern.

bernardnormier (commented):
After further review, I think a "defensive" copy in the `public RequestContextFeature(IDictionary<string, string> dictionary)` constructor implementation is the best solution.

pepone:
> After further review, I think a "defensive" copy in the public

The same `RequestContextFeature` can be concurrently used by multiple requests, an interceptor that wants to update the context still needs to replace the context feature with a create a new dictionary.

I don't think this extra copy on the constructor improve things.

bernardnormier (commented):
> The same RequestContextFeature can be concurrently used by multiple requests,

I don't see this as a valid usage. A mutable feature should not be shared between concurrent requests. 

This is not unique to RequestContextFeature. For example, you can't (shouldn't) share an IServiceRequestFeature between requests.

On the server-side, the RequestContext middleware always creates a per-request RequestContextFeature. On the client-side, the RequestContextFeature is up to the application, so the application could incorrectly share a mutable RequestContextFeature between several requests. We should document "don't do that". 

Then, it's indeed questionable whether this defensive copy in the constructor is useful (vs simply documenting that we don't copy and adopt as is). The adopt-as-is has also the advantage of being able to adopt a read-only dictionary.

bernardnormier (commented):
> I don't see this as a valid usage. A mutable feature should not be shared between concurrent requests.

Needs more thinking. 

A good example is:
 - middleware creates RequestContextFeature
 - dispatch implementation sends multiple concurrent outgoing requests and passes along the IRequestContextFeature
 - interceptors update this IRequestContextFeature

The dispatch implementation would consider the IRequestContextFeature to be safe to pass along when with my rule it's not.

pepone:
Maybe the clear pattern is to update the feature getter to return `IReadOnlyDictionary`, no much point in having just a `get` that returns a mutable dictionary.

Issue #4416, opened 2026-04-15
[Audit-Low] Reserved and unknown compression-format values are silent…


pepone:
> **AI-generated audit finding** — this issue was opened from an automated security/correctness audit. It **has not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

### Low: Reserved and unknown compression-format values are silently ignored — **CONFIRMED**

Affected code:

- [src/IceRpc.Compressor/CompressorInterceptor.cs:78-98](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Compressor/CompressorInterceptor.cs#L78) — only `Brotli` and `Deflate` handled; `// else nothing to do`
- [src/IceRpc.Compressor/CompressorMiddleware.cs:56-70](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Compressor/CompressorMiddleware.cs#L56) — same pattern
- Slice definition: `slice/IceRpc/CompressionFormat.slice:8` — `NotCompressed = 0` reserved

Verification:

Confirmed. A peer can attach `CompressionFormat.NotCompressed` (value `0`, reserved by the Slice definition) or any unknown enum value to a request or response; the implementation silently accepts the payload as-is. The peer is thereby permitted to set a malformed field without any protocol-level rejection.

The concrete security impact is limited because an attacker who can craft a malformed field can generally also craft an absent field, but the behavior hides a protocol bug instead of surfacing it — any subtle client/server mismatch or enum-numbering regression becomes invisible.

Status: **Valid**. Severity: Low.

Recommendation:

- Validate the decoded format: reject `NotCompressed` and any value outside the supported set with an `InvalidDataException` or an RPC-level error.
- Add tests for `NotCompressed` and unknown enum values on both request and response paths.

---

**Source report:** [src-IceRpc.Compressor-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.Compressor-audit-2026-04-14.md) (finding `Reserved and unknown compression-format values are silently ignored — **CONFIRMED**`)

**Severity (auditor-assigned):** Low

pepone:
Closing as not planned.

The silent pass-through on the receive side is by design:

[...]

PR #4545, opened 2026-04-23
Make send-side compression format switch explicit

pepone:
## Summary
- Replace `else { Debug.Assert(_compressionFormat == Deflate); ... }` with `else if (_compressionFormat == Deflate) { ... }` in both `CompressorInterceptor.InvokeAsync` and `CompressorMiddleware.DispatchAsync`.
- The previous pattern silently treats any future `CompressionFormat` value as Deflate if someone updates the constructor's whitelist but forgets to update the send-side switch. The explicit `else if` makes the omission a no-op instead of a silent wrong compression.
- No behavior change — constructor still rejects unsupported formats upfront.

Noticed while reviewing #4416 (which is unrelated and closed as not planned: the receive-side pass-through is by design).

## Test plan
- [x] `dotnet build src/IceRpc.Compressor/IceRpc.Compressor.csproj` (clean)
- [x] `dotnet test tests/IceRpc.Compressor.Tests/IceRpc.Compressor.Tests.csproj` — 16/16 passed

### DIFF
--- a/src/IceRpc.Compressor/CompressorInterceptor.cs
+++ b/src/IceRpc.Compressor/CompressorInterceptor.cs
@@ -3,7 +3,6 @@
 using IceRpc.Extensions.DependencyInjection;
 using IceRpc.Features;
 using System.Buffers;
-using System.Diagnostics;
 using System.IO.Compression;
 using System.IO.Pipelines;
 using ZeroC.Slice.Codec;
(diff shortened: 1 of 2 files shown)


src/IceRpc.Compressor/CompressorInterceptor.cs
```diff
@@ -61,9 +60,8 @@ public async Task<IncomingResponse> InvokeAsync(OutgoingRequest request, Cancell
             {
                 request.Use(next => PipeWriter.Create(new BrotliStream(next.AsStream(), _compressionLevel)));
             }
-            else
+            else if (_compressionFormat == CompressionFormat.Deflate)
             {
-                Debug.Assert(_compressionFormat == CompressionFormat.Deflate);
```
pepone, line 66: CompressionFormat is unchecked, I think is better to assume unknown formats can be handle elsewhere in the pipeline. This will silently use deflate for an unknown format in Release.

It should be also more consistent with what we do for decompressing.

externl (approved):

bernardnormier (commented):
This PR is not correct.

_compressionFormat can only have two values. See the constructor of CompressorInterceptor / CompressorMiddleware.

pepone:
> _compressionFormat can only have two values. See the constructor of CompressorInterceptor / CompressorMiddleware.


You are right, I missed the constructor check, closed.

PR #4555, opened 2026-04-24
Parse protoc diagnostics reliably in ProtocTask

### LINKED ISSUE #4451: [Audit-Medium] protoc diagnostic parsing can throw inside the MSBuild…

### Medium: `protoc` diagnostic parsing can throw inside the MSBuild task instead of reporting a normal compiler error — **CONFIRMED**

Affected code:

- [src/IceRpc.Protobuf.Tools/ProtocTask.cs:146-166](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Protobuf.Tools/ProtocTask.cs#L146) — `LogEventsFromTextOutput`
- [src/IceRpc.Protobuf.Tools/ProtocTask.cs:149](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Protobuf.Tools/ProtocTask.cs#L149) — `int colonCount = singleLine.Count(c => c == ':');`
- [src/IceRpc.Protobuf.Tools/ProtocTask.cs:154-158](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Protobuf.Tools/ProtocTask.cs#L154) — `Split([':'], 4)` + `int.Parse(parts[1])` + `int.Parse(parts[2])`

Verification:

Confirmed. The parser assumes that any stderr line with ≥ 3 colons is a `file:line:column: message` diagnostic. On Windows, a typical absolute path already has one colon (`C:\…`), so a perfectly normal `C:\work\greeter.proto:9:1: syntax error` line has 4 colons. `Split([':'], 4)` then produces:

- `parts[0] = "C"`
- `parts[1] = "\\work\\greeter.proto"`
- `parts[2] = "9"`
- `parts[3] = "1: syntax error"`

`int.Parse(parts[1], CultureInfo.InvariantCulture)` at [line 156](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Protobuf.Tools/ProtocTask.cs#L156) throws `FormatException`, which the `ToolTask` surface then rethrows — the *schema error becomes a task crash*, losing the original diagnostic text.

Even on Unix, any message that happens to contain a colon in its body can mislead the column parse (e.g. an error mentioning "expected `;`: got `}`"). Less common but equally brittle.

Impact:

- Ordinary schema errors surface as task crashes on Windows instead of readable MSBuild diagnostics.
- Build failures become harder to diagnose and can bypass normal error reporting pipelines (IDE squiggles, etc.).

Recommendation:

- Use a regex anchored on `<source>:<line>:<column>:` where `<line>` and `<column>` are `\d+`, and fall back to the literal-log path on no match.
- Even in the structured branch, wrap the `int.Parse` calls in `TryParse` and fall back to `Log.LogError(singleLine)` on failure.
- Add regression tests with Windows drive-letter paths and messages whose body contains a colon.

---

**Source report:** [src-IceRpc.Protobuf.Tools-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.Protobuf.Tools-audit-2026-04-14.md) (finding ``protoc` diagnostic parsing can throw inside the MSBuild task instead of reporting a normal compiler error — **CONFIRMED**`)

pepone:
## Summary
- The previous colon-counting parser crashed on Windows drive-letter paths: a normal `C:\work\foo.proto:9:1: msg` has four colons, so `int.Parse("\\work\\foo.proto")` threw and the schema error surfaced as a task crash instead of a clickable diagnostic.
- Pass `--error_format=msvs` to protoc and rewrite its output (`file(line) : error|warning in column=N: msg`) into MSBuild's canonical form (`file(line,N): error|warning: msg`), then hand off to `base.LogEventsFromTextOutput`. The base `ToolTask` logger does the error/warning categorization via `CanonicalError`, so the override is a single `Regex.Replace` + delegation.

Fixes #4451


externl (approved):

src/IceRpc.Protobuf.Tools/ProtocTask.cs
```diff
@@ -14,7 +13,7 @@ namespace IceRpc.Protobuf.Tools;
 
 /// <summary>A MSBuild task to generate code from Protobuf files using <c>protoc</c> C# built-in generator and
 /// <c>protoc-gen-icerpc-csharp</c> generator.</summary>
-public class ProtocTask : ToolTask
+public partial class ProtocTask : ToolTask
```
pepone, line 16: partial is required for `GeneratedRegex` source generator.

src/IceRpc.Protobuf.Tools/ProtocTask.cs
```diff
@@ -141,29 +144,18 @@ protected override string GenerateFullPathToTool()
...
-        else
-        {
-            Log.LogError(singleLine);
-        }
-    }
+    // Rewrites protoc's --error_format=msvs output ("file(line) : error|warning in column=N: msg") into MSBuild's
+    // canonical diagnostic format ("file(line,N): error|warning: msg") so the base ToolTask logger can parse it via
+    // CanonicalError. Non-matching lines pass through unchanged for the base class to handle.
+    [GeneratedRegex(
+        @"^(?<file>.+)\((?<line>\d+)\)\s*:\s*(?<severity>error|warning) in column=(?<column>\d+):\s*(?<message>.*)$")]
```
InsertCreativityHere: This should probably be a lazy match, not a greedy match (what it is right now).
```suggestion
    [GeneratedRegex(
        @"^(?<file>.+?)\((?<line>\d+)\)\s*:\s*(?<severity>error|warning) in column=(?<column>\d+):\s*(?<message>.*)$")]
```
But in the case where the message contains text which could match against earlier parts of the regex, neither is perfect. If there is any ambiguity, with a greedy match, extra stuff will end up in `file`, with a lazy match, extra stuff will end up in `message`.

Doing a greedy match will match the entire string against `.+`, then back-track it's cursor as little as needed to match the rest of the string. This can cause `file` to end up with extra stuff if the `message` contains text which looks like the other matches that come before it.

Doing a lazy match means it will only match as few characters as necessary (left-to-right). And since `(` is an uncommon path character, this should work very reliably. Also, if there are any ambiguities like I mentioned above, the extra cruff will be a part of `message`, not `file`, which is preferable.

InsertCreativityHere (approved):
Looks good to me!

I couldn't find the `protoc` documentation describing `"file(line) : error|warning in column=N: msg"`, but if it works, it works!

src/IceRpc.Protobuf.Tools/ProtocTask.cs
pepone: Reverted to lazy `.+?` in 3061138b0. Realistic paths with parens still parse correctly thanks to backtracking, and on ambiguous inputs lazy keeps the cruft in `message` rather than corrupting `file` — better failure mode for IDE jump-to-error.

PR #4556, opened 2026-04-27
Report diagnostics on malformed service interfaces instead of crashing

### LINKED ISSUE #4460: [Audit-Medium] malformed or version-skewed service interfaces can cra…

### Medium: malformed or version-skewed service interfaces can crash the generator instead of producing diagnostics — **CONFIRMED**

Affected code:

- [src/IceRpc.ServiceGenerator/Internal/ProtobufServiceMethod.cs:48-76](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/Internal/ProtobufServiceMethod.cs#L48) — `Debug.Assert(items.Length == 1)`; `method.Parameters[0].Type`; `Debug.Assert(genericType.TypeArguments.Length == 1)`; unchecked `(INamedTypeSymbol)method.ReturnType` cast
- [src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs:170-186](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs#L170) — `requestClass?.GetMembers()…FirstOrDefault` returns `null`; `Debug.Assert(decodeArgsMethod is not null)`; `decodeArgsMethod!.ReturnType` with null-forgiving operator
- [src/IceRpc.ServiceGenerator/Internal/SliceServiceMethod.cs:218-234](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/Internal/SliceServiceMethod.cs#L218) — same pattern

Verification:

Confirmed. In release builds `Debug.Assert` is a no-op and the null-forgiving `!` operator doesn't add a runtime check. For example:

- A hand-written `[RpcMethod("Op")]` on a parameterless method triggers `IndexOutOfRangeException` at [ProtobufServiceMethod.cs:56](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/Internal/ProtobufServiceMethod.cs#L56) (`method.Parameters[0]`).
- A Slice/Ice interface that doesn't have the generator-produced nested `Request` class triggers `NullReferenceException` at [SliceServiceMethod.cs:234](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/Internal/SliceServiceMethod.cs#L234) / [IceServiceMethod.cs:186](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/Internal/IceServiceMethod.cs#L186).
- A version-skewed attribute with a different constructor argument count hits the `items.Length == 1` assert and then crashes on the index access at [line 51](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/Internal/ProtobufServiceMethod.cs#L51).

Each crash propagates out of the generator and manifests as an opaque "source generator threw" MSBuild failure, not an actionable C# diagnostic.

Impact:

- Build breaks for hand-authored service interfaces or under package-version skew with no clear diagnostic.
- Developer experience degrades sharply when the code *almost* matches the expected IDL shape.

Recommendation:

- Replace `Debug.Assert` and naked casts with explicit validation; on shape mismatch emit a `Diagnostic` with an IceRPC-specific descriptor and return early (the factory should return `false` or the parser should skip the offending class).
- Add a `DiagnosticDescriptors` entry for each shape mismatch so users see a targeted message.

---

**Source report:** [src-IceRpc.ServiceGenerator-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.ServiceGenerator-audit-2026-04-14.md) (finding `malformed or version-skewed service interfaces can crash the generator instead of producing diagnostics — **CONFIRMED**`)

pepone:
## Summary
- The three `ServiceMethod` constructors (Protobuf/Ice/Slice) used `Debug.Assert` and the null-forgiving operator for shape validation. In release builds `Debug.Assert` is a no-op and `!` adds no runtime check, so a hand-authored `[RpcMethod]`/`[IceOperation]`/`[SliceOperation]` on a wrong-shape method, or a version mismatch between this generator and the IDL-side generator, propagated as an opaque ""source generator threw"" MSBuild error.
- Convert each constructor into a static `TryCreate` factory that validates up front, reports a targeted diagnostic, and returns `null` on shape mismatch. Plumb an `Action<Diagnostic>` reporter through `ServiceMethodFactory.TryCreate` so the parser simply skips offending methods.
- Three new descriptors: `IRSG0002` (unexpected operation attribute shape), `IRSG0003` (unexpected method signature), `IRSG0004` (missing generator-produced `Request` class).

Fixes #4460


pepone:
I think this adds quite a lot of complexity for an unlikely scenario, we depend on a exact version of the service-generator, and this errors are likely only possible during development when you are updating the genertors.

PR #4558, opened 2026-04-27
Migrate Telemetry examples from Zipkin to OTLP

pepone:
## Summary
- `OpenTelemetry.Exporter.Zipkin 1.14.0` trips a `NU1902` build break in CI for [GHSA-g94r-2vxg-569j](https://github.com/advisories/GHSA-g94r-2vxg-569j) (vuln lives in a transitive `OpenTelemetry.Api`). The fix is in 1.15.x, which also marks `AddZipkinExporter` as `[Obsolete]` — and the example projects use `TreatWarningsAsErrors`, so a plain bump turns the security warning into a deprecation error.
- Replace `OpenTelemetry.Exporter.Zipkin` with `OpenTelemetry.Exporter.OpenTelemetryProtocol 1.15.3` across the four Telemetry example projects (Slice + Protobuf, client + server) and use `AddOtlpExporter`.
- Update both READMEs to point the reader at a Jaeger all-in-one container, which accepts OTLP on `:4317` natively and exposes the trace UI on `:16686`.
- Drop the now-orphaned `zipkin.png` screenshots.


externl (approved):
Looks good!

externl (approved):

PR #4560, opened 2026-04-27
Dedup partial service-class declarations in the generator

### LINKED ISSUE #4458: [Audit-Low] partial service classes with [Service] declared on multip…

### Low: partial service classes with `[Service]` declared on multiple parts produce duplicate `AddSource` entries

Affected code:

- [src/IceRpc.ServiceGenerator/ServiceGenerator.cs:20-25](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/ServiceGenerator.cs#L20) — `ForAttributeWithMetadataName` produces one `ClassDeclarationSyntax` per attribute application
- [src/IceRpc.ServiceGenerator/ServiceGenerator.cs:47](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/ServiceGenerator.cs#L47) — `classes.Distinct()` uses reference equality on syntax nodes, so two partial declarations of the same type are distinct
- [src/IceRpc.ServiceGenerator/ServiceGenerator.cs:50-54](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/ServiceGenerator.cs#L50) — the hint name derives from `serviceClass.FullName`, which is the same for every partial declaration

Description:

If a user writes `[Service]` on more than one part of a partial class (which the attribute target `AttributeTargets.Class` permits), the generator iterates both declarations, both produce `ServiceClass` objects with the same fully-qualified name, and `context.AddSource("Same.g.cs", …)` is called twice with the same hint name. Roslyn treats duplicate source hint names as an `ArgumentException` from the generator.

With `[Service]` declared on only one partial part (the normal case) this does not fire — but no diagnostic steers users away from the multi-declaration pattern. The symptom is an opaque Roslyn exception in the build.

Impact:

- Edge-case, but produces a confusing build failure when it triggers.

Recommendation:

- Group `classes` by their declared `INamedTypeSymbol` (via the semantic model) and process each symbol once. The parser already retrieves the symbol at [Parser.cs:55](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/Internal/Parser.cs#L55); dedup on that.
- Or emit a diagnostic directing the user to declare `[Service]` on only one partial part.

---

**Source report:** [src-IceRpc.ServiceGenerator-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.ServiceGenerator-audit-2026-04-14.md) (finding `partial service classes with `[Service]` declared on multiple parts produce duplicate `AddSource` entries`)

pepone:
## Summary
- `ForAttributeWithMetadataName` emits one `ClassDeclarationSyntax` per attribute application. If a user writes `[Service]` on more than one part of a partial class, the parser previously processed every syntax node and the generator called `context.AddSource` twice with the same hint name, which Roslyn rejects with an opaque generator-threw `ArgumentException` at build time.
- Track classes by their resolved `INamedTypeSymbol` (using `SymbolEqualityComparer.Default`) and skip ones we've already processed. The parts together are one class — silent dedup is the natural behavior.

Fixes #4458

### DIFF
--- a/src/IceRpc.ServiceGenerator/Internal/Parser.cs
+++ b/src/IceRpc.ServiceGenerator/Internal/Parser.cs
@@ -43,6 +43,10 @@ internal sealed class Parser
         }
 
         var serviceDefinitions = new List<ServiceClass>();
+        // Multiple partial declarations of the same class with [Service] applied to more than one part each yield
+        // their own ClassDeclarationSyntax from ForAttributeWithMetadataName. Process each class symbol once so we
+        // don't AddSource the same hint name twice.
+        var seenClassSymbols = new HashSet<INamedTypeSymbol>(SymbolEqualityComparer.Default);
         // we enumerate by syntax tree, to minimize the need to instantiate semantic models (since they're expensive)
         foreach (IGrouping<SyntaxTree, ClassDeclarationSyntax> group in classes.GroupBy(x => x.SyntaxTree))
         {
@@ -53,7 +57,7 @@ internal sealed class Parser
 
                 SemanticModel semanticModel = _compilation.GetSemanticModel(classDeclaration.SyntaxTree);
                 INamedTypeSymbol? classSymbol = semanticModel.GetDeclaredSymbol(classDeclaration, _cancellationToken);
-                if (classSymbol is null)
+                if (classSymbol is null || !seenClassSymbols.Add(classSymbol))
                 {
                     continue;
                 }


pepone:
Closing as not applicable.

The audit's premise relies on the user being able to apply [Service] on multiple partial parts of the same class. ServiceAttribute is declared as `[AttributeUsage(AttributeTargets.Class, Inherited = false)]` — AllowMultiple defaults to false, so the C# compiler itself rejects this with CS0579 before the generator runs (verified empirically with a test that failed to compile). The duplicate-AddSource scenario is therefore unreachable, making the dedup defensive belt-and-braces at the wrong layer. See #4458.

PR #4565, opened 2026-04-28
Reject control chars in request headers

### LINKED ISSUE #4432: [Audit-Medium] LoggerMiddleware logs peer-controlled operation names…

### Medium: `LoggerMiddleware` logs peer-controlled operation names and `icerpc` request paths without sanitization — **CONFIRMED**

Affected code:

- [src/IceRpc.Logger/LoggerMiddleware.cs:39-44](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Logger/LoggerMiddleware.cs#L39) — logs `request.Path` and `request.Operation` verbatim on success
- [src/IceRpc.Logger/LoggerMiddleware.cs:49-52](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Logger/LoggerMiddleware.cs#L49) — same values logged on exception
- [src/IceRpc/IncomingRequest.cs](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc/IncomingRequest.cs) — `Operation` is an unconstrained wire-decoded string; `Path` on icerpc is a mutable string from the request header
- [src/IceRpc/Internal/IceRpcProtocolConnection.cs:1069](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1069) — assigns decoded header values directly to the request
- [src/IceRpc/Internal/IceProtocolConnection.cs:1486](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc/Internal/IceProtocolConnection.cs#L1486) — assigns decoded operation directly

Verification:

Confirmed. The middleware's `LoggerMessage` templates interpolate `{Operation}` and `{Path}` into the rendered message text. For console/file sinks — the most common logger back ends — this produces literal output containing any control characters the peer sent. A peer can craft an operation name like `"op\r\nINFO: fake admin login by root"` and produce a forged log line.

The `ice` path goes through an `Identity` round-trip that normalizes the path, so only the operation is peer-controlled there. For icerpc, both path and operation are attacker-shaped strings.

Impact:

- Log injection / log forging against text-based sinks.
- Corrupted audit trails and misleading operational diagnostics.
- Parser confusion in downstream log processors that assume one event per line.

Recommendation:

- Escape control characters in `Path` and `Operation` before logging. At minimum, replace CR/LF/tab and other non-printable ASCII with `\r` / `\n` / `\t` escape sequences.
- Consider a shared `SafeLogValue(string)` helper to apply the same rule uniformly across the codebase.
- Add regression tests that log requests with embedded newlines in `Operation`, and for icerpc, in `Path`.

---

**Source report:** [src-IceRpc.Logger-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.Logger-audit-2026-04-14.md) (finding ``LoggerMiddleware` logs peer-controlled operation names and `icerpc` request paths without sanitization — **CONFIRMED**`)

pepone:
Validate the path and operation name of incoming requests at decode time so peer-controlled control characters cannot reach loggers and other downstream text-based sinks.

Fix #4432

### DIFF
--- a/src/IceRpc/Internal/IceRpcProtocolConnection.cs
+++ b/src/IceRpc/Internal/IceRpcProtocolConnection.cs
@@ -1227,6 +1227,20 @@ internal sealed class IceRpcProtocolConnection : IProtocolConnection
         {
             var decoder = new SliceDecoder(buffer);
             var header = new IceRpcRequestHeader(ref decoder);
+
+            // Reject paths and operation names that contain characters outside the printable ASCII range. This
+            // matches the constraints already enforced by ServiceAddress and by the Slice identifier grammar, and
+            // prevents control characters (notably CR/LF/NUL) from flowing into loggers and other downstream sinks.
+            try
+            {
+                ServiceAddress.CheckPath(header.Path);
+                ServiceAddress.CheckOperation(header.Operation);
+            }
+            catch (FormatException exception)
+            {
+                throw new InvalidDataException(exception.Message, exception);
+            }
+
             (IDictionary<RequestFieldKey, ReadOnlySequence<byte>> fields, PipeReader? pipeReader) =
                 DecodeFieldDictionary(
                     ref decoder,
--- a/src/IceRpc/ServiceAddress.cs
+++ b/src/IceRpc/ServiceAddress.cs
@@ -490,7 +490,7 @@ public sealed record class ServiceAddress
         if (path.Length == 0 || path[0] != '/' || !IsValid(path, _notValidInPath))
         {
             throw new FormatException(
-                $"Invalid path '{path}'; a valid path starts with '/' and contains only unreserved characters, '%', and reserved characters other than '?' and '#'.");
+                $"Invalid path '{Sanitize(path)}'; a valid path starts with '/' and contains only unreserved characters, '%', and reserved characters other than '?' and '#'.");
         }
     }
 
@@ -500,6 +500,23 @@ public sealed record class ServiceAddress
     /// <returns><see langword="true" /> if <paramref name="value" /> is a valid parameter value; otherwise,
     /// <see langword="false" />.</returns>
     internal static bool IsValidParamValue(string value) => IsValid(value, _notValidInParamValue);
+
+    /// <summary>Checks if <paramref name="operation" /> is a valid operation name. A valid operation name contains
+    /// only printable ASCII characters (in the inclusive range <c>0x21</c> to <c>0x7E</c>) and may be empty.
+    /// </summary>
+    /// <param name="operation">The operation name to check.</param>
+    /// <exception cref="FormatException">Thrown if the operation name is not valid.</exception>
+    /// <remarks>This range covers every legal Slice identifier and rejects all C0 control characters as well as DEL
+    /// and any non-ASCII content. Empty operation names are accepted because <see cref="OutgoingRequest.Operation" />
+    /// defaults to the empty string.</remarks>
+    internal static void CheckOperation(string operation)
+    {
+        if (operation.AsSpan().IndexOfAnyExceptInRange(FirstValidChar, LastValidChar) != -1)
+        {
+            throw new FormatException(
+                $"Invalid operation name '{Sanitize(operation)}'; an operation name contains only printable ASCII characters.");
+        }
+    }
 
     /// <summary>"unchecked" constructor used by the Ice decoder when decoding a service address.
     /// </summary>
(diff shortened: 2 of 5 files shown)


externl (approved):

src/IceRpc/Internal/IceRpcProtocolConnection.cs
```diff
@@ -1227,6 +1227,20 @@ async Task<OutgoingResponse> PerformDispatchRequestAsync(
         {
             var decoder = new SliceDecoder(buffer);
             var header = new IceRpcRequestHeader(ref decoder);
+
+            // Reject paths and operation names that contain characters outside the printable ASCII range. This
+            // matches the constraints already enforced by ServiceAddress and by the Slice identifier grammar, and
+            // prevents control characters (notably CR/LF/NUL) from flowing into loggers and other downstream sinks.
+            try
+            {
+                ServiceAddress.CheckPath(header.Path);
```
externl, line 1236: I'm not sure that IceRPC has the same restriction as Ice here. i.e. ascii only characters.

pepone: For Path we already have this check in ServiceAddress, so it makes sense we don't accept paths that are not valid according to the ServiceAddress rules.

For Operation I guess we didn't really specify what a valid operation names is in the "icerpc" protocol, at the protocol level is just a string. And this should probably be handled in the integration layer if at all, for example Slice operations are derived from Slice identifiers and have the same limitations that the identifier has. Same for Protobuf, they are derived from the Protobuf IDL and have its own rules.

I think is probably wrong to check the operation name here, as the protocol layer doesn't specify what the operation name is, checking the path seems correct because ServiceAddress format is specified in the protocol.

externl: Ok, I couldn't find anywhere in the docs were we way we only accept ASCII chars (but maybe its there?). 

Is this check really required here? ServiceAddress' Path init function already does this check, no?

pepone: > Is this check really required here? ServiceAddress' Path init function already does this check, no?

ServiceAddress is not used here, we decode the Path as a string, using IceRpcRequestHeader https://github.com/icerpc/icerpc-csharp/blob/0909b4b5a3758f16b75fb358ac74fb8afcc80da1/slice/IceRpc/Internal/IceRpcDefinitions.slice#L21-L26

Then the path string is assigned to `IncomingRequest.Path`, another string without further validation.

https://github.com/icerpc/icerpc-csharp/blob/0909b4b5a3758f16b75fb358ac74fb8afcc80da1/src/IceRpc/IncomingRequest.cs#L43

Maybe is clear to move the validation to `IncomingRequest.Path` init

externl: Right. Yeah it would make sense to also have it there. 

Reminds me of https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate

src/IceRpc/ServiceAddress.cs
```diff
@@ -490,7 +490,7 @@ internal static void CheckPath(string path)
         if (path.Length == 0 || path[0] != '/' || !IsValid(path, _notValidInPath))
         {
             throw new FormatException(
-                $"Invalid path '{path}'; a valid path starts with '/' and contains only unreserved characters, '%', and reserved characters other than '?' and '#'.");
+                "Invalid path; a valid path starts with '/' and contains only unreserved characters, '%', and reserved characters other than '?' and '#'.");
```
InsertCreativityHere, line 493: I think printing the 'invalid path' (like we did before) would be a valuable piece of information for debugging.

I recognize there's a tradeoff where: if there's control characters, the message might look weird. but I think the benefit of seeing "oh, `𒀗𒀝/𒀱` was the bad path" outweighs the very unlikely situation where you've gotten control or formatting characters in here.

InsertCreativityHere (approved):
Code changes look good to me!

But I think I will blog about what comprises a valid operation name in Slice.
I also don't think checking it at the IceRpc level is quite right, but it fine for now.

I'd prefer operation names to just be "strings comprised of alphanumeric characters".
And not limit them to purely printable ASCII characters. The fact Slice identifiers are ASCII only is more a reflection of my inability to write a super cool parser : v) not of the Slice spec.

PR #4566, opened 2026-04-29
Escape cref value in generated <see> doc comments

### LINKED ISSUE #4494: [Audit-Low] DocCommentFormatter.FormatEntityCref doesn't escape name/…

### Low: `DocCommentFormatter.FormatEntityCref` doesn't escape name/namespace in the `cref` attribute value

Affected code:

- [src/ZeroC.Slice.Generator/DocCommentFormatter.cs:55-79](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/ZeroC.Slice.Generator/DocCommentFormatter.cs#L55) — builds `global::{entityNamespace}.{name}` and returns it unchanged
- [src/ZeroC.Slice.Generator/DocCommentFormatter.cs:46-53](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/ZeroC.Slice.Generator/DocCommentFormatter.cs#L46) — `FormatInlineLink` emits `$"""<see cref="{FormatEntityCref(...)}" />"""` with no escape

Description:

`FormatEntityCref` concatenates `entity.Namespace` and a computed `name`. Both can originate from CS-attribute values (via `cs::identifier` on the module for namespace, or `cs::type` on a `CustomType` for the type name itself). The returned string is then placed inside an XML `cref="..."` attribute without XML-escaping. If either value contains `"`, `<`, `>`, or `&`, the resulting XML doc comment is malformed.

The `XmlEscape` helper *is* defined in the same file and *is* used in several sibling paths (`FormatOverview`, `FormatInlineLink` for `UnresolvedCommentLink` and `TypeAlias`). `FormatEntityCref` is the outlier.

Impact:

- Generated XML doc comments break under unusual CS-attribute values.
- DocFX / `GenerateDocumentationFile=true` fail to parse.

Recommendation:

- Apply `XmlEscape` (or `System.Security.SecurityElement.Escape`) to both `entityNamespace` and `name` before splicing.
- Add a test with a `cs::type("MyClass<int>")` on a `CustomType` whose doc comment references it.

Severity: Low.

---

**Source report:** [src-ZeroC.Slice.Generator-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-ZeroC.Slice.Generator-audit-2026-04-14.md) (finding ``DocCommentFormatter.FormatEntityCref` doesn't escape name/namespace in the `cref` attribute value`)

pepone:
Fix #4494

### DIFF
--- a/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
+++ b/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
@@ -36,7 +36,22 @@ internal static class DocCommentFormatter
 
         foreach (CommentLink link in seeTags)
         {
-            if (link is ResolvedCommentLink r)
+            if (link is not ResolvedCommentLink r)
+            {
+                continue;
+            }
+
+            if (r.Entity is TypeAlias alias)
+            {
+                // Type aliases don't generate a C# type; map to the underlying C# type. We can't reliably
+                // produce a seealso for a generic mapped type without C# parsing, so skip those.
+                string mapped = alias.UnderlyingType.Type.ToTypeString(currentNamespace);
+                if (!mapped.Contains('<', StringComparison.Ordinal))
+                {
+                    yield return new CommentTag("seealso", "cref", mapped, "");
+                }
+            }
+            else
             {
                 yield return new CommentTag("seealso", "cref", FormatEntityCref(r.Entity, currentNamespace), "");
             }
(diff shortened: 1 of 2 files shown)


externl (approved):

InsertCreativityHere (commented):
I don't think this is correct.

The string that we generate here is fully contained within double quotes:
`<see cref="..." />` So where's the need to escape XML characters?
The only way to break the XML is to put a literal `\"` character in their `cs::type` or `cs::identifier`, which means they've gone insane. Not something we should protect against :)

Assuming sanity, there's no reason for `&`, `<`, `>` or `"` to ever appear in `cs::identifier`.
I don't think that `&` or `"` can ever validly be part of a type-name in C#, can they?
`<` and `>` can appear in `cs::type` for a generic type, but according to the docs, the correct behavior here isn't to XML escape them, it's to replace them with `{` and `}`:

externl:
> I don't think this is correct.
> 
> The string that we generate here is fully contained within double quotes: `<see cref="..." />` So where's the need to escape XML characters? The only way to break the XML is to put a literal `\"` character in their `cs::type` or `cs::identifier`, which means they've gone insane. Not something we should protect against :)
> 
> Assuming sanity, there's no reason for `&`, `<`, `>` or `"` to ever appear in `cs::identifier`. I don't think that `&` or `"` can ever validly be part of a type-name in C#, can they? `<` and `>` can appear in `cs::type` for a generic type, but according to the docs, the correct behavior here isn't to XML escape them, it's to replace them with `{` and `}`:

Seems XML is even more complicated than this. For an attribute value, `<` is bad, `>` is fine.

InsertCreativityHere:
If you search for `Dictionary{` on this page:
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#cref-attribute

So, instead of XML escaping, we should be ignoring `"` and `&`, and we should replace `<` and `>` with `{` and `}`.
The `{}` syntax is probably because DocFx wants to avoid this specific headache with XML : v)

pepone:
> Seems XML is even more complicated than this. For an attribute value, < is bad, > is fine.

I guess this is not a general XML rule, but a C# documentation one, isn't it?

externl:
See also: https://learn.microsoft.com/en-us/dotnet/desktop/xaml-services/xml-character-entities

pepone:
Replaced by #4573

PR #4570, opened 2026-04-29
Fail loudly on duplicate scoped identifiers in SymbolConverter

### LINKED ISSUE #4500: [Audit-Low] named.TryAdd silently drops duplicate scoped identifiers

### Low: `_named.TryAdd` silently drops duplicate scoped identifiers

Affected code:

- [src/ZeroC.Slice.Symbols/SymbolConverter.cs:53-59](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/ZeroC.Slice.Symbols/SymbolConverter.cs#L53) — the constructor builds `_named` via `TryAdd`; conflicts are discarded without notice

Description:

Two symbols with the same fully-scoped identifier (same `moduleScope::id`) in the `_named` index map to the same key. `TryAdd` keeps the first and drops the second silently. `slicec` shouldn't emit duplicates — it's the compiler's job to catch name collisions — but if two files share a scope and one defines a type already defined in the other, the second definition disappears from the symbol tree used by generators.

No diagnostic is emitted. The downstream generator sees a different tree from what the user wrote.

Impact:

- Silent drop of duplicate named types in the face of `slicec` bugs or multi-file scope collisions.
- No operator-visible signal when this happens.

Recommendation:

- Detect collisions explicitly and emit a `Diagnostic` at `DiagnosticLevel.Error`.
- Or assert that `TryAdd` must succeed (throwing `InvalidOperationException` with context) since this library operates on already-validated `slicec` output.

Severity: Low.

---

**Source report:** [src-ZeroC.Slice.Symbols-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-ZeroC.Slice.Symbols-audit-2026-04-14.md) (finding ``_named.TryAdd` silently drops duplicate scoped identifiers`)

pepone:
Fix #4500

### DIFF
--- a/src/ZeroC.Slice.Symbols/SymbolConverter.cs
+++ b/src/ZeroC.Slice.Symbols/SymbolConverter.cs
@@ -54,7 +54,7 @@ internal sealed class SymbolConverter
             {
                 if (GetNamedIdentifier(symbol) is string id)
                 {
-                    // Scoped IDs are warned to be unique by the compiler, we can safely use them as dictionary keys.
+                    // Scoped IDs are guaranteed to be unique by the compiler, we can safely use them as dictionary keys.
                     _named.Add($"{moduleScope}::{id}", (file, symbol));
                 }
             }
--- a/src/ZeroC.Slice.Symbols/SymbolConverter.cs
+++ b/src/ZeroC.Slice.Symbols/SymbolConverter.cs
@@ -54,7 +54,8 @@ internal sealed class SymbolConverter
             {
                 if (GetNamedIdentifier(symbol) is string id)
                 {
-                    _named.TryAdd($"{moduleScope}::{id}", (file, symbol));
+                    // Scoped IDs are warned to be unique by the compiler, we can safely use them as dictionary keys.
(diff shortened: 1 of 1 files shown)


src/ZeroC.Slice.Symbols/SymbolConverter.cs
```diff
@@ -54,7 +54,8 @@ private SymbolConverter(IEnumerable<Compiler.SliceFile> allFiles)
             {
                 if (GetNamedIdentifier(symbol) is string id)
                 {
-                    _named.TryAdd($"{moduleScope}::{id}", (file, symbol));
+                    // Scoped IDs are warned to be unique by the compiler, we can safely use them as dictionary keys.
```
externl:
```suggestion
                    // Scoped IDs are guaranteed to be unique by the compiler, we can safely use them as dictionary keys.
```

externl (approved):

InsertCreativityHere (approved):
Looks good to me!

PR #4571, opened 2026-04-29
Document whitespace-suppression behavior on CodeBlock

pepone:
Clarify `WriteLine` and `AddBlock` behavior for empty line, empty block


externl (approved):
Looks good!

PR #4574, opened 2026-04-30
Fix link resolution in symbols converter

pepone:
This PR reworks how doc comment links are handled in the symbols converter. We cannot resolve links until all symbols are converted, the eager resolution assume that there were not circular dependencies which is true for Slice definitions but not for links in the doc copmments.


externl (approved):
Looks good to me!

InsertCreativityHere (approved):
All looks good to me!

PR #4573, opened 2026-04-30
Map TypeAlias doc links to the underlying C# type

pepone:
Fix #4494

PR #4577, opened 2026-05-04
Correctly Generate Doc-Links to Generic Types, Instead of Code-Formatting

InsertCreativityHere:
This PR follows in the footsteps of #4573. When we generate a link to a generic type, we just use code-formatting instead to avoid all the headaches we found in #4566.

This PR updates our behavior to always generate a valid link, by massaging generic types with a little parser.
It replaces `<` and `>` with `{` and `}`, and replaces the actual type parameters with `T0`, `T1`, etc., up to `TN` where `N` is the number of commas in the type-string. This is the expected syntax for linking to generic types with `cref` in DocFX.

----

Before:
```csharp
///  The <see cref="string" /> alias maps to a non-generic type while <c>global::System.Collections.Generic.IList&lt;string&gt;</c>,
///  <c>global::System.Collections.Generic.IDictionary&lt;string, MySession&gt;</c>, and <c>ZeroC.Slice.Result&lt;MySession, string&gt;</c> map to generic types.
```

After:
```csharp
///  The <see cref="string" /> alias maps to a non-generic type while <see cref="global::System.Collections.Generic.IList{T0}" />,
///  <see cref="global::System.Collections.Generic.IDictionary{T0, T1}" />, and <see cref="ZeroC.Slice.Result{T0, T1}" /> map to generic types.
```

These are easier to read (no XML escaping) and are clickable links.
(This snippet was generated from `tests/IceRpc.Slice.Generator.Tests/DocumentationTests.slice`)

### DIFF
--- a/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
+++ b/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
@@ -70,7 +70,7 @@ internal static class DocCommentFormatter
                 switch (c)
                 {
                     case ',' when nestingLevel == 1: // Only count commas within the first level of '<...>'.
-                        sanitizedTypeString = $"{commaCount}{sanitizedTypeString}, T";
+                        sanitizedTypeString = $"{sanitizedTypeString}{commaCount}, T";
                         commaCount += 1;
                         break;
                     case '<':
--- a/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
+++ b/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
@@ -92,7 +92,8 @@ internal static class DocCommentFormatter
         if (entity is TypeAlias alias)
         {
             // Type aliases don't generate a C# type; link to the underlying C# type instead.
-            string mapped = alias.Attributes.FindAttribute(CSAttributes.CSType)?.Args[0] ?? alias.UnderlyingType.Type.ToTypeString(currentNamespace);
+            string mapped = alias.UnderlyingType.Attributes.FindAttribute(CSAttributes.CSType)?.Args[0]
+                ?? alias.UnderlyingType.Type.ToTypeString(currentNamespace);
             return FormatTypeString(mapped);
         }
 
--- a/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
+++ b/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
@@ -68,13 +68,40 @@ internal static class DocCommentFormatter
 
     private static string FormatTypeAliasInline(TypeAlias alias, string currentNamespace)
     {
-        // Type aliases don't generate a C# type; link to the underlying C# type instead. For generic mapped
-        // types we emit the type name as inline code (XML-escaped) to avoid constructing cref-friendly forms
-        // like IList{T} or IDictionary{T, U}.
+        // Type aliases don't generate a C# type; link to the underlying C# type instead.
         string mapped = alias.UnderlyingType.Type.ToTypeString(currentNamespace);
-        return mapped.Contains('<', StringComparison.Ordinal)
-            ? $"<c>{CommentTag.XmlEscape(mapped)}</c>"
-            : $"""<see cref="{mapped}" />""";
+
+        // For generic mapped we have to convert them to their cref-friendly forms (ex: IList{T0}, IDictionary{T0, T1}).
+        var start = mapped.IndexOf('<', StringComparison.Ordinal);
+        if (start != -1)
+        {
+            // Get the type-name without any generics, then append a generic parameter 'T0'. There must be at least one.
+            string sanitizedTypeString = string.Concat(mapped.AsSpan(0, start), "{T0");
+
+            // Add an extra type parameter for each top-level comma we see, skipping over any commas within nested
+            // generic types, since they don't correspond to type parameters of the outer type.
+            int commaCount = 0;
+            int nestingLevel = 0;
+            foreach(char c in mapped)
+            {
+                switch (c)
+                {
+                    case ',' when nestingLevel == 1: // Only count commas within the first level of '<...>'.
+                        commaCount += 1;
+                        sanitizedTypeString += ", T" + commaCount;
+                        break;
+                    case '<':
+                        nestingLevel += 1;
+                        break;
+                    case '>':
+                        nestingLevel -= 1;
+                        break;
+                }
+            }
+            mapped = sanitizedTypeString + "}";
+        }
+
+        return $"""<see cref="{mapped}" />""";
     }
 
     private static string FormatEntityCref(Entity entity, string currentNamespace)
--- a/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
+++ b/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
@@ -98,6 +98,7 @@ internal static class DocCommentFormatter
 
         if (entity is CustomType custom)
         {
+            // Custom types don't generate a C# type; link to their mapped C# type instead.
             if (custom.Attributes.FindAttribute(CSAttributes.CSType)?.Args[0] is string mapped)
             {
                 return FormatTypeString(mapped);
--- a/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
+++ b/src/ZeroC.Slice.Generator/DocCommentFormatter.cs
@@ -19,8 +19,8 @@ internal static class DocCommentFormatter
 
         return string.Concat(overview.Select(c => c switch
         {
-            CommentText t => CommentTag.XmlEscape(t.Value),
-            CommentInlineLink l => FormatInlineLink(l.Target, currentNamespace),
+            CommentText text => CommentTag.XmlEscape(text.Value),
+            CommentInlineLink link => FormatInlineLink(link.Target, currentNamespace),
             _ => ""
         })).TrimEnd();
     }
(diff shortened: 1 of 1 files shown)


src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -68,13 +68,40 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
...
+            // generic types, since they don't correspond to type parameters of the outer type.
+            int commaCount = 0;
+            int nestingLevel = 0;
+            foreach(char c in mapped)
+            {
+                switch (c)
+                {
+                    case ',' when nestingLevel == 1: // Only count commas within the first level of '<...>'.
+                        commaCount += 1;
+                        sanitizedTypeString += ", T" + commaCount;
```
pepone: We usually prefer string interpolation for this sort of thing.

```suggestion
                        sanitizedTypeString = $"{sanitizedTypeString}, T{commaCount}";
```

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -68,13 +68,40 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
...
+                        break;
+                    case '>':
+                        nestingLevel -= 1;
+                        break;
+                }
+            }
+            mapped = sanitizedTypeString + "}";
+        }
+
+        return $"""<see cref="{mapped}" />""";
```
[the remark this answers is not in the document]
pepone: We should update seealso handling too, which is currently skip.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
pepone: I guess CSType can be generic too.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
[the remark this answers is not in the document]
InsertCreativityHere: I thought it was already using this method... I'll update everywhere I can find to use this!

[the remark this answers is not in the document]
InsertCreativityHere: I moved this logic into `FormatEntityCref` which is called by both the `@link` and the `@see` paths.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -68,13 +68,40 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
 
     private static string FormatTypeAliasInline(TypeAlias alias, string currentNamespace)
     {
-        // Type aliases don't generate a C# type; link to the underlying C# type instead. For generic mapped
-        // types we emit the type name as inline code (XML-escaped) to avoid constructing cref-friendly forms
-        // like IList{T} or IDictionary{T, U}.
+        // Type aliases don't generate a C# type; link to the underlying C# type instead.
         string mapped = alias.UnderlyingType.Type.ToTypeString(currentNamespace);
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed the logic to take `cs::type` into account.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
InsertCreativityHere: Used string interpolation instead of concatenation.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -36,22 +36,7 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
 
         foreach (CommentLink link in seeTags)
         {
-            if (link is not ResolvedCommentLink r)
-            {
-                continue;
-            }
-
-            if (r.Entity is TypeAlias alias)
```
InsertCreativityHere, line 44: TypeAlias handling was moved inside of `FormatEntityCref`, so callers don't need a special case for TypeAliases anymore.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -60,29 +45,66 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
 
     private static string FormatInlineLink(CommentLink link, string currentNamespace) => link switch
     {
-        ResolvedCommentLink { Entity: TypeAlias alias } => FormatTypeAliasInline(alias, currentNamespace),
```
InsertCreativityHere, line 63: Same. The logic was moved into `FormatEntityCref`. So now `FormatEntityCref` works for all entities, not just non-type-aliases. I think this is cleaner.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -60,29 +45,68 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
...
+
+        return typeString;
     }
 
     private static string FormatEntityCref(Entity entity, string currentNamespace)
     {
+        if (entity is TypeAlias alias)
+        {
+            // Type aliases don't generate a C# type; link to the underlying C# type instead.
+            string mapped = alias.Attributes.FindAttribute(CSAttributes.CSType)?.Args[0] ?? alias.UnderlyingType.Type.ToTypeString(currentNamespace);
```
InsertCreativityHere: For type-aliases, we check for `cs::type`, and use that type-string if present.
Otherwise we get the type-string for it's underlying type and use that instead.

One thing not clear to me, will `UnderlyingType.Type.ToTypeString` take `cs::type`s into account?
If I had, for example:
```
typealias Foo = [cs::type("CsFoo")] RandomStruct
typealias Bar = [cs::type("CsBar")] Foo
```

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -60,29 +45,68 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
...
+            return FormatTypeString(mapped);
+        }
+
+        if (entity is CustomType custom)
+        {
+            if (custom.Attributes.FindAttribute(CSAttributes.CSType)?.Args[0] is string mapped)
+            {
+                return FormatTypeString(mapped);
```
InsertCreativityHere, line 111: Also reworked this PR to handle custom types properly.
With the old code, mapping a custom type to a generic would generate bad-XML in the link.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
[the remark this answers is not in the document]
InsertCreativityHere: We're checking the underlying type's attributes.
Indeed, these are different from the attributes on the alias itself.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -60,29 +45,70 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
...
-        ResolvedCommentLink { Entity: TypeAlias alias } => FormatTypeAliasInline(alias, currentNamespace),
         ResolvedCommentLink r => $"""<see cref="{FormatEntityCref(r.Entity, currentNamespace)}" />""",
         UnresolvedCommentLink u => $"<c>{CommentTag.XmlEscape(u.Identifier)}</c>",
         _ => ""
     };
 
-    private static string FormatTypeAliasInline(TypeAlias alias, string currentNamespace)
+    /// <summary>Converts a C# type string into a form that can be safely used in cref attributes. Specifically, it
+    /// converts generic type parameters into their cref-friendly forms (e.g. IList{T0}, IDictionary{T0, T1}).</summary>
+    private static string FormatTypeString(string typeString)
```
pepone, line 55: We can make this local to FormatEntityCref if we are not going to use anywhere else

pepone (approved):

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -36,22 +36,7 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
...
-                // Type aliases don't generate a C# type; map to the underlying C# type. We can't reliably
-                // produce a seealso for a generic mapped type without C# parsing, so skip those.
-                string mapped = alias.UnderlyingType.Type.ToTypeString(currentNamespace);
-                if (!mapped.Contains('<', StringComparison.Ordinal))
-                {
-                    yield return new CommentTag("seealso", "cref", mapped, "");
-                }
-            }
-            else
+            if (link is ResolvedCommentLink r)
```
bernardnormier: This should be a full variable name, not `r`.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -60,29 +45,70 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
...
-        string mapped = alias.UnderlyingType.Type.ToTypeString(currentNamespace);
-        return mapped.Contains('<', StringComparison.Ordinal)
-            ? $"<c>{CommentTag.XmlEscape(mapped)}</c>"
-            : $"""<see cref="{mapped}" />""";
+        // For generic types we have to convert them to their cref-friendly forms (e.g. IList{T0}, IDictionary{T0, T1}).
+        var start = typeString.IndexOf('<', StringComparison.Ordinal);
+        if (start != -1)
+        {
+            // Get the type-name without any generics, then append a generic parameter 'T0'. There must be at least one.
+            string sanitizedTypeString = string.Concat(typeString.AsSpan(0, start), "{T0");
```
bernardnormier: Can we just use `T` here?

The standard for single type parameter generics is T, not T0. E.g. https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.ilist-1?view=net-10.0

bernardnormier (approved):

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
```diff
@@ -60,29 +45,70 @@ internal static IEnumerable<CommentTag> FormatSeeAlsoTags(Comment? comment, stri
...
+                        break;
+                    case '<':
+                        nestingLevel += 1;
+                        break;
+                    case '>':
+                        nestingLevel -= 1;
+                        break;
+                }
+            }
+            typeString = sanitizedTypeString + "}";
```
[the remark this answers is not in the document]
InsertCreativityHere: This is possible to encounter, but seems like an even bigger pain to deal with, and the odds of someone linking to something like this are very low IMO. I think what we have right now is good enough.

And if someone encounters this, they'll get a warning, tell us about it, and then we can address.

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
InsertCreativityHere: This was leftover from the original code.
I updated all the single-letter names in this file (except for `char c` which I think is fine).

src/ZeroC.Slice.Generator/DocCommentFormatter.cs
InsertCreativityHere: I reworked the code so that we generate `{T}` when there's only one parameter, but `{T0, T1, ...}` if there's multiple parameters.

PR #4578, opened 2026-05-05
Bound Slic outbound buffering with PauseWriterThreshold on the connec…

pepone:
Replaces #4532


src/IceRpc/Transports/Slic/SlicTransportOptions.cs
```diff
@@ -60,6 +60,21 @@ public int MaxStreamFrameSize
             value;
     }
 
+    /// <summary>Gets or sets the pause writer threshold. It defines the maximum amount of data that Slic buffers
+    /// locally on the connection's outbound pipe before <see cref="System.IO.Pipelines.PipeWriter.FlushAsync" /> on the
+    /// underlying pipe starts blocking. This bounds the memory used by the connection's outbound buffering when a peer
+    /// is slow or not reading.</summary>
+    /// <value>The pause writer threshold in bytes. It can't be less than <c>1</c> KB. Defaults to <c>1</c> MB.</value>
```
bernardnormier: It's not clear why we picked this 1MB default and 1KB min. 

I would just use the Pipe defaults, and allow the user to configure `0` to get no flow control at this level (Slic's current behavior).

src/IceRpc/Transports/Slic/SlicTransportOptions.cs
```diff
@@ -60,6 +60,27 @@ public int MaxStreamFrameSize
             value;
     }
 
+    /// <summary>Gets or sets the pause writer threshold. It defines the maximum amount of data that Slic buffers
```
bernardnormier: I would write "the per-connection pause writer threshold" to make it immediately clear this configuration is per connection.

src/IceRpc/Transports/Slic/SlicTransportOptions.cs
```diff
@@ -60,6 +60,27 @@ public int MaxStreamFrameSize
...
+    /// locally on the connection's outbound pipe before <see cref="System.IO.Pipelines.PipeWriter.FlushAsync" /> on the
+    /// underlying pipe starts blocking. This bounds the memory used by the connection's outbound buffering when a peer
+    /// is slow or not reading.</summary>
+    /// <value>The pause writer threshold in bytes. Set to <c>0</c> to disable this flow control mechanism. Otherwise,
+    /// it can't be less than <c>1</c> KB. Defaults to <c>64</c> KB (the <see cref="System.IO.Pipelines.Pipe" />
+    /// default).</value>
+    public int PauseWriterThreshold
+    {
+        get => _pauseWriterThreshold;
```
bernardnormier: Can we use the new field keyword?

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-14.0/field-keyword

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
```diff
@@ -1430,6 +1430,269 @@ public async Task Dispose_connection_when_duplex_connection_shutdown_write_hangs
...
+        writeCts.Cancel();
+
+        // Assert
+        Assert.That(
+            async () => await writeTask,
+            Throws.InstanceOf<OperationCanceledException>());
+    }
+
+    /// <summary>Verifies that a write blocked on the connection-level pipe pause (PauseWriterThreshold reached)
+    /// can be canceled.</summary>
```
bernardnormier, line 1562: Also interesting: such a cancellation does not abort the connection nor does it abort the stream.

That's because the write itself is actually not cancelled. We cancel the FlushAsync (waiting for the data to be read and pushed down to the duplex connection).  

Maybe we should add a test that verifies the stream still works fine after a cancellation?

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -370,28 +389,41 @@ public async Task CloseAsync(MultiplexedConnectionCloseError closeError, Cancell
         bool waitForWriterShutdown = false;
         if (TryClose(new IceRpcException(IceRpcError.OperationAborted), "The connection was closed."))
         {
-            lock (_mutex)
+            await _writeSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
```
bernardnormier: You should use AcquireAsync and our helper SemaphoreSlim code.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -88,6 +91,15 @@ internal class SlicConnection : IMultiplexedConnection
     private int _unidirectionalStreamCount;
     private SemaphoreSlim? _unidirectionalStreamSemaphore;
 
+    // Serializes writes to _duplexConnectionWriter so that frame bytes are appended to the outbound pipe in order and
+    // the pipe's pauseWriterThreshold is observed strictly. This async lock is held across the FlushAsync call, so a
+    // single parked flush blocks all other connection writers until the background writer task drains enough data.
+    // Not disposed because background fire-and-forget writes (e.g. StreamWindowUpdate from sync code paths) may attempt
+    // to acquire it after DisposeAsync is called.
```
bernardnormier: We have a similar (?) writeSemaphore in IceProtocolConnection.

src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionWriter.cs
```diff
@@ -50,17 +49,24 @@ async Task PerformDisposeAsync()
...
-        // (MaxBidirectionalStreams + MaxUnidirectionalStreams) * PeerPauseWriterThreshold bytes.
         _pipe = new Pipe(new PipeOptions(
             pool: pool,
             minimumSegmentSize: minimumSegmentSize,
-            pauseWriterThreshold: 0,
+            pauseWriterThreshold: pauseWriterThreshold,
+            // When pauseWriterThreshold is non-zero, resume writes once buffered data drops to half of the threshold.
+            // Without this override, the PipeOptions default of 32 KB would exceed pauseWriterThreshold for small
+            // thresholds and throw. When pauseWriterThreshold is 0, leave both at 0 to disable pausing.
+            resumeWriterThreshold: pauseWriterThreshold == 0 ? 0 : pauseWriterThreshold / 2,
```
externl, line 69: Might be worth mentioning that we're just using same algorithm used by PipeOptions when use the defaults.

https://github.com/dotnet/runtime/blob/108d82802754e81e972079fc17c5bcab283031b6/src/libraries/System.IO.Pipelines/src/System/IO/Pipelines/PipeOptions.cs#L62-L65

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1181,9 +1282,17 @@ async Task ReadCloseFrameAsync(int size, CancellationToken cancellationToken)
             if (notAlreadyClosed && !IsServer)
             {
                 // DisposeAsync waits for the reads frames task to complete before disposing the writer.
```
bernardnormier, line 1292: I find the comment confusing. 

Here, we acquire the writeSemaphore because we use this writeSemaphore to serialize access to _duplexConnectionWriter (at least everything exception WriterTask).

It also begs the question: why do we also lock the mutex?

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1391,13 +1503,22 @@ await ReadFrameAsync(header.Value.Type, header.Value.Size, header.Value.StreamId
                 // server-side.
 
                 // DisposeAsync waits for the reads frames task to complete before disposing the writer.
```
bernardnormier, line 1506: Same here, the comment is confusing.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -585,18 +619,32 @@ internal SlicConnection(
...
-            {
-                WriteConnectionFrame(FrameType.Ping, new PingBody(payload).Encode);
-            }
-            catch (IceRpcException)
-            {
-                // Expected if the connection is closed.
-            }
-            catch (Exception exception)
+            // Fire-and-forget: timer callbacks cannot await. The synchronous prefix runs up to the WaitAsync inside
+            // WriteConnectionFrameAsync, which atomically queues this writer.
```
externl: Maybe we should move this comment and logic (fire-and-forget)  up a few levels to where we actually provide it to the timer. 

It's not so obvious when you look at this function that it's really only indented to be used in a Timer.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1181,9 +1282,17 @@ async Task ReadCloseFrameAsync(int size, CancellationToken cancellationToken)
...
+                    {
+                        _duplexConnectionWriter.Shutdown();
+                    }
+                }
+                finally
                 {
-                    _duplexConnectionWriter.Shutdown();
+                    _writeSemaphore.Release();
                 }
                 await _duplexConnectionWriter.WriterTask.WaitAsync(cancellationToken).ConfigureAwait(false);
```
bernardnormier, line 1298: Ideally, we'd have a more encapsulated API ... not give direct access to this WriterTask.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1391,13 +1503,22 @@ await ReadFrameAsync(header.Value.Type, header.Value.Size, header.Value.StreamId
...
+                    {
+                        _duplexConnectionWriter.Shutdown();
 
-                    // Make sure that CloseAsync doesn't call Write on the writer if it's called shortly after the peer
-                    // shutdown its side of the connection (which triggers ReadFrameHeaderAsync to return null).
-                    _writerIsShutdown = true;
+                        // Make sure that CloseAsync doesn't call Write on the writer if it's called shortly after the
+                        // peer shutdown its side of the connection (which triggers ReadFrameHeaderAsync to return
+                        // null).
+                        _writerIsShutdown = true;
```
bernardnormier: It's possible the writeSemaphore could also protect _writerIsShutdown.

bernardnormier (approved):
Looks good.

I would further review the interactions between _mutex and the new writeSemaphore.

externl (approved):
Looks good!

src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionDecorator.cs
```diff
@@ -97,11 +97,16 @@ async ValueTask PerformWriteAsync()
...
-    internal SlicDuplexConnectionDecorator(IDuplexConnection decoratee, Action sendReadPing, Action sendWritePing)
+    internal SlicDuplexConnectionDecorator(
+        IDuplexConnection decoratee,
+        Func<Task> sendReadPing,
+        Func<Task> sendWritePing)
         : this(decoratee)
     {
-        _readTimer = new Timer(_ => sendReadPing());
-        _writeTimer = new Timer(_ => sendWritePing());
+        // Timer callbacks cannot await; we fire-and-forget the returned Task. The send-ping callers are responsible
```
bernardnormier: I know you implemented Joe's suggestion, but I don't find this an improvement. For me, Action are much cleaner here.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -88,10 +96,19 @@ internal class SlicConnection : IMultiplexedConnection
     private int _unidirectionalStreamCount;
     private SemaphoreSlim? _unidirectionalStreamSemaphore;
 
+    // Serializes writes to _duplexConnectionWriter so that frame bytes are appended to the outbound pipe in order and
+    // the pipe's pauseWriterThreshold is observed strictly. This async lock is held across the FlushAsync call, so a
+    // single parked flush blocks all other connection writers until the background writer task drains enough data.
+    // Not disposed because background fire-and-forget writes (e.g. StreamWindowUpdate from sync code paths) may attempt
```
bernardnormier: Should also say why it's ok to not dispose it.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -370,27 +394,26 @@ public async Task CloseAsync(MultiplexedConnectionCloseError closeError, Cancell
         bool waitForWriterShutdown = false;
         if (TryClose(new IceRpcException(IceRpcError.OperationAborted), "The connection was closed."))
         {
-            lock (_mutex)
+            using SemaphoreLock _ = await _writeSemaphore.AcquireAsync(cancellationToken).ConfigureAwait(false);
```
bernardnormier: I think we should be consistent and write it as
```cs
using (...)
{
    ...
}
```

like you did elsewhere.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -524,6 +547,12 @@ await Task.WhenAll(
             {
             }
 
+            // Acquire (and never release) the write semaphore so no in-flight fire-and-forget writer (e.g. a stream
+            // frame parked on FlushAsync due to PauseWriterThreshold) can race with the writer disposal below. The
+            // _closedCts cancellation we triggered above unblocks parked WaitAsync/FlushAsync calls in those writers,
+            // so this wait is bounded.
+            await _writeSemaphore.WaitAsync(CancellationToken.None).ConfigureAwait(false);
```
bernardnormier, line 574: So: we don't dispose the _writeSemaphore before background tasks may wait for it.

We don't have a similar concern for CTS that we dispose below, such as the disposedCts and closedCts?

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -710,25 +748,58 @@ internal void WriteStreamFrame(
...
+                return;
             }
-            if (frameType == FrameType.StreamLast)
+
+            using (semaphoreLock)
             {
-                // Notify the stream that the last stream frame is considered sent at this point. This will close
-                // writes on the stream and allow the stream to be released if reads are also closed.
-                stream.WroteLastStreamFrame();
+                lock (_mutex)
```
bernardnormier, line 801: Should we release the mutex after checking isClosed? If not, what are we protecting?

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -678,22 +709,29 @@ internal void ThrowIfClosed()
...
         Debug.Assert(frameType < FrameType.Stream);
 
+        using SemaphoreLock _ = await _writeSemaphore.AcquireAsync(cancellationToken).ConfigureAwait(false);
         lock (_mutex)
         {
             if (_isClosed)
             {
                 throw new IceRpcException(_peerCloseError ?? IceRpcError.ConnectionAborted, _closedMessage);
             }
             WriteFrame(frameType, streamId: null, encode);
```
bernardnormier, line 757: I don't think WriteFrame needs to be / should be in the mutex lock. It's only for isClosed.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -804,60 +875,67 @@ internal async ValueTask<FlushResult> WriteStreamDataFrameAsync(
...
+                            }
+                            else
+                            {
+                                AddStream(_nextUnidirectionalId, stream);
+                                _nextUnidirectionalId += 4;
+                            }
                         }
-                        else
+
+                        // Notify the stream that we're consuming sendSize credit. It's important to call this before
```
bernardnormier: Why we do keep mutex locked here?

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -563,13 +608,14 @@ internal SlicConnection(
         // Only the client-side sends pings to keep the connection alive when idle timeout (set later) is not infinite.
         _duplexConnection = IsServer ?
             new SlicDuplexConnectionDecorator(duplexConnection) :
-            new SlicDuplexConnectionDecorator(duplexConnection, SendReadPing, SendWritePing);
+            new SlicDuplexConnectionDecorator(duplexConnection, SendReadPingAsync, SendWritePingAsync);
```
externl: Maybe we can move the fire-and-forget to here?

PR #4579, opened 2026-05-05
Use IReadOnlyDictionary for IRequestContextFeature.Value

### LINKED ISSUE #4454: [Audit-Low] outbound request context is encoded from a live mutable d…

pepone:
Replaces #4542
Fixes #4454


src/IceRpc/Features/IRequestContextFeature.cs
```diff
@@ -2,11 +2,19 @@
 
 namespace IceRpc.Features;
 
-/// <summary>Represents a feature that holds a <see cref="IDictionary{TKey, TValue}"/> of strings. This feature can be
-/// transmitted as a <see cref="RequestFieldKey.Context" /> field with both ice and icerpc.</summary>
+/// <summary>Represents a feature that holds a read-only dictionary of strings. This feature can be transmitted as a
+/// <see cref="RequestFieldKey.Context" /> field with both ice and icerpc.</summary>
+/// <remarks>The outgoing request context is encoded lazily, when the request is sent. An interceptor that needs to
```
bernardnormier: I don't see the value of describing the implementation: "The outgoing request context is encoded lazily, when the request is sent."

src/IceRpc/Features/RequestContextFeature.cs
```diff
@@ -10,32 +10,36 @@ namespace IceRpc.Features;
...
-        get => Value[key];
-        set => Value[key] = value;
+        get => _dictionary[key];
+        set => _dictionary[key] = value;
     }
 
-    /// <summary>Constructs an empty writeable request context feature.</summary>
-    public RequestContextFeature() => Value = new Dictionary<string, string>();
+    /// <summary>Constructs an empty request context feature.</summary>
+    public RequestContextFeature() => _dictionary = new Dictionary<string, string>();
```
bernardnormier: I don't find the changes in this file correct.

A few questions through:
- what should be the type of the main field?

It could be an IReadOnlyDictionary - the one supplied by the user through the primary constructor.
Or it could be an ImmutableDictionary to make sure there is no unintentional post-creation mutation.

- what do we do in the parameterless constructor?

Just use `ImmutableDictionary<string, string>.Empty`.

docfx/examples/IceRpc.RequestContext.Examples/RequestContextInterceptorExamples.cs
```diff
@@ -48,4 +48,26 @@ public static async Task UseRequestContext()
         #endregion
         }
     }
+
+    public static void UpdateRequestContextInInterceptor()
+    {
+        #region UpdateRequestContextInInterceptor
+        // An interceptor that adds an entry to the request context. The pattern is to build a new
+        // dictionary from any existing entries and install a new feature, rather than attempting to
```
bernardnormier: We should update the comment; you don't "attempt to mutate" a read-only dictionary.

docfx/examples/IceRpc.RequestContext.Examples/RequestContextInterceptorExamples.cs
```diff
@@ -48,4 +48,26 @@ public static async Task UseRequestContext()
...
+        // mutate the existing one.
+        _ = new Pipeline()
+            .Use(next => new InlineInvoker((request, cancellationToken) =>
+            {
+                IRequestContextFeature? feature = request.Features.Get<IRequestContextFeature>();
+                var updated = feature is null
```
bernardnormier: I would write it as:
```
var dictionary = feature?.Value?.ToImmutableDictionary() ?? ImmutableDictionary<string, string>.Empty;

dictionary = dictionary.Add("CorrelationId", Guid.NewGuid().ToString());
...
```

Especially since the logic here is to add an entry, not add-or-update.

src/IceRpc/Features/RequestContextFeature.cs
pepone: I kept the indexer / `Add` to support the collection initializer pattern, which is convenient when constructing the feature.

Interceptors and other consumers only see `IRequestContextFeature`, and there the contract is clear: `Value` is an `IReadOnlyDictionary<string, string>`, so to add or remove entries you have to create a new feature.

The concrete `RequestContextFeature` is still mutable post-construction via the indexer / `Add`, but that feels like an acceptable tradeoff for the ergonomics — the read-only invariant lives on the interface that consumers actually depend on.

Would you rather back the implementation with an immutable dictionary and drop the collection initializer?

bernardnormier: I lean towards making `_dictionary` an `IReadOnlyDictionary<string, string>`, and using `IReadOnlyDictionary<string, string>` as the parameter for the primary constructor.

An alternative is no primary constructor and making Value get-init.

> support the collection initializer pattern, 

I don't think it's that important for RequestContext.

pepone: Fixed as suggested.

src/IceRpc/Features/RequestContextFeature.cs
```diff
@@ -1,41 +1,22 @@
...
-
-    /// <inheritdoc />
-    public IEnumerator<KeyValuePair<string, string>> GetEnumerator() => Value.GetEnumerator();
-
-    /// <inheritdoc />
-    IEnumerator IEnumerable.GetEnumerator() => ((IEnumerable)Value).GetEnumerator();
+    /// <summary>Constructs a request context feature wrapping the specified dictionary.</summary>
+    /// <param name="value">The dictionary held by the new feature. It is copied to an
+    /// <see cref="ImmutableDictionary{TKey, TValue}" /> unless it is already one, so post-construction mutations
+    /// to <paramref name="value" /> (and casts of the feature's <c>Value</c> back to a mutable dictionary) cannot
```
bernardnormier: I would suppress this parenthesis: you can't cast Value to a mutable dictionary!

docfx/examples/IceRpc.RequestContext.Examples/RequestContextInterceptorExamples.cs
```diff
@@ -48,4 +49,25 @@ public static async Task UseRequestContext()
         #endregion
         }
     }
+
+    public static void UpdateRequestContextInInterceptor()
+    {
+        #region UpdateRequestContextInInterceptor
+        // An interceptor that adds an entry to the request context. Since the existing context is
+        // read-only, the pattern is to build a new dictionary that includes any existing entries
```
bernardnormier: Needs to be updated. The pattern is to retrieve the existing feature's ImmutableDictionary, add the new entry, then set a new feature with the new ImmutableDictionary.

bernardnormier (approved):

docfx/examples/IceRpc.RequestContext.Examples/RequestContextInterceptorExamples.cs
```diff
@@ -22,11 +23,11 @@ public static async Task UseRequestContext()
 
         // Create a feature collection holding an IRequestContextFeature.
         IFeatureCollection features = new FeatureCollection().With<IRequestContextFeature>(
-            new RequestContextFeature
+            new RequestContextFeature(new Dictionary<string, string>
             {
                 ["UserId"] = Environment.UserName.ToLowerInvariant(),
                 ["MachineName"] = Environment.MachineName
-            });
+            }.ToImmutableDictionary()));
```
externl, line 30: Is ToImmutableDictionary necessary? We've not doing this in [examples/protobuf/RequestContext/Client/Program.cs‎](https://github.com/icerpc/icerpc-csharp/pull/4579/changes#diff-ba84a55b22b86b57e56284002149c895155732a8002f0070bf64891b91900fd1)

docfx/examples/IceRpc.RequestContext.Examples/RequestContextInterceptorExamples.cs
pepone: This doesn't sound good the API IRequestContextFeature exposes IReadOnlyDictionary, should we update the feature to use ImmutableDictionary?

externl (approved):

docfx/examples/IceRpc.RequestContext.Examples/RequestContextInterceptorExamples.cs
externl: Using `ImmutableDictionary` is probably the safer choice.

bernardnormier: I was just saying "the comment needs to be updated" and describing the update knowing the feature is most likely implemented using the default feature implementation.

I think `IRequestContextFeature` should keep returning an `IReadOnlyDictionary`. That's all you need to read or encode the feature. `ImmutableDictionary` is an implementation detail. We could use `IImmutableDictionary` but I'd rather stick to `IReadOnlyDictionary` for the interface.

PR #4581, opened 2026-05-05
Fix IceRpc.ServiceGenerator silently dropping derived operations that collide with base operations (#4461)

### LINKED ISSUE #4461: [Audit-Medium] derived service operations can be silently dropped whe…

### Medium: derived service operations can be silently dropped when a base service already uses the same operation name — **CONFIRMED**

Affected code:

- [src/IceRpc.ServiceGenerator/Internal/Parser.cs:61-87](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/Internal/Parser.cs#L61) — `GetServiceMethods(classSymbol.AllInterfaces).Except(baseServiceMethods)` then the duplicate-name diagnostic loop
- [src/IceRpc.ServiceGenerator/Internal/ServiceMethod.cs:18-33](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/Internal/ServiceMethod.cs#L18) — `Equals` and `GetHashCode` both defined solely in terms of `OperationName`

Verification:

Confirmed. Two problems stacked on each other:

1. `ServiceMethod.Equals` is `OperationName == other.OperationName` only — two methods with the same operation name are equal even if they dispatch to different C# methods with different IDL shapes.
2. `Except(baseServiceMethods)` therefore strips *any* derived method whose operation name collides with a base method, silently. The duplicate-name diagnostic at [line 75-83](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.ServiceGenerator/Internal/Parser.cs#L75) runs on the already-filtered set, so the collision is never reported. The derived dispatcher then omits the operation entirely, and requests for that operation dispatch to the base class's implementation.

For a developer who meant to override a base operation with a differently-shaped IDL method (e.g. Slice override of an Ice operation), the generated dispatch quietly ignores their override.

Impact:

- Dispatch tables diverge from the declared service contract.
- Override and collision mistakes are hidden instead of diagnosed.

Recommendation:

- Check for duplicate operation names across *both* derived and inherited sets before the `Except` filter, and report `DuplicateOperationNames` for any collision.
- Make `ServiceMethod.Equals` consider the full method identity (concrete subtype + declaring interface + signature) so that only genuinely inherited methods get removed by `Except`.

---

**Source report:** [src-IceRpc.ServiceGenerator-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.ServiceGenerator-audit-2026-04-14.md) (finding `derived service operations can be silently dropped when a base service already uses the same operation name — **CONFIRMED**`)

bernardnormier:
Fixes #4461.

### Problem

When a derived `[Service]` class declares an operation whose IDL name collides with an operation already declared by its base service class (via a different interface), the generator silently drops the derived operation. Incoming requests for that operation name then dispatch to the base implementation, diverging from the declared service contract — and no diagnostic is emitted.

Root cause: in `Parser.GetServiceDefinitions`, the duplicate-name check ran on the result of `derivedServiceMethods.Except(baseServiceMethods)`. Because `ServiceMethod` equality is by `OperationName`, `Except` filtered out the derived method first, so the collision never reached the diagnostic loop.

### Note on the audit recommendation

The audit suggested making `ServiceMethod.Equals` consider the full method identity (subtype + interface + signature) so both methods would be kept. That isn't right for this generator: the dispatcher routes by operation-name string, so two methods sharing an op name in one dispatcher are fundamentally undispatchable regardless of C# signature. The correct outcome is a diagnostic, not preserving both. So `Equals`/`GetHashCode` keyed on `OperationName` is left as-is.

### Fix

In `Parser.GetServiceDefinitions`:

- Partition `classSymbol.AllInterfaces` into interfaces inherited from the base service class vs. interfaces new in this class (using `SymbolEqualityComparer.Default`). Only the new interfaces contribute operations to dispatch from this class.
- Seed the duplicate-name `HashSet` with the base service's operation names, so collisions between an operation added by this class and a base operation are now reported as `IRSG0001` instead of silently dropped.

This also addresses the broader concern in the issue that two methods with different names mapping to the same operation name in the same dispatcher should be reported as a diagnostic, not silently merged.

### Tests

Added `MixedServiceTests.Derived_service_with_operation_name_colliding_with_base_reports_diagnostic`, which uses `CSharpGeneratorDriver` to run `ServiceGenerator` against a small in-memory source where a derived `[Service]` class declares an operation (`"ping"`) whose name collides with a base operation declared via a different interface, and asserts that `IRSG0001` is reported.

Verified that reverting only the `Parser.cs` change makes the new test fail, and that the existing `IceRpc.Ice.Generator.Tests` (which exercise legitimate inheritance via shared interfaces) still pass — they were the original signal that the first cut of the fix was over-eager.

### Files changed

- `src/IceRpc.ServiceGenerator/Internal/Parser.cs` — partition interfaces and seed the duplicate-name set.
- `tests/IceRpc.ServiceGenerator.Tests/MixedServiceTests.cs` — new collision test.
- `tests/IceRpc.ServiceGenerator.Tests/IceRpc.ServiceGenerator.Tests.csproj` — add `Microsoft.CodeAnalysis.CSharp` reference and drop `ReferenceOutputAssembly="false"` on the generator project ref so the test can instantiate `ServiceGenerator` directly while still wiring it as an analyzer.

### DIFF
--- a/src/IceRpc.ServiceGenerator/Internal/Parser.cs
+++ b/src/IceRpc.ServiceGenerator/Internal/Parser.cs
@@ -58,22 +58,36 @@ internal sealed class Parser
                     continue;
                 }
 
+                ImmutableArray<INamedTypeSymbol> baseInterfaces = [];
                 IReadOnlyList<ServiceMethod> baseServiceMethods = [];
+
                 INamedTypeSymbol? baseServiceClass = GetBaseServiceClass(classSymbol);
+
                 if (baseServiceClass is not null)
                 {
-                    baseServiceMethods = GetServiceMethods(baseServiceClass.AllInterfaces);
+                    baseInterfaces = baseServiceClass.AllInterfaces;
+                    baseServiceMethods = GetServiceMethods(baseInterfaces);
                 }
 
-                IEnumerable<ServiceMethod> serviceMethods =
-                    GetServiceMethods(classSymbol.AllInterfaces).Except(baseServiceMethods);
+                // Partition the interfaces into those inherited from the base service class and those new in this
+                // class. Methods from inherited interfaces are already dispatched by the base; methods from new
+                // interfaces are what this class adds.
+                var newInterfaces = classSymbol.AllInterfaces
+                    // The explicit type argument is required because SymbolEqualityComparer.Default is typed as
+                    // IEqualityComparer<ISymbol>, which would otherwise infer Except's element type as ISymbol.
+                    .Except<INamedTypeSymbol>(baseInterfaces, SymbolEqualityComparer.Default)
+                    .ToImmutableArray();
 
-                // We check for duplicates only once per class.
-                var operationNames = new HashSet<string>();
+                IEnumerable<ServiceMethod> serviceMethods = GetServiceMethods(newInterfaces);
+
+                // We check for duplicates only once per class. Seed the set with the base service operation names so
+                // we also report collisions between an operation added by this class and a base operation.
+                var operationNames = new HashSet<string>(baseServiceMethods.Select(m => m.OperationName));
                 foreach (ServiceMethod method in serviceMethods)
                 {
                     if (!operationNames.Add(method.OperationName))
                     {
+                        // The diagnostic report is not fatal: we keep going.
                         _reportDiagnostic(
                             Diagnostic.Create(
                                 DiagnosticDescriptors.DuplicateOperationNames,
@@ -83,7 +97,7 @@ internal sealed class Parser
                     }
                 }
 
-                // Suppress duplicates, if any.
+                // Suppress duplicates, if any. Any such duplicate was reported above as an error.
                 serviceMethods = serviceMethods.Distinct();
 
                 string containingNamespace = classSymbol.ContainingNamespace.GetFullName();
--- a/tests/IceRpc.ServiceGenerator.Tests/IceRpc.ServiceGenerator.Tests.csproj
+++ b/tests/IceRpc.ServiceGenerator.Tests/IceRpc.ServiceGenerator.Tests.csproj
@@ -11,6 +11,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="5.0.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.0.1" />
     <PackageReference Include="NUnit.Analyzers" Version="4.11.2">
       <PrivateAssets>all</PrivateAssets>
@@ -25,7 +26,7 @@
     <ProjectReference Include="../../src/IceRpc.Ice/IceRpc.Ice.csproj" />
     <ProjectReference Include="../../src/IceRpc.Slice/IceRpc.Slice.csproj" />
     <ProjectReference Include="../../src/IceRpc.Protobuf/IceRpc.Protobuf.csproj" />
-    <ProjectReference Include="../../src/IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
+    <ProjectReference Include="../../src/IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.csproj" OutputItemType="Analyzer" />
     <ProjectReference Include="../IceRpc.Tests.Common/IceRpc.Tests.Common.csproj" />
     <!-- Reference the built-in Slice files -->
     <SliceDirectory Include="../../slice" />
--- a/tests/IceRpc.ServiceGenerator.Tests/MixedServiceTests.cs
+++ b/tests/IceRpc.ServiceGenerator.Tests/MixedServiceTests.cs
@@ -67,9 +67,8 @@ public partial class MixedServiceTests
             CancellationToken cancellationToken) => new(message);
     }
 
-    /// <summary>Verifies that the service generator reports IRSG0001 when a derived service class adds an
-    /// operation whose name collides with an operation declared by its base service class (via a different
-    /// interface). Without the fix, the colliding derived operation would be silently dropped.</summary>
+    /// <summary>Verifies that the service generator reports IRSG0001 when a derived service class adds an operation
+    /// whose name collides with an operation declared by its base service class (via a different interface).</summary>
     [Test]
     public void Derived_service_with_operation_name_colliding_with_base_reports_diagnostic()
     {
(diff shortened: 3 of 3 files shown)


tests/IceRpc.ServiceGenerator.Tests/IceRpc.ServiceGenerator.Tests.csproj
```diff
@@ -25,7 +26,7 @@
     <ProjectReference Include="../../src/IceRpc.Ice/IceRpc.Ice.csproj" />
     <ProjectReference Include="../../src/IceRpc.Slice/IceRpc.Slice.csproj" />
     <ProjectReference Include="../../src/IceRpc.Protobuf/IceRpc.Protobuf.csproj" />
-    <ProjectReference Include="../../src/IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
+    <ProjectReference Include="../../src/IceRpc.ServiceGenerator/IceRpc.ServiceGenerator.csproj" OutputItemType="Analyzer" />
```
bernardnormier, line 29: Explanation from Claude:

The new test instantiates the generator type directly:

```csharp
.Create(new global::IceRpc.ServiceGenerator.ServiceGenerator())
```

With `ReferenceOutputAssembly="false"`, the project reference only feeds the generator DLL to Roslyn as an analyzer at build time — the test assembly never gets a compile-time reference to `IceRpc.ServiceGenerator.dll`, so the `ServiceGenerator` type is unknown to the C# compiler. Removing that attribute (defaulting it back to `true`) adds the assembly as a normal reference *as well*, so the test code can name the type, while the `OutputItemType="Analyzer"` still wires it as a source generator for the existing `MixedServiceTests` to keep working.

tests/IceRpc.ServiceGenerator.Tests/MixedServiceTests.cs
```diff
@@ -61,4 +66,81 @@ public ValueTask<ProtoMessage> OpProtoWithArgsAsync(
...
+            assemblyName: "DuplicateOpTest",
+            syntaxTrees: [syntaxTree],
+            references: references,
+            options: new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary));
+
+        // Act
+        GeneratorDriver driver = CSharpGeneratorDriver
+            .Create(new global::IceRpc.ServiceGenerator.ServiceGenerator())
+            .RunGeneratorsAndUpdateCompilation(compilation, out _, out ImmutableArray<Diagnostic> diagnostics);
+        _ = driver;
```
InsertCreativityHere: `_ = driver;`
Why do we need to "manually drop" the driver here? Does it only run lazily or something?

tests/IceRpc.ServiceGenerator.Tests/MixedServiceTests.cs
```diff
@@ -61,4 +66,80 @@ public ValueTask<ProtoMessage> OpProtoWithArgsAsync(
...
+            using System.Threading;
+            using System.Threading.Tasks;
+
+            namespace TestNs;
+
+            public interface IBaseTestService
+            {
+                [RpcMethod("ping")]
+                ValueTask<string> BasePingAsync(
+                    string input, IFeatureCollection? features, CancellationToken cancellationToken);
```
pepone, line 89: I would remove the input parameter, and use ValueTask for the return, my understanding is we don't need any parameters or return values to test this.

pepone (approved):

tests/IceRpc.ServiceGenerator.Tests/MixedServiceTests.cs
bernardnormier: With Protobuf, we always have an input message and an output message. I am actually not sure how it works here with a `string`.

InsertCreativityHere (approved):
Looks good to me!

tests/IceRpc.ServiceGenerator.Tests/MixedServiceTests.cs
```diff
@@ -61,4 +66,80 @@ public ValueTask<ProtoMessage> OpProtoWithArgsAsync(
             IFeatureCollection? features,
             CancellationToken cancellationToken) => new(message);
     }
+
+    /// <summary>Verifies that the service generator reports IRSG0001 when a derived service class adds an
+    /// operation whose name collides with an operation declared by its base service class (via a different
+    /// interface). Without the fix, the colliding derived operation would be silently dropped.</summary>
```
InsertCreativityHere: I don't see the value of talking about "the fix" and describing what the old broken code did.
And if we do keep it, should at least link to this PR or the issue, so one can understand what "the fix" was 3 years from now.

tests/IceRpc.ServiceGenerator.Tests/MixedServiceTests.cs
bernardnormier: See https://github.com/icerpc/icerpc-csharp/blob/a312787ef63823f25713409e77916f376111995b/src/IceRpc.ServiceGenerator/Internal/ProtobufServiceMethod.cs#L56

We expect a specific shape for these methods, with one input arg and a `ValueTask<T>` return. It works for `string`, it wouldn't work for "void".

tests/IceRpc.ServiceGenerator.Tests/MixedServiceTests.cs
bernardnormier: Fixed

src/IceRpc.ServiceGenerator/Internal/Parser.cs
```diff
@@ -58,22 +58,36 @@ internal IReadOnlyList<ServiceClass> GetServiceDefinitions(IEnumerable<ClassDecl
...
 
-                IEnumerable<ServiceMethod> serviceMethods =
-                    GetServiceMethods(classSymbol.AllInterfaces).Except(baseServiceMethods);
+                // Partition the interfaces into those inherited from the base service class and those new in this
+                // class. Methods from inherited interfaces are already dispatched by the base; methods from new
+                // interfaces are what this class adds.
+                var newInterfaces = classSymbol.AllInterfaces
+                    // The explicit type argument is required because SymbolEqualityComparer.Default is typed as
+                    // IEqualityComparer<ISymbol>, which would otherwise infer Except's element type as ISymbol.
+                    .Except<INamedTypeSymbol>(baseInterfaces, SymbolEqualityComparer.Default)
```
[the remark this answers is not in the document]
bernardnormier: This is incorrect. The code works.

tests/IceRpc.ServiceGenerator.Tests/MixedServiceTests.cs
```diff
@@ -61,4 +64,81 @@ public ValueTask<ProtoMessage> OpProtoWithArgsAsync(
...
+            }
+            """;
+
+        SyntaxTree syntaxTree = CSharpSyntaxTree.ParseText(source);
+
+        // Reference all assemblies currently loaded so the test source can resolve IceRpc types and BCL types.
+        ImmutableArray<MetadataReference> references = AppDomain.CurrentDomain.GetAssemblies()
+            .Where(a => !a.IsDynamic && !string.IsNullOrEmpty(a.Location))
+            .Select(a => (MetadataReference)MetadataReference.CreateFromFile(a.Location))
+            .ToImmutableArray();
```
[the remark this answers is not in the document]
bernardnormier: I tried running the test on its own and it works fine.

Issue #4439, opened 2026-04-15
[Audit-Medium] detached streaming payloads can leak transport resourc…


pepone:
> **AI-generated audit finding** — this issue was opened from an automated security/correctness audit. It **has not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

### Medium: detached streaming payloads can leak transport resources when the returned or provided `IAsyncEnumerable` is ignored — **CONFIRMED**

Affected code:

- [src/IceRpc.Protobuf/RpcMethods/IncomingRequestExtensions.cs:70-73](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Protobuf/RpcMethods/IncomingRequestExtensions.cs#L70) — `DispatchClientStreamingAsync` detaches the payload and converts to `IAsyncEnumerable<TInput>`
- [src/IceRpc.Protobuf/RpcMethods/IncomingRequestExtensions.cs:140](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Protobuf/RpcMethods/IncomingRequestExtensions.cs#L140) — same pattern in `DispatchBidiStreamingAsync`
- [src/IceRpc.Protobuf/RpcMethods/InvokerExtensions.cs:257](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Protobuf/RpcMethods/InvokerExtensions.cs#L257) — client side `ReceiveStreamingResponseAsync` detaches the response payload
- [src/IceRpc.Protobuf/RpcMethods/Internal/PipeReaderExtensions.cs:43-84](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Protobuf/RpcMethods/Internal/PipeReaderExtensions.cs#L43) — `ToAsyncEnumerable` completes the reader in `finally`, but only if enumeration is driven at all
- [src/IceRpc/IncomingFrameExtensions.cs:15](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc/IncomingFrameExtensions.cs#L15) — detach transfers ownership; request/response disposal can no longer clean up

Verification:

[...]

PR #4582, opened 2026-05-06
IceRpc.Slice: prevent leak of detached streaming `PipeReader` when `I…

### LINKED ISSUE #4466: [Audit-Medium] detached streaming PipeReader leaks when the returned…

### Medium: detached streaming `PipeReader` leaks when the returned `IAsyncEnumerable<T>` is ignored — **CONFIRMED**

Affected code:

- [src/IceRpc.Slice/Operations/PipeReaderExtensions.cs:38-46](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Slice/Operations/PipeReaderExtensions.cs#L38) — fixed-size `ToAsyncEnumerable` transfers reader ownership
- [src/IceRpc.Slice/Operations/PipeReaderExtensions.cs:102-109](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Slice/Operations/PipeReaderExtensions.cs#L102) — variable-size `ToAsyncEnumerable` does the same
- [src/IceRpc.Slice/Operations/PipeReaderExtensions.cs:137-190](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.Slice/Operations/PipeReaderExtensions.cs#L137) — the private `async IAsyncEnumerable<T>` only calls `reader.Complete()` in its `finally`
- [src/IceRpc/IncomingFrameExtensions.cs:11-18](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc/IncomingFrameExtensions.cs#L11) — `DetachPayload` transfers ownership and replaces `incoming.Payload` with `InvalidPipeReader.Instance`
- [src/IceRpc/IncomingRequest.cs:78-84](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc/IncomingRequest.cs#L78), [src/IceRpc/OutgoingRequest.cs:65-72](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc/OutgoingRequest.cs#L65), and [src/IceRpc/IncomingResponse.cs:114-117](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc/IncomingResponse.cs#L114) — request/response disposal only completes the frame-owned payloads, not a detached one

Verification:

Confirmed. The leak is real and the ownership transfer is explicit.

`DetachPayload` removes the payload from the owning frame by replacing `incoming.Payload` with `InvalidPipeReader.Instance`. After that point, disposing the `IncomingRequest`, `OutgoingRequest`, or `IncomingResponse` no longer reaches the detached reader. Cleanup is delegated entirely to the compiler-generated async-enumerable state machine in `PipeReaderExtensions.ToAsyncEnumerable`, and that state machine only enters its `try/finally` after `GetAsyncEnumerator()` is called.

As a result, a stream payload is cleaned up in these already-tested cases:

- normal full enumeration,
- partial enumeration with `break`,
- iteration canceled via `WithCancellation`, and
- decode failures after enumeration has started.

But if a caller simply receives the `IAsyncEnumerable<T>` and drops it without ever starting enumeration, `reader.Complete()` is never reached. This can happen on both sides of the stack:

- a server handler validates the request and returns early without reading the streamed argument, or
- a client receives a streamed return value and abandons it before the first `await foreach`.

Impact:

- Per-connection resource leak under abandoned streaming payloads.
- Peer-reachable amplification on the server side when handlers legitimately return early without consuming a streamed argument.

Recommendation:

- Replace the compiler-generated `async IAsyncEnumerable<T>` wrapper with a custom enumerable/enumerator type that owns the `PipeReader` and can complete it deterministically even when iteration never starts.
- Or keep ownership on the frame side and expose a higher-level abstraction that does not require detaching the raw `PipeReader`.
- Add regression tests for both overloads where the returned enumerable is never enumerated and the underlying reader must still be completed.

---

**Source report:** [src-IceRpc.Slice-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.Slice-audit-2026-04-14.md) (finding `detached streaming `PipeReader` leaks when the returned `IAsyncEnumerable<T>` is ignored — **CONFIRMED**`)

bernardnormier:
…AsyncEnumerable<T>` is ignored

### Background

When a Slice operation returns or receives a streamed value, the runtime detaches the relevant section of the request/response payload as a `PipeReader` and hands it to the user wrapped in an `IAsyncEnumerable<T>`. The `PipeReader` is completed by the iterator's `finally` block.

That works only **if the iterator runs**. If the user assigns the returned `IAsyncEnumerable<T>` to a variable and then ignores it (e.g. due to an early return on an error path), the iterator's state machine is never started, the `finally` never executes, and the underlying `PipeReader` is leaked. See #4466.

### Approach

Introduce a new public interface in the `IceRpc` namespace:

```csharp
public interface IAsyncStream<out T> : IAsyncEnumerable<T>, IDisposable
    where T : allows ref struct
{
}
```

…and use it as the return/parameter type for *received* Slice streams (sent streams remain `IAsyncEnumerable<T>`). The interface is encoding-agnostic: it lives in `IceRpc` rather than `IceRpc.Slice` so the upcoming Protobuf fix (#4439) can share it without forcing users to add a `using IceRpc.Slice;` to reach a commonly-used type.

`Dispose` either:
- completes the underlying `PipeReader` directly when no enumeration was started, or
- cancels an internal `CancellationTokenSource` to unblock any pending `ReadAsync`, letting the iterator's `finally` complete the reader from the consumer thread.

The interface is intentionally `IDisposable` rather than `IAsyncDisposable`: there is nothing async about completing a `PipeReader`, and a `using` statement at the call site is the idiomatic protection against the leak.

### Behavior

- An `IAsyncStream<T>` can be enumerated **at most once**. A second call to `GetAsyncEnumerator` throws `InvalidOperationException`.
- `Dispose` is idempotent; calling it after iteration completes is a no-op.
- Cancelling the caller-supplied `CancellationToken` (via `await foreach (... .WithCancellation(ct))`) gracefully ends the iteration and completes the reader.
- The contract is **not** thread-safe: callers must not race `Dispose` against iteration, nor enumerate from multiple threads.

### API changes (breaking; targets `main`)

| Before | After |
| --- | --- |
| `Task<IAsyncEnumerable<T>> OpAsync(...)` (proxy) | `Task<IAsyncStream<T>> OpAsync(...)` |
| `ValueTask OpAsync(IAsyncEnumerable<T> p, ...)` (service) | `ValueTask OpAsync(IAsyncStream<T> p, ...)` |
| `PipeReaderExtensions.ToAsyncEnumerable<T>` | `PipeReaderExtensions.ToAsyncStream<T>` |

The send side (`AsyncEnumerableExtensions.ToPipeReader` and outgoing streamed parameters) is unchanged.

### Files of note

- `src/IceRpc/IAsyncStream.cs` — new public interface.
- `src/IceRpc.Slice/Operations/Internal/AsyncStream.cs` — the default implementation wrapping a `PipeReader` + decode iterator.
- `src/IceRpc.Slice/Operations/PipeReaderExtensions.cs` — `ToAsyncEnumerable` → `ToAsyncStream`, returning `IAsyncStream<T>`.
- `src/IceRpc.Slice.Generator/OperationExtensions.cs` — split into `GetIncomingStreamTypeString` (`IAsyncStream<T>`) / `GetOutgoingStreamTypeString` (`IAsyncEnumerable<T>`).
- `src/IceRpc.Slice.Generator/{Proxy,Service}Generator.cs` — emit the new types and the renamed factory.
- `examples/slice/Stream/**`, `examples/slice/Thermostat/**` — updated to use `using IAsyncStream<T> ... = await ...`
- `tests/IceRpc.Slice.Tests/AsyncStreamTests.cs` — new tests covering dispose-without-iteration, idempotent dispose, full iteration, break-mid-loop, dispose-during-pending-read, caller-token cancellation, second-enumerator rejection, and post-dispose enumeration.
- `tests/IceRpc.Slice.Generator.Tests/StreamTests.cs` — existing call sites switched to `using IAsyncStream<T> ... = ...` so the disposable contract is visible at the call site.

### Out of scope

- The equivalent fix for IceRpc.Protobuf (#4439) is a follow-up PR. The Protobuf-side `ToAsyncEnumerable` is unchanged here; the new `IceRpc.IAsyncStream<T>` interface is positioned so that PR can adopt it without further public-API churn.

Closes #4466.


examples/slice/Thermostat/Server/ThermoBridge.cs
```diff
@@ -29,7 +29,8 @@ public ValueTask ReportAsync(
         _deviceConnection.SetInvoker(dispatchInfo.ConnectionContext.Invoker);
 
         // Notify the ThermoFacade its device is connected. We let the async-iteration over readings execute in the
-        // background.
+        // background. This calls transfers the ownership of the readings stream to the ThermoFacade, which will
```
externl:
```suggestion
        // background. This call transfers the ownership of the readings stream to the ThermoFacade, which will
```

externl (approved):
Looks good!

tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
```diff
@@ -0,0 +1,201 @@
...
+        stream.Dispose();
+
+        Assert.That(trackingReader.CompleteCallCount, Is.EqualTo(1));
+    }
+
+    [Test]
+    public async Task Full_iteration_completes_reader()
+    {
+        // Three int32s = 12 bytes.
+        byte[] payload = new byte[12];
```
bernardnormier: We should encode using a SliceEncoder, not with BitConverter.

tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
```diff
@@ -0,0 +1,201 @@
...
+        BitConverter.TryWriteBytes(payload.AsSpan(0, 4), 1);
+        BitConverter.TryWriteBytes(payload.AsSpan(4, 4), 2);
+        BitConverter.TryWriteBytes(payload.AsSpan(8, 4), 3);
+        var trackingReader = new TrackingPipeReader(PipeReader.Create(new ReadOnlySequence<byte>(payload)));
+        IAsyncStream<int> stream = trackingReader.ToAsyncStream(
```
bernardnormier: These tests show that the C# analyzer is not very good at issuing CA2000 diagnostics:
 - stream is disposable
 - we don't dispose it
 - no diagnostics

Also, no diag if I comment out the iteration.

So, the API is nicer/cleaner, but don't count on the C# analyzer to remind you to call Dispose.

src/IceRpc.Slice.Generator/ProxyGenerator.cs
```diff
@@ -571,8 +571,8 @@ private static CodeBlock GenerateProxyDecoderExtensions(string name, string acce
                 $"""Provides an extension method for <see cref="SliceDecoder" /> to decode a <see cref="{proxyName}" />.""")
             .AddBlock(
                 new FunctionBuilder(
-                    $"{accessModifier} static", 
-                    proxyName, $"Decode{proxyName}", 
+                    $"{accessModifier} static",
+                    proxyName, $"Decode{proxyName}",
```
pepone: not changed by this PR but these two params should be each in its own lne.

src/IceRpc.Slice/Operations/Internal/AsyncStream.cs
```diff
@@ -0,0 +1,130 @@
...
+                    }
+                    if (readResult.Buffer.IsEmpty)
+                    {
+                        Debug.Assert(readResult.IsCompleted);
+                        yield break;
+                    }
+                }
+                catch (OperationCanceledException) when (linkedToken.IsCancellationRequested)
+                {
+                    // Canceling the cancellation token (caller token or our dispose token) is a normal way to
```
pepone, line 113: seems this was the existing behavior but I find a bit surprising that cancelation just result in normal ending the iteration. I think it might be desirable not differentiate between you iterate to the end, and iteration was canceled.

src/IceRpc.Slice/Operations/Internal/AsyncStream.cs
```diff
@@ -0,0 +1,130 @@
...
+
+    public void Dispose()
+    {
+        if (_disposed)
+        {
+            return;
+        }
+        _disposed = true;
+
+        if (_iterationStarted)
```
pepone, line 40: Did you consider simplifying and always dispose from Dipose and no need to track _iterationStarted?

bernardnormier: That's not much simpler and in my view undesirable: you want to release the source as soon as possible, and this includes end-of-iteration.

src/IceRpc.Slice/Operations/Internal/AsyncStream.cs
bernardnormier: Why is it surprising?

In the case where you can the async-iteration yourself (via Dispose or the `WithCancellation` mechanism), what would you expect?

pepone: I will expect that the iteration throws, not that it silently handle the cancelation that is a much expected behavior when an operation is canceled. Same for dispose, if you dispose mid iteration the iteration should terminate with ObjectDisposedException not handle it internally and exit like nothing happen.

tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
pepone: Previous analysis assuming this not run was bogus, probably related to https://github.com/dotnet/roslyn-analyzers/issues/5330

pepone (approved):

src/IceRpc.Slice/Operations/Internal/AsyncStream.cs
bernardnormier: Created #4584 to review this behavior.

src/IceRpc.Slice.Generator/ProxyGenerator.cs
bernardnormier: Fixed

src/IceRpc.Slice/Operations/Internal/AsyncStream.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Diagnostics;
+using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+
+namespace IceRpc.Slice.Operations.Internal;
+
+/// <summary>The default <see cref="IAsyncStream{T}" /> implementation. It wraps a <see cref="PipeReader" /> and
+/// decodes its bytes into elements of type <typeparamref name="T"/> using a read function and a decode function.
+/// </summary>
+internal sealed class AsyncStream<T> : IAsyncStream<T>
```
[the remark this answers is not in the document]
bernardnormier: I think Copilot is mistaken here. Nevertheless I removed the "allows ref struct" constraint as we don't need to allow ref structs for these stream elements.

InsertCreativityHere (approved):
Looks good to me!

src/IceRpc/IAsyncStream.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) ZeroC, Inc.
+
+namespace IceRpc;
+
+/// <summary>Represents an incoming stream of elements decoded from a request or response payload.</summary>
+/// <typeparam name="T">The type of the streamed elements.</typeparam>
+/// <remarks>An <see cref="IAsyncStream{T}" /> owns the underlying transport
+/// <see cref="System.IO.Pipelines.PipeReader" /> from which elements are decoded. It is the caller's responsibility to
+/// dispose the stream — typically with a <c>using</c> statement — to complete this reader.</remarks>
```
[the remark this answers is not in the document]
bernardnormier: Fixed

src/IceRpc.Slice/Operations/Internal/AsyncStream.cs
```diff
@@ -0,0 +1,131 @@
...
+{
+    private readonly PipeReader _reader;
+    private readonly Func<PipeReader, CancellationToken, ValueTask<ReadResult>> _readFunc;
+    private readonly Func<ReadOnlySequence<byte>, IEnumerable<T>> _decodeBufferFunc;
+
+    // Canceled by Dispose when iteration has started, to unblock any pending ReadAsync. We never dispose this CTS:
+    // disposing it could race with a still-active linked CTS holding a registration on its token.
+    private readonly CancellationTokenSource _disposeCts = new();
```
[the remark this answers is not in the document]
bernardnormier: Codex answer:

Your instinct is right: the review comment is mostly overblown.

Why this is not a real leak in practice:

1. CancellationTokenSource does not hold an OS handle by default.  
It allocates its WaitHandle lazily only if someone touches Token.WaitHandle. Your code path does not do that.

2. The linked CTS registration is already disposed.  
In EnumerateAsync, linkedCts is wrapped in using, so registrations to _disposeCts.Token are cleaned up when iteration ends.

3. Lifetime is bounded by AsyncStream lifetime.  
If caller drops the stream, GC collects AsyncStream and its _disposeCts together. That is normal managed-memory reclamation, not an unbounded resource leak.

So this is not a meaningful leak concern unless:
- code accesses WaitHandle, or
- many streams are retained for a long time after completion.

If you wanted deterministic cleanup anyway, you can do it, but it adds edge-case handling:
- dispose `_disposeCts` in iterator finally,
- and make Dispose tolerate ObjectDisposedException when calling Cancel after iteration has already finished.

Given your current design goals (simple, no extra synchronization), keeping it as-is is a reasonable choice.

tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
```diff
@@ -0,0 +1,201 @@
...
+            (ref SliceDecoder decoder) => decoder.DecodeInt32(),
+            elementSize: 4);
+
```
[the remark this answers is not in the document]
bernardnormier: Will fix. Unfortunately this test and other tests demonstrate that there is no analyzer warning.

src/IceRpc.Slice.Generator/OperationExtensions.cs
```diff
@@ -58,9 +58,9 @@ internal static class OperationExtensions
         /// </summary>
         internal string EncodedReturnPayloadName => op.StreamedReturn?.Name == "Payload" ? "Payload_" : "Payload";
 
-        /// <summary>Returns the C# type string for a streamed field (parameter or return).
-        /// Non-optional stream uint8 maps to PipeReader, all others to IAsyncEnumerable&lt;T&gt;.</summary>
-        internal static string GetStreamTypeString(Field streamField, string currentNamespace)
+        /// <summary>Returns the C# type string for an outgoing (sent) streamed field. Non-optional stream uint8
+        /// maps to PipeReader, all others to IAsyncEnumerable{T}.</summary>
```
InsertCreativityHere: After some discussion: https://github.com/icerpc/icerpc-csharp/pull/4585
```suggestion
        /// maps to <c>PipeReader</c>, all others to <c>IAsyncEnumerable&lt;T&gt;</c>.</summary>
```
This should fit on one line just fine.

src/IceRpc.Slice/Operations/PipeReaderExtensions.cs
```diff
@@ -24,18 +23,19 @@ public static class PipeReaderExtensions
...
+    /// <returns>The async stream to decode and return the streamed elements.</returns>
     /// <exception cref="ArgumentException">Thrown if <paramref name="elementSize" /> is equal of inferior to
     /// <c>0</c>.</exception>
-    /// <remarks>The reader ownership is transferred to the returned async enumerable. The caller should no longer use
-    /// the reader after this call.</remarks>
-    public static IAsyncEnumerable<T> ToAsyncEnumerable<T>(
+    /// <remarks>The reader ownership is transferred to the returned async stream. The caller should no longer use
+    /// the reader after this call, and must dispose the returned async stream when done to release the reader.
+    /// </remarks>
+    public static IAsyncStream<T> ToAsyncStream<T>(
```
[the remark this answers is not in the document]
bernardnormier: We're in the IceRpc.Slice namespace, so that's not an issue.

src/IceRpc.Slice.Generator/OperationExtensions.cs
```diff
@@ -70,6 +70,19 @@ internal static string GetStreamTypeString(Field streamField, string currentName
             return $"global::System.Collections.Generic.IAsyncEnumerable<{elemType}>";
         }
 
+        /// <summary>Returns the C# type string for an incoming (received) streamed field. Non-optional stream uint8
+        /// maps to PipeReader, all others to IceRpc.IAsyncStream{T} — a disposable wrapper that releases
```
InsertCreativityHere: This may require slight reformatting...
```suggestion
        /// maps to <c>PipeReader</c>, all others to <c>IceRpc.IAsyncStream&lt;T&gt;</c> — a disposable wrapper that releases
```

src/IceRpc/IAsyncStream.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) ZeroC, Inc.
+
+namespace IceRpc;
+
+/// <summary>Represents an incoming stream of elements decoded from a request or response payload.</summary>
+/// <typeparam name="T">The type of the streamed elements.</typeparam>
+/// <remarks>An <see cref="IAsyncStream{T}" /> owns the underlying transport
+/// <see cref="System.IO.Pipelines.PipeReader" /> from which elements are decoded. It is the caller's responsibility to
+/// to complete this reader by disposing the stream, or iterating over its elements, or doing both. Calling
```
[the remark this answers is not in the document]
bernardnormier: Fixed.

src/IceRpc.Slice.Generator/OperationExtensions.cs
bernardnormier: Fixed

src/IceRpc.Slice.Generator/OperationExtensions.cs
bernardnormier: Fixed

src/IceRpc.Slice/Operations/Internal/AsyncStream.cs
```diff
@@ -0,0 +1,130 @@
...
+            // iterator may still hold a registration on _disposeCts.Token.
+            _disposeCts.Cancel();
+        }
+        else
+        {
+            // No enumerator was ever created, hence no pending read can exist. Safe to complete the reader
+            // directly from this thread.
+            _reader.Complete();
+            _disposeCts.Dispose();
+        }
```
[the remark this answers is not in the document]
bernardnormier: This is indeed a valid edge case. Fixed.

src/IceRpc.Slice/Operations/Internal/AsyncStream.cs
```diff
@@ -0,0 +1,141 @@
...
+            // iterator may still hold a registration on _disposeCts.Token.
+            _disposeCts.Cancel();
+        }
+        else
+        {
+            // No iteration has started (either no enumerator was created, or one was created but never started),
+            // hence no pending read can exist. Safe to complete the reader directly from this thread.
+            _reader.Complete();
+            _disposeCts.Dispose();
+        }
```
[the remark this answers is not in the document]
bernardnormier: If you create an enumerator, you need to dispose it. If you don't dispose it, you may leak. No need to complicate the AsyncStream implementation to make sure AsyncStream.Dispose works properly if you forget to dispose your enumerator.

src/IceRpc/IAsyncStream.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace IceRpc;
+
+/// <summary>Represents an incoming stream of elements decoded from a request or response payload.</summary>
+/// <typeparam name="T">The type of the streamed elements.</typeparam>
+/// <remarks>An <see cref="IAsyncStream{T}" /> owns the underlying transport
+/// <see cref="System.IO.Pipelines.PipeReader" /> from which elements are decoded. It is the caller's responsibility to
+/// complete this reader by disposing the stream, or iterating over its elements, or doing both. Calling
+/// <see cref="IDisposable.Dispose"/> on a stream while an async iteration is in progress is safe and cancels the
+/// ongoing iteration.</remarks>
```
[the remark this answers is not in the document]
bernardnormier: I don't think this clarification is necessary.

PR #4585, opened 2026-05-07
Fix How we Reference Generic Types in Doc-Comments

InsertCreativityHere:
We should use code-formatting for the places which are actually code, instead of leaving them in plain-text formatting.

#### Old, and no longer relevant:

The correct syntax is to use '{' and '}' instead of '<' and '>', even if they're XML escaped to '&gt;' or '&lt;'.
As we figured out after some research on #4566.

### DIFF
--- a/src/IceRpc.Slice.Generator/OperationExtensions.cs
+++ b/src/IceRpc.Slice.Generator/OperationExtensions.cs
@@ -131,7 +131,7 @@ internal static class OperationExtensions
             return fn.Build();
         }
 
-        /// <summary>Returns the C# return type for an operation (Task, Task&lt;T&gt;, or Task&lt;tuple&gt;).
+        /// <summary>Returns the C# return type for an operation (Task, Task{T}, or Task{tuple}).
         /// Stream returns are included in the tuple with their stream type.</summary>
         internal string GetClientReturnType(string currentNamespace) =>
             BuildReturnType("global::System.Threading.Tasks.Task", op, currentNamespace, fieldType: false);
--- a/src/IceRpc.Slice.Generator/OperationExtensions.cs
+++ b/src/IceRpc.Slice.Generator/OperationExtensions.cs
@@ -131,8 +131,8 @@ internal static class OperationExtensions
             return fn.Build();
         }
 
-        /// <summary>Returns the C# return type for an operation (Task, Task{T}, or Task{tuple}).
-        /// Stream returns are included in the tuple with their stream type.</summary>
+        /// <summary>Returns the C# return type for an operation (<c>Task</c>, <c>Task&lt;T&gt;</c>, or
+        /// <c>Task&lt;tuple&gt;</c>). Stream returns are included in the tuple with their stream type.</summary>
         internal string GetClientReturnType(string currentNamespace) =>
             BuildReturnType("global::System.Threading.Tasks.Task", op, currentNamespace, fieldType: false);
 
--- a/src/IceRpc/Transports/DuplexConnectionOptions.cs
+++ b/src/IceRpc/Transports/DuplexConnectionOptions.cs
@@ -21,7 +21,7 @@ public record class DuplexConnectionOptions
     /// <value>A pool of memory blocks used for buffer management. Defaults to <see cref="MemoryPool{T}.Shared"
     /// />.</value>
     /// <remarks>The built-in TCP transport requires the pool to return array-backed memory blocks, because its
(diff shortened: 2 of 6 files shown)


src/ZeroC.Slice.Generator/FieldExtensions.cs
```diff
@@ -254,7 +254,7 @@ internal static CodeBlock GenerateEncodeBody(
         /// <summary>Gets a value indicating whether this field has the cs::readonly attribute.</summary>
         internal bool IsReadonly => value.Attributes.HasAttribute(CSAttributes.CSReadonly);
 
-        /// <summary>Gets a value indicating whether this field maps to <c>ReadOnlyMemory&lt;T&gt;</c> for outgoing
+        /// <summary>Gets a value indicating whether this field maps to <c>ReadOnlyMemory{T}</c> for outgoing
```
InsertCreativityHere: This one I'm not sure about. Should we keep the '<' and '>' since this is code-formatted?
Or should we drop the code-formatting all together? Or is what I've done optimal?

src/IceRpc.Slice.Generator/OperationExtensions.cs
```diff
@@ -131,7 +131,7 @@ internal CodeBlock BuildEncodeStreamMethod(
             return fn.Build();
         }
 
-        /// <summary>Returns the C# return type for an operation (Task, Task&lt;T&gt;, or Task&lt;tuple&gt;).
+        /// <summary>Returns the C# return type for an operation (Task, Task{T}, or Task{tuple}).
```
pepone: We should use the code tag for those.

src/ZeroC.Slice.Generator/FieldExtensions.cs
pepone: I think we should keep the html escaped version for code tags.

src/ZeroC.Slice.Generator/TypeRefExtensions.cs
```diff
@@ -55,7 +55,7 @@ internal static string GetEncodeLambda(this TypeRef typeRef, bool isOptional, st
     }
 
     /// <summary>Returns the C# type string for an incoming parameter (decode target). Sequences map to arrays,
-    /// dictionaries map to Dictionary&lt;K,V&gt;. Respects cs::type attribute.</summary>
+    /// dictionaries map to Dictionary{K,V};. Respects cs::type attribute.</summary>
```
pepone: Dictionary and cs::type should be render inside code tag IMO

src/ZeroC.Slice.Generator/TypeRefExtensions.cs
```diff
@@ -74,8 +74,8 @@ DictionaryType _ when typeRef.Attributes.FindAttribute(CSAttributes.CSType) is S
     }
 
     /// <summary>Returns the C# type string for an outgoing parameter (encode source). Sequences of fixed-size
-    /// primitives map to ReadOnlyMemory&lt;T&gt;, other sequences to IEnumerable&lt;T&gt;, dictionaries to
-    /// IEnumerable&lt;KeyValuePair&lt;K,V&gt;&gt;.</summary>
+    /// primitives map to ReadOnlyMemory{T}, other sequences to IEnumerable{T}, dictionaries to
+    /// IEnumerable{KeyValuePair{K,V}}.</summary>
```
pepone: Ditto use code tag for the code bits

InsertCreativityHere:
Okay, I wrapped all of the generics in code-tags, instead of using the curly-brace syntax.
You're right, we're really showing the code-snippets that these map to, we have no interest in linking to the type itself.

src/IceRpc/Transports/DuplexConnectionOptions.cs
```diff
@@ -21,7 +21,7 @@ public int MinSegmentSize
     /// <value>A pool of memory blocks used for buffer management. Defaults to <see cref="MemoryPool{T}.Shared"
     /// />.</value>
     /// <remarks>The built-in TCP transport requires the pool to return array-backed memory blocks, because its
-    /// multi-segment write path uses <see cref="System.Net.Sockets.Socket.SendAsync(System.Collections.Generic.IList{ArraySegment{byte}},
```
InsertCreativityHere, line 24: While looking through everywhere we reference a generic type with curly braces (to catch any others that should use code-formatting) I came across a couple places which the IDE said were over-qualified.

pepone (approved):

bernardnormier (approved):

PR #4586, opened 2026-05-07
Enable StyleCop analyzers on test projects

pepone:
Enable StyleCop analyzers on test projects and fix reported issues.


externl (approved):

bernardnormier (approved):
Looks good.

PR #4599, opened 2026-05-10
Use consistent spelling for file name (fileName/FileName)

bernardnormier:
This tiny PR fixes the spelling for file name in a few files. The convention in C#/MSBuild is to use two words.


pepone (approved):

pepone:
Note MSBuild is not consistent here for example:

- `%(Filename)` in https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-well-known-item-metadata?view=visualstudio
- `MSBuildThisFileName` in https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-reserved-and-well-known-properties?view=visualstudio

PR #4604, opened 2026-05-11
Reject duplicate icerpc header field keys as InvalidDataException

### LINKED ISSUE #4601: [Audit-Medium] duplicate icerpc header fields escape InvalidDataException classification

### Medium: duplicate icerpc header fields escape `InvalidDataException` classification

Affected code:

- [src/IceRpc/Internal/IceRpcProtocolConnection.cs:794](https://github.com/icerpc/icerpc-csharp/blob/main/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L794) — `DecodeFieldDictionary<TKey>` builds a `Dictionary<TKey, ReadOnlySequence<byte>>` directly
- [src/IceRpc/Internal/IceRpcProtocolConnection.cs:857](https://github.com/icerpc/icerpc-csharp/blob/main/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L857) — `fields.Add(key, value)` on peer-decoded keys
- [src/IceRpc/Internal/IceRpcProtocolConnection.cs:523](https://github.com/icerpc/icerpc-csharp/blob/main/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L523) — response header decoding only catches `InvalidDataException`
- [src/IceRpc/Internal/IceRpcProtocolConnection.cs:1040](https://github.com/icerpc/icerpc-csharp/blob/main/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1040), [:1059](https://github.com/icerpc/icerpc-csharp/blob/main/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1059), [:1147](https://github.com/icerpc/icerpc-csharp/blob/main/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1147) — request dispatch only special-cases `InvalidDataException`; other exceptions fall through to the unobserved-task path

### Verification

Confirmed. `Dictionary<TKey, TValue>.Add` throws `ArgumentException` on a duplicate key. \`DecodeFieldDictionary\` calls \`fields.Add(key, value)\` directly inside the per-field loop, and the surrounding \`try { ... } catch { pipe.Reader.Complete(); throw; }\` block rethrows the exception unchanged. The duplicate-key path therefore escapes as \`ArgumentException\` rather than the \`InvalidDataException\` that other malformed-header cases use (e.g. \`OverflowException\` -> wrapped, value-extends-past-buffer -> thrown as \`InvalidDataException\`, \`CheckEndOfBuffer\` -> \`InvalidDataException\`).

Downstream classification:
- Request side (\`DispatchRequestAsync\`): \`InvalidDataException\` is converted into a clean \`IceRpcException(IceRpcError.IceRpcError, \"Received invalid icerpc request header.\")\`. \`ArgumentException\` is not caught there, so it falls through to the \`catch (Exception) when (_taskExceptionObserver is not null)\` branch (when configured) or to the outer \`catch (Exception)\` at line 1147 which fires \`Debug.Fail\` and rethrows as an unobserved task exception.
- Response side (\`InvokeAsync\`): a single \`catch (InvalidDataException)\` wraps it as \`IceRpcException(IceRpcError.IceRpcError, \"Received an icerpc response with an invalid header.\")\`. A duplicate-key \`ArgumentException\` is not caught there and surfaces to the caller as an unexpected exception type.

### Relationship to #4455 / #4518

#4455 reported the same bug class for the codec-level dictionary decoders. #4518 fixed it inside \`SliceDecoderExtensions.DecodeDictionary\` and \`IceDecoderExtensions.DecodeDictionary\` (used for Slice payload decoding such as \`RequestContextMiddleware\`). This issue is the protocol-layer cousin: \`IceRpcProtocolConnection.DecodeFieldDictionary\` is a hand-rolled decoder that builds its own dictionary and was not touched by #4518.

### Impact

- A trivially malformed peer frame (duplicate field key in a request or response header) is reported as an implementation bug instead of a clean wire-protocol error.
- On the server, without a task exception observer this triggers an unobserved-task exception via \`Debug.Fail\`.
- On the client, callers see an unexpected exception type instead of \`IceRpcException(IceRpcError.IceRpcError)\`.

### Recommendation

- Wrap the \`fields.Add(key, value)\` call in \`DecodeFieldDictionary\` so a duplicate key throws \`InvalidDataException\` (preserving the original \`ArgumentException\` as inner exception, ideally with a message identifying the duplicate key).
- Add tests on both the request and response decode paths that send a header with duplicate field keys and assert the connection surfaces it as the normal invalid-wire-data error.

pepone:
Fixes #4601.

### DIFF
--- a/tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs
+++ b/tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs
@@ -446,6 +446,76 @@ public sealed class IceRpcProtocolConnectionTests
             Throws.Nothing);
     }
 
+    /// <summary>Verifies that a request whose header contains duplicate field keys is rejected as a clean
+    /// wire-protocol error (<see cref="IceRpcError.IceRpcError" />), rather than surfacing the underlying
+    /// <see cref="System.ArgumentException" /> from <see cref="Dictionary{TKey, TValue}.Add" />.</summary>
+    [Test]
+    public async Task Request_with_duplicate_field_key_is_refused()
+    {
+        // Arrange
+        var taskExceptionObserver = new TestTaskExceptionObserver();
+
+        await using ServiceProvider provider = new ServiceCollection()
+            .AddProtocolTest(Protocol.IceRpc)
+            .AddTestMultiplexedTransportDecorator()
+            .AddSingleton<ITaskExceptionObserver>(taskExceptionObserver)
+            .BuildServiceProvider(validateScopes: true);
+
+        ClientServerProtocolConnection sut = provider.GetRequiredService<ClientServerProtocolConnection>();
+        await sut.ConnectAsync();
+
+        var clientTransport = provider.GetRequiredService<TestMultiplexedClientTransportDecorator>();
+        IMultiplexedConnection clientTransportConnection = clientTransport.LastCreatedConnection;
+        IMultiplexedStream stream = await clientTransportConnection.CreateStreamAsync(bidirectional: true, default);
+
+        byte[] requestBytes = EncodeRequestWithDuplicateFieldKey();
+
+        // Act - write the manufactured request directly to the multiplexed stream.
+        stream.Output.Write(requestBytes);
+        await stream.Output.FlushAsync();
+        stream.Output.CompleteOutput(success: true);
+
+        // Assert
+        Assert.That(
+            async () => await taskExceptionObserver.DispatchRefusedException,
+            Is.InstanceOf<IceRpcException>()
+                .With.Property("IceRpcError").EqualTo(IceRpcError.IceRpcError)
+                .And.Message.Contains("Received invalid icerpc request header."));
+
+        // The server doesn't write a response — the request header decode fails first — so the client
+        // observes the stream being aborted with TruncatedData when it tries to read the response.
+        Assert.That(
+            async () => await stream.Input.ReadAsync(),
+            Throws.InstanceOf<IceRpcException>().With.Property("IceRpcError").EqualTo(IceRpcError.TruncatedData));
+
+        static byte[] EncodeRequestWithDuplicateFieldKey()
+        {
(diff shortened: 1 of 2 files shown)


tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs
```diff
@@ -446,6 +446,70 @@ public async Task Invalid_request_refused(byte[] invalidRequestBytes, bool succe
...
+        await stream.Output.FlushAsync();
+        stream.Output.CompleteOutput(success: true);
+
+        // Assert
+        Assert.That(
+            async () => await taskExceptionObserver.DispatchRefusedException,
+            Is.InstanceOf<IceRpcException>()
+                .With.Property("IceRpcError").EqualTo(IceRpcError.IceRpcError)
+                .And.Message.Contains("Received invalid icerpc request header."));
+
```
bernardnormier, line 484: Can we add an assert for what the client (stream.Input) receives? Do we get StatusCode.InvalidData?

bernardnormier (approved):

tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs
pepone: The client receives `IceRpcError.TruncatedData`, in this case we don't send a response we abort the stream.

bernardnormier: I would still add an assert for this TruncatedData. It was not obvious to me.

externl (approved):

tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs
pepone: Added as suggested

PR #4608, opened 2026-05-11
Document that AddIceRpcServer requires a unique optionsName

### LINKED ISSUE #4423: [Audit-Medium] named server registration can cross-wire dispatchers w…

### Medium: named server registration can cross-wire dispatchers when the same `optionsName` is reused — **CONFIRMED**

Affected code:

- [src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs:98-99](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs#L98) — `AddOptions<ServerOptions>(optionsName).Configure(o => o.ConnectionOptions.Dispatcher = dispatcher)`
- [src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs:124-125](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs#L124) — direct mutation in the singleton factory: `options.ConnectionOptions.Dispatcher = dispatcherBuilder.Build();`
- [src/IceRpc/Server.cs](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc/Server.cs) — `Server` stores the supplied `ServerOptions` reference without defensive copy; listeners and accepted protocol connections snapshot `ConnectionOptions.Dispatcher` from that shared object

Verification:

Confirmed. Two failure modes arise from the same root cause (shared mutable `ServerOptions` instance per `optionsName`):

1. `AddIceRpcServer(optionsName, IDispatcher)` ([line 93-101](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs#L93)) wires the dispatcher through `IConfigureOptions<ServerOptions>` at [line 99](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs#L99). Two calls with the same name install two `Configure` callbacks on the same named options; the last one wins when the options are built.
2. `AddIceRpcServer(optionsName, Action<IDispatcherBuilder>)` ([line 113-132](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs#L113)) resolves the `ServerOptions` from `IOptionsMonitor<ServerOptions>.Get(optionsName)` — which returns the *same* cached instance on every call — and mutates `options.ConnectionOptions.Dispatcher` at [line 125](https://github.com/icerpc/icerpc-csharp/blob/d6354eca1dc6df4b58f2115b01848c7b66a7bed4/src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs#L125) directly. Two such registrations produce two `Server` singletons that share one `ServerOptions` / `ConnectionOptions`; both listeners end up routing into whichever dispatcher wrote last.

Because DI containers commonly return the *last* registration for a non-enumerable `GetRequiredService<Server>()`, the first `AddIceRpcServer` call's factory might never run at resolve time — but any code that enumerates (`GetServices<Server>()`, tests, IHostedService registration of both) runs both factories and leaves the options in the last-written state.

Impact:

- Requests dispatched to the wrong service graph on the wrong listener.
- Endpoint exposure can drift silently from the configured server address, since dispatchers may encode routing/authorization assumptions.
- The bug is invisible at registration time and only surfaces at runtime.

Recommendation:

- Clone `ServerOptions` (and its nested `ConnectionOptions`) before mutating `Dispatcher` inside the factory. `Server` could also defensively copy options in its constructor, but the DI package owns the named-options contract so the fix belongs here.
- Reject duplicate server registrations for the same `optionsName`, or document/enforce that names must be unique. A simple `TryAdd`-style guard would catch the common case.
- Add tests covering two server registrations with the same name and asserting that each receives its own dispatcher.

---

**Source report:** [src-IceRpc.Extensions.DependencyInjection-audit-2026-04-14.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/src-IceRpc.Extensions.DependencyInjection-audit-2026-04-14.md) (finding `named server registration can cross-wire dispatchers when the same `optionsName` is reused — **CONFIRMED**`)

bernardnormier:
Fixes #4423.

When two `AddIceRpcServer` registrations share the same `optionsName`, both
`Server` singletons resolve the same cached `ServerOptions` from
`IOptionsMonitor<ServerOptions>`. The second registration then mutates that
shared instance (e.g. setting `ConnectionOptions.Dispatcher`), which silently
cross-wires the first server's dispatcher.

This PR (option A from the issue) makes the contract explicit in the
doc-comments of the three `optionsName`-taking overloads of `AddIceRpcServer`:
each `Server` registered in a service collection must use a distinct options
name.

The requirement is **documentation only — it is not enforced at runtime**.
Rationale:

- `Server` already copies the values it needs out of `ServerOptions` in its
  constructor, so no defensive `Clone()` is required inside `Server` itself.
- Sharing a single `ServerOptions` instance between multiple `Server`
  instances *without* DI is legitimate (e.g. when `ServerAddress.Port == 0`),
  so there is no reason to disallow it at the `ServerOptions` level.
- The DI misuse is uncommon (most apps register one or two servers) and the
  fix would otherwise require additional runtime bookkeeping that doesn't
  carry its weight.

Also adds a test that registers two servers with distinct named options and
verifies each server dispatches to its own dispatcher.

### DIFF
--- a/src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs
+++ b/src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs
@@ -140,7 +140,7 @@ public static class ServerServiceCollectionExtensions
     /// <param name="optionsName">The name of the options instance. Each <see cref="Server" /> registered in
     /// <paramref name="services" /> must use a distinct options name.</param>
     /// <returns>The service collection.</returns>
-    /// <remarks>You need to set a least the dispatcher in the injected options.</remarks>
+    /// <remarks>You need to set at least the dispatcher in the injected options.</remarks>
     /// <seealso cref="AddIceRpcServer(IServiceCollection, string, IDispatcher)" />
     public static IServiceCollection AddIceRpcServer(this IServiceCollection services, string optionsName) =>
         services
--- a/src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs
+++ b/src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs
@@ -80,7 +80,7 @@ public static class ServerServiceCollectionExtensions
     /// named <paramref name="optionsName" />.</summary>
     /// <param name="services">The service collection to add services to.</param>
     /// <param name="optionsName">The name of the options instance. Each <see cref="Server" /> registered in
-    /// <paramref name="services" /> must use a distinct options name.</param>
+    /// <paramref name="services" /> must use a unique options name.</param>
     /// <param name="dispatcher">The dispatch pipeline of the server.</param>
     /// <returns>The service collection.</returns>
     /// <example>
@@ -106,7 +106,7 @@ public static class ServerServiceCollectionExtensions
     /// <see cref="ServerOptions" /> named <paramref name="optionsName" />.</summary>
     /// <param name="services">The service collection to add services to.</param>
     /// <param name="optionsName">The name of the options instance. Each <see cref="Server" /> registered in
-    /// <paramref name="services" /> must use a distinct options name.</param>
+    /// <paramref name="services" /> must use a unique options name.</param>
     /// <param name="configure">The action to configure the dispatch pipeline using an
     /// <see cref="IDispatcherBuilder" />.</param>
     /// <returns>The service collection.</returns>
@@ -138,7 +138,7 @@ public static class ServerServiceCollectionExtensions
     /// </summary>
     /// <param name="services">The service collection to add services to.</param>
     /// <param name="optionsName">The name of the options instance. Each <see cref="Server" /> registered in
-    /// <paramref name="services" /> must use a distinct options name.</param>
+    /// <paramref name="services" /> must use a unique options name.</param>
     /// <returns>The service collection.</returns>
     /// <remarks>You need to set a least the dispatcher in the injected options.</remarks>
     /// <seealso cref="AddIceRpcServer(IServiceCollection, string, IDispatcher)" />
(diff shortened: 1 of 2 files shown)


pepone (approved):

src/IceRpc.Extensions.DependencyInjection/ServerServiceCollectionExtensions.cs
```diff
@@ -104,7 +105,8 @@ public static IServiceCollection AddIceRpcServer(
     /// server; you can specify the server's options by injecting an <see cref="IOptionsMonitor{T}" /> of
     /// <see cref="ServerOptions" /> named <paramref name="optionsName" />.</summary>
     /// <param name="services">The service collection to add services to.</param>
-    /// <param name="optionsName">The name of the options instance.</param>
+    /// <param name="optionsName">The name of the options instance. Each <see cref="Server" /> registered in
+    /// <paramref name="services" /> must use a distinct options name.</param>
```
externl: I think `unique` is better way to phrase this.

externl (approved):

Issue #4498, opened 2026-04-15
[Audit-Medium] CsAttributeValidator.Validate is called-through a disa…


pepone:
> **AI-generated audit finding** — this issue was opened from an automated security/correctness audit. It **has not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

### Medium: `CsAttributeValidator.Validate` is called-through a disabled branch, so no CS-attribute validation runs

Affected code:

- [src/ZeroC.Slice.Generator/GeneratorDriver.cs:34-39](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/ZeroC.Slice.Generator/GeneratorDriver.cs#L34) — the `Validate` call is commented out behind a `TODO: enable validation once slicec correctly handles the diagnostics reported by the generators.`, and `diagnostics` is initialized to an empty list
- [src/ZeroC.Slice.Generator/CSAttributeValidator.cs:33-50](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/ZeroC.Slice.Generator/CSAttributeValidator.cs#L33) — `Validate` is fully implemented and would reject mis-targeted CS attributes, wrong argument counts, and unknown `cs::` directives if invoked

Description:

The validator class exists and is non-trivial: it knows the target restrictions for `cs::attribute`, `cs::encodedReturn`, `cs::identifier`, `cs::public`, `cs::readonly`, and `cs::type`, and it rejects unknown `cs::` directives outright. None of that runs in the current generator pipeline. The call-site in `GeneratorDriver.RunCore` is commented out, and the comment says validation was removed because `slicec` isn't yet forwarding generator diagnostics correctly.

Consequences:

- `cs::type` on an unrelated target (e.g. an enumerator) reaches the code generator, which either silently ignores it or splices it into an unexpected position.
- `cs::attribute` with zero arguments — which `Validate` would reject — reaches `AddCSAttributes` at [BuilderExtensions.cs:37](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/ZeroC.Slice.Generator/BuilderExtensions.cs#L37) and throws an `ArgumentOutOfRangeException` from `Args[0]`.
- Unknown `cs::` directives (typos, future attributes) are silently accepted.

[...]

PR #4615, opened 2026-05-12
Use new Slice-Compiler-Definitions (Mostly Diagnostics)

InsertCreativityHere:
This PR pulls in 2 changes from `slicec`:
- Switching `Options` from a `Dictionary<string, string>` to a `Sequence<CodeGeneratorOption>`
- The new API for reporting diagnostics

The first is pretty straightforward, I just updated the decoding logic and the signatures that used it.
Instead of a `Dictionary<string, string>`, I chose a `KeyValuePair<string, string>[]`.
If there is a better choice of type, please let me know. I am not a master of C# type conventions!

The second involved creating a new wrapper type `Diagnostic`, with a handful of factory functions on it to create specific types of diagnostics. The validation (duplicate files and attributes) was updated to use these factory functions, instead of the old API which just took a `string` and a `level`.

----

Alot of the diff is just generated code. I promise this PR isn't as scary as it looks ;-;


src/ZeroC.Slice.Generator/GeneratorDriver.cs
```diff
@@ -29,14 +29,11 @@ internal static async Task RunAsync(
...
+        Task<GeneratorResponse>
+        TransformAsync(ImmutableList<SliceFile> symbolFiles, KeyValuePair<string, string>[] additionalOptions)
         {
             // Validate CS attributes before generation.
-            //
-            // List<Diagnostic> diagnostics = CsAttributeValidator.Validate(symbolFiles);
-            // TODO: enable validation once slicec correctly handles the diagnostics reported by the
-            // generators.
-            var diagnostics = new List<Diagnostic>();
+            List<Diagnostic> diagnostics = CsAttributeValidator.Validate(symbolFiles);
```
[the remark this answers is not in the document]
InsertCreativityHere: I just changed the name, it being unused was already the case.
Also, marking an unused variable as "High" seems crazy to me.

slice/Compiler/CodeGenerator.slice
InsertCreativityHere, line 1: This file was just synchronized with `slicec`; don't need to review it.

slice/Compiler/Diagnostics.slice
InsertCreativityHere, line 1: This file was just synchronized with `slicec`; don't need to review it.

slice/Ice/ReplyStatus.ice
```diff
@@ -69,5 +69,10 @@ module Ice
         /// The caller is not authorized to access the requested resource.
```
InsertCreativityHere, line 69: For completeness, I also updated the commit IDs for `ice` and `icerpc-slice`.
Apparently the `*.ice` files were slightly out of date.

src/IceRpc.Slice.BuildTelemetry/Program.cs
```diff
@@ -27,7 +27,7 @@
 
 static async Task<GeneratorResponse> BuildResponseAsync(
     ImmutableList<SliceFile> symbolFiles,
-    Dictionary<string, string> options)
+    KeyValuePair<string, string>[] additionalOptions)
```
InsertCreativityHere: The new mapping for the additional options you provide a plugin.

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
```diff
@@ -37,8 +37,8 @@ internal static List<Diagnostic> Validate(ImmutableList<SliceFile> files)
 
         foreach (SliceFile file in files)
         {
```
InsertCreativityHere, line 39: We pass along the file-name/scoped-identifier to report it back to `slicec` as a `DiagnosticSource`.

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
```diff
@@ -74,37 +74,36 @@ private static void ValidateSymbol(ISymbol symbol, List<Diagnostic> diagnostics)
...
                 {
                     ValidateOperation(op, diagnostics);
                 }
                 break;
 
             case CustomType c:
-                ValidateAttributes(c.Attributes, Target.CustomType, diagnostics);
+                ValidateAttributes(c.Attributes, c.ScopedIdentifier, Target.CustomType, diagnostics);
                 if (!c.Attributes.HasAttribute(CSAttributes.CSType))
                 {
```
InsertCreativityHere, line 92: The new API for creating diagnostics.
In addition to switching to this new API, I also tried to add notes to the errors, explaining the real problem, and in some cases, what users should/must do to fix it.

InsertCreativityHere: Note that the notes are always provided as **lower-case non-sentences**.

I know this is not the C# convention, but these notes are not for C#. They are for `slicec`.
And `slicec` reports diagnostics as lower-case non-sentences.

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
```diff
@@ -161,139 +158,141 @@ private static void ValidateOperation(Operation op, List<Diagnostic> diagnostics
         }
 
         // cs::encodedReturn: only valid on operations with non-streamed return values.
```
InsertCreativityHere, line 160: I changed the logic here.
With the old code, if I had multiple invalid `cs::encodedReturn`s, we'd only report an error for the first one. Now, we report each of them separately (which is more correct).

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
```diff
@@ -161,139 +158,141 @@ private static void ValidateOperation(Operation op, List<Diagnostic> diagnostics
...
         List<Diagnostic> diagnostics)
     {
-        foreach (Attribute attr in attributes)
+        for (int i = 0; i < attributes.Count; i++)
         {
-            if (attr.Directive.StartsWith("cs::", StringComparison.Ordinal))
+            if (attributes[i].Directive.StartsWith("cs::", StringComparison.Ordinal))
             {
-                ValidateCSAttribute(attr, target, diagnostics);
+                string attrSource = $"{source}::$attributes::{i}";
```
InsertCreativityHere, line 205: Where we splice the `$attributes::<index>` onto the source string.
Maybe I should make `$attributes` a constant?

src/ZeroC.Slice.Generator/CSAttributeValidator.cs
```diff
@@ -161,139 +158,141 @@ private static void ValidateOperation(Operation op, List<Diagnostic> diagnostics
...
 
     private static void ValidateCSAttribute(
         Attribute attr,
+        string source,
         Target target,
         List<Diagnostic> diagnostics)
     {
         switch (attr.Directive)
         {
             case CSAttributes.CSAttribute:
```
InsertCreativityHere, line 219: The meat of the changes.

src/ZeroC.Slice.Generator/VariantEnumGenerator.cs
```diff
@@ -142,7 +142,6 @@ private static CodeBlock GenerateVariantRecord(
             .AddDocCommentSeeAlso(variant.Comment, currentNamespace)
             .AddDeprecatedAttribute(variant.Attributes)
             .AddBase(parentIdentifier)
-            .AddCSAttributes(variant.Attributes)
```
InsertCreativityHere, line 145: We were incorrectly generating code for `cs::attribute` on variants.

Note that we have a separate code-path for dealing with enumerators (which is allowed).

src/ZeroC.Slice.Symbols/Compiler/CodeGenerator.cs
InsertCreativityHere, line 1: This is generated code from the compiler definitions. No need to review.

src/ZeroC.Slice.Symbols/Compiler/Diagnostics.cs
InsertCreativityHere, line 1: This is generated code from the compiler definitions. No need to review.

slice.toml
```diff
@@ -1,11 +1,11 @@
 # slice.toml — provenance for vendored Slice files
 #
 # Each [source.<name>] section describes an upstream repository and which files to
-# vendor into this repo.  Run `python tools/sync-slice.py` to synchronize.
```
InsertCreativityHere, line 4: We deleted the `tools` directory and moved the script to `build` some time ago.

tests/ZeroC.Slice.Generator.Tests/VariantEnumTests.cs
```diff
@@ -166,16 +166,6 @@ public void Decode_compact_variant_enum_with_optional_field(string? name, uint m
             Assert.That(decoder.Consumed, Is.EqualTo(1 + 1 + 1 + name.Length + 4 + 4));
         }
     }
-
-    [Test]
```
InsertCreativityHere, line 170: This test was incorrectly testing that we allowed `cs::attribute` on variants.

src/ZeroC.Slice.Generator/GeneratorDriver.cs
```diff
@@ -29,14 +29,11 @@ internal static async Task RunAsync(
 
         await Symbols.Generator.RunAsync(reader, writer, TransformAsync).ConfigureAwait(false);
 
-        Task<GeneratorResponse> TransformAsync(ImmutableList<SliceFile> symbolFiles, Dictionary<string, string> options)
+        Task<GeneratorResponse>
+        TransformAsync(ImmutableList<SliceFile> symbolFiles, KeyValuePair<string, string>[] additionalOptions)
         {
             // Validate CS attributes before generation.
```
InsertCreativityHere, line 35: We're actually performing the C# attribute validation now!

src/ZeroC.Slice.Generator/GeneratorDriver.cs
```diff
@@ -57,14 +54,9 @@ Task<GeneratorResponse> TransformAsync(ImmutableList<SliceFile> symbolFiles, Dic
...
                 {
-                    diagnostics.Add(new Diagnostic
-                    {
-                        Level = DiagnosticLevel.Error,
-                        Message =
-                            $"Multiple source files have the same file name '{fileName}': " +
-                            $"'{previousPath}' and '{file.Path}'. " +
-                            "Generated files are written to a common directory, so source files must have unique file names.",
-                    });
+                    string message = $"Multiple source files have the same file name '{fileName}': '{previousPath}' and '{file.Path}'. " +
```
InsertCreativityHere, line 57: For the record, this fixes the bug you had the other day @bernardnormier.

src/ZeroC.Slice.Symbols/Diagnostic.cs
```diff
@@ -3,14 +3,67 @@
 namespace ZeroC.Slice.Symbols;
 
 /// <summary>Represents a diagnostic message produced during code generation.</summary>
```
InsertCreativityHere, line 5: The new API for `Diagnostics`.
This is the handwritten domain type for `Symbols`, not generated code.

It's no longer `readonly` since you can conditionally attach notes to it, further describing the error.

src/ZeroC.Slice.Symbols/Generator.cs
```diff
@@ -43,10 +43,11 @@ public static async Task RunAsync(
         Compiler.SliceFile[] referenceFiles =
             decoder.DecodeSequence((ref decoder) => new Compiler.SliceFile(ref decoder));
 
-        Dictionary<string, string> options = decoder.DecodeDictionary(
-            count => new Dictionary<string, string>(count),
-            (ref decoder) => decoder.DecodeString(),
-            (ref decoder) => decoder.DecodeString());
+        KeyValuePair<string, string>[] additionalOptions = decoder.DecodeSequence((ref decoder) =>
```
InsertCreativityHere: I settled on `KeyValuePair`, since that's the correct semantics here, and I used an array since that's what we used for the `SliceFile` arguments.

If there's a better type for me to use here, please let me know!

src/ZeroC.Slice.Symbols/Diagnostic.cs
```diff
@@ -3,14 +3,67 @@
...
+    /// <summary>Gets any notes that should be reported along with this diagnostic.</summary>
+    internal IList<Compiler.DiagnosticNote> Notes { get; init; }
+
+    /// <summary>Creates a new diagnostic to describe a general error.</summary>
+    public static Diagnostic Error(string message, string? source = null) => new()
+    {
+        Kind = new Compiler.DiagnosticKind.Error(message),
+        Source = source,
+        Notes = [],
+    };
```
pepone, line 23: I find this style confusing, I would indent this, to better distinguish from a 
```suggestion
        {
            Kind = new Compiler.DiagnosticKind.Error(message),
            Source = source,
            Notes = [],
        };
```

and same for others below.

src/ZeroC.Slice.Generator/GeneratorDriver.cs
pepone, line 36: I assigned  https://github.com/icerpc/icerpc-csharp/issues/4498 to you, given that you already fixed it here!

pepone (approved):

slice/Ice/ReplyStatus.ice
bernardnormier: This requires update to the C# code as well.

src/ZeroC.Slice.Symbols/Generator.cs
bernardnormier: I would either define a new struct in Symbols or use a tuple with named fields.

src/ZeroC.Slice.Generator/GeneratorDriver.cs
InsertCreativityHere: Cool, I didn't even see it yet! 😆
Should really look through the remaining audits sometime.

bernardnormier (approved):

src/ZeroC.Slice.Symbols/Diagnostic.cs
InsertCreativityHere: Will fix!
My knowledge about C# formatting has definitely been degraded over the years...

slice/Ice/ReplyStatus.ice
InsertCreativityHere: Okay, I will update this too while I'm at it.

externl (approved):
Looks good to me!

slice/Ice/ReplyStatus.ice
bernardnormier: You can also create an issue to make this update in a follow-up PR.

src/ZeroC.Slice.Symbols/Generator.cs
InsertCreativityHere: Updated to `(string key, string value)`.

src/ZeroC.Slice.Symbols/Diagnostic.cs
InsertCreativityHere: .Net format complains about the indented syntax and tells me to dedent it:
https://github.com/icerpc/icerpc-csharp/actions/runs/26453087442/job/77879180269?pr=4615

slice/Ice/ReplyStatus.ice
pepone, line 1: Is better to do this in its own PR, seems totally unrelated.

InsertCreativityHere: I undid this and will redo it in my next PR.

But I'm not sure if I agree with the premise. It feels safer to pro-actively sync Slice sources to catch things like here where `NotSupported` seems to of been forgotten. If every PR only updates the minimum sources it needs, I worry we will never properly synchronize these files and things will fall out of alignment.

Maybe there's a way to update our vendored files action to better warn us about when files get out of sync?
As it stands right now, the `verify` check doesn't seem very useful to me TBH.

PR #4618, opened 2026-05-13
Backport #4541 to 0.5.x

pepone:
Backport of #4541 — bound Slic stream frame size and fix peer window overflow check.

## Summary
- Bounds the peer-advertised `MaxStreamFrameSize` (prevents memory exhaustion from a malicious peer advertising a huge value).
- Fixes the peer window overflow check: `Interlocked.Add` followed by `> int.MaxValue` was unreachable after wrap; now uses a pre-add check that actually detects overflow.

## Test plan
- [x] `dotnet test --filter "FullyQualifiedName~SlicTransportTests"` — 67/67 pass.

## Backport notes
Cherry-pick of f90a19f17 with two mechanical adaptations to the new test cases:
- `listener.TransportAddress` → `listener.ServerAddress` (0.5.x still uses `ServerAddress`).
- `new SliceEncoder(writer)` → `new SliceEncoder(writer, SliceEncoding.Slice2)` (0.5.x still has the two-encoding constructor).

### DIFF
--- a/src/IceRpc/Transports/Slic/SlicTransportOptions.cs
+++ b/src/IceRpc/Transports/Slic/SlicTransportOptions.cs
@@ -31,29 +31,37 @@ public sealed record class SlicTransportOptions
             throw new ArgumentException(
                 $"The {nameof(InitialStreamWindowSize)} value cannot be less than 1 KB.",
                 nameof(value)) :
-            value > MaxWindowSize ?
-            throw new ArgumentException(
-                $"The {nameof(InitialStreamWindowSize)} value cannot be larger than {MaxWindowSize}.",
-                nameof(value)) :
             value;
     }
 
     /// <summary>Gets or sets the maximum stream frame size in bytes.</summary>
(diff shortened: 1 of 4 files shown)


src/IceRpc/Transports/Slic/SlicTransportOptions.cs
```diff
@@ -31,29 +31,37 @@ public int InitialStreamWindowSize
...
-            throw new ArgumentException(
-                $"The {nameof(InitialStreamWindowSize)} value cannot be larger than {MaxWindowSize}.",
-                nameof(value)) :
             value;
     }
 
     /// <summary>Gets or sets the maximum stream frame size in bytes.</summary>
-    /// <value>The maximum stream frame size in bytes. It can't be less than <c>1</c> KB. Defaults to <c>32</c>
-    /// KB.</value>
+    /// <value>The maximum stream frame size in bytes. It can't be less than <c>1</c> KB or greater than
```
bernardnormier: It would be clearer for the doc-comment to use the same units (bytes or KB), not mix them.

bernardnormier (approved):

externl (approved):

PR #4593, opened 2026-05-08
Fix/protobuf envelope validation

pepone:
Fix #4437

PR #4553, opened 2026-04-24
Reject trailing bytes in unary Protobuf payloads

pepone:
## Summary
- Unary `DecodeProtobufMessageAsync` now performs a final read after the single expected message and throws `InvalidDataException` if trailing bytes remain, so concatenated frames or junk after the message surface as a framing error instead of being silently discarded.
- Adds tests for a valid message followed by garbage bytes and for two concatenated framed messages.

Fixes #4438

PR #4624, opened 2026-05-13
Backport #4593 to 0.5.x

pepone:
Backport of #4593 — Protobuf envelope validation.

## Summary
- Decoder treats the envelope length as **unsigned** big-endian; previously a value with the high bit set was interpreted as a negative `int`, slipping past the `> maxMessageLength` check and propagating into `ReadAtLeastAsync` / buffer slicing (classic parser-DoS primitive).
- The compression flag is now strictly validated: `0` = uncompressed, `1` = compressed (rejected as `NotSupportedException`), any other byte rejected as `InvalidDataException`. Previously flags `2..0xFF` were silently accepted as uncompressed.
- `ParseFrom` exceptions (`InvalidProtocolBufferException`) are now wrapped as `InvalidDataException` so the protocol layer surfaces them as `StatusCode.InvalidData` rather than `InternalError`.
- Adds `StatusCode.NotSupported` and maps `NotSupportedException` at the icerpc protocol layer so the compressed-flag rejection above is reported correctly.

## Test plan
- [x] `dotnet test tests/IceRpc.Protobuf.Tests` — 51/51 pass, covering the 5 new envelope-validation tests.

## Backport notes
Cherry-pick of 5d8c146cb with these adaptations:
- Source path differs: 0.5.x uses `src/IceRpc.Protobuf/Internal/PipeReaderExtensions.cs` (not `…/RpcMethods/Internal/…`).
- Removed the now-redundant post-`DecodeMessageLength` "exceeds maximum value" check (the new `DecodeMessageLength` does it before returning).
- Resolved the test-file conflict by keeping only the 5 new tests added by #4593. Two trailing-bytes / concatenated-messages tests that travelled along the diff context belong to #4553 (unary trailing-byte rejection) and will be added when #4553 is backported.
- Dropped the incidental `slice.toml` rev bump and the `slice/IceRpc/Ice/Object.ice` docstring drift (not part of the security fix; would pull in unrelated upstream slice changes). These files are not tracked on 0.5.x anyway.

Includes the additions to `StatusCode.slice` (`NotSupported`) and the matching mapping in `IceRpcProtocolConnection.cs`, which together preserve the protocol behavior the new compression-flag path expects.

### DIFF
--- a/tests/IceRpc.Protobuf.Tests/PipeReaderExtensionsTests.cs
+++ b/tests/IceRpc.Protobuf.Tests/PipeReaderExtensionsTests.cs
@@ -38,10 +38,7 @@ public partial class PipeReaderExtensionsTests
     {
         // Arrange
         var pipe = new Pipe();
-        pipe.Writer.Write(new byte[] { 1 });
-        Span<byte> lengthBytes = pipe.Writer.GetSpan(4);
-        BinaryPrimitives.WriteInt32BigEndian(lengthBytes, 0);
-        pipe.Writer.Advance(4);
+        WriteLengthPrefixedMessage(pipe.Writer, message: null, compressionFlag: 1);
         pipe.Writer.Complete();
 
         // Act & Assert
@@ -62,10 +59,7 @@ public partial class PipeReaderExtensionsTests
     {
         // Arrange
         var pipe = new Pipe();
-        pipe.Writer.Write(new byte[] { compressionFlag });
-        Span<byte> lengthBytes = pipe.Writer.GetSpan(4);
-        BinaryPrimitives.WriteInt32BigEndian(lengthBytes, 0);
-        pipe.Writer.Advance(4);
+        WriteLengthPrefixedMessage(pipe.Writer, message: null, compressionFlag: compressionFlag);
         pipe.Writer.Complete();
 
         // Act & Assert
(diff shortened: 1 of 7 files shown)


tests/IceRpc.Protobuf.Tests/PipeReaderExtensionsTests.cs
```diff
@@ -26,4 +29,116 @@ await pipeReader.DecodeProtobufMessageAsync(
...
+        // Arrange: serialize a valid StringValue, claim one extra byte in the envelope length, and append a
+        // single 0xFF — a varint with the continuation bit set and no follow-up byte, which Protobuf treats
+        // as a truncated tag.
+        var validMessage = new StringValue { Value = "hi" };
+        int actualLength = validMessage.CalculateSize();
+
+        var pipe = new Pipe();
+        pipe.Writer.Write(new byte[] { 0 });
+        Span<byte> lengthBytes = pipe.Writer.GetSpan(4);
+        BinaryPrimitives.WriteInt32BigEndian(lengthBytes, actualLength + 1);
```
bernardnormier: It would be preferable to use the helper, and fix the helper.

tests/IceRpc.Protobuf.Tests/PipeReaderExtensionsTests.cs
```diff
@@ -26,4 +29,116 @@ await pipeReader.DecodeProtobufMessageAsync(
...
+    /// silently accepted as if uncompressed.</summary>
+    [TestCase((byte)2)]
+    [TestCase((byte)0xFF)]
+    public void Decode_message_throws_invalid_data_exception_for_reserved_compression_flag(byte compressionFlag)
+    {
+        // Arrange
+        var pipe = new Pipe();
+        pipe.Writer.Write(new byte[] { compressionFlag });
+        Span<byte> lengthBytes = pipe.Writer.GetSpan(4);
+        BinaryPrimitives.WriteInt32BigEndian(lengthBytes, 0);
```
bernardnormier: Same here: use helper

bernardnormier (approved):

PR #4626, opened 2026-05-13
Dispose ColocListener when the address is already in use

pepone:
## Summary
When `ColocServerTransport.Listen` constructs a `ColocListener` but `_listeners.TryAdd` fails because the address is already taken, the newly-constructed listener becomes unreachable. Its `CancellationTokenSource` and bounded `Channel` leak (and the captured `onDispose` delegate keeps the server transport alive longer than necessary). Dispose the listener before throwing `AddressInUse`.

Flagged in code review of the backport (#4621); applying it here first so main and 0.5.x stay consistent.

## Test plan
- [x] `dotnet test --filter "FullyQualifiedName~ColocTransportTests"` — 14/14 pass.

### DIFF
--- a/src/IceRpc.Transports.Coloc/Internal/ColocServerTransport.cs
+++ b/src/IceRpc.Transports.Coloc/Internal/ColocServerTransport.cs
@@ -47,6 +47,8 @@ internal class ColocServerTransport : IDuplexServerTransport
 
         if (!_listeners.TryAdd(key, listener))
         {
+            // The listener was never published; release its resources before reporting the collision.
+            listener.DisposeAsync().AsTask().GetAwaiter().GetResult();
             throw new IceRpcException(IceRpcError.AddressInUse);
         }
         return listener;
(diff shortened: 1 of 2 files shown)


src/IceRpc.Transports.Coloc/Internal/ColocServerTransport.cs
```diff
@@ -47,6 +47,8 @@ public IListener<IDuplexConnection> Listen(
 
         if (!_listeners.TryAdd(key, listener))
         {
+            // The listener was never published; release its resources before reporting the collision.
```
bernardnormier, line 50: I don't like this code at all:
 - we don't use GetAwaiter().GetResult() - a big code smell - anywhere in the code base
 - ColocListener.DisposeAsync always completes synchronously. The Async comes from the interface

Here, it could make sense to make ColocListener (*not* IListener) IDisposable and simply call `Dispose`.

pepone: Reworked as suggested

bernardnormier (approved):

PR #4630, opened 2026-05-13
Backport #4553 to 0.5.x

pepone:
Backport of #4553 — reject trailing bytes in unary Protobuf payloads.

## Summary
A unary Protobuf payload silently accepting bytes after the framed message is a smuggling primitive (middleware vs. dispatcher parsing divergence). `DecodeProtobufMessageAsync` now reads once more after extracting the message and rejects any remaining bytes (and any non-completed read) with `InvalidDataException`.

## Test plan
- [x] `dotnet test tests/IceRpc.Protobuf.Tests` — 53/53 pass (51 + 2 new tests).

## Backport notes
Cherry-pick of 80ff266a3 with one conflict resolution in `tests/IceRpc.Protobuf.Tests/PipeReaderExtensionsTests.cs`: the file was modified by the already-merged [#4624](https://github.com/icerpc/icerpc-csharp/pull/4624) (backport of #4593, envelope validation), which introduced a richer `WriteLengthPrefixedMessage` helper. Kept all of #4593's tests + the richer helper from 0.5.x, and appended the two new #4553 tests (`Decode_message_throws_invalid_data_exception_when_payload_has_trailing_bytes`, `Decode_message_throws_invalid_data_exception_when_payload_has_concatenated_messages`).

Now unblocked since [#4624](https://github.com/icerpc/icerpc-csharp/pull/4624) merged.

### DIFF
--- a/src/IceRpc.Protobuf/Internal/PipeReaderExtensions.cs
+++ b/src/IceRpc.Protobuf/Internal/PipeReaderExtensions.cs
@@ -26,15 +26,31 @@ internal static class PipeReaderExtensions
         this PipeReader reader,
         MessageParser<T> parser,
         int maxMessageLength,
-        CancellationToken cancellationToken) where T : IMessage<T>
+        CancellationToken cancellationToken) where T : class, IMessage<T>
     {
-        T? message = await ReadMessageAsync(
+        T message = await ReadMessageAsync(
             reader,
             parser,
             maxMessageLength,
-            cancellationToken).ConfigureAwait(false);
+            cancellationToken).ConfigureAwait(false) ??
+            throw new InvalidDataException("The payload is empty; expected a Protobuf message.");
 
-        Debug.Assert(message is not null);
+        // A unary payload must contain exactly one message; any trailing bytes indicate a framing error.
+        ReadResult readResult = await reader.ReadAsync(cancellationToken).ConfigureAwait(false);
+        // We never call CancelPendingRead; an interceptor or middleware can but it's not correct.
+        if (readResult.IsCanceled)
+        {
+            throw new InvalidOperationException("Unexpected call to CancelPendingRead.");
+        }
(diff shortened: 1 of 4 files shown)


src/IceRpc.Protobuf/Internal/PipeReaderExtensions.cs
```diff
@@ -35,6 +35,23 @@ internal static async ValueTask<T> DecodeProtobufMessageAsync<T>(
             cancellationToken).ConfigureAwait(false);
 
         Debug.Assert(message is not null);
```
bernardnormier: Do we have a separate backport for this Debug.Assert? It could be mixed-in the AsyncStream work.

bernardnormier (approved):

src/IceRpc.Protobuf/Internal/PipeReaderExtensions.cs
pepone: Good catch — that's actually the only line in main that meaningfully closes this surface in Release builds (the existing `Debug.Assert` is a no-op there, so a crafted empty payload currently slips through and returns `null` to the unary caller on 0.5.x).

The replacement landed in #4590 as part of the AsyncStream rewrite, but the assert → `?? throw new InvalidDataException(...)` swap is logically separable. I've folded it into this PR (force-pushed):

- `DecodeProtobufMessageAsync<T>` now uses main's `?? throw new InvalidDataException("The payload is empty; expected a Protobuf message.")`, with `where T : class, IMessage<T>`.
- The `class` constraint cascades to the public `InvokerExtensions` / `IncomingRequestExtensions` entry points (8 sites in each file). Whole solution builds clean; Protobuf-generated messages are always classes so no real caller is affected.
- Added `Decode_message_throws_invalid_data_exception_on_empty_payload` to lock in the new behaviour (verified passing in both Debug and Release configs).

PR #4544, opened 2026-04-23
Gap-check the first remote Slic stream ID

pepone:
## Summary
- The gap check in `ReadStreamDataFrameAsync` used `streamId > _lastRemote...StreamId + 4`, which is a no-op when `_lastRemote...` is null (first remote stream). A peer could open its first bidi stream at any ID, e.g. `0x4000000`, and subsequent frames addressed to never-opened lower IDs were silently ignored or read-and-discarded.
- Replaced the check with an explicit expected next stream ID: `last + 4`, or the initial ID for the role/direction (0/1 for bidi, 2/3 for uni) when no remote stream has opened yet.
- Added two regression cases to `ConnectionProtocolErrors` covering bogus initial bidirectional (`streamId=4`) and unidirectional (`streamId=6`) stream IDs.

Closes #4412

## Test plan
- [x] `dotnet build tests/IceRpc.Tests/IceRpc.Tests.csproj` (clean)
- [x] `dotnet test --filter "FullyQualifiedName~Connection_protocol_errors"` - 11/11 passed (9 pre-existing + 2 new)
- [x] `dotnet test --filter "FullyQualifiedName~Slic"` - 186/186 passed

PR #4633, opened 2026-05-13
Backport #4544 to 0.5.x

pepone:
Backport of #4544 — gap-check the first remote Slic stream ID.

## Summary
When `_lastRemoteBidirectionalStreamId is null` (i.e., we haven't yet seen a remote bidirectional stream), the gap check `streamId > _lastRemoteBidirectionalStreamId + 4` evaluated with `null + 4 == null`, and `streamId > null` is `false`, so **any** first stream ID (e.g. `4`, `6`, or `2^62 - 1`) was accepted. That let a peer skew the server's stream-id accounting and bypass the gap invariant. The fix initialises `_lastRemoteBidirectionalStreamId`/`_lastRemoteUnidirectionalStreamId` to a sentinel that makes the existing check work uniformly for the first stream and subsequent streams.

## Test plan
- [x] `dotnet test --filter "FullyQualifiedName~SlicTransportTests"` — 69/69 pass (2 new gap-check cases).

## Backport notes
Clean cherry-pick of 453157988.

**Sibling Tier B PR sharing this file:** [#4632](https://github.com/icerpc/icerpc-csharp/pull/4632) (backport of #4533, Slic idle timeout) — whichever lands first, the other will need a trivial rebase.

### DIFF
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -1459,7 +1459,10 @@ internal class SlicConnection : IMultiplexedConnection
 
             if (isBidirectional)
             {
-                if (streamId > _lastRemoteBidirectionalStreamId + 4)
+                ulong expectedStreamId = _lastRemoteBidirectionalStreamId is ulong lastId
+                    ? lastId + 4
+                    : (IsServer ? 0ul : 1ul);
+                if (streamId != expectedStreamId)
                 {
                     throw new InvalidDataException("Invalid stream ID.");
                 }
@@ -1474,7 +1477,10 @@ internal class SlicConnection : IMultiplexedConnection
             }
             else
             {
-                if (streamId > _lastRemoteUnidirectionalStreamId + 4)
+                ulong expectedStreamId = _lastRemoteUnidirectionalStreamId is ulong lastId
+                    ? lastId + 4
+                    : (IsServer ? 2ul : 3ul);
+                if (streamId != expectedStreamId)
                 {
                     throw new InvalidDataException("Invalid stream ID.");
                 }
--- a/tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
+++ b/tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
@@ -67,6 +67,26 @@ public class SlicTransportTests
                     $"Received {frameType} frame for unknown stream.");
             }
 
+            // Stream frame opening a new remote bidirectional stream with a non-initial stream ID.
+            // The first remote bidirectional stream ID the server expects from the client is 0.
+            yield return new TestCaseData(
+                (IDuplexConnection connection) => WriteStreamFrameAsync(
+                    connection,
+                    FrameType.Stream,
+                    streamId: 4ul,
+                    (ref SliceEncoder encoder) => encoder.EncodeBool(false)),
+                "Invalid stream ID.");
+
+            // Stream frame opening a new remote unidirectional stream with a non-initial stream ID.
+            // The first remote unidirectional stream ID the server expects from the client is 2.
+            yield return new TestCaseData(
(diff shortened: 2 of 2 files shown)


src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1459,7 +1459,14 @@ private async Task ReadStreamDataFrameAsync(
...
             {
-                if (streamId > _lastRemoteBidirectionalStreamId + 4)
+                // The next expected remote bidirectional stream ID is the last one plus 4, or the initial remote
+                // bidirectional stream ID (0 for the server, 1 for the client) if no remote bidirectional stream has
+                // been opened yet. This check also rejects a bogus first stream ID, which would otherwise slip
+                // through because `null + 4` evaluates to null.
+                ulong expectedStreamId = _lastRemoteBidirectionalStreamId is ulong lastId
+                    ? lastId + 4
+                    : (IsServer ? 0ul : 1ul);
+                if (streamId > expectedStreamId)
```
bernardnormier: Why is this check `>` as opposed to `!=` ? We're creating a new stream in this block and it's ok to get a stream ID less than the expected stream ID?

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1459,7 +1459,14 @@ private async Task ReadStreamDataFrameAsync(
 
             if (isBidirectional)
             {
-                if (streamId > _lastRemoteBidirectionalStreamId + 4)
+                // The next expected remote bidirectional stream ID is the last one plus 4, or the initial remote
+                // bidirectional stream ID (0 for the server, 1 for the client) if no remote bidirectional stream has
+                // been opened yet. This check also rejects a bogus first stream ID, which would otherwise slip
+                // through because `null + 4` evaluates to null.
```
bernardnormier: I find this comment over verbose. It only make sense when you read the previous code.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1474,7 +1481,14 @@ private async Task ReadStreamDataFrameAsync(
             }
             else
             {
-                if (streamId > _lastRemoteUnidirectionalStreamId + 4)
+                // The next expected remote unidirectional stream ID is the last one plus 4, or the initial remote
+                // unidirectional stream ID (2 for the server, 3 for the client) if no remote unidirectional stream
+                // has been opened yet. This check also rejects a bogus first stream ID, which would otherwise slip
+                // through because `null + 4` evaluates to null.
+                ulong expectedStreamId = _lastRemoteUnidirectionalStreamId is ulong lastId
```
bernardnormier, line 1480: Same as above

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
pepone: This `>` is pre-existing behavior but I think you are right `!=` should be clear, SLIC opens streams in order.

pepone: Changed both checks to `!=`. SlicTransportTests 69/69 still pass; full IceRpc.Tests suite 950/950 pass (29 skipped). Force-pushed.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
pepone: Removed the comment — with the explicit `expectedStreamId` variable and the `!=` check, the code reads on its own.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
pepone: Same — removed.

bernardnormier (approved):

externl (approved):

PR #4548, opened 2026-04-24
Wrap URI-based server address decoding errors as InvalidDataException

pepone:
Fixes #4427.

PR #4634, opened 2026-05-13
Backport #4548 to 0.5.x

pepone:
Backport of #4548 — wrap URI-based server address decoding errors as `InvalidDataException`.

## Summary
`new ServerAddress(new Uri(decoder.DecodeString()))` could throw `UriFormatException` (malformed scheme) or `ArgumentException` (unsupported protocol) on a crafted payload, which escaped as an unhandled parser exception (DoS / misclassified dispatch) rather than the documented `InvalidDataException`. The fix routes both `TransportCode.Uri` branches through a new `DecodeUriServerAddress` helper that catches those and wraps them.

## Test plan
- [x] `dotnet test --filter "FullyQualifiedName~ProxyTests"` — 38/38 pass (2 new tests, parameterised on ice + icerpc).

## Backport notes
Cherry-pick of 88514baad. The original lands in `src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs` (a file that exists only on main, where `IceRpc.Ice` is split out). On 0.5.x the same two `TransportCode.Uri` call sites live in `src/IceRpc.Slice/ServiceAddressSliceDecoderExtensions.cs`. Ported by hand:
- Replaced both `new ServerAddress(new Uri(decoder.DecodeString()))` call sites with the new `DecodeUriServerAddress` helper.
- Added the helper near the bottom of the class.
- Dropped the cherry-pick's `tests/IceRpc.Ice.Tests/ProxyTests.cs` additions and re-implemented the two new tests inside `tests/IceRpc.Slice.Tests/ProxyTests.cs` using 0.5.x's unified `SliceEncoder`/`SliceDecoder` + `PingableProxy` types and explicit `SliceEncoding.Slice1`.

**Sibling Tier B PR sharing this file:** [#4531 backport](https://github.com/icerpc/icerpc-csharp/pull/4531) (escape adapter-id) will be sent next — whichever lands first, the other will need a trivial rebase.

### DIFF
--- a/src/IceRpc.Slice/ServiceAddressSliceDecoderExtensions.cs
+++ b/src/IceRpc.Slice/ServiceAddressSliceDecoderExtensions.cs
@@ -110,7 +110,7 @@ public static class ServiceAddressSliceDecoderExtensions
                         break;
 
                     case TransportCode.Uri:
-                        serverAddress = new ServerAddress(new Uri(decoder.DecodeString()));
+                        serverAddress = DecodeUriServerAddress(decoder.DecodeString());
                         if (serverAddress.Value.Protocol != protocol)
                         {
                             throw new InvalidDataException(
@@ -151,7 +151,7 @@ public static class ServiceAddressSliceDecoderExtensions
             else if (transportCode == TransportCode.Uri)
             {
                 // The server addresses of Slice1 encoded icerpc proxies only use TransportCode.Uri.
-                serverAddress = new ServerAddress(new Uri(decoder.DecodeString()));
+                serverAddress = DecodeUriServerAddress(decoder.DecodeString());
                 if (serverAddress.Value.Protocol != protocol)
                 {
                     throw new InvalidDataException(
@@ -177,6 +177,20 @@ public static class ServiceAddressSliceDecoderExtensions
         }
 
         return serverAddress.Value;
+
+        static ServerAddress DecodeUriServerAddress(string uriString)
+        {
+            try
+            {
+                return new ServerAddress(new Uri(uriString));
+            }
+            catch (Exception exception) when (exception is UriFormatException or ArgumentException)
+            {
+                throw new InvalidDataException(
+                    $"Received invalid server address URI '{uriString}'.",
+                    exception);
+            }
+        }
     }
 
     /// <summary>Decodes a service address encoded with Slice1.</summary>
--- a/tests/IceRpc.Slice.Tests/ProxyTests.cs
+++ b/tests/IceRpc.Slice.Tests/ProxyTests.cs
@@ -89,6 +89,73 @@ public partial class ProxyTests
         Assert.That(decoded.ServiceAddress, Is.EqualTo(expected));
     }
 
+    // Using transport=quic forces TransportCode.Uri on the ice side (only tcp/ssl have dedicated codes);
+    // icerpc always uses TransportCode.Uri.
+    [TestCase("ice://hello.zeroc.com/hello?transport=quic", "ice:", SliceEncoding.Slice1)]
+    [TestCase("icerpc://hello.zeroc.com/hello", "icerpc:", SliceEncoding.Slice1)]
+    [TestCase("icerpc://hello.zeroc.com/hello", "icerpc:", SliceEncoding.Slice2)]
+    public void Decode_proxy_with_unparsable_server_address_uri_throws_invalid_data(
+        ServiceAddress value,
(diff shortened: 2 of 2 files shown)


externl (approved):

.vscode/cspell.json
```diff
@@ -70,6 +70,7 @@
     "uncategorized",
     "unflushed",
     "uninstantiated",
+    "unparseable",
```
bernardnormier: We should use unparsable (no e).

src/IceRpc.Slice/ServiceAddressSliceDecoderExtensions.cs
```diff
@@ -299,4 +299,19 @@ private static ServerAddress DecodeTcpServerAddressBody(this ref SliceDecoder de
                 exception);
         }
     }
+
+    /// <summary>Decodes a server address from its URI string representation.</summary>
```
bernardnormier: It would be nicer to make this method a local static function.

tests/IceRpc.Slice.Tests/ProxyTests.cs
```diff
@@ -89,6 +89,71 @@ public void Decode_proxy(ServiceAddress value, ServiceAddress? expected, SliceEn
         Assert.That(decoded.ServiceAddress, Is.EqualTo(expected));
     }
 
+    // Using transport=quic forces TransportCode.Uri on the ice side (only tcp/ssl have dedicated codes);
+    // icerpc always uses TransportCode.Uri.
+    [TestCase("ice://hello.zeroc.com/hello?transport=quic", "ice:", SliceEncoding.Slice1)]
+    [TestCase("icerpc://hello.zeroc.com/hello", "icerpc:", SliceEncoding.Slice1)]
```
bernardnormier, line 95: We should use Slice2 otherwise we don't test the Slice2 path.

tests/IceRpc.Slice.Tests/ProxyTests.cs
```diff
@@ -89,6 +89,71 @@ public void Decode_proxy(ServiceAddress value, ServiceAddress? expected, SliceEn
...
+                var decoder = new SliceDecoder(buffer.AsMemory(0, length), encoding: encoding);
+                _ = decoder.DecodePingableProxy();
+            },
+            Throws.InstanceOf<InvalidDataException>());
+    }
+
+    // Using transport=quic forces TransportCode.Uri on the ice side (only tcp/ssl have dedicated codes);
+    // icerpc always uses TransportCode.Uri.
+    [TestCase("ice://hello.zeroc.com/hello?transport=quic", "ice:", "ftp:", SliceEncoding.Slice1)]
+    [TestCase("icerpc://hello.zeroc.com/hello", "icerpc:", "gopher:", SliceEncoding.Slice1)]
```
bernardnormier, line 127: Same here

.vscode/cspell.json
pepone: Done — renamed everywhere (cspell dictionary, test method names, source comment).

src/IceRpc.Slice/ServiceAddressSliceDecoderExtensions.cs
pepone: Done — moved `DecodeUriServerAddress` to a local `static` function inside `DecodeServerAddress` (the only caller). Both call sites still go through it.

tests/IceRpc.Slice.Tests/ProxyTests.cs
pepone: Added a `SliceEncoding.Slice2` case for the icerpc proxy (kept the `Slice1` icerpc case too so the second call site in `DecodeServerAddress` — the `else if (transportCode == TransportCode.Uri)` branch — stays covered).

tests/IceRpc.Slice.Tests/ProxyTests.cs
pepone: Same — added a `Slice2` icerpc case alongside the existing `Slice1` one.

.vscode/cspell.json
pepone: Follow-up: verified that cspell accepts `unparsable` natively without any dictionary entry, so the `unparseable` entry was dead weight to begin with. Simplified the PR to just **delete** the entry (instead of renaming it to `unparsable`). Force-pushed.

bernardnormier (approved):

Issue #4456, opened 2026-04-15
[Audit-Medium] request-context decoding accepts trailing bytes instea…


pepone:
> **AI-generated audit finding** — this issue was opened from an automated security/correctness audit. It **has not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

### Medium: request-context decoding accepts trailing bytes instead of rejecting malformed input — **CONFIRMED**

Affected code:

- [src/IceRpc.RequestContext/RequestContextMiddleware.cs:30-44](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc.RequestContext/RequestContextMiddleware.cs#L30) — creates `IceDecoder`/`SliceDecoder` directly over `value` and calls `DecodeDictionary` without end-of-buffer checks
- [src/IceRpc/Ice/Codec/ReadOnlySequenceExtensions.cs:17](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/IceRpc/Ice/Codec/ReadOnlySequenceExtensions.cs#L17), [src/ZeroC.Slice.Codec/ReadOnlySequenceExtensions.cs:18](https://github.com/icerpc/icerpc-csharp/blob/150d6ee01de9609eb037963cf2897df190c23809/src/ZeroC.Slice.Codec/ReadOnlySequenceExtensions.cs#L18) — the idiomatic whole-buffer helpers (`DecodeIceBuffer` / `DecodeSliceBuffer`) that *do* call `CheckEndOfBuffer`

Verification:

Confirmed. Both branches in `DispatchAsync` construct a decoder directly, decode a dictionary, and return. Neither branch calls `decoder.CheckEndOfBuffer()`, and neither uses the convenience helper that would. A peer can append any extra bytes after the valid dictionary encoding inside the `Context` field payload and the middleware accepts the field as well-formed.

Impact:

- Malformed request metadata is accepted instead of rejected.
- Weakens the malformed-input contract consistently applied to other request-field decode paths.

Recommendation:

- Switch both branches to the whole-buffer helpers (`value.DecodeIceBuffer(...)` / `value.DecodeSliceBuffer(...)`) so end-of-buffer validation happens automatically.
- Or call `decoder.CheckEndOfBuffer()` explicitly before handing the dictionary to the feature.
[...]

PR #4636, opened 2026-05-13
Backport #4521 to 0.5.x

pepone:
Backport of #4521 (fix [#4456](https://github.com/icerpc/icerpc-csharp/issues/4456)) — reject trailing bytes in the request context field.

## Summary
`RequestContextMiddleware` decoded the `RequestFieldKey.Context` dictionary and silently ignored any trailing bytes. That's a smuggling primitive across middleware boundaries (the middleware sees one dictionary, a downstream component re-decoding the same field could observe extra bytes). The middleware now throws `InvalidDataException` when the decoder hasn't drained the buffer after decoding the dictionary.

## Test plan
- [x] `dotnet test tests/IceRpc.RequestContext.Tests` — 5/5 pass (3 original + 2 new trailing-bytes cases for ice and icerpc).

## Backport notes
Cherry-pick of dd980a1d1. Adaptations:
- Main's fix uses the post-split `IceDecoder` / `SliceDecoder` types and the `DecodeIceBuffer`/`DecodeSliceBuffer` extension helpers (which internally check for trailing bytes). 0.5.x keeps a unified `SliceDecoder` carrying the `SliceEncoding` parameter. Implemented the check directly with `if (decoder.Remaining != 0) throw new InvalidDataException(...)` after `DecodeDictionary`.
- The new test used the main-only `IceEncoder` (and `IceRpc.Ice.Codec` namespace). Rewrote it on top of 0.5.x's unified `SliceEncoder` constructed with `SliceEncoding.Slice1` (for ice) or `SliceEncoding.Slice2` (for icerpc); the rest of the test is unchanged.

### DIFF
--- a/src/IceRpc.RequestContext/RequestContextMiddleware.cs
+++ b/src/IceRpc.RequestContext/RequestContextMiddleware.cs
@@ -32,6 +32,14 @@ public class RequestContextMiddleware : IDispatcher
                 size => new Dictionary<string, string>(size),
                 keyDecodeFunc: (ref SliceDecoder decoder) => decoder.DecodeString(),
                 valueDecodeFunc: (ref SliceDecoder decoder) => decoder.DecodeString());
+
+            // The context field's buffer must be fully consumed; reject trailing bytes after the dictionary.
+            if (decoder.Remaining != 0)
+            {
+                throw new InvalidDataException(
+                    $"Unexpected trailing {decoder.Remaining} byte(s) in request context field.");
+            }
+
             if (context.Count > 0)
             {
                 request.Features = request.Features.With<IRequestContextFeature>(new RequestContextFeature(context));
--- a/tests/IceRpc.RequestContext.Tests/RequestContextMiddlewareTests.cs
+++ b/tests/IceRpc.RequestContext.Tests/RequestContextMiddlewareTests.cs
@@ -57,4 +57,43 @@ public sealed class RequestContextMiddlewareTests
             return pipe.Reader;
         }
     }
+
+    /// <summary>Verifies that a context field with trailing bytes is rejected with
+    /// <see cref="InvalidDataException" />.</summary>
+    [TestCase("icerpc")]
+    [TestCase("ice")]
+    public void Context_field_with_trailing_bytes_is_rejected(string protocolName)
+    {
+        Protocol protocol = Protocol.Parse(protocolName);
+        var pipe = new Pipe();
+        var encoder = new SliceEncoder(
+            pipe.Writer,
+            protocol == Protocol.Ice ? SliceEncoding.Slice1 : SliceEncoding.Slice2);
+        encoder.EncodeDictionary(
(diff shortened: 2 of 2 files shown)


src/IceRpc.RequestContext/RequestContextMiddleware.cs
```diff
@@ -32,6 +32,15 @@ public ValueTask<OutgoingResponse> DispatchAsync(IncomingRequest request, Cancel
                 size => new Dictionary<string, string>(size),
                 keyDecodeFunc: (ref SliceDecoder decoder) => decoder.DecodeString(),
                 valueDecodeFunc: (ref SliceDecoder decoder) => decoder.DecodeString());
+
+            // Reject trailing bytes after the dictionary: a request-context field that decodes to a valid
+            // dictionary followed by extra bytes is a smuggling primitive across middleware boundaries.
```
externl: This sounds a little off: `followed by extra bytes is a smuggling primitive across middleware boundaries`

externl (approved):

src/IceRpc.RequestContext/RequestContextMiddleware.cs
pepone: Good call, that phrasing was unique to this 0.5.x backport (main's fix uses `DecodeIceBuffer`/`DecodeSliceBuffer` helpers and has no inline comment). Reworded the comment to match the doc-string on those helpers — short, factual, no jargon. Force-pushed.

tests/IceRpc.RequestContext.Tests/RequestContextMiddlewareTests.cs
```diff
@@ -57,4 +57,44 @@ static PipeReader EncodeContextField(Dictionary<string, string> context)
...
+        var pipe = new Pipe();
+        var encoder = new SliceEncoder(
+            pipe.Writer,
+            protocol == Protocol.Ice ? SliceEncoding.Slice1 : SliceEncoding.Slice2);
+        encoder.EncodeDictionary(
+            new Dictionary<string, string> { ["Foo"] = "Bar" },
+            (ref SliceEncoder encoder, string key) => encoder.EncodeString(key),
+            (ref SliceEncoder encoder, string value) => encoder.EncodeString(value));
+        // Append extra trailing bytes.
+        pipe.Writer.GetMemory(4).Span.Clear();
```
bernardnormier: It would clearer to write 4 bytes. It's not obvious that:
```cs
pipe.Writer.GetMemory(4).Span.Clear();
pipe.Writer.Advance(4);
```

means "write 4 bytes set to 0".

bernardnormier (approved):

tests/IceRpc.RequestContext.Tests/RequestContextMiddlewareTests.cs
pepone: Fair point — switched to `pipe.Writer.Write(new byte[] { 0, 0, 0, 0 });` so it's obvious. Force-pushed.

PR #4547, opened 2026-04-23
Clean up DeadlineMiddleware/Interceptor edge cases

pepone:
Two related edge-case fixes on the deadline middleware/interceptor. Kept in one PR because they touch the same files and share the same theme.

## Commit 1 — DateTime.MinValue sentinel (closes #4419)
`DeadlineMiddleware` used the decoded `DateTime.MinValue` as its "field absent" sentinel. A peer encoding \`ticks=0\` decodes to exactly that value, so an explicit MinValue deadline was indistinguishable from an absent field and bypassed deadline enforcement entirely.

Switched to explicit \`TryGetValue\` on the fields dictionary, then decode + enforce. Added a regression test encoding \`new DateTime(0, DateTimeKind.Utc)\`.

## Commit 2 — extreme future deadline clamp (closes #4420)
\`CancellationTokenSource.CancelAfter(TimeSpan)\` rejects values greater than \`int.MaxValue\` ms (~24.8 days). A peer-encoded deadline thousands of years in the future (or a caller-supplied \`IDeadlineFeature\` near \`DateTime.MaxValue\`) produced a TimeSpan that \`CancelAfter\` rejected with \`ArgumentOutOfRangeException\`, leaking as a generic \`InternalError\` response.

- `DeadlineMiddleware`: clamp the computed timeout before `PerformDispatchAsync`.
- `DeadlineInterceptor`: clamp the timeout in `PerformInvokeAsync`; reject configured `defaultTimeout` beyond the `CancelAfter` max at construction.

At the clamp bound (~24.8 days) the deadline is effectively infinite for RPC purposes; matches the precedent set by `HttpClient.Timeout`.

Closes #4419
Closes #4420

## Test plan
- [x] `dotnet build src/IceRpc.Deadline/IceRpc.Deadline.csproj` (clean)
- [x] `dotnet test tests/IceRpc.Deadline.Tests/IceRpc.Deadline.Tests.csproj` — 17/17 passed
- [x] All four new tests fail when the production fixes are stashed (verified before/after)

PR #4517, opened 2026-04-16
Normalize deadline feature DateTime to UTC

pepone:
Fixes #4421.

PR #4655, opened 2026-05-15
Backport #4517 + #4547 + #4668 to 0.5.x

pepone:
Bundled backport of three related deadline-handling fixes from main, squashed because [#4547](https://github.com/icerpc/icerpc-csharp/pull/4547) introduces clamping behavior that [#4668](https://github.com/icerpc/icerpc-csharp/pull/4668) then reworks to clean rejection — landing them as a single backport gives 0.5.x the final intended shape directly.

## Summary
- **[#4517](https://github.com/icerpc/icerpc-csharp/pull/4517)** — Normalize the `IDeadlineFeature` `DateTime` to UTC. A caller setting `IDeadlineFeature` with a `Local` or `Unspecified` `DateTime` currently encodes the wrong instant on the wire.
- **[#4547](https://github.com/icerpc/icerpc-csharp/pull/4547)** — Validate `defaultTimeout` / `IDeadlineFeature.FromTimeout` inputs at construction; handle near-`MaxValue` deadlines and a peer-encoded `DateTime.MinValue` (was ambiguous with the "field absent" sentinel).
- **[#4668](https://github.com/icerpc/icerpc-csharp/pull/4668)** — Reject deadlines beyond the supported maximum (~24.8 days) with `NotSupportedException` (interceptor) or `StatusCode.NotSupported` (middleware) instead of silently clamping. Also cleans up doc-comments to not reference `CancellationTokenSource.CancelAfter` implementation detail.

## Test plan
- [x] `dotnet test tests/IceRpc.Deadline.Tests` — 17/17 pass.
- [x] `dotnet test tests/IceRpc.Tests --filter "FullyQualifiedName~DeadlineFeatureTests"` — 2/2 pass.

## Backport notes
`DeadlineMiddleware` on main uses the `value.DecodeSliceBuffer(...)` extension helper from `ZeroC.Slice.Codec`; neither exists on 0.5.x. The 0.5.x port replaces both with `new SliceDecoder(value, SliceEncoding.Slice2)` and `using ZeroC.Slice`. Same adaptation applied to the test files.

Surfaced by the [2026-05-15 follow-up audit](https://github.com/icerpc/icerpc-csharp-audit/blob/main/0.5.x-backport-audit-2026-05-15.md); reworked per [Bernard's review](https://github.com/icerpc/icerpc-csharp/pull/4655#pullrequestreview-...) before merging.

### DIFF
--- a/src/IceRpc.Deadline/DeadlineInterceptor.cs
+++ b/src/IceRpc.Deadline/DeadlineInterceptor.cs
@@ -25,6 +25,10 @@ namespace IceRpc.Deadline;
 /// <seealso cref="DeadlineInvokerBuilderExtensions"/>
 public class DeadlineInterceptor : IInvoker
 {
+    // The maximum supported timeout (int.MaxValue ms, ~24.8 days). This is the maximum delay
+    // CancellationTokenSource.CancelAfter accepts.
+    private static readonly TimeSpan _maxSupportedTimeout = TimeSpan.FromMilliseconds(int.MaxValue);
+
     private readonly bool _alwaysEnforceDeadline;
     private readonly IInvoker _next;
     private readonly TimeSpan _defaultTimeout;
@@ -32,17 +36,28 @@ public class DeadlineInterceptor : IInvoker
 
     /// <summary>Constructs a Deadline interceptor.</summary>
     /// <param name="next">The next invoker in the invocation pipeline.</param>
-    /// <param name="defaultTimeout">The default timeout. When not infinite, the interceptor adds a deadline to requests
-    /// without a deadline.</param>
-    /// <param name="timeProvider">The optional time provider used to obtain the current time. If <see langword="null"/>, it uses
-    /// <see cref="TimeProvider.System"/>.</param>
+    /// <param name="defaultTimeout">The default timeout applied to requests without a deadline. Must be a positive
+    /// value not exceeding ~24.8 days, or <see cref="Timeout.InfiniteTimeSpan" /> to disable the default timeout
+    /// entirely.</param>
+    /// <param name="timeProvider">The optional time provider used to obtain the current time. If
+    /// <see langword="null"/>, it uses <see cref="TimeProvider.System"/>.</param>
     /// <param name="alwaysEnforceDeadline">When <see langword="true" /> and the request carries a deadline, the
     /// interceptor always creates a cancellation token source to enforce this deadline. When <see langword="false" />
     /// and the request carries a deadline, the interceptor creates a cancellation token source to enforce this deadline
     /// only when the invocation's cancellation token cannot be canceled. The default value is <see langword="false" />.
     /// </param>
+    /// <exception cref="ArgumentException">Thrown if <paramref name="defaultTimeout" /> is neither
+    /// <see cref="Timeout.InfiniteTimeSpan" /> nor a positive value within the supported range.</exception>
     public DeadlineInterceptor(IInvoker next, TimeSpan defaultTimeout, bool alwaysEnforceDeadline, TimeProvider? timeProvider = null)
     {
+        if (defaultTimeout != Timeout.InfiniteTimeSpan &&
+            (defaultTimeout <= TimeSpan.Zero || defaultTimeout > _maxSupportedTimeout))
+        {
+            throw new ArgumentException(
+                $"The {nameof(defaultTimeout)} value must be Timeout.InfiniteTimeSpan or a positive value not exceeding {_maxSupportedTimeout}.",
+                nameof(defaultTimeout));
+        }
+
         _next = next;
         _alwaysEnforceDeadline = alwaysEnforceDeadline;
         _defaultTimeout = defaultTimeout;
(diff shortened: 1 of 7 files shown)


src/IceRpc.Deadline/DeadlineInterceptor.cs
```diff
@@ -25,24 +25,48 @@ namespace IceRpc.Deadline;
...
     /// <summary>Constructs a Deadline interceptor.</summary>
     /// <param name="next">The next invoker in the invocation pipeline.</param>
-    /// <param name="defaultTimeout">The default timeout. When not infinite, the interceptor adds a deadline to requests
-    /// without a deadline.</param>
-    /// <param name="timeProvider">The optional time provider used to obtain the current time. If <see langword="null"/>, it uses
-    /// <see cref="TimeProvider.System"/>.</param>
+    /// <param name="defaultTimeout">The default timeout applied to requests without a deadline. Must be positive and
+    /// must not exceed <see cref="int.MaxValue" /> milliseconds (~24.8 days), the maximum supported by
+    /// <see cref="CancellationTokenSource.CancelAfter(TimeSpan)" />, or <see cref="Timeout.InfiniteTimeSpan" /> to
+    /// disable this behavior.</param>
```
bernardnormier: Something is not correct with this sentence.

src/IceRpc.Deadline/DeadlineInterceptor.cs
```diff
@@ -25,24 +25,48 @@ namespace IceRpc.Deadline;
...
+    /// <param name="timeProvider">The optional time provider used to obtain the current time. If
+    /// <see langword="null"/>, it uses <see cref="TimeProvider.System"/>.</param>
     /// <param name="alwaysEnforceDeadline">When <see langword="true" /> and the request carries a deadline, the
     /// interceptor always creates a cancellation token source to enforce this deadline. When <see langword="false" />
     /// and the request carries a deadline, the interceptor creates a cancellation token source to enforce this deadline
     /// only when the invocation's cancellation token cannot be canceled. The default value is <see langword="false" />.
     /// </param>
+    /// <exception cref="ArgumentException">Thrown if <paramref name="defaultTimeout" /> is not
+    /// <see cref="Timeout.InfiniteTimeSpan" />, not positive, or exceeds the maximum supported value.</exception>
```
bernardnormier: This should be rewritten in a clearer manner.

src/IceRpc.Deadline/DeadlineMiddleware.cs
```diff
@@ -2,17 +2,24 @@
...
 using IceRpc.Features;
+using System.Buffers;
 using ZeroC.Slice;
 
 namespace IceRpc.Deadline;
 
 /// <summary>Represents a middleware that decodes deadline fields into deadline features. When the decoded deadline
 /// expires, this middleware cancels the dispatch and returns an <see cref="OutgoingResponse" /> with status code
 /// <see cref="StatusCode.DeadlineExceeded" />.</summary>
+/// <remarks>A peer-encoded deadline whose computed remaining timeout exceeds the
```
bernardnormier: I don't like this middleware behavior. The middleware should not clamp the deadline provided by the client due to implementation limits.
You send me deadline 5 min in the future? Sorry, my implementation only supports 3 min, so you get 3 min. This doesn't make sense.

Here, I'd rather reject the request with a NotSupported exception.

One complicating factor is the request could be a one-way request where throwing NotSupported is not super useful. At least you get the log.

src/IceRpc/Features/DeadlineFeature.cs
```diff
@@ -5,10 +5,32 @@ namespace IceRpc.Features;
...
 public sealed class DeadlineFeature : IDeadlineFeature
 {
+    // The maximum delay CancellationTokenSource.CancelAfter(TimeSpan) accepts.
+    private static readonly TimeSpan MaxCancelAfterDelay = TimeSpan.FromMilliseconds(int.MaxValue);
+
     /// <summary>Creates a deadline from a timeout.</summary>
-    /// <param name="timeout">The timeout.</param>
+    /// <param name="timeout">The timeout. Must be positive and must not exceed
+    /// <see cref="int.MaxValue" /> milliseconds (~24.8 days), the maximum supported by
+    /// <see cref="CancellationTokenSource.CancelAfter(TimeSpan)" />.</param>
```
bernardnormier: Explaining in the doc-comment that int.MaxValue milliseconds is the maximum supported by CancelAfter is more confusing than anything else. We should just document the max, not where it's coming from.

bernardnormier (commented):
I think we should rework the code and comments on main before backporting it to 0.5.x.

src/IceRpc.Deadline/DeadlineMiddleware.cs
pepone: We can throw NotSupportedException, this large deadlines are non-sensical (~24.8 days is the max supported by the implementation, which is already way to big to make sense as a deadline value).

bernardnormier (approved):

PR #4662, opened 2026-05-15
Cover the regression fixed by #4285 with a wire-level test

pepone:
## Summary
The tests added by [#4285](https://github.com/icerpc/icerpc-csharp/pull/4285) (\`Response_reply_status_maps_to_expected_status_code\`) round-trip \`StatusCode\` values through IceRpc's own encoder, which never produces a reply-status byte the client doesn't recognize — so they don't actually exercise the \`DecodeResponseHeader\` fix.

[Bernard](https://github.com/icerpc/icerpc-csharp/pull/4654#discussion_r3247225623) and [Copilot](https://github.com/icerpc/icerpc-csharp/pull/4654#discussion_r3247225418) flagged this on the 0.5.x backport ([#4654](https://github.com/icerpc/icerpc-csharp/pull/4654)).

This PR adds \`Invocation_succeeds_with_unknown_reply_status\`:

- A client-side \`ReadDecorator\` scans each read for the request-ID marker \`0x01,0x00,0x00,0x00\` (the first invocation always uses request ID 1) and overwrites the byte immediately after — the reply-status — with an unknown value (99).
- The client should return an \`IncomingResponse\` with \`StatusCode.InternalError\` (the default mapping for unknown system exceptions in \`DecodeResponseHeader\`), **not** throw \`InvalidDataException\`.

## Test plan
- [x] \`dotnet test --filter "FullyQualifiedName~Invocation_succeeds_with_unknown_reply_status"\` — passes with the fix in place.
- [x] Verified the test **fails** if the fix is reverted to \`((int)buffer.FirstSpan[0]).AsReplyStatus()\`.
- [x] Full \`IceProtocolConnectionTests\` class — 48/48 pass.

## Follow-up
Once this lands on main, I'll backport it to 0.5.x alongside the existing \`DecodeResponseHeader\` fix that's already there via [#4654](https://github.com/icerpc/icerpc-csharp/pull/4654).

### DIFF
--- a/tests/IceRpc.Tests/IceProtocolConnectionTests.cs
+++ b/tests/IceRpc.Tests/IceProtocolConnectionTests.cs
@@ -208,6 +208,67 @@ public sealed class IceProtocolConnectionTests
         Assert.That(response.ErrorMessage, Is.EqualTo(expectedErrorMessage));
     }
 
+    /// <summary>Verifies that an Ice reply with a reply-status value the client does not know is decoded as
+    /// <see cref="StatusCode.InternalError" /> rather than failing with <see cref="InvalidDataException" />.
+    /// This is the regression covered by the direct-cast fix in <c>DecodeResponseHeader</c>: the generated
+    /// <c>AsReplyStatus()</c> converter throws on unrecognized values, so without the fix a peer sending a
+    /// reply-status the client doesn't yet know about would surface as a protocol error rather than a
+    /// normal failed response.</summary>
+    [Test]
+    public async Task Invocation_succeeds_with_unknown_reply_status()
+    {
+        // Arrange
+        // The dispatcher result is irrelevant; the client-side ReadDecorator below rewrites the reply-status
+        // byte on the wire.
+        var dispatcher = new InlineDispatcher((request, cancellationToken) =>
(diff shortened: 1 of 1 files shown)


tests/IceRpc.Tests/IceProtocolConnectionTests.cs
```diff
@@ -208,6 +208,67 @@ public async Task Response_reply_status_maps_to_expected_status_code(
         Assert.That(response.ErrorMessage, Is.EqualTo(expectedErrorMessage));
     }
 
+    /// <summary>Verifies that an Ice reply with a reply-status value the client does not know is decoded as
+    /// <see cref="StatusCode.InternalError" /> rather than failing with <see cref="InvalidDataException" />.
+    /// This is the regression covered by the direct-cast fix in <c>DecodeResponseHeader</c>: the generated
+    /// <c>AsReplyStatus()</c> converter throws on unrecognized values, so without the fix a peer sending a
+    /// reply-status the client doesn't yet know about would surface as a protocol error rather than a
+    /// normal failed response.</summary>
```
bernardnormier, line 216: We should not include "This is the regression ..." in the test comment or doc-comment. Way too verbose.

bernardnormier (commented):
I'd rather have no test for this very simple code than add such a complicated test.

PR #4676, opened 2026-05-19
Accept empty payloads in Protobuf-encoded responses

bernardnormier:
This is a follow-up to #4675.

When IceRpc.Protobuf receives a non-streaming IncomingResponse, it expects the response's payload to contain a length-prefixed Protobuf message.

Prior to this PR, if the response payload is empty - which is what happens with a oneway request - the decoding fails.

This PR updates the logic to accept an empty payload in this situation (non-streaming response): empty is treated like an empty (zero byte) Protobuf message. 

Such a Protobuf message is decoded as default-initialized message. This works for google.protobuf.empty and any other messages. You just get a default initialized message.

### DIFF
--- a/src/IceRpc.Protobuf/RpcMethods/InvokerExtensions.cs
+++ b/src/IceRpc.Protobuf/RpcMethods/InvokerExtensions.cs
@@ -230,6 +230,7 @@ public static class InvokerExtensions
                 return await response.Payload.DecodeProtobufMessageAsync(
                     messageParser,
                     protobufFeature.MaxMessageLength,
+                    acceptEmptyPayload: true,
                     cancellationToken).ConfigureAwait(false);
             }
             else
--- a/src/IceRpc.Protobuf/RpcMethods/InvokerExtensions.cs
+++ b/src/IceRpc.Protobuf/RpcMethods/InvokerExtensions.cs
@@ -230,7 +230,9 @@ public static class InvokerExtensions
                 return await response.Payload.DecodeProtobufMessageAsync(
                     messageParser,
                     protobufFeature.MaxMessageLength,
-                    acceptEmptyPayload: true,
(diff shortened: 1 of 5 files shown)


tests/IceRpc.Protobuf.Tests/OperationTests.cs
```diff
@@ -36,6 +36,46 @@ public async Task Unary_rpc()
...
+        var client = new MyOperationsClient(pipeline);
+
+        var message = new InputMessage()
+        {
+            P1 = "P1",
+            P2 = 2,
+        };
+
+        // Act
+        var response = await client.UnaryOpAsync(message);
```
InsertCreativityHere, line 71: It's very convenient that decoding `[]` gets us the default representation,
but I find it a little weird that firing a oneway request will return a default-constructed message.

Unfortunately, I don't have enough familiarity here to suggest anything :)

InsertCreativityHere (approved):

src/IceRpc.Protobuf/RpcMethods/InvokerExtensions.cs
```diff
@@ -230,6 +230,7 @@ private static async Task<TOutput> ReceiveResponseAsync<TOutput>(
                 return await response.Payload.DecodeProtobufMessageAsync(
                     messageParser,
                     protobufFeature.MaxMessageLength,
+                    acceptEmptyPayload: true,
```
pepone: Shouldn't we set this base on IsOneway?

pepone (approved):

src/IceRpc.Protobuf/RpcMethods/InvokerExtensions.cs
bernardnormier: Fixed.

PR #4679, opened 2026-05-20
Add support for Protobuf editions

### LINKED ISSUE #4677: Add support for Protobuf editions to the Protobuf plug-ins
See https://protobuf.dev/editions/implementation/#generators

As I understand it, i's just a matter of setting the supported_features field in the CodeGeneratorResponse.

bernardnormier:
Fixes #4677 

I also asked Claude if we should update the examples to use edition:

My take: **keep `syntax = "proto3"` in the examples**, at least for now.

Reasons:

1. **Examples are teaching material, not a showcase.** People landing on the Protobuf examples are almost certainly coming from a proto3 background — that's still what every tutorial, every existing `.proto` file in the wild, and every Google doc uses. Greeting them with `edition = "2024"` adds a "wait, what's this?" speed bump before they even get to the IceRPC-specific parts, which is what the examples are actually meant to demonstrate.

2. **Editions are still very new.** Edition 2023 only shipped in late 2023 and 2024 only recently. Tooling support outside of `protoc` itself (IDEs, linters, third-party generators, docs) is uneven. Examples that "just work" when copy-pasted are more valuable than examples that look modern.

3. **Signaling matters.** Using `edition = "2024"` in *all* examples implicitly suggests "this is what IceRPC recommends / requires." Since you explicitly don't want to require any version, the conservative default sends the right message: *bring your own proto dialect, we don't care*.

4. **The interesting bit doesn't change.** Editions don't unlock anything that makes the Greeter example more illustrative. The .proto file is essentially the same shape either way — you'd just be swapping a header line.

When I *would* reconsider:

- If/when you add a feature example that genuinely benefits from an edition-only feature (e.g. `features.field_presence`, the new enum defaults, the open/closed enum distinction), put **that one** example on `edition = "2024"`. That doubles as documentation: "here's why editions exist."
- Once edition 2024 has been the protoc default for a year or two and proto3 starts feeling legacy, flip the default.

A middle ground worth considering: leave the examples on proto3, but add a short note in README.md (one or two sentences) saying "IceRPC's Protobuf integration supports proto2, proto3, and editions — pick whichever your project uses." That gets you the "we support modern Protobuf" signal without making every example look unusual.

### DIFF
--- a/examples/protobuf/README.md
+++ b/examples/protobuf/README.md
@@ -10,7 +10,7 @@ you can use proto2, proto3, or editions (e.g. `edition = "2024"`) in your own pr
 | [Deadline](./Deadline/)                | Shows how to use the deadline interceptor to set the invocation deadline.                                                           |
 | [GenericHost](./GenericHost/)          | Shows how to create client and server applications using Microsoft's Dependency Injection container.                                |
 | [Greeter](./Greeter/)                  | Shows how to call and implement a canonical Greeter service using the IceRPC + Protobuf integration.                                |
-| [Logger](./Loggger/)                   | Shows how to enable logging.                                                                                                        |
+| [Logger](./Logger/)                    | Shows how to enable logging.                                                                                                        |
 | [Metrics](./Metrics/)                  | Shows how to use the metrics interceptor and middleware.                                                                            |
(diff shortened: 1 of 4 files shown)


InsertCreativityHere (approved):
Looks good to me!

pepone (approved):

externl (approved):

PR #4684, opened 2026-05-21
Update structure of Slice directories in examples and templates

bernardnormier:
This is the counterpart of #4683 for Ice and Slice.

Note that we are already using this module-derived directory structure in the top-level `slice` directory, with one small exception: IceRpc/Ice/Object.ice.


pepone (approved):

InsertCreativityHere (approved):
Looks good to me!

externl (approved):

Issue #783, opened 2022-02-03
Allow field values to use any encoding


bernardnormier:
We currently receive request/response fields as `IReadOnlyDictionary<int, ReadOnlyMemory<byte>>`:

https://github.com/zeroc-ice/icerpc-csharp/blob/main/src/IceRpc/IncomingFrame.cs#L22

but we send them as `Dictionary<int, EncodeAction>`, where the encoder given to the encode action is using Slice 2.0

https://github.com/zeroc-ice/icerpc-csharp/blob/main/src/IceRpc/OutgoingFrame.cs#L16

There are several issues here:

1. the send sand receive APIs are inconsistent

2. on the send side, we force the application to  use the Slice 2.0 encoding for the field value
It makes sense most of the time but occasionally it may not. See for example: https://github.com/zeroc-ice/icerpc-csharp/blob/main/src/IceRpc/TelemetryInterceptor.cs#L113

3. encoding through an  encode action executed later can be more expensive than providing a pre-encoded buffer immediately and later copying this buffer. Take for example:
https://github.com/zeroc-ice/icerpc-csharp/blob/main/src/IceRpc/Internal/OutgoingFrameExtensions.cs#L24

We're allocating an object (the delegate) for every single call here.

4. some simple fields like idempotent would always have no value. It would be nicer to encode them with a simple:
```
    request.Fields.Add(FieldKey.Idempotent, default); // with idempotent, the value is always empty (default = empty ROM)
```

**Proposal**

replace Fields in both IncomingFrame and OutgoingFrame by `IDictionary<int, ReadOnlyMemory<byte>> Fields { get;  init; } `. The dictionaries are read-write for both. 

And remove OutgoingFrame.FieldDefaults - it's just the initial value of Fields.

PR #4690, opened 2026-05-26
Remove File-Overwriting Check (now done by `slicec`)

InsertCreativityHere:
Now that `slicec` has the logic to detect file-overwrites (see #783), this check in C# is redundant.

### DIFF
--- a/src/ZeroC.Slice.Symbols/Diagnostic.cs
+++ b/src/ZeroC.Slice.Symbols/Diagnostic.cs
@@ -14,14 +14,6 @@ public record struct Diagnostic
     /// <summary>Gets any notes that should be reported along with this diagnostic.</summary>
     internal IList<Compiler.DiagnosticNote> Notes { get; init; }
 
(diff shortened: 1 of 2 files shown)


src/ZeroC.Slice.Symbols/Diagnostic.cs
```diff
@@ -14,14 +14,6 @@ public record struct Diagnostic
     /// <summary>Gets any notes that should be reported along with this diagnostic.</summary>
     internal IList<Compiler.DiagnosticNote> Notes { get; init; }
 
-    /// <summary>Creates a new diagnostic to describe a general error.</summary>
-    public static Diagnostic Error(string message, string? source = null) => new()
```
InsertCreativityHere, line 18: This was only used in the generator-driver.

While it's a useful way to report a general-purpose error, I don't think we should leave it here just in the hope that it's useful one day.

pepone (approved):

PR #4692, opened 2026-05-27
Revert 'Diagnostic' API to Include General Error

InsertCreativityHere:
Also fix `IsError` to be a computed property instead of a method.

### DIFF
--- a/src/ZeroC.Slice.Symbols/Diagnostic.cs
+++ b/src/ZeroC.Slice.Symbols/Diagnostic.cs
@@ -15,8 +15,8 @@ public record struct Diagnostic
     internal IList<Compiler.DiagnosticNote> Notes { get; init; }
 
     /// <summary>Returns whether this diagnostic represents an error.</summary>
-    public bool IsError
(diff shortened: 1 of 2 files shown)


src/ZeroC.Slice.Symbols/Diagnostic.cs
```diff
@@ -14,6 +14,18 @@ public record struct Diagnostic
     /// <summary>Gets any notes that should be reported along with this diagnostic.</summary>
     internal IList<Compiler.DiagnosticNote> Notes { get; init; }
 
+    /// <summary>Returns whether this diagnostic represents an error.</summary>
+    public bool IsError
+        => Kind is not Compiler.DiagnosticKind.Info and not Compiler.DiagnosticKind.Warning;
```
bernardnormier: In our C# code, the `=>` is usually on the first line, as in:
```cs
public bool IsError =>
        ...

```

bernardnormier (approved):

src/ZeroC.Slice.Symbols/Diagnostic.cs
InsertCreativityHere: Fixed.

PR #4695, opened 2026-05-27
Fix protoc logging

bernardnormier:
Fixes the ProtocTask logging to proper log plug-in errors.

Note that protoc plug-ins can only report errors; they can report other diagnostics such as warnings.

### DIFF
--- a/src/IceRpc.Protobuf.Tools/ProtocTask.cs
+++ b/src/IceRpc.Protobuf.Tools/ProtocTask.cs
@@ -145,8 +145,7 @@ public partial class ProtocTask : ToolTask
         Match diagnosticMatch = DiagnosticRegex().Match(singleLine);
         if (diagnosticMatch.Success)
         {
-            // Hand off to the base implementation in canonical MSBuild format so its CanonicalError parser picks up
-            // the file, line, column, severity and message.
+            // Construct a string in the canonical MSBuild diagnostic format to pass to the base implementation.
             base.LogEventsFromTextOutput(
                 $"{diagnosticMatch.Groups["file"].Value}({diagnosticMatch.Groups["line"].Value},{diagnosticMatch.Groups["column"].Value}): {diagnosticMatch.Groups["severity"].Value}: {diagnosticMatch.Groups["message"].Value}",
                 messageImportance);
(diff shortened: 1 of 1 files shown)


src/IceRpc.Protobuf.Tools/ProtocTask.cs
```diff
@@ -133,11 +133,46 @@ protected override string GenerateFullPathToTool()
...
+                file: null,
+                lineNumber: 0,
+                columnNumber: 0,
+                endLineNumber: 0,
+                endColumnNumber: 0,
+                message: pluginMatch.Groups["message"].Value);
+            return;
+        }
+
+        // Fallback: any other protoc output (informational messages, unrecognized diagnostics) is passed through as a
```
bernardnormier, line 171: In practice, this doesn't work very well. You need to build with `-v:n -tl:off` to see it.

If there are other protoc error messages we want to handle properly, we'll need to add another regex Log.LogError.

src/IceRpc.Protobuf.Tools/ProtocTask.cs
```diff
@@ -133,11 +133,46 @@ protected override string GenerateFullPathToTool()
...
-        base.LogEventsFromTextOutput(
-            DiagnosticRegex().Replace(singleLine, "${file}(${line},${column}): ${severity}: ${message}"),
-            messageImportance);
+    protected override void LogEventsFromTextOutput(string singleLine, MessageImportance messageImportance)
+    {
+        Match diagnosticMatch = DiagnosticRegex().Match(singleLine);
+        if (diagnosticMatch.Success)
+        {
+            // Hand off to the base implementation in canonical MSBuild format so its CanonicalError parser picks up
+            // the file, line, column, severity and message.
```
InsertCreativityHere: I don't find this comment super readable, apart from saying that we're constructing a "canonical MSBuild format" string. Maybe something simpler like?:
```suggestion
            // Construct a string in the canonical MSBuild diagnostic format to pass to the base implementation.
```

InsertCreativityHere (approved):
Looks good to me!

pepone (approved):

PR #4697, opened 2026-05-28
Improve the Diagnostic Reporting API

InsertCreativityHere:
Makes `Diagnostic` a `readonly record struct`, adds a private constructor to reduce verbosity, and a new `ToCompilerDiagnostic` method, which lets us make all the fields `private` and `readonly`.

### DIFF
--- a/src/ZeroC.Slice.Symbols/Diagnostic.cs
+++ b/src/ZeroC.Slice.Symbols/Diagnostic.cs
@@ -47,7 +47,7 @@ public readonly record struct Diagnostic
 
     /// <summary>Creates a new diagnostic to describe an attribute with an incorrect number of arguments.</summary>
     /// <param name="directive">The directive of the attribute that had an incorrect number of arguments.</param>
-    /// <param name="expected">The correct number of arguments that should of been supplied to this attribute.</param>
+    /// <param name="expected">The correct number of arguments that should have been supplied to this attribute.</param>
     /// <param name="actual">The actual number of arguments that were supplied to this attribute.</param>
     /// <param name="source">The source location of the attribute which with an incorrect number of arguments.</param>
     /// <returns>A new diagnostic.</returns>
--- a/src/ZeroC.Slice.Symbols/Diagnostic.cs
+++ b/src/ZeroC.Slice.Symbols/Diagnostic.cs
@@ -49,7 +49,7 @@ public readonly record struct Diagnostic
     /// <param name="directive">The directive of the attribute that had an incorrect number of arguments.</param>
     /// <param name="expected">The correct number of arguments that should have been supplied to this attribute.</param>
     /// <param name="actual">The actual number of arguments that were supplied to this attribute.</param>
-    /// <param name="source">The source location of the attribute which with an incorrect number of arguments.</param>
+    /// <param name="source">The source location of the attribute with an incorrect number of arguments.</param>
     /// <returns>A new diagnostic.</returns>
     public static Diagnostic IncorrectAttributeArgumentCount(string directive, int expected, int actual, string source)
     {
--- a/src/ZeroC.Slice.Symbols/Diagnostic.cs
+++ b/src/ZeroC.Slice.Symbols/Diagnostic.cs
@@ -6,41 +6,51 @@ namespace ZeroC.Slice.Symbols;
 public readonly record struct Diagnostic
 {
     /// <summary>Returns whether this diagnostic represents an error.</summary>
-    public bool IsError => Kind is not Compiler.DiagnosticKind.Info and not Compiler.DiagnosticKind.Warning;
+    public bool IsError => _kind is not Compiler.DiagnosticKind.Info and not Compiler.DiagnosticKind.Warning;
 
-    /// <summary>Gets the exact kind of diagnostic.</summary>
-    private Compiler.DiagnosticKind Kind { get; }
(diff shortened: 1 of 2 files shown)


src/ZeroC.Slice.Symbols/Diagnostic.cs
```diff
@@ -3,67 +3,55 @@
 namespace ZeroC.Slice.Symbols;
 
 /// <summary>Represents a diagnostic message produced during code generation.</summary>
-public record struct Diagnostic
+public readonly record struct Diagnostic
 {
+    /// <summary>Returns whether this diagnostic represents an error.</summary>
+    public bool IsError => Kind is not Compiler.DiagnosticKind.Info and not Compiler.DiagnosticKind.Warning;
```
InsertCreativityHere: Just moved to the top. `public` properties should come before `private` properties I imagine.

src/ZeroC.Slice.Symbols/Diagnostic.cs
```diff
@@ -3,67 +3,55 @@
...
+
     /// <summary>Gets the exact kind of diagnostic.</summary>
-    internal readonly Compiler.DiagnosticKind Kind { get; init; }
+    private Compiler.DiagnosticKind Kind { get; }
```
bernardnormier: I would use private readonly fields instead:
```cs
private readonly Compiler.DiagnosticKind _kind;
```

same for the remaining private properties.

src/ZeroC.Slice.Symbols/Diagnostic.cs
```diff
@@ -3,67 +3,55 @@
...
+    private string? Source { get; }
 
     /// <summary>Gets any notes that should be reported along with this diagnostic.</summary>
-    internal IList<Compiler.DiagnosticNote> Notes { get; init; }
+    private IList<Compiler.DiagnosticNote> Notes { get; }
 
-    /// <summary>Returns whether this diagnostic represents an error.</summary>
-    public bool IsError =>
-        Kind is not Compiler.DiagnosticKind.Info and not Compiler.DiagnosticKind.Warning;
+    private Diagnostic(Compiler.DiagnosticKind kind, string? source)
```
bernardnormier: private methods, including constructors, go after other methods, including static methods.

src/ZeroC.Slice.Symbols/Diagnostic.cs
bernardnormier: It's unclear why you don't have a `using Compiler` in this file.

bernardnormier: I don't find the logic sound here.

`Kind` is a value we set locally. It should (obviously) be limited to values we know about.

If we keep this Compiler.DiagnosticKind field, IsError means kind is Error.

src/ZeroC.Slice.Symbols/Diagnostic.cs
```diff
@@ -3,67 +3,55 @@
...
+    {
+        Kind = kind;
+        Source = source;
+        Notes = [];
+    }
 
     /// <summary>Creates a new diagnostic to describe a general error.</summary>
```
bernardnormier, line 20: The doc-comment is missing param and return elements.

src/ZeroC.Slice.Symbols/Diagnostic.cs
InsertCreativityHere: > It's unclear why you don't have a using Compiler in this file.

To avoid the massive ambiguity between the many types in `Compiler` and `Symbols` which have the exact same name. Even this struct, `Diagnostic` is a type in both `Compiler` and `ZeroC.Slice.Symbols`.

----

> Kind is a value we set locally. It should (obviously) be limited to values we know about.
> If we keep this Compiler.DiagnosticKind field, IsError means kind is Error.

I don't understand this comment, `IsError` is limited to error kinds.
Maybe what's being missed is that `Error` is just a general error message, but the more specific attribute errors (like `InvalidAttribute`) are also 'errors' which should cause us to not generate code.
And it's easier to just filter out warning/info, than the list out `Error` + all the more-specific attribute error kinds.

src/ZeroC.Slice.Symbols/Diagnostic.cs
InsertCreativityHere: I added some.

But I was basing my style off of the other code in the Symbol/Generator namespaces, where most of the time, we don't include `param` and `return` for simple functions. If it's a bug here, then many of the functions in these namespaces also need to be updated to include more tags.

src/ZeroC.Slice.Symbols/Diagnostic.cs
InsertCreativityHere: Moved the constructor to the bottom (it was the only private method).

For the record, my original choice was based off the [official StyleCop recommendation](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md), where 'constructors' aren't treated equal to other 'methods'.

src/ZeroC.Slice.Symbols/Diagnostic.cs
InsertCreativityHere: Removed the properties and replaced them with fields.

bernardnormier (approved):

PR #4706, opened 2026-05-28
Add RouterExtensions.Map overload for service class

bernardnormier:
Most of the time, a service class implements a single Slice interface or Protobuf service. Implementing multiple services is rare; furthermore, interface inheritance in Slice is not very common, and not available at all in Protobuf.

In this context, it's unfortunate to require the user to specify the generated IxxxService interface when mapping a service to its default path, as in:
```cs
// the default service path is retrieve from the interface; we require an explicit interface type param
// for the uncommon case where chatbot implements multiple interfaces
router.Map<IGreeterService>(chatbot);
```

> [!NOTE]
> In the code above, chatbot is an IDispatcher, while IGreeterService is a generated Ice/Slice/Protobuf interface that doesn't implement IDispatcher. For the dispatch to work later, chatbot must implement IGreeterService as well, but  it's not a compile-time requirement - chatbot just needs to be an IDispatcher.

The solution implemented by this PR is to add a new non-generic Map extension method:

```cs
  public static Router Map(this Router router, IDispatcher service)
```

You use this overload as follows:
```cs
router.Map(chatbot);
```

and it works fine as long as chatbot implements exactly one interface with an DefaultServicePatchAttribute, which is very common. This is based on the runtime type of chatbot, so it works even if chatbot is passed as a plain IDispatcher.

 If there is an ambiguity, you get a runtime exception. And in this case, you should either:
 - specify the path you want with a string
 - call `Map<ISomeService>(chatbot)` to specify which interface provides the default service path


src/IceRpc/RouterExtensions.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) ZeroC, Inc.
 
 using IceRpc.Features;
+using System.Reflection;
```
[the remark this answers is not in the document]
bernardnormier: You're mistaken: it's required by the `<exception cref="AmbiguousMatchException">`

src/IceRpc/RouterExtensions.cs
```diff
@@ -9,13 +10,33 @@ public static class RouterExtensions
...
     /// <returns>The router being configured.</returns>
     /// <exception cref="InvalidOperationException">Thrown if <see cref="IDispatcher.DispatchAsync" /> was already
     /// called on this router.</exception>
+    /// <exception cref="ArgumentException">Thrown if the interface(s) implemented by <typeparamref name="TService"/>
+    /// do not have a <see cref="DefaultServicePathAttribute"/> attribute.</exception>
+    /// <exception cref="AmbiguousMatchException">Thrown if <typeparamref name="TService"/> is a class that
+    /// implements multiple interfaces with a <see cref="DefaultServicePathAttribute"/> attribute.</exception>
+    public static Router Map<TService>(this Router router, TService service)
+        where TService : class, IDispatcher =>
+        router.Map(typeof(TService).GetDefaultServicePath(), service);
```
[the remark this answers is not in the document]
bernardnormier: Fixed

src/IceRpc/TypeExtensions.cs
```diff
@@ -1,35 +1,81 @@
...
     /// </summary>
-    /// <param name="type">The type of an interface.</param>
+    /// <param name="type">The interface with the <see cref="DefaultServicePathAttribute"/> attribute, or a class
+    /// that implements such an interface.</param>
     /// <returns>The default service path.</returns>
-    /// <exception cref="ArgumentException">Thrown if <paramref name="type" /> is not an interface, or if
-    /// it does not have a <see cref="DefaultServicePathAttribute" /> attribute.</exception>
+    /// <exception cref="ArgumentException">Thrown if <paramref name="type" /> is neither a class nor an interface, or
+    /// if it does not have a <see cref="DefaultServicePathAttribute" /> attribute.</exception>
+    /// <exception cref="AmbiguousMatchException">Thrown if <paramref name="type" /> is a class that implements multiple
```
pepone: I don't see the point of using a different exception for this case, I would use ArgumentException for both with an appropriate message

pepone (approved):

src/IceRpc/RouterExtensions.cs
```diff
@@ -9,13 +10,29 @@ public static class RouterExtensions
...
-    /// <typeparam name = "TService" > An interface with the DefaultServicePathAttribute attribute. The path of the
-    /// mapped service corresponds to the value of this attribute.</typeparam>
     /// <param name="router">The router being configured.</param>
     /// <param name="service">The target service of this route.</param>
     /// <returns>The router being configured.</returns>
     /// <exception cref="InvalidOperationException">Thrown if <see cref="IDispatcher.DispatchAsync" /> was already
     /// called on this router.</exception>
+    /// <exception cref="ArgumentException">Thrown if the interface(s) implemented by <paramref name="service" /> do not
+    /// have a <see cref="DefaultServicePathAttribute"/> attribute.</exception>
+    /// <exception cref="AmbiguousMatchException">Thrown if <paramref name="service" /> implements multiple interfaces
```
externl: C# has an exception for everything!

externl (approved):

src/IceRpc/TypeExtensions.cs
bernardnormier: Ok, will fix.

PR #4708, opened 2026-05-29
Reject unknown options from plugin/generator

bernardnormier:


src/ZeroC.Slice.Generator/GeneratorDriver.cs
```diff
@@ -46,6 +46,11 @@ internal static async Task RunAsync(
                 ?? assembly.GetName().Version?.ToString()
                 ?? "unknown";
 
+            if (options.Length > 0)
+            {
+                diagnostics.Add(Diagnostic.Error("Generator does not support any options"));
```
bernardnormier: slicec does not currently specify from which generator an error comes from; as a result this error message is ambiguous. The fix belongs to slicec, not here.

pepone (approved):

src/ZeroC.Slice.Generator/GeneratorDriver.cs
```diff
@@ -46,6 +46,11 @@ internal static async Task RunAsync(
                 ?? assembly.GetName().Version?.ToString()
                 ?? "unknown";
 
+foreach ((string key, string value) in options)
+{
+    diagnostics.Add(Diagnostic.Error($"Unknown option: '{key}'"));
```
InsertCreativityHere: The error messages `slicec` uses are lower-case non-sentences.
```suggestion
    diagnostics.Add(Diagnostic.Error($"unknown option: '{key}'"));
```

InsertCreativityHere (approved):
Looks good to me!

src/ZeroC.Slice.Generator/GeneratorDriver.cs
InsertCreativityHere: Also this entire block is completely dedented? Thought it was just the view at first, but even checking the raw file, it's missing indentation.

PR #4713, opened 2026-06-03
Code coverage cleanup

### LINKED ISSUE #4651: Coverage output when running dotnet test
dotnet test always produces and shows coverage reports, when it shouldn't:

```
Test summary: total: 2293, failed: 0, succeeded: 2293, skipped: 0, duration: 44.8s
Build succeeded in 51.6s

Attachments:
  C:\Users\bernard\source\repos\icerpc-csharp\tests\IceRpc.Extensions.DependencyInjection.Tests\TestResults\8c2dae0f-7339-4c69-bbab-eb93d0dcb180\coverage.cobertura.xml
[...]
```

bernardnormier:
Fixes #4651.

### DIFF
--- a/build/Tests.runsettings
+++ b/build/Tests.runsettings
@@ -7,12 +7,14 @@
     </NUnit>
     <DataCollectionRunSettings>
         <DataCollectors>
-            <DataCollector friendlyName="XPlat code coverage">
+            <!-- Disabled by default; the coverage CI job opts in via `collect:"XPlat Code Coverage"`,
+                 which re-enables this collector while preserving the configuration below. -->
+            <DataCollector friendlyName="XPlat code coverage" enabled="False">
                 <Configuration>
                     <!-- Exclude test assemblies and the IceRpc.ServiceGenerator source generator (loaded as an
                          analyzer into the test process; its behavior is exercised through generated code, not by
                          tracking which lines of the generator itself execute) -->
-                    <Exclude>[*]IceRpc.Conformance.Tests*,[*]IceRpc.Tests.Common*,[IceRpc.ServiceGenerator]*</Exclude>
+                    <Exclude>[IceRpc.Conformance.Tests]*,[IceRpc.Tests.Common]*,[IceRpc.ServiceGenerator]*</Exclude>
                     <!-- Filter out generated files -->
                     <ExcludeByFile>**/generated/**/*.cs,**/LoggerMessage.g.cs</ExcludeByFile>
                 </Configuration>
--- a/build/Tests.runsettings
+++ b/build/Tests.runsettings
@@ -7,8 +7,6 @@
     </NUnit>
     <DataCollectionRunSettings>
         <DataCollectors>
-            <!-- Disabled by default; the coverage CI job opts in via `collect:"XPlat Code Coverage"`,
-                 which re-enables this collector while preserving the configuration below. -->
             <DataCollector friendlyName="XPlat code coverage" enabled="False">
                 <Configuration>
                     <!-- Exclude test assemblies and the IceRpc.ServiceGenerator source generator (loaded as an


build/Tests.runsettings
```diff
@@ -7,12 +7,14 @@
...
-            <DataCollector friendlyName="XPlat code coverage">
+            <!-- Disabled by default; the coverage CI job opts in via `collect:"XPlat Code Coverage"`,
+                 which re-enables this collector while preserving the configuration below. -->
+            <DataCollector friendlyName="XPlat code coverage" enabled="False">
                 <Configuration>
                     <!-- Exclude test assemblies and the IceRpc.ServiceGenerator source generator (loaded as an
                          analyzer into the test process; its behavior is exercised through generated code, not by
                          tracking which lines of the generator itself execute) -->
-                    <Exclude>[*]IceRpc.Conformance.Tests*,[*]IceRpc.Tests.Common*,[IceRpc.ServiceGenerator]*</Exclude>
+                    <Exclude>[IceRpc.Conformance.Tests]*,[IceRpc.Tests.Common]*,[IceRpc.ServiceGenerator]*</Exclude>
```
bernardnormier, line 15: That's the Claude suggestion from my previous PR.

build/Tests.runsettings
```diff
@@ -7,12 +7,14 @@
     </NUnit>
     <DataCollectionRunSettings>
         <DataCollectors>
-            <DataCollector friendlyName="XPlat code coverage">
+            <!-- Disabled by default; the coverage CI job opts in via `collect:"XPlat Code Coverage"`,
+                 which re-enables this collector while preserving the configuration below. -->
+            <DataCollector friendlyName="XPlat code coverage" enabled="False">
```
bernardnormier, line 10: enabled="False" is the fix for #4651.

externl, line 10: Googling suggests that false is the default. If true we shouldn't extra disable it.

build/Tests.runsettings
```diff
@@ -7,12 +7,14 @@
     </NUnit>
     <DataCollectionRunSettings>
         <DataCollectors>
-            <DataCollector friendlyName="XPlat code coverage">
+            <!-- Disabled by default; the coverage CI job opts in via `collect:"XPlat Code Coverage"`,
+                 which re-enables this collector while preserving the configuration below. -->
```
externl: I would not include this. It's just a detail and easily prone to breakage.

build/Tests.runsettings
bernardnormier: Here is where this comes from:

IceRpc.Tests.props sets `<RunSettingsFilePath>` to Tests.runsettings, so every test project automatically picks up that file on every `dotnet test`. And Tests.runsettings unconditionally declares an `XPlat code coverage` `<DataCollector>` — the mere presence of a `<DataCollector>` element enables it. There's no `enabled="False"` and no opt-in switch, so coverage is collected every time, producing those `coverage.cobertura.xml` attachments.

The CI workflow coverage.yml passes `--collect:"XPlat Code Coverage"` explicitly, but because the runsettings already enables it, the flag is redundant — coverage was being collected for every developer too.

### Fix options

1. **Split the runsettings** (cleanest): keep Tests.runsettings with just the NUnit block, move the `<DataCollectionRunSettings>` block into a new `Coverage.runsettings`, and change the coverage workflow to `dotnet test --settings build/Coverage.runsettings --collect:"XPlat Code Coverage"`.

2. **Inline-disable**: add `enabled="False"` to the `<DataCollector>` element. Then `--collect:"XPlat Code Coverage"` in CI re-enables it while preserving the `<Exclude>` / `<ExcludeByFile>` config. Quick, but a bit subtle.

externl (approved):

build/Tests.runsettings
pepone: How do you enable it?

pepone (approved):

build/Tests.runsettings
bernardnormier: It's enabled through `--collect:"XPlat Code Coverage"` as documented in BUILDING.md. That's also what CI does.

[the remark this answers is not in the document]
bernardnormier: Removed comment.

build/Tests.runsettings
bernardnormier: Removed comment.

PR #4716, opened 2026-06-03
Update Microsoft.CodeAnalysis.CSharp dependency

bernardnormier:
This PR updates (reduces) the Microsoft.CodeAnalysis.CSharp to 5.0.0 for compatibility with dotnet 10.0.10x.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -22,6 +22,8 @@
          See: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3187 -->
     <PackageVersion Include="StyleCop.Analyzers" Version="1.2.0-beta.556" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="5.3.0" />
+
+    <!-- 5.0.0 to maintain compatibility with .NET SDK 10.0.10x -->
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="5.0.0" />
 
     <!-- MSBuild SDK packages (used by IceRpc.*.Tools) -->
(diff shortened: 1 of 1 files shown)


Directory.Packages.props
```diff
@@ -22,7 +22,7 @@
          See: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3187 -->
     <PackageVersion Include="StyleCop.Analyzers" Version="1.2.0-beta.556" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="5.3.0" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="5.3.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="5.0.0" />
```
pepone, line 27: We should add a comment here regarding the downgrade, the analyzers should probably use the same version.  Google suggests something like:

```
<ItemGroup>
  <!-- The actual compiler API (Hard locked to Ubuntu's .NET 10 base version) -->
  <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="5.0.0" PrivateAssets="all" />

  <!-- Rules for writing analyzers (Kept at 5.0.0 to stay compatible with the host) -->
  <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="5.0.0" PrivateAssets="all" 
                    DevelopmentDependency="true" />
</ItemGroup>
```

Would be nice to also test this Ubuntu version in CI, we run CI on Ubuntu but never noticed this because we install the latest .NET 10.x SDK

externl, line 27: We also need to add an exclusion to `.github/dependabot.yml`

bernardnormier: We don't need to change `Microsoft.CodeAnalysis.Analyzers`, only `Microsoft.CodeAnalysis.CSharp`.

externl: Actually let's not update dependabot. We can merge this for the 0.6.x release branch. Then let dependabot update it again next week so that main uses the latest.

bernardnormier: BTW changing the version for `Microsoft.CodeAnalysis.Analyzers` breaks the build.

pepone (approved):

externl (approved):

PR #4718, opened 2026-06-04
Update version in generated code

bernardnormier:


src/ZeroC.Slice.Generator/GeneratorDriver.cs
```diff
@@ -35,16 +35,9 @@ internal static async Task RunAsync(
...
-            // the <Version> MSBuild property. Fall back to the assembly version (e.g., "0.6.0.0") if
-            // the informational version attribute is not present.
             Assembly assembly = Assembly.GetEntryAssembly()!;
             string generatorName = assembly.GetName().Name ?? "unknown";
-            string version =
-                CustomAttributeExtensions.GetCustomAttribute<AssemblyInformationalVersionAttribute>(
-                    assembly)?.InformationalVersion
-                ?? assembly.GetName().Version?.ToString()
-                ?? "unknown";
+            string version = assembly.GetName().Version?.ToString(3) ?? "unknown";
```
bernardnormier, line 40: the 3-fields limit is what we already use for Protobuf generated code

externl (approved):

pepone (approved):

src/ZeroC.Slice.Generator/GeneratorDriver.cs
pepone: The downside of this PR is we no longer know if the generated code was build from the release version. The informational version is based on the version used by the MSBuild build which will include the preview modifier etc.

bernardnormier: I don't think this is critical.

PR #4724, opened 2026-06-04
Update version on main to 0.7.0-preview.1

bernardnormier:


.lychee.toml
```diff
@@ -29,7 +29,7 @@ exclude = [
     # GitHub edit/blame links (may require auth)
     "github\\.com/.*/edit/",
     "github\\.com/.*/blame/",
-    # TODO: remove once 0.6 packages are published.
+    # TODO: remove once 0.7 packages are published.
```
externl, line 32: I think technically this should still be 0.6, but it doesn't matter so long as we remember to actually do this.

externl (approved):

pepone (approved):

PR #4741, opened 2026-06-09
Stop using floating versions for Microsoft.Extensions packages

bernardnormier:
It's unclear why we are using floating versions.  This PR switches to `10.0.0` instead of `10.0.*`.

### DIFF
--- a/docfx/examples/Directory.Packages.props
+++ b/docfx/examples/Directory.Packages.props
@@ -2,9 +2,6 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
-    <!-- Allow floating versions (e.g. 10.0.*) in <PackageVersion>. NuGet rejects floating versions in central package
-         lists by default. -->
-    <CentralPackageFloatingVersionsEnabled>true</CentralPackageFloatingVersionsEnabled>
     <!-- Pin transitive dependencies to the versions listed below (when listed). Lets us force-upgrade a vulnerable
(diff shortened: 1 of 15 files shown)


externl (approved):
Looks good. Dependabot will update these to their latest versions on its next update.

docfx/examples/Directory.Packages.props
```diff
@@ -26,8 +23,8 @@
     <PackageVersion Include="IceRpc.Slice" Version="$(IceRpcVersion)" />
     <PackageVersion Include="IceRpc.Slice.Tools" Version="$(IceRpcVersion)" />
     <PackageVersion Include="IceRpc.Telemetry" Version="$(IceRpcVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Hosting" Version="10.0.*" />
-    <PackageVersion Include="Microsoft.Extensions.Logging" Version="10.0.*" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="10.0.*" />
+    <PackageVersion Include="Microsoft.Extensions.Hosting" Version="10.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="10.0.0" />
```
pepone, line 27: Note the 0.6.0 package we ship requires 10.0.8, we can probably upgrade this versions in a follow up PR or let dependanbot take care as @externl  mentioned.

https://www.nuget.org/packages/IceRpc#dependencies-body-tab

pepone (approved):

PR #4742, opened 2026-06-09
Better slice size validation during decoding of an Ice class/exception

bernardnormier:
This PR improves the validation of slice sizes during the decoding of Ice classes and exceptions.

### DIFF
--- a/src/IceRpc/Ice/Codec/IceDecoder.Class.cs
+++ b/src/IceRpc/Ice/Codec/IceDecoder.Class.cs
@@ -406,7 +406,7 @@ public ref partial struct IceDecoder
     /// SkipIndirectionTable itself.</summary>
     private void SkipIndirectionTable()
     {
-        // We never skip an exception's indirection table
+        // We never skip an exception's indirection table since we don't preserve exception slices.
         Debug.Assert(_classContext.Current.InstanceType == InstanceType.Class);
 
         int tableSize = DecodeSize();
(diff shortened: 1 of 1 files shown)


InsertCreativityHere (approved):
Looks good to me!

src/IceRpc/Ice/Codec/IceDecoder.Class.cs
```diff
@@ -342,14 +343,18 @@ private IceClass DecodeInstance(int index)
     }
 
     /// <summary>Decodes the size of the current slice.</summary>
-    /// <returns>The slice of the current slice, not including the size length.</returns>
-    private int DecodeSliceSize()
+    /// <param name="sliceHasTaggedFields">Indicates whether the slice has tagged fields.</param>
+    /// <returns>The size of the current slice, not including the size length.</returns>
```
pepone, line 347: Is this comment accurate?

> The byte count includes the four bytes occupied by the count itself, so an empty slice has a byte count of four and no data.

https://docs.zeroc.com/ice/3.8/cpp/basic-data-encoding#BasicDataEncoding-SliceSize

pepone (approved):

src/IceRpc/Ice/Codec/IceDecoder.Class.cs
bernardnormier: Yes. Note that this helper function removes "4" from the size before returning it.

PR #4747, opened 2026-06-10
Fix bug in ServiceAddress.ToString

### LINKED ISSUE #4744: [Audit-Medium] ServiceAddress.ToString emits a second '?' — malformed URI and silent round-trip corruption

## Summary

After appending the main server address, `ToString` computes `firstOption = serverAddress.Params.Count == 0`, but `AppendServerAddress` already emitted `?transport=<name>` whenever `Transport` is non-null — independently of `Params`. With `Transport` set, empty `Params`, at least one alt server address, and `OriginalUri == null` (any programmatically built or `with`-modified address), the output is malformed:

```csharp
var sa = new ServiceAddress(Protocol.IceRpc)
{
    ServerAddress = new ServerAddress(new Uri("icerpc://host1?transport=tcp")),
    AltServerAddresses = [new ServerAddress(new Uri("icerpc://host2"))]
};
sa.ToString(); // icerpc://host1/?transport=tcp?alt-server=host2
```

Re-parsing splits query options only on `&`, so `Transport` becomes `"tcp?alt-server=host2"` (the `?` passes `IsValidParamValue`) and the alt servers are silently lost. The malformed string also flows into `ToUri()`. The two adjacent combinations (Transport+Params, no-Transport+Params) are correct; this is precisely the one broken case.

## Source references

- [src/IceRpc/ServiceAddress.cs#L402](https://github.com/icerpc/icerpc-csharp/blob/ed5f6b8843d95e96cbc0251329bf8301eb1e520b/src/IceRpc/ServiceAddress.cs#L402) — `firstOption` ignores `Transport`
- [src/IceRpc/ServiceAddress.cs#L444-L449](https://github.com/icerpc/icerpc-csharp/blob/ed5f6b8843d95e96cbc0251329bf8301eb1e520b/src/IceRpc/ServiceAddress.cs#L444-L449) — second `?` appended
- [src/IceRpc/Internal/ServerAddressExtensions.cs#L55-L60](https://github.com/icerpc/icerpc-csharp/blob/ed5f6b8843d95e96cbc0251329bf8301eb1e520b/src/IceRpc/Internal/ServerAddressExtensions.cs#L55-L60) — `?transport=` emitted independently of `Params`
- [src/IceRpc/Internal/UriExtensions.cs#L26](https://github.com/icerpc/icerpc-csharp/blob/ed5f6b8843d95e96cbc0251329bf8301eb1e520b/src/IceRpc/Internal/UriExtensions.cs#L26) — query parsing splits only on `&`

## Recommendation

`firstOption = serverAddress.Params.Count == 0 && serverAddress.Transport is null;`

## Audit context

Severity: **Medium** (malformed URI + silent data loss on round-trip; one-line fix).

bernardnormier:
Fixes #4744 as suggested. Also added test.

### DIFF
--- a/src/IceRpc/ServiceAddress.cs
+++ b/src/IceRpc/ServiceAddress.cs
@@ -399,7 +399,7 @@ public sealed record class ServiceAddress
         if (ServerAddress is ServerAddress serverAddress)
         {
             sb.AppendServerAddress(serverAddress, Path);
-            firstOption = serverAddress.Params.Count == 0;
+            firstOption = serverAddress.Params.Count == 0 && serverAddress.Transport is null;
         }
         else
         {
--- a/tests/IceRpc.Tests/ServiceAddressTests.cs
+++ b/tests/IceRpc.Tests/ServiceAddressTests.cs
@@ -143,6 +143,12 @@ public class ServiceAddressTests
         get
         {
             // Service address with alt servers
+            var serviceAddressWithTransport = new ServiceAddress(Protocol.IceRpc)
+            {
+                ServerAddress = new ServerAddress(new Uri("icerpc://host1?transport=tcp")),
+                AltServerAddresses = [new ServerAddress(new Uri("icerpc://host2"))]
+            };
+
             var serviceAddressWithAltServerAddresses = new ServiceAddress(
                 new Uri("ice://localhost:8080/foo?abc=123#bar"));
             serviceAddressWithAltServerAddresses = serviceAddressWithAltServerAddresses with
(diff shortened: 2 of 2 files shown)


src/IceRpc/ServiceAddress.cs
```diff
@@ -399,7 +399,7 @@ public override string ToString()
         if (ServerAddress is ServerAddress serverAddress)
         {
             sb.AppendServerAddress(serverAddress, Path);
-            firstOption = serverAddress.Params.Count == 0;
+            firstOption = serverAddress.Params.Count == 0 && serverAddress.Transport is null;
```
pepone, line 402: Did you consider adding `out bool firstOption` to AppendServerAddress instead?

pepone (approved):
Looks good

src/IceRpc/ServiceAddress.cs
bernardnormier: No, but I just looked into it.

I don't find this super convenient: we call AppendServerAddress 3 times and only need this firstOption once. AppendServerServer has also lots of parameters (with default values) - don't really want to add an extra out param.

PR #4756, opened 2026-06-11
Reject empty Slice segment in the middle of an element stream

### LINKED ISSUE #4755: [Audit-Low] Zero-size Slice segment mid-stream silently ends an element stream

## Summary

`IsCompleteSegment` deliberately returns an empty buffer with `isCompleted: false` for a wire segment whose size prefix is `0` when more data follows. That result flows through `ReadSliceSegmentAsync` to `AsyncStream<T>.EnumerateAsync`, whose `if (readResult.Buffer.IsEmpty) { Debug.Assert(readResult.IsCompleted); yield break; }` assumes empty ⇒ completed:

- debug builds — a `Debug.Assert` fires on peer-controlled data;
- release builds — the element stream silently ends early and all remaining elements are discarded (the reader is completed in the `finally`).

The C# encoder never emits zero-size segments, but they are exercised as legal by the repository's own `IceRpc.Slice.Tests/PipeReaderTests`, so a different-language (or buggy) peer that emits one gets silent data loss instead of an `InvalidDataException`. Either "skip" or "reject" would be correct; "treat as end of stream" is not.

## Source references

- [src/IceRpc/Internal/PipeReaderExtensions.cs#L182-L192](https://github.com/icerpc/icerpc-csharp/blob/ed5f6b8843d95e96cbc0251329bf8301eb1e520b/src/IceRpc/Internal/PipeReaderExtensions.cs#L182-L192) — `IsCompleteSegment` zero-size path
- [src/IceRpc.Slice/Operations/Internal/AsyncStream.cs#L111-L115](https://github.com/icerpc/icerpc-csharp/blob/ed5f6b8843d95e96cbc0251329bf8301eb1e520b/src/IceRpc.Slice/Operations/Internal/AsyncStream.cs#L111-L115) — empty buffer treated as end-of-stream

## Recommendation

In `AsyncStream.EnumerateAsync`, distinguish "empty + completed" (end of stream) from "empty + not completed" (skip, or throw `InvalidDataException`).

## Audit context

Finding 13 of [general-correctness-audit-2026-06-10.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-06-10.md); confirmed by Codex ([cross-review](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-codex-review-2026-06-10.md)).

Severity: **Low** (protocol-validation hole; requires a non-C# peer).

pepone:
Fixes #4755.

`AsyncStream<T>.EnumerateAsync` treated any empty buffer as the end of the stream. A zero-size Slice segment followed by more data also produces an empty buffer with `IsCompleted` false (see `PipeReaderExtensions.IsCompleteSegment`): a `Debug.Assert` fired in debug builds, and in release builds the element stream ended silently, dropping the remaining elements.

The only expected empty segment is the one that ends the stream. `EnumerateAsync` now throws `InvalidDataException` when it reads an empty segment in the middle of an element stream.

Changes:
- `AsyncStream.EnumerateAsync`: distinguish "empty + completed" (end of stream) from "empty + not completed" (protocol violation, `InvalidDataException`).
- New test `Zero_size_segment_in_the_middle_of_the_stream_throws_InvalidDataException`: written first against the unfixed code, where it failed with `Debug.Fail`; also checks that elements decoded before the invalid segment are still yielded and that the reader is completed exactly once.
- New test `Zero_size_segment_at_the_end_of_the_stream_completes_iteration`: pins the legal case, a trailing zero-size segment ends the stream normally.

The C# encoder (`AsyncEnumerablePipeReader`) never emits zero-size segments, so this only affects streams received from a peer that does.

### DIFF
--- a/tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
+++ b/tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
@@ -62,8 +62,7 @@ public class AsyncStreamTests
     public void Zero_size_segment_in_the_middle_of_the_stream_throws_InvalidDataException()
     {
         // A zero-size segment is never valid: the end of a stream is signaled by the reader completing with no
-        // bytes, not by a zero-size segment. See https://github.com/icerpc/icerpc-csharp/issues/4755. We use a
-        // variable-size element (string) because only variable-size element streams are encoded with segments.
+        // bytes, not by a zero-size segment.
         byte[] payload = EncodeStringSegments(["a", "b"], [], ["c"]);
         var trackingReader = new TrackingPipeReader(PipeReader.Create(new ReadOnlySequence<byte>(payload)));
         using IAsyncStream<string> stream = trackingReader.ToAsyncStream(
--- a/tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
+++ b/tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
@@ -59,6 +59,59 @@ public class AsyncStreamTests
     }
 
     [Test]
+    public void Zero_size_segment_in_the_middle_of_the_stream_throws_InvalidDataException()
+    {
+        // A zero-size segment is never valid: the end of a stream is signaled by the reader completing with no
+        // bytes, not by a zero-size segment. See https://github.com/icerpc/icerpc-csharp/issues/4755.
+        byte[] payload = EncodeInt32Segments([1, 2], [], [3]);
+        var trackingReader = new TrackingPipeReader(PipeReader.Create(new ReadOnlySequence<byte>(payload)));
+        using IAsyncStream<int> stream = trackingReader.ToAsyncStream(
+            (ref SliceDecoder decoder) => decoder.DecodeInt32());
+
+        var items = new List<int>();
+
+        Assert.That(
+            async () =>
+            {
+                await foreach (int item in stream)
+                {
+                    items.Add(item);
+                }
+            },
+            Throws.TypeOf<InvalidDataException>());
+
+        Assert.That(items, Is.EqualTo(new[] { 1, 2 }));
+        Assert.That(trackingReader.CompleteCallCount, Is.EqualTo(1));
+    }
+
+    [Test]
+    public void Zero_size_segment_at_the_end_of_the_stream_throws_InvalidDataException()
+    {
+        // A zero-size segment is never valid, including at the end of the stream: the end of a stream is signaled
+        // by the reader completing with no bytes, not by a zero-size segment.
+        // See https://github.com/icerpc/icerpc-csharp/issues/4755.
+        byte[] payload = EncodeInt32Segments([1, 2, 3], []);
+        var trackingReader = new TrackingPipeReader(PipeReader.Create(new ReadOnlySequence<byte>(payload)));
+        using IAsyncStream<int> stream = trackingReader.ToAsyncStream(
+            (ref SliceDecoder decoder) => decoder.DecodeInt32());
+
+        var items = new List<int>();
+
+        Assert.That(
+            async () =>
+            {
+                await foreach (int item in stream)
+                {
+                    items.Add(item);
+                }
+            },
+            Throws.TypeOf<InvalidDataException>());
+
+        Assert.That(items, Is.EqualTo(new[] { 1, 2, 3 }));
+        Assert.That(trackingReader.CompleteCallCount, Is.EqualTo(1));
+    }
+
+    [Test]
     public async Task Break_during_iteration_completes_reader()
     {
         byte[] payload = EncodeInt32Values(1, 2, 3);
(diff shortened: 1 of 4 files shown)


bernardnormier:
I think we should actually reject 0 segment sizes in general, in `IsCompleteSegment`.

A Slice segment with an encoded 0-size is not correct:
 - if it corresponds to non-streamed args or return value, it must be at least 1 (for the tag end marker)
 - if it corresponds to streamed elements, it must also be at least 1 (the size is the number of bytes in the segment)

The "0" case is the end of stream; we don't encode a size 0.

Note that we currently have various tests that encode a 0 Slice segment. They'll need to be updated.

tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
```diff
@@ -58,6 +58,59 @@ public async Task Full_iteration_completes_reader()
         Assert.That(trackingReader.CompleteCallCount, Is.EqualTo(1));
     }
 
+    [Test]
+    public void Zero_size_segment_in_the_middle_of_the_stream_throws_InvalidDataException()
+    {
+        // A zero-size segment is never valid: the end of a stream is signaled by the reader completing with no
+        // bytes, not by a zero-size segment. See https://github.com/icerpc/icerpc-csharp/issues/4755.
+        byte[] payload = EncodeInt32Segments([1, 2], [], [3]);
```
bernardnormier: This doesn't look like a good test: a stream of int32 is not encoded in segments. It would be better to pick a variable-size element like a string.

tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
```diff
@@ -58,6 +58,59 @@ public async Task Full_iteration_completes_reader()
...
+        Assert.That(items, Is.EqualTo(new[] { 1, 2 }));
+        Assert.That(trackingReader.CompleteCallCount, Is.EqualTo(1));
+    }
+
+    [Test]
+    public void Zero_size_segment_at_the_end_of_the_stream_throws_InvalidDataException()
+    {
+        // A zero-size segment is never valid, including at the end of the stream: the end of a stream is signaled
+        // by the reader completing with no bytes, not by a zero-size segment.
+        // See https://github.com/icerpc/icerpc-csharp/issues/4755.
```
bernardnormier: Same here.

bernardnormier (approved):

tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
pepone: 🤖 **Claude:** Done in 415bc1b68. Converted both zero-size-segment tests to stream a variable-size element (`string`) via the segment-based `ToAsyncStream` overload. You're right that this matters: fixed-size elements like `int32` use the `elementSize` overload, whose `ReadAsync` calls `ReadAtLeastAsync` and never reads segments — so a zero-size *segment* can only occur in a variable-size element stream. The previous `int32` + no-`elementSize` pairing tested a combination that does not happen on the wire. Verified the new string tests fail against the unfixed `IsCompleteSegment` and the full Slice + IceRpc suites pass.

tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
pepone: 🤖 **Claude:** Done in 415bc1b68 — same change applied to this test (now streams `string`). See the reply above for the rationale.

tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
```diff
@@ -58,6 +58,61 @@ public async Task Full_iteration_completes_reader()
         Assert.That(trackingReader.CompleteCallCount, Is.EqualTo(1));
     }
 
+    [Test]
+    public void Zero_size_segment_in_the_middle_of_the_stream_throws_InvalidDataException()
+    {
+        // A zero-size segment is never valid: the end of a stream is signaled by the reader completing with no
+        // bytes, not by a zero-size segment. See https://github.com/icerpc/icerpc-csharp/issues/4755. We use a
+        // variable-size element (string) because only variable-size element streams are encoded with segments.
```
pepone:
```suggestion
        // A zero-size segment is never valid: the end of a stream is signaled by the reader completing with no
        // bytes, not by a zero-size segment.
```

tests/IceRpc.Slice.Tests/AsyncStreamTests.cs
```diff
@@ -58,6 +58,60 @@ public async Task Full_iteration_completes_reader()
...
+        Assert.That(trackingReader.CompleteCallCount, Is.EqualTo(1));
+    }
+
+    [Test]
+    public void Zero_size_segment_at_the_end_of_the_stream_throws_InvalidDataException()
+    {
+        // A zero-size segment is never valid, including at the end of the stream: the end of a stream is signaled
+        // by the reader completing with no bytes, not by a zero-size segment.
+        // See https://github.com/icerpc/icerpc-csharp/issues/4755. We use a variable-size element (string) because
+        // only variable-size element streams are encoded with segments.
```
pepone:
```suggestion
        // A zero-size segment is never valid, including at the end of the stream: the end of a stream is signaled
        // by the reader completing with no bytes, not by a zero-size segment.
```

PR #4758, opened 2026-06-11
Fail Slic connection on fire-and-forget write failure

### LINKED ISSUE #4757: [Audit-Low] Slic duplex-write failure does not close the connection; flush exception escapes a fire-and-forget task

## Summary

If `_connection.WriteAsync` fails in `SlicDuplexConnectionWriter`'s background write task, the task only does `_pipe.Reader.Complete(exception)` — it never calls `TryClose`, so `_isClosed` stays false and the connection looks healthy until the read side eventually errors. Until then, every stream writer parked on `FlushAsync` repeatedly trips over the rethrown exception while `AcceptStreamAsync`/`CreateStreamAsync` and the close machinery don't react to the dead write side. Additionally, the fire-and-forget `WriteStreamFrameAsync` local function (`_ = WriteStreamFrameAsync()`) catches only `OperationCanceledException` and `InvalidOperationException` around `FlushAsync` — the rethrown transport `IceRpcException` escapes into the unawaited task as an unobserved task exception, and the error that should fail the connection is swallowed.

## Source references

- [src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionWriter.cs#L72-L103](https://github.com/icerpc/icerpc-csharp/blob/ed5f6b8843d95e96cbc0251329bf8301eb1e520b/src/IceRpc/Transports/Slic/Internal/SlicDuplexConnectionWriter.cs#L72-L103) — failure path completes the pipe only
- [src/IceRpc/Transports/Slic/Internal/SlicConnection.cs#L784-L834](https://github.com/icerpc/icerpc-csharp/blob/ed5f6b8843d95e96cbc0251329bf8301eb1e520b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs#L784-L834) — fire-and-forget task with narrow catches
- [src/IceRpc/Transports/Slic/Internal/SlicConnection.cs#L1683](https://github.com/icerpc/icerpc-csharp/blob/ed5f6b8843d95e96cbc0251329bf8301eb1e520b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs#L1683) — `_isClosed` changes only through `TryClose`

## Recommendation

Fail the connection (`TryClose`) from the background writer's failure path, and widen the fire-and-forget task's catch to observe transport exceptions.

## Audit context

Finding 20 of [general-correctness-audit-2026-06-10.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-06-10.md); confirmed by Codex ([cross-review](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-codex-review-2026-06-10.md)).

Severity: **Low**.

pepone:
When the SlicDuplexConnectionWriter background task fails to write on the duplex connection, it completes its pipe reader with the transport exception. The fire-and-forget WriteStreamFrame path only caught OperationCanceledException and InvalidOperationException, so the rethrown IceRpcException escaped as an unobserved task exception and the connection was not closed from the write side.

Catch the IceRpcException and fail the connection via TryClose so the dead write side is surfaced promptly instead of relying on the read frames task.

Fixes #4757

### DIFF
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -830,6 +830,19 @@ internal class SlicConnection : IMultiplexedConnection
                 {
                     // The pipe writer was completed (Shutdown called) — connection is going away.
                 }
+                catch (IceRpcException exception)
+                {
+                    // The duplex connection write failed. Since this fire-and-forget task has no caller to observe
+                    // the failure, close the connection.
+                    TryClose(exception, "The connection was lost.", IceRpcError.ConnectionAborted);
+                }
+                catch (Exception exception)
+                {
+                    // A duplex connection is only expected to fail a write with an IceRpcException. Rethrow to
+                    // generate an unobserved task exception.
+                    Debug.Fail($"The stream frame write failed with an unexpected exception: {exception}");
+                    throw;
+                }
             }
         }
     }
--- a/tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
+++ b/tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
@@ -1113,6 +1113,59 @@ public class SlicTransportTests
         duplexClientConnection.Operations.Hold = DuplexTransportOperations.None;
     }
 
+    /// <summary>Verifies that when a write on the underlying duplex connection fails, a subsequent fire-and-forget
+    /// stream frame (here the StreamLast frame sent when the stream's output is completed) closes the Slic connection
+    /// instead of letting the transport exception go unobserved.</summary>
+    [Test]
+    public async Task Duplex_connection_write_failure_from_fire_and_forget_stream_frame_closes_connection()
+    {
+        // Arrange
+        await using ServiceProvider provider = new ServiceCollection()
+            .AddSlicTest()
+            .AddTestDuplexTransportDecorator()
+            .BuildServiceProvider(validateScopes: true);
+        var sut = provider.GetRequiredService<ClientServerMultiplexedConnection>();
+        await sut.AcceptAndConnectAsync();
+
+        var duplexClientConnection =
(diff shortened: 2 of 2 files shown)


src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -830,6 +830,18 @@ async Task WriteStreamFrameAsync()
                 {
                     // The pipe writer was completed (Shutdown called) — connection is going away.
                 }
+                catch (IceRpcException exception)
+                {
+                    // The background writer task failed and completed the pipe with this exception. Fail the
```
bernardnormier: I find this comment confusing. "This" task is the background writer task. And it's the awaited FlushAsync that throws this exception.

Better no comment than a confusing comment.

bernardnormier: It's also unclear where the pipe comes from. _duplexConnectionWriter is not a pipe.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -830,6 +830,18 @@ async Task WriteStreamFrameAsync()
...
+                    // connection since this fire-and-forget task has no caller to observe the failure.
+                    TryClose(exception, "The connection was lost.", IceRpcError.ConnectionAborted);
+                }
+                catch (Exception exception)
+                {
+                    // A duplex connection is only expected to fail a write with an IceRpcException.
+                    Debug.Fail($"The stream frame write failed with an unexpected exception: {exception}");
+                    TryClose(exception, "The connection was lost.", IceRpcError.ConnectionAborted);
+                }
```
bernardnormier: We should rethrow to generate an UTE. That's what we do in IceRpcProtocolConnection.

bernardnormier (approved):

PR #4761, opened 2026-06-11
Better protocolMajor validation when decoding Ice service address

bernardnormier:
A minor fix to the protocolMajor decoding.

### DIFF
--- a/src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs
+++ b/src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs
@@ -211,21 +211,21 @@ public static class IceProxyIceDecoderExtensions
         byte protocolMinor = decoder.DecodeByte();
         decoder.Skip(2); // skip encoding major and minor
 
-        if (protocolMajor == 0)
-        {
-            throw new InvalidDataException("Received service address with protocol set to 0.");
-        }
+        Protocol protocol = protocolMajor == Protocol.Ice.ByteValue ? Protocol.Ice :
+            (protocolMajor == Protocol.IceRpc.ByteValue ? Protocol.IceRpc :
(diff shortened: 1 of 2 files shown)


src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs
```diff
@@ -211,21 +211,21 @@ private static ServiceAddress DecodeServiceAddressCore(this ref IceDecoder decod
...
+            (protocolMajor == Protocol.IceRpc.ByteValue ? Protocol.IceRpc :
+                throw new InvalidDataException(
+                    $"Received service address with invalid protocolMajor value: {protocolMajor}"));
+
         if (protocolMinor != 0)
         {
             throw new InvalidDataException(
-                $"Received service address with invalid protocolMinor value: {protocolMinor}.");
+                $"Received service address with invalid protocolMinor value: {protocolMinor}");
         }
```
pepone, line 223: Maybe we can use a switch expression:

```
Protocol protocol = protocolMajor switch
{
    Protocol.Ice.ByteValue => Protocol.Ice,
    Protocol.IceRpc.ByteValue => Protocol.IceRpc,
    _ => throw new InvalidDataException(
        $"Received service address with invalid protocolMajor value: {protocolMajor}.")
};

if (protocolMinor != 0)
{
        $"Received service address with invalid protocolMinor value: {protocolMinor}.");
}
```

pepone (approved):

src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs
bernardnormier: I am pretty sure you need constant values. `Protocol.Ice.ByteValue` is not a constant.

PR #4764, opened 2026-06-12
Write Slic Pong frames in the background and cap queued Pongs

pepone:
Fixes icerpc/icerpc-csharp-audit#17.
Fixes icerpc/icerpc-csharp-audit#29.

Since #4578, the Slic read frames loop awaits the Pong frame write that responds to a Ping frame. This write acquires `_writeSemaphore`, which another writer can hold for an arbitrarily long time while parked on `FlushAsync` when the connection's outbound pipe reaches `PauseWriterThreshold` (peer not reading). While the loop is blocked on the semaphore, no further frames are read and the idle timeout cannot fire, since the idle abort is only armed during `ReadAsync`: the connection hangs until it's disposed.

### Changes

- The Pong frame is now written by a background task, so the read frames loop keeps processing incoming frames (and the idle timeout keeps working) when writes are slow or stuck.
- New `SlicTransportOptions.MaxOutstandingPongs` option (default: 100, minimum: 1) that bounds the number of Pong frames queued for sending. The connection is aborted with a protocol error when a Ping frame is received while this limit is reached, so a peer can't queue Pong writes without limit by flooding the connection with Ping frames.

### Tests

- `Read_frames_loop_is_not_blocked_by_pong_write_when_a_write_is_parked_on_pipe_threshold`: parks a server stream write on the connection-level pipe pause (small `PauseWriterThreshold` plus a held duplex write) and verifies, using a raw duplex client, that the server still reads frames and accepts a new stream after receiving a Ping frame. Verified that this test fails (the accept times out) without the `SlicConnection` fix.
- `Connection_is_aborted_when_max_outstanding_pongs_is_exceeded`: with `MaxOutstandingPongs = 3` and a parked write, a fourth Ping frame aborts the connection with `Received a Ping frame while 3 Pong frames are already queued for sending.`
- Default value and validation tests for the new option.

- `Ping_received_after_local_close_does_not_fail_graceful_close`: writing the Pong frame in the background (with connection-closed exceptions swallowed) also fixes audit #29 — a Ping frame received after a local CloseAsync no longer faults the read frames loop. Verified that this test fails without the fix with the exact audit #29 failure mode (the graceful close throws `IceRpcException` from the Pong write).

### DIFF
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -1317,15 +1317,14 @@ internal class SlicConnection : IMultiplexedConnection
                 (ref SliceDecoder decoder) => new PingBody(ref decoder),
                 cancellationToken).ConfigureAwait(false);
 
-            // Only the read frames loop increments _outstandingPongCount, so the check before the increment doesn't
-            // race with other increments; concurrent decrements can only make the count smaller.
-            if (Volatile.Read(ref _outstandingPongCount) >= _maxOutstandingPongs)
+            if (Interlocked.Increment(ref _outstandingPongCount) > _maxOutstandingPongs)
             {
+                // Roll back the increment so the count keeps matching the number of queued Pong frames.
+                Interlocked.Decrement(ref _outstandingPongCount);
                 throw new IceRpcException(
                     IceRpcError.IceRpcError,
                     $"Received a {nameof(FrameType.Ping)} frame while {_maxOutstandingPongs} {nameof(FrameType.Pong)} frames are already queued for sending.");
             }
-            Interlocked.Increment(ref _outstandingPongCount);
 
             // Return a pong frame with the ping payload, written in the background: writing it from the read frames
             // loop would block the loop when another writer holds _writeSemaphore while parked on a full outbound
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -76,6 +76,7 @@ internal class SlicConnection : IMultiplexedConnection
     private ulong? _lastRemoteUnidirectionalStreamId;
     private readonly TimeSpan _localIdleTimeout;
     private readonly int _maxBidirectionalStreams;
+    private readonly int _maxOutstandingPongs;
     private readonly int _maxStreamFrameSize;
     private readonly int _maxUnidirectionalStreams;
     // _mutex ensure the assignment of _lastRemoteXxx members and the addition of the stream to _streams is
@@ -83,8 +84,15 @@ internal class SlicConnection : IMultiplexedConnection
     private readonly Lock _mutex = new();
     private ulong _nextBidirectionalId;
     private ulong _nextUnidirectionalId;
+
+    // The number of Pong frames queued for sending (in response to Ping frames) but not yet written to the duplex
+    // connection. The connection is aborted when a Ping frame is received while this count has reached
+    // _maxOutstandingPongs.
+    private int _outstandingPongCount;
     private IceRpcError? _peerCloseError;
     private TimeSpan _peerIdleTimeout = Timeout.InfiniteTimeSpan;
+
+    // The number of Ping frames sent to the peer that have not been answered yet by a Pong frame.
     private int _pendingPongCount;
     private Task? _readFramesTask;
 
@@ -600,6 +608,7 @@ internal class SlicConnection : IMultiplexedConnection
         InitialStreamWindowSize = slicOptions.InitialStreamWindowSize;
         PauseWriterThreshold = slicOptions.PauseWriterThreshold;
         _localIdleTimeout = slicOptions.IdleTimeout;
+        _maxOutstandingPongs = slicOptions.MaxOutstandingPongs;
         _maxStreamFrameSize = slicOptions.MaxStreamFrameSize;
 
         _acceptStreamChannel = Channel.CreateUnbounded<IMultiplexedStream>(new UnboundedChannelOptions
@@ -1308,11 +1317,43 @@ internal class SlicConnection : IMultiplexedConnection
                 (ref SliceDecoder decoder) => new PingBody(ref decoder),
                 cancellationToken).ConfigureAwait(false);
 
-            // Return a pong frame with the ping payload.
-            await WriteConnectionFrameAsync(
-                FrameType.Pong,
-                new PongBody(pingBody.Payload).Encode,
-                cancellationToken).ConfigureAwait(false);
+            if (Interlocked.Increment(ref _outstandingPongCount) > _maxOutstandingPongs)
+            {
+                throw new InvalidDataException(
+                    $"Received a {nameof(FrameType.Ping)} frame while {_maxOutstandingPongs} {nameof(FrameType.Pong)} frames are already queued for sending.");
+            }
+
+            // Return a pong frame with the ping payload, written in the background: writing it from the read frames
+            // loop would block the loop when another writer holds _writeSemaphore while parked on a full outbound
+            // pipe, suppressing further reads (and idle timeout detection) until the pipe drains.
+            _ = WritePongFrameAsync(pingBody.Payload);
+        }
+
+        async Task WritePongFrameAsync(long payload)
+        {
+            try
+            {
+                await WriteConnectionFrameAsync(
+                    FrameType.Pong,
+                    new PongBody(payload).Encode,
+                    _closedCancellationToken).ConfigureAwait(false);
+            }
+            catch (IceRpcException)
+            {
+                // Expected if the connection is closed.
+            }
+            catch (OperationCanceledException)
+            {
+                // Expected if the connection is closed.
+            }
+            catch (Exception exception)
+            {
+                Debug.Fail($"The sending of a Pong frame failed with an unexpected exception: {exception}");
+            }
+            finally
+            {
+                Interlocked.Decrement(ref _outstandingPongCount);
+            }
         }
 
         async Task ReadPongFrameAsync(int size, CancellationToken cancellationToken)
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -1319,8 +1319,6 @@ internal class SlicConnection : IMultiplexedConnection
 
             if (Interlocked.Increment(ref _outstandingPongCount) > _maxOutstandingPongs)
             {
-                // Roll back the increment so the count keeps matching the number of queued Pong frames.
-                Interlocked.Decrement(ref _outstandingPongCount);
                 throw new IceRpcException(
                     IceRpcError.IceRpcError,
                     $"Received a {nameof(FrameType.Ping)} frame while {_maxOutstandingPongs} {nameof(FrameType.Pong)} frames are already queued for sending.");
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -1317,11 +1317,15 @@ internal class SlicConnection : IMultiplexedConnection
                 (ref SliceDecoder decoder) => new PingBody(ref decoder),
                 cancellationToken).ConfigureAwait(false);
 
-            if (Interlocked.Increment(ref _outstandingPongCount) > _maxOutstandingPongs)
+            // Only the read frames loop increments _outstandingPongCount, so the check before the increment doesn't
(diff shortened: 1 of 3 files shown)


src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1321,11 +1317,43 @@ async Task ReadPingFrameAndWritePongFrameAsync(int size, CancellationToken cance
...
                 cancellationToken).ConfigureAwait(false);
 
-            // Return a pong frame with the ping payload.
-            await WriteConnectionFrameAsync(
-                FrameType.Pong,
-                new PongBody(pingBody.Payload).Encode,
-                cancellationToken).ConfigureAwait(false);
+            if (Interlocked.Increment(ref _outstandingPongCount) > _maxOutstandingPongs)
+            {
+                throw new InvalidDataException(
```
pepone: Not clear this is the correct exception here. InvalidDataException is typically when you receive data with doesn't follow the expected encoding, but sending an additional frame seems something else.

What do we do when you send more data than allow by Slice window size?

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1321,11 +1317,43 @@ async Task ReadPingFrameAndWritePongFrameAsync(int size, CancellationToken cance
...
+            {
+                // Expected if the connection is closed.
+            }
+            catch (OperationCanceledException)
+            {
+                // Expected if the connection is closed.
+            }
+            catch (Exception exception)
+            {
+                Debug.Fail($"The sending of a Pong frame failed with an unexpected exception: {exception}");
```
pepone, line 1352: We should `throw;` so the exception is not swallow in release builds and can be presented to the application as an unobserved exceptions.

src/IceRpc/Transports/Slic/SlicTransportOptions.cs
```diff
@@ -42,6 +42,23 @@ public int InitialStreamWindowSize
             value;
     }
 
+    /// <summary>Gets or sets the maximum number of Pong frames that can be queued for sending. A Pong frame is queued
+    /// for sending when a Ping frame is received, and remains queued until it is written to the underlying duplex
+    /// connection. The connection is aborted when a Ping frame is received while this limit is reached; this protects
+    /// against a peer that sends Ping frames faster than the connection can write the Pong frames.</summary>
```
pepone: Isn't this more about protecting from a peer that doesn't read, and keep sending pings which in turn will keep queueing write pong tasks?

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
pepone: 🤖 **Claude:** Good point — switched to `IceRpcException(IceRpcError.IceRpcError)`, matching the flow-control violation (`Received more data than flow control permits.` in `SlicPipeReader`) and the max stream count check: this is a resource-limit violation, not malformed data. The exception now surfaces directly from `AcceptStreamAsync` instead of being wrapped as a Slic protocol error.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
pepone: 🤖 **Claude:** Done — added `throw;` after the `Debug.Fail` so release builds surface it as an unobserved task exception.

src/IceRpc/Transports/Slic/SlicTransportOptions.cs
pepone: 🤖 **Claude:** Right, reworded: the limit protects against a peer that doesn't read from the connection but keeps sending Ping frames, which would otherwise queue Pong frame writes without limit.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1308,11 +1317,43 @@ async Task ReadPingFrameAndWritePongFrameAsync(int size, CancellationToken cance
...
-            // Return a pong frame with the ping payload.
-            await WriteConnectionFrameAsync(
-                FrameType.Pong,
-                new PongBody(pingBody.Payload).Encode,
-                cancellationToken).ConfigureAwait(false);
+            if (Interlocked.Increment(ref _outstandingPongCount) > _maxOutstandingPongs)
+            {
+                throw new InvalidDataException(
+                    $"Received a {nameof(FrameType.Ping)} frame while {_maxOutstandingPongs} {nameof(FrameType.Pong)} frames are already queued for sending.");
+            }
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Reworked: the count is now checked before incrementing (only the read frames loop increments, so check-then-increment doesn't race), so it stays exact on the abort path.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1308,11 +1317,43 @@ async Task ReadPingFrameAndWritePongFrameAsync(int size, CancellationToken cance
...
+                // Expected if the connection is closed.
+            }
+            catch (OperationCanceledException)
+            {
+                // Expected if the connection is closed.
+            }
+            catch (Exception exception)
+            {
+                Debug.Fail($"The sending of a Pong frame failed with an unexpected exception: {exception}");
+            }
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Not applicable to this path: `WriteConnectionFrameAsync` re-checks `_isClosed` under `_writeSemaphore` after acquiring it, and `_duplexConnectionWriter.Shutdown()` is only ever called while holding the semaphore with `_isClosed` already true — so a pong write can't observe a shut-down writer. `SendPingAsync` uses the same path without an `InvalidOperationException` catch. The `WriteStreamFrameAsync` catch exists for its fire-and-forget callers; if an unexpected exception does occur here it now propagates (see the `throw;` added per the other comment) instead of being silently swallowed.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1308,11 +1317,51 @@ async Task ReadPingFrameAndWritePongFrameAsync(int size, CancellationToken cance
...
-                cancellationToken).ConfigureAwait(false);
+            // Only the read frames loop increments _outstandingPongCount, so the check before the increment doesn't
+            // race with other increments; concurrent decrements can only make the count smaller.
+            if (Volatile.Read(ref _outstandingPongCount) >= _maxOutstandingPongs)
+            {
+                throw new IceRpcException(
+                    IceRpcError.IceRpcError,
+                    $"Received a {nameof(FrameType.Ping)} frame while {_maxOutstandingPongs} {nameof(FrameType.Pong)} frames are already queued for sending.");
+            }
+            Interlocked.Increment(ref _outstandingPongCount);
```
bernardnormier: Simplification: fold the increment into the limit check and drop the comment — it's more noise than signal here. Exceeding the limit is fatal (the throw aborts the read frames loop via `TryClose`), so the over-increment-by-1 on the throw path is never observed; no need to explain it.

```suggestion
            if (Interlocked.Increment(ref _outstandingPongCount) > _maxOutstandingPongs)
            {
                throw new IceRpcException(
                    IceRpcError.IceRpcError,
                    $"Received a {nameof(FrameType.Ping)} frame while {_maxOutstandingPongs} {nameof(FrameType.Pong)} frames are already queued for sending.");
            }
```

pepone: 🤖 **Claude (Opus 4.8):** Applied in `00c50f9d`, with one small deviation from the exact suggestion. I folded the increment into the limit check as you proposed (dropping the separate `Volatile.Read`), but added a rollback on the reject path so the counter stays exact:

```csharp
if (Interlocked.Increment(ref _outstandingPongCount) > _maxOutstandingPongs)
{
    // Roll back the increment so the count keeps matching the number of queued Pong frames.
    Interlocked.Decrement(ref _outstandingPongCount);
    throw new IceRpcException(
        IceRpcError.IceRpcError,
        $"Received a {nameof(FrameType.Ping)} frame while {_maxOutstandingPongs} {nameof(FrameType.Pong)} frames are already queued for sending.");
}
```

You're right that the over-increment is never observed (the throw aborts the read frames loop via `TryClose`), so the rollback is belt-and-suspenders — but it preserves the exact-count property added earlier in this PR for the Copilot off-by-one thread, while still collapsing to a single atomic op. Build + 210 Slic tests pass; a codex cross-check confirmed the admission bound is unchanged and the rollback can't double-decrement or drift against the background `finally` decrements. Happy to drop the rollback and go with your exact form if you prefer the simpler version.

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
```diff
@@ -1784,6 +1745,213 @@ public async Task Server_dispose_does_not_deadlock_when_close_is_blocked_on_pipe
         }
     }
 
+    /// <summary>Verifies that the read frames loop is not blocked by the Pong frame write that responds to a Ping
+    /// frame when another write holds the connection's write semaphore while parked on the connection-level pipe
+    /// pause (PauseWriterThreshold reached). See icerpc/icerpc-csharp-audit#17.</summary>
```
pepone:
```suggestion
    /// pause (PauseWriterThreshold reached).</summary>
```

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
```diff
@@ -1784,6 +1745,213 @@ public async Task Server_dispose_does_not_deadlock_when_close_is_blocked_on_pipe
...
+            streamId: 4ul,
+            (ref SliceEncoder encoder) => encoder.EncodeBool(false));
+
+        // Assert. The WaitAsync margin exceeds CI's --blame-hang-timeout (60 s) so that a hang is detected and
+        // dumped by the hang detection first; the timeout is only a fallback that fails this test instead of
+        // hanging the test run when hang detection is not enabled.
+        Assert.That(
+            async () => await nextAcceptStreamTask.AsTask().WaitAsync(TimeSpan.FromMinutes(2)),
+            Throws.Nothing);
+
```
pepone, line 1818: We should assert here the write didn't complete, this ensures the read loop continue working while write was block.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1321,11 +1317,50 @@ async Task ReadPingFrameAndWritePongFrameAsync(int size, CancellationToken cance
...
 
-            // Return a pong frame with the ping payload.
-            await WriteConnectionFrameAsync(
-                FrameType.Pong,
-                new PongBody(pingBody.Payload).Encode,
-                cancellationToken).ConfigureAwait(false);
+            if (Interlocked.Increment(ref _outstandingPongCount) > _maxOutstandingPongs)
+            {
+                // Roll back the increment so the count keeps matching the number of queued Pong frames.
+                Interlocked.Decrement(ref _outstandingPongCount);
```
pepone:
```suggestion
```

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
pepone: 🤖 **Claude (Opus 4.8):** Done in `7ad6a1ff`. Added right after the second stream is accepted, before the cleanup releases the hold:

```csharp
Assert.That(serverWriteTask.IsCompleted, Is.False);
```

`serverWriteTask` is the right proxy — it parked on `FlushAsync` while holding the write semaphore, so as long as it's incomplete the queued Pong write is still blocked too. Asserting it here confirms the read frames loop accepted the second stream *while* the write was genuinely still parked, rather than because the write quietly drained. Targeted test passes.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -1308,11 +1317,48 @@ async Task ReadPingFrameAndWritePongFrameAsync(int size, CancellationToken cance
...
-            await WriteConnectionFrameAsync(
-                FrameType.Pong,
-                new PongBody(pingBody.Payload).Encode,
-                cancellationToken).ConfigureAwait(false);
+            if (Interlocked.Increment(ref _outstandingPongCount) > _maxOutstandingPongs)
+            {
+                throw new IceRpcException(
+                    IceRpcError.IceRpcError,
+                    $"Received a {nameof(FrameType.Ping)} frame while {_maxOutstandingPongs} {nameof(FrameType.Pong)} frames are already queued for sending.");
+            }
```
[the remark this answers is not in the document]
bernardnormier: And this all fine. We never read again _outstandingPongCount in the abort path.

bernardnormier (approved):
Looks good.

Issue #26, opened 2020-10-16
idempotent and tuple return type


bernardnormier:
This currently does not work on master - I get a strange syntax error.

For example:
```
idempotent (EndpointDataSeq endpoints, StringSeq newLocation) resolveLocation();
```

PR #4765, opened 2026-06-12
Fix Slic stream lifecycle bugs triggered by canceled writes

pepone:
Fixes icerpc/icerpc-csharp-audit#25.
Fixes icerpc/icerpc-csharp-audit#26.

Two bugs left a Slic connection degraded when an invocation was canceled at the wrong time:

- **Stream-count permit leak (audit #25):** a stream created by `CreateStreamAsync` acquires a stream-count semaphore permit that was only released for *started* streams (`TrySetState` → `ReleaseStream`, gated on `IsStarted`). A stream abandoned before its first write — e.g. invocation canceled while waiting on payload data or send credit — burned its permit forever; after `MaxBidirectionalStreams`/`MaxUnidirectionalStreams` such leaks, every new `CreateStreamAsync` on an otherwise healthy connection blocked forever. The permit is now released when an unstarted local stream reaches its terminal state (new `SlicConnection.ReleaseUnstartedStream`).

- **`_writesClosePending` never rolled back (audit #26):** `SlicStream.WriteStreamFrameAsync` set `_writesClosePending` before the endStream write and never rolled it back when that write failed before the `StreamLast` frame was queued (cancellation while blocked on send credit or the write semaphore). The later `Complete(exception)` on the stream output was then suppressed by the `CloseWrites` guard: neither `StreamLast` nor `StreamWritesClosed` was sent, and the peer hung reading the payload until connection closure. `_writesClosePending` is now rolled back when the endStream write fails before the `StreamLast` frame was queued — detected via `_writesClosedTcs`, which `WroteLastStreamFrame` completes exactly when the frame is queued — so `CloseWrites` sends the `StreamWritesClosed` frame and the peer observes `TruncatedData`. The bundled `_closeReadsOnWritesClosure` is preserved as well since it is never cleared on this path.

### Tests

- `Completing_unstarted_stream_releases_the_stream_count_permit` (bidirectional and unidirectional): with max streams = 1, completes an unstarted stream and verifies a new stream can still be created.
- `Canceled_end_stream_write_does_not_lose_the_writes_closed_frame`: cancels an endStream write blocked on send credit, completes the output with an exception, and verifies the peer observes `TruncatedData` instead of waiting for data forever.
- `End_stream_write_canceled_after_frame_queued_preserves_graceful_peer_reads`: pins the opposite case — when the `StreamLast` frame was already queued (write parked on the connection-level pipe pause), no `StreamWritesClosed` frame is sent after it and the peer reads all the data with a graceful end of stream. This guards against an over-eager rollback; it passes before and after the fix.

Verified the first three fail without the fix (all time out on their 2 min fallback guards); the `WaitAsync` margins follow the convention that CI's `--blame-hang-timeout` (60 s) detects and dumps a hang first.

Full `IceRpc.Tests` suite passes (991 passed / 0 failed), `dotnet build` and `dotnet format --verify-no-changes` clean.

### DIFF
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -703,9 +703,7 @@ internal class SlicConnection : IMultiplexedConnection
     /// <param name="stream">The released stream.</param>
     internal void ReleaseStream(SlicStream stream)
     {
-        // A stream that was created but never started is always local; only started streams are registered in _streams.
-        Debug.Assert(stream.IsStarted || !stream.IsRemote);
-
+        // Only a started stream has an Id and is registered in _streams.
         if (stream.IsStarted)
         {
             _streams.Remove(stream.Id, out SlicStream? _);
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -703,9 +703,13 @@ internal class SlicConnection : IMultiplexedConnection
     /// <param name="stream">The released stream.</param>
     internal void ReleaseStream(SlicStream stream)
     {
-        Debug.Assert(stream.IsStarted);
+        // A stream that was created but never started is always local; only started streams are registered in _streams.
+        Debug.Assert(stream.IsStarted || !stream.IsRemote);
 
-        _streams.Remove(stream.Id, out SlicStream? _);
+        if (stream.IsStarted)
+        {
+            _streams.Remove(stream.Id, out SlicStream? _);
+        }
 
         if (stream.IsRemote)
         {
@@ -719,26 +723,6 @@ internal class SlicConnection : IMultiplexedConnection
             }
         }
         else if (!_isClosed)
-        {
-            if (stream.IsBidirectional)
-            {
-                _bidirectionalStreamSemaphore!.Release();
-            }
-            else
-            {
-                _unidirectionalStreamSemaphore!.Release();
-            }
-        }
-    }
-
-    /// <summary>Releases the stream-count semaphore permit acquired by <see cref="CreateStreamAsync" /> for a local
-    /// stream that was created but never started.</summary>
-    /// <param name="stream">The unstarted stream.</param>
-    internal void ReleaseUnstartedStream(SlicStream stream)
-    {
-        Debug.Assert(!stream.IsRemote && !stream.IsStarted);
-
-        if (!_isClosed)
         {
             if (stream.IsBidirectional)
             {
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -731,6 +731,26 @@ internal class SlicConnection : IMultiplexedConnection
         }
     }
 
+    /// <summary>Releases the stream-count semaphore permit acquired by <see cref="CreateStreamAsync" /> for a local
+    /// stream that was created but never started.</summary>
+    /// <param name="stream">The unstarted stream.</param>
+    internal void ReleaseUnstartedStream(SlicStream stream)
+    {
+        Debug.Assert(!stream.IsRemote && !stream.IsStarted);
+
+        if (!_isClosed)
+        {
+            if (stream.IsBidirectional)
+            {
+                _bidirectionalStreamSemaphore!.Release();
+            }
+            else
+            {
+                _unidirectionalStreamSemaphore!.Release();
+            }
+        }
+    }
+
     /// <summary>Throws the connection closure exception if the connection is already closed.</summary>
     internal void ThrowIfClosed()
     {
--- a/src/IceRpc/Transports/Slic/Internal/SlicStream.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicStream.cs
@@ -359,7 +359,7 @@ internal class SlicStream : IMultiplexedStream
     /// <param name="endStream"><see langword="true" /> to write a <see cref="FrameType.StreamLast" /> frame; otherwise,
     /// <see langword="false" />.</param>
     /// <param name="cancellationToken">A cancellation token that receives the cancellation requests.</param>
-    internal ValueTask<FlushResult> WriteStreamFrameAsync(
+    internal async ValueTask<FlushResult> WriteStreamFrameAsync(
         ReadOnlySequence<byte> source1,
         ReadOnlySequence<byte> source2,
         bool endStream,
@@ -375,13 +375,31 @@ internal class SlicStream : IMultiplexedStream
             }
         }
 
-        return _connection.WriteStreamDataFrameAsync(
-            this,
-            source1,
-            source2,
-            endStream,
-            writeReadsClosedFrame,
-            cancellationToken);
+        try
+        {
+            return await _connection.WriteStreamDataFrameAsync(
+                this,
+                source1,
+                source2,
+                endStream,
+                writeReadsClosedFrame,
+                cancellationToken).ConfigureAwait(false);
+        }
+        catch
+        {
+            // If the write failed before the StreamLast frame was queued on the connection (WroteLastStreamFrame
+            // completes _writesClosedTcs when the frame is queued), roll back _writesClosePending so that completing
+            // the stream output can still close writes and notify the peer with a StreamWritesClosed frame. The
+            // bundled reads closure isn't lost either since _closeReadsOnWritesClosure wasn't cleared.
+            if (endStream && !_writesClosedTcs.Task.IsCompleted)
+            {
+                lock (_mutex)
+                {
+                    _writesClosePending = false;
+                }
+            }
+            throw;
+        }
     }
 
     /// <summary>Notifies the stream that the <see cref="FrameType.StreamLast" /> was written by the
@@ -426,6 +444,12 @@ internal class SlicStream : IMultiplexedStream
                 if (IsStarted)
                 {
                     _connection.ReleaseStream(this);
+                }
+                else if (!IsRemote)
+                {
+                    // The stream was created but never started: release the stream-count semaphore permit acquired by
+                    // CreateStreamAsync.
+                    _connection.ReleaseUnstartedStream(this);
                 }
             }
             return true;
--- a/src/IceRpc/Transports/Slic/Internal/SlicStream.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicStream.cs
@@ -452,13 +452,8 @@ internal class SlicStream : IMultiplexedStream
             if (newState.HasFlag(State.ReadsClosed | State.WritesClosed))
             {
                 // The stream reads and writes are closed, it's time to release the stream to either allow creating or
-                // accepting a new stream. We don't release an unstarted remote stream (which can briefly exist when
-                // AddStream fails on a closing connection) since it never acquired any resource. Releasing an unstarted
-                // local stream returns the stream-count semaphore permit acquired by CreateStreamAsync.
-                if (IsStarted || !IsRemote)
-                {
-                    _connection.ReleaseStream(this);
-                }
+                // accepting a new stream.
+                _connection.ReleaseStream(this);
             }
             return true;
         }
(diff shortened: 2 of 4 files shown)


tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
```diff
@@ -1619,6 +1619,157 @@ public async Task Stream_write_cancellation_while_blocked_on_peer_credit()
...
+        // instead of hanging the test run when hang detection is not enabled.
+        Assert.That(
+            async () =>
+            {
+                while (true)
+                {
+                    ReadResult readResult = await streams.Remote.Input.ReadAsync(default)
+                        .AsTask().WaitAsync(TimeSpan.FromMinutes(2));
+                    streams.Remote.Input.AdvanceTo(readResult.Buffer.End);
+                }
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Correct — applied. The loop now asserts `IsCompleted` is false after each read, so a graceful end of stream fails the test immediately instead of spinning.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -722,6 +722,26 @@ internal void ReleaseStream(SlicStream stream)
...
+            if (stream.IsBidirectional)
+            {
+                _bidirectionalStreamSemaphore!.Release();
+            }
+            else
+            {
+                _unidirectionalStreamSemaphore!.Release();
+            }
+        }
+    }
```
bernardnormier: Optional simplification: instead of a separate `ReleaseUnstartedStream` that duplicates the local-stream semaphore-release block, fold the unstarted case into `ReleaseStream`. The only thing in `ReleaseStream` that's unsafe for an unstarted stream is `_streams.Remove(stream.Id)` — `Id` throws when `_id == ulong.MaxValue` — so guarding that one line is enough; an unstarted stream is local and already lands in the `else if (!_isClosed)` release branch.

```csharp
internal void ReleaseStream(SlicStream stream)
{
    if (stream.IsStarted)
    {
        _streams.Remove(stream.Id, out SlicStream? _);
    }

    if (stream.IsRemote)
    {
        if (stream.IsBidirectional)
        {
            Interlocked.Decrement(ref _bidirectionalStreamCount);
        }
        else
        {
            Interlocked.Decrement(ref _unidirectionalStreamCount);
        }
    }
    else if (!_isClosed)
    {
        if (stream.IsBidirectional)
        {
            _bidirectionalStreamSemaphore!.Release();
        }
        else
        {
            _unidirectionalStreamSemaphore!.Release();
        }
    }
}
```

and the call site in `SlicStream.TrySetState` collapses to:

```csharp
if (IsStarted || !IsRemote)
{
    _connection.ReleaseStream(this);
}
```

This drops `ReleaseUnstartedStream` and the duplicated release block with it. Behavior is unchanged across the case table: started (local/remote) → same as today; unstarted local → skips the `Remove`, releases the permit; unstarted remote can't occur (remote streams get their ID at creation) and `IsStarted || !IsRemote` is `false` for it anyway, matching the current `else if (!IsRemote)`. `ReleaseStream` has a single caller, so widening its contract is contained.

Heads-up: this doesn't fix the failing spellcheck by itself — the method name and its doc comment go away, but the `Completing_unstarted_stream_…` test name still trips cspell, so `"unstarted"` still needs adding to `.vscode/cspell.json` (or renaming the test).

src/IceRpc/Transports/Slic/Internal/SlicStream.cs
```diff
@@ -375,13 +375,31 @@ internal ValueTask<FlushResult> WriteStreamFrameAsync(
...
+            // bundled reads closure isn't lost either since _closeReadsOnWritesClosure wasn't cleared.
+            if (endStream && !_writesClosedTcs.Task.IsCompleted)
+            {
+                lock (_mutex)
+                {
+                    _writesClosePending = false;
+                }
+            }
+            throw;
+        }
```
bernardnormier: Two refinements here — one correctness-rationale, one perf.

**Keep the early `_writesClosePending = true`; don't be tempted to move it into `WroteLastStreamFrame` to drop the rollback.** The early set is *atomic with the `writeReadsClosedFrame` capture* under `_mutex`, and that atomicity is what keeps `StreamReadsClosed` sent exactly once when `CloseReads` races the endStream write on a remote bidirectional stream. If the flag were set only when the frame is queued:

1. write captures `writeReadsClosedFrame = _closeReadsOnWritesClosure` (`false`), pending still `false`;
2. read side finishes → `CloseReads(graceful)` sees `!_writesClosePending` → defers (`_closeReadsOnWritesClosure = true`), expecting to bundle with the writes-closure frame;
3. write queues `StreamLast` with the stale `writeReadsClosedFrame = false` (no bundled frame), then sets pending;
4. `Complete` → `CloseWrites` sees pending == `true` → sends nothing, and `_closeReadsOnWritesClosure` is never consumed → **`StreamReadsClosed` is lost**.

The current early-atomic set closes that window (a later `CloseReads` sees pending and sends the frame itself; an earlier one gets folded into the capture). So the localized rollback is the right design.

**But the new `async` doesn't need to be on the hot path.** Only the once-per-stream endStream write needs the bookkeeping; the bulk `endStream == false` path can forward the inner `ValueTask` directly, as it did before this PR:

```csharp
internal ValueTask<FlushResult> WriteStreamFrameAsync(
    ReadOnlySequence<byte> source1,
    ReadOnlySequence<byte> source2,
    bool endStream,
    CancellationToken cancellationToken)
{
    if (!endStream)
    {
        // Hot path: no writes-closure bookkeeping — forward without an async frame.
        return _connection.WriteStreamDataFrameAsync(
            this, source1, source2, endStream: false, writeReadsClosedFrame: false, cancellationToken);
    }

    return WriteLastStreamFrameAsync();

    async ValueTask<FlushResult> WriteLastStreamFrameAsync()
    {
        bool writeReadsClosedFrame;
        lock (_mutex)
        {
            writeReadsClosedFrame = _closeReadsOnWritesClosure;
            _writesClosePending = true;
        }

        try
        {
            return await _connection.WriteStreamDataFrameAsync(
                this, source1, source2, endStream: true, writeReadsClosedFrame, cancellationToken)
                .ConfigureAwait(false);
        }
        catch when (!_writesClosedTcs.Task.IsCompleted)
        {
            // The StreamLast frame wasn't queued; roll back so CloseWrites sends StreamWritesClosed.
            lock (_mutex) { _writesClosePending = false; }
            throw;
        }
    }
}
```

The `catch when (…)` filter also reads a bit cleaner than `catch { if (…) …; throw; }` — it only enters the handler when there's actually something to undo.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
pepone: 🤖 **Claude (Opus 4.8):** Done in `093708af`. Folded `ReleaseUnstartedStream` into `ReleaseStream`: guarded `_streams.Remove(stream.Id, …)` with `if (stream.IsStarted)` (since `Id` throws when unstarted), replaced the `Debug.Assert(stream.IsStarted)` with the combined precondition `Debug.Assert(stream.IsStarted || !stream.IsRemote)`, and deleted `ReleaseUnstartedStream` + its doc. The call site in `TrySetState` collapses to `if (IsStarted || !IsRemote) { _connection.ReleaseStream(this); }`. Verified behavior-preserving across the truth table: started local/remote unchanged; unstarted local skips the `Remove` and releases the permit (matches the old method); unstarted remote (only possible when `AddStream` fails on a closing connection) is correctly skipped, same as the old `else if (!IsRemote)`. Also added `unstarted` to `.vscode/cspell.json` for the cspell CI. Build + 210 Slic tests pass; codex confirmed the fold is behavior-preserving.

src/IceRpc/Transports/Slic/Internal/SlicStream.cs
pepone: 🤖 **Claude (Opus 4.8):** Done in `093708af`. `WriteStreamFrameAsync` is now non-async: the hot `!endStream` path forwards the inner `ValueTask` from `WriteStreamDataFrameAsync` directly (no async state machine), and the endStream bookkeeping lives in a nested `async ValueTask<FlushResult> WriteLastStreamFrameAsync()` that captures `writeReadsClosedFrame` + sets `_writesClosePending = true` under `_mutex` (kept early/atomic as you note), then rolls back via `catch when (!_writesClosedTcs.Task.IsCompleted)`. Verified the filter is equivalent to the old `catch { if (endStream && !…) …; throw; }` including the cancellation path, and the direct forward is safe here since `WriteStreamDataFrameAsync` is itself `async` (no synchronous-throw path) and the sole caller awaits it. Build + 210 Slic tests pass; codex confirmed behavior preservation.

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
```diff
@@ -703,9 +703,13 @@ internal ValueTask FillBufferWriterAsync(
     /// <param name="stream">The released stream.</param>
     internal void ReleaseStream(SlicStream stream)
     {
-        Debug.Assert(stream.IsStarted);
+        // A stream that was created but never started is always local; only started streams are registered in _streams.
+        Debug.Assert(stream.IsStarted || !stream.IsRemote);
```
bernardnormier: `ReleaseStream` already handles unstarted streams safely — `_streams.Remove` is guarded by `IsStarted` — so we can drop the assert and let the single caller invoke it unconditionally:

```suggestion
        // Only a started stream has an Id and is registered in _streams.
```

The unconditional call now also runs for an unstarted *remote* stream (the brief state when `AddStream` throws at line 1686, before it assigns the Id). That reaches the `if (stream.IsRemote)` decrement below, which is **correct** — the remote count is incremented at line 1659/1677 *before* `AddStream` — and **effectively a no-op**: `AddStream` only throws when `_isClosed` (line 1025), and once the connection is closed the stream counts are never read again (they only gate accepting new remote streams, line 1653). So the decrement has no observable effect; it just keeps the bookkeeping balanced.

src/IceRpc/Transports/Slic/Internal/SlicStream.cs
```diff
@@ -422,8 +452,10 @@ private bool TrySetState(State state)
...
                 // The stream reads and writes are closed, it's time to release the stream to either allow creating or
-                // accepting a new stream.
-                if (IsStarted)
+                // accepting a new stream. We don't release an unstarted remote stream (which can briefly exist when
+                // AddStream fails on a closing connection) since it never acquired any resource. Releasing an unstarted
+                // local stream returns the stream-count semaphore permit acquired by CreateStreamAsync.
+                if (IsStarted || !IsRemote)
                 {
                     _connection.ReleaseStream(this);
                 }
```
bernardnormier: And the call site collapses to an unconditional call (safe for the unstarted-remote case — see the `ReleaseStream` comment):

```suggestion
                // The stream reads and writes are closed, it's time to release the stream to either allow creating or
                // accepting a new stream.
                _connection.ReleaseStream(this);
```

src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
pepone: 🤖 **Claude (Opus 4.8):** Done in `013a835c5`. Dropped the `Debug.Assert` and simplified the comment as suggested — `_streams.Remove` stays guarded by `IsStarted`, so the unstarted case is safe. Verified across the truth table: started local/remote and unstarted-local are unchanged; the only newly-covered case is an unstarted **remote** stream (the brief state when `AddStream` throws on a closing connection), where the `IsRemote` decrement balances the increment from the accept path. As a small bonus the old guarded path left that increment in place — a harmless bookkeeping imbalance on an already-closing connection — so the unconditional call also keeps the count balanced. 218/218 Slic tests pass; an adversarial pass (reachability / count-balance / exactly-once / started-equivalence) found no counterexample.

src/IceRpc/Transports/Slic/Internal/SlicStream.cs
pepone: 🤖 **Claude (Opus 4.8):** Done in `013a835c5`. Collapsed the call site to an unconditional `_connection.ReleaseStream(this)`. It still runs exactly once per stream — the atomic `Interlocked.Or` transition into `ReadsClosed | WritesClosed` is observed by a single caller — so there is no double decrement and no double semaphore release (`IsRemote` streams never touch the local stream-count semaphore). Build + 218 Slic tests pass.

bernardnormier (approved):

PR #4766, opened 2026-06-12
Cache canceled read results in SlicPipeReader

pepone:
Fixes icerpc/icerpc-csharp-audit#30.

`ReadAsync` and `TryRead` returned canceled read results without caching them in `_readResult`. `AdvanceTo` computes the consumed and examined offsets against `_readResult.Buffer`, so advancing a canceled read with positions from the returned buffer computed the offsets against a stale buffer — or the default read result when the first read is canceled: `GetOffset` either throws `ArgumentOutOfRangeException` or produces garbage offsets that corrupt the window accounting (`_examined`/`_windowSize`) fed to the StreamWindowUpdate frames. The repository's own conformance tests advance canceled results, but only in arrangements where the stale offsets happen to be benign, which is why this stayed latent.

Canceled read results are now cached like regular ones, so `AdvanceTo` operates on the buffer the application actually received.

### Tests

- `Canceled_read_with_buffered_data_then_resume_reading` (`ReadAsync` and `TryRead` variants): cancels a read that returns buffered data, advances with positions from the returned buffer (examine all, consume nothing), and verifies reading resumes correctly through the end of the stream. **Both variants fail without the fix** with `ArgumentOutOfRangeException` from the stale-buffer offset computation.
- `Canceled_read_advance_does_not_corrupt_window_accounting`: wire-level test (raw duplex client) pinning the flow-control invariant that the sum of `StreamWindowUpdate` increments sent to the peer never exceeds the bytes written on the stream, across a cancel/advance/resume sequence. This one is invariant-pinning rather than regression-detecting: with the unfixed code its particular advance positions happen to produce benign offsets.
- The `ReadFrameAsync` test helper copied and consumed the **whole** buffered sequence into a frame-sized array, which throws (and would drop trailing frames) when several frames are buffered back to back — single-frame arrangements never hit this. It now slices exactly one frame per call, which the wire-level test relies on to read consecutive window updates.

Full `IceRpc.Tests` suite passes (990 passed / 0 failed), `dotnet build` and `dotnet format --verify-no-changes` clean.

### DIFF
--- a/src/IceRpc/Transports/Slic/Internal/SlicPipeReader.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicPipeReader.cs
@@ -222,26 +222,19 @@ internal class SlicPipeReader : PipeReader
                 }
             }
             // else: the application called CancelPendingRead, return the canceled read result as-is.
-
-            // Cache the read result returned to the application: AdvanceTo computes the consumed and examined offsets
-            // against the cached buffer. Without this, AdvanceTo called with positions from the returned buffer would
-            // compute the offsets against a stale buffer and corrupt the window accounting.
-            _readResult = result;
         }
-        else
+        else if (result.IsCompleted)
         {
-            // Cache the read result for the implementation of AdvanceTo that needs to figure out how much data got
-            // examined and consumed.
-            _readResult = result;
-
             // All the data from the peer is considered read at this point. It's time to close reads on the stream. This
             // will write the StreamReadsClosed frame to the peer and allow it to release the stream semaphore.
-            if (result.IsCompleted)
-            {
-                _stream.CloseReads(graceful: true);
-            }
+            _stream.CloseReads(graceful: true);
         }
 
+        // Cache the read result returned to the application: AdvanceTo computes the consumed and examined offsets
+        // against the cached buffer. Without this, AdvanceTo called with positions from the returned buffer would
+        // compute the offsets against a stale buffer and corrupt the window accounting. The throw above is the only
+        // path that skips this cache, and it hands the caller an exception rather than a result to advance.
+        _readResult = result;
         return result;
     }
 
--- a/src/IceRpc/Transports/Slic/Internal/SlicPipeReader.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicPipeReader.cs
@@ -77,7 +77,13 @@ internal class SlicPipeReader : PipeReader
         ReadResult result = await _pipe.Reader.ReadAsync(cancellationToken).ConfigureAwait(false);
         if (result.IsCanceled)
         {
-            return GetReadResult(result);
+            result = GetReadResult(result);
+
+            // Cache the read result returned to the application: AdvanceTo computes the consumed and examined offsets
+            // against the cached buffer. Without this, AdvanceTo called with positions from the returned buffer would
+            // compute the offsets against a stale buffer and corrupt the window accounting.
+            _readResult = result;
+            return result;
         }
 
         // Cache the read result for the implementation of AdvanceTo that needs to figure out how much data got examined
@@ -108,6 +114,11 @@ internal class SlicPipeReader : PipeReader
             if (result.IsCanceled)
             {
                 result = GetReadResult(result);
+
+                // Cache the read result returned to the application: AdvanceTo computes the consumed and examined
+                // offsets against the cached buffer. Without this, AdvanceTo called with positions from the returned
+                // buffer would compute the offsets against a stale buffer and corrupt the window accounting.
+                _readResult = result;
                 return true;
             }
 
--- a/src/IceRpc/Transports/Slic/Internal/SlicPipeReader.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicPipeReader.cs
@@ -2,7 +2,6 @@
 
 using IceRpc.Internal;
 using IceRpc.Transports.Internal;
-using System.Diagnostics;
 using System.IO.Pipelines;
 
 namespace IceRpc.Transports.Slic.Internal;
@@ -74,30 +73,7 @@ internal class SlicPipeReader : PipeReader
             _stream.ThrowIfConnectionClosed();
         }
 
-        ReadResult result = await _pipe.Reader.ReadAsync(cancellationToken).ConfigureAwait(false);
-        if (result.IsCanceled)
-        {
-            result = GetReadResult(result);
-
-            // Cache the read result returned to the application: AdvanceTo computes the consumed and examined offsets
-            // against the cached buffer. Without this, AdvanceTo called with positions from the returned buffer would
-            // compute the offsets against a stale buffer and corrupt the window accounting.
-            _readResult = result;
-            return result;
-        }
-
-        // Cache the read result for the implementation of AdvanceTo that needs to figure out how much data got examined
-        // and consumed.
-        _readResult = result;
-
-        // All the data from the peer is considered read at this point. It's time to close reads on the stream. This
-        // will write the StreamReadsClosed frame to the peer and allow it to release the stream semaphore.
-        if (result.IsCompleted)
-        {
-            _stream.CloseReads(graceful: true);
-        }
-
-        return result;
+        return ProcessReadResult(await _pipe.Reader.ReadAsync(cancellationToken).ConfigureAwait(false));
     }
 
     public override bool TryRead(out ReadResult result)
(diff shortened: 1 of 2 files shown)


src/IceRpc/Transports/Slic/Internal/SlicPipeReader.cs
```diff
@@ -77,7 +77,13 @@ public override async ValueTask<ReadResult> ReadAsync(CancellationToken cancella
         ReadResult result = await _pipe.Reader.ReadAsync(cancellationToken).ConfigureAwait(false);
         if (result.IsCanceled)
         {
-            return GetReadResult(result);
+            result = GetReadResult(result);
```
bernardnormier: Follow-up cleanup (heads-up: not a one-click suggestion — it touches `ReadAsync`, `TryRead`, and `GetReadResult`, most of which is outside this PR's diff hunks, so it needs to be applied by hand).

Now that the canceled path caches `_readResult` too, the only per-method difference left is the `GetReadResult` transform vs. the `IsCompleted`→`CloseReads` call — caching and return are identical and now duplicated across both branches of both methods. Folding that into one helper (and inlining `GetReadResult`, which then has no other caller) removes all of it:

```csharp
private ReadResult ProcessReadResult(ReadResult result)
{
    if (result.IsCanceled)
    {
        // _pipe.Reader returns a canceled result either because the application called CancelPendingRead or because
        // the connection was closed while data was being written to _pipe.Writer.
        if (_state.HasFlag(State.PipeWriterCompleted))
        {
            // The connection was closed while the pipe writer was in use: throw the completion exception, or report
            // a graceful completion when there is none.
            if (_exception is not null)
            {
                throw ExceptionUtil.Throw(_exception);
            }
            result = new ReadResult(result.Buffer, isCanceled: false, isCompleted: true);
        }
        // else: the application called CancelPendingRead; return the canceled result as-is.
    }
    else if (result.IsCompleted)
    {
        // All the data from the peer is considered read. Close reads on the stream: this writes the StreamReadsClosed
        // frame to the peer and lets it release the stream semaphore.
        _stream.CloseReads(graceful: true);
    }

    // Cache the result handed to the application: AdvanceTo computes consumed/examined offsets against this buffer.
    _readResult = result;
    return result;
}
```

`ReadAsync` / `TryRead` collapse to:

```csharp
public override async ValueTask<ReadResult> ReadAsync(CancellationToken cancellationToken = default)
{
    ThrowIfCompleted();
    if (_exception is not null)
    {
        _stream.ThrowIfConnectionClosed();
    }
    return ProcessReadResult(await _pipe.Reader.ReadAsync(cancellationToken).ConfigureAwait(false));
}

public override bool TryRead(out ReadResult result)
{
    ThrowIfCompleted();
    if (_exception is not null)
    {
        _stream.ThrowIfConnectionClosed();
    }
    if (_pipe.Reader.TryRead(out result))
    {
        result = ProcessReadResult(result);
        return true;
    }
    return false;
}
```

Two notes:
- `GetReadResult` disappears, and `Debug.Assert` was its only use of `System.Diagnostics` — drop that `using` (line 5) so `dotnet format` stays clean.
- Behavior is unchanged, including the `else if`: a canceled result that is also `IsCompleted` does **not** call `CloseReads` (it matches today's early-return on the canceled path); the graceful close still fires on the next non-canceled read, or via `Complete()` → `CloseReads(graceful: false)`.

pepone: 🤖 **Claude (Opus 4.8):** Applied in 21d6cc2. Extracted `ProcessReadResult(ReadResult)`, collapsed `ReadAsync`/`TryRead` to call it, inlined+removed `GetReadResult` (no other caller), and dropped the now-unused `using System.Diagnostics;` (the `Debug.Assert` was its only use).

One deviation from the snippet: the helper keeps the non-canceled completed path as `_readResult = result;` **before** `_stream.CloseReads(graceful: true)`, matching the original ordering exactly (the snippet cached after `CloseReads`). This is the one observable side-effect difference if `CloseReads` ever threw, so I preserved the original order while still folding the duplication. Behavior is otherwise unchanged, including the `else if`: a canceled+`IsCompleted` result still does not call `CloseReads`. Build + 209 Slic tests pass.

pepone: 🤖 **Claude (Opus 4.8):** Follow-up in `ee32e39d`: hoisted the caching out to a single `_readResult = result;` just before the `return` — the `throw` (PipeWriterCompleted + exception) is the only path that skips it, and it hands the caller an exception rather than a result to advance. This is exactly your original snippet structure; I went back to it after confirming the cache-vs-`CloseReads` ordering I'd flagged is immaterial here: `_readResult` is private to `SlicPipeReader` so `CloseReads` can't read it, and `AdvanceTo` (the only consumer) can't run until `ProcessReadResult` returns. Build + 209 Slic tests pass.

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
```diff
@@ -1619,6 +1619,156 @@ public async Task Stream_write_cancellation_while_blocked_on_peer_credit()
             Throws.InstanceOf<OperationCanceledException>());
     }
 
+    /// <summary>Verifies that a read returning a canceled result with buffered data can be advanced with positions
+    /// from the returned buffer, and that reading resumes correctly afterwards. See
+    /// icerpc/icerpc-csharp-audit#30.</summary>
+    [Test]
```
pepone, line 1624:
```suggestion
    /// <summary>Verifies that a read returning a canceled result with buffered data can be advanced with positions
    /// from the returned buffer, and that reading resumes correctly afterwards.</summary>
    [Test]
```

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
```diff
@@ -1619,6 +1619,156 @@ public async Task Stream_write_cancellation_while_blocked_on_peer_credit()
...
+        readResult = await streams.Remote.Input.ReadAsync(default).AsTask().WaitAsync(TimeSpan.FromMinutes(2));
+        Assert.That(readResult.IsCanceled, Is.False);
+        Assert.That(readResult.IsCompleted, Is.True);
+        Assert.That(readResult.Buffer, Has.Length.EqualTo(1025));
+        streams.Remote.Input.AdvanceTo(readResult.Buffer.End);
+    }
+
+    /// <summary>Verifies that AdvanceTo following a canceled read computes the window accounting against the buffer
+    /// returned by the canceled read: the total of the StreamWindowUpdate increments sent to the peer never exceeds
+    /// the number of bytes the peer wrote on the stream. See icerpc/icerpc-csharp-audit#30.</summary>
```
pepone:
```suggestion
    /// the number of bytes the peer wrote on the stream.</summary>
```

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
```diff
@@ -1619,6 +1619,157 @@ public async Task Stream_write_cancellation_while_blocked_on_peer_credit()
...
```
pepone:
```suggestion
    /// the number of bytes the peer wrote on the stream.</summary>
```

src/IceRpc/Transports/Slic/Internal/SlicPipeReader.cs
```diff
@@ -231,33 +198,44 @@ await _stream.FillBufferWriterAsync(
...
         {
-            // The application called CancelPendingRead, return the read result as-is.
-            return readResult;
+            // All the data from the peer is considered read at this point. It's time to close reads on the stream. This
+            // will write the StreamReadsClosed frame to the peer and allow it to release the stream semaphore.
+            _stream.CloseReads(graceful: true);
         }
+
+        // Cache the read result returned to the application: AdvanceTo computes the consumed and examined offsets
+        // against the cached buffer. Without this, AdvanceTo called with positions from the returned buffer would
```
bernardnormier: I find this comment too verbose. I would drop the whole "Without this, .... The throws above." - you can read the code.

bernardnormier (approved):

src/IceRpc/Transports/Slic/Internal/SlicPipeReader.cs
pepone: 🤖 **Claude (Opus 4.8):** Done in `9c5421be` — trimmed it down to just the first sentence:

```csharp
// Cache the read result returned to the application: AdvanceTo computes the consumed and examined offsets
// against the cached buffer.
```

Issue #32, opened 2019-11-04
Remove Value factories, protected keyword and postMarshal/perUnmarshal callbacks


bernardnormier:
The goal is to make classes and class instances simpler.

Affects all language mappings:
- [ ] C++
- [x] C#
- [ ] Java
- [ ] JavaScript/TypeScript
- [ ] MATLAB
- [ ] Python
- [ ] Swift

PR #4767, opened 2026-06-12
Fix two ice protocol connection bugs

pepone:
Fixes icerpc/icerpc-csharp-audit#19.
Fixes icerpc/icerpc-csharp-audit#32.

Two small fixes in `IceProtocolConnection`:

- **Missing lower-bound check on the ice frame size (audit #19):** `ReadFramesAsync` validated only `FrameSize > _maxFrameSize`. A Request/Reply prologue with `FrameSize` smaller than the prologue size (14) made `frameSize - PrologueSize` negative, which threw `ArgumentOutOfRangeException` from `ReadOnlySequence.Slice` and reached the `Debug.Fail` in the generic exception handler — an assertion failure on untrusted input in debug builds. The lower bound is now rejected as `InvalidDataException` (malformed peer input), like the upper bound.

- **Reply frame reader leak on a cancellation race (audit #32):** when the read frames loop completes `responseCompletionSource` with the reply frame reader (transferring ownership — the loop no longer completes it) and the invocation's `WaitAsync(invocationCts.Token)` observes cancellation in that same window, the reader was discarded without `Complete()`: the frame reader pipe's rented `MemoryPool<byte>` segments were never returned. The invocation's `finally` now completes the reader when the completion source holds a reader that was never retrieved (`!responseCreated && frameReader is null`). This also covers the read-loop-wins micro-race against the `finally`'s own `TrySetResult`.

### Tests

- `Receiving_a_frame_smaller_than_the_prologue_size_aborts_the_connection`: a server-side duplex read decorator rewrites the incoming request's frame size to 0; the connection aborts with `ConnectionAborted`. Without the fix this test fails with the exact failure mode from the audit: `ArgumentOutOfRangeException` from `Slice` reaching `Debug.Fail` in `ReadFramesAsync`.
- The #32 race window (response landing in the completion source in the same instant the invocation observes cancellation) cannot be reached deterministically from a test — the fix is in the `finally`'s already-completed branch and is covered by review plus the existing suite (all response/cancellation paths still pass).

Full `IceRpc.Tests` suite passes (988 passed / 0 failed), `dotnet build` and `dotnet format --verify-no-changes` clean.

### DIFF
--- a/src/IceRpc/Internal/IceProtocolConnection.cs
+++ b/src/IceRpc/Internal/IceProtocolConnection.cs
@@ -419,8 +419,7 @@ internal sealed class IceProtocolConnection : IProtocolConnection
                     Payload = frameReader
                 };
 
-                frameReader = null; // response now owns frameReader
-                responseCreated = true;
+                responseCreated = true; // the response now owns frameReader
                 return response;
             }
             catch (OperationCanceledException)
@@ -442,15 +441,9 @@ internal sealed class IceProtocolConnection : IProtocolConnection
                 {
                     try
                     {
-                        PipeReader reader = await responseCompletionSource.Task.ConfigureAwait(false);
-
-                        // The read frames loop set the response frame reader (and transferred its ownership) but this
-                        // invocation was canceled before retrieving it: complete the reader here, no other code
-                        // completes it.
-                        if (!responseCreated && frameReader is null)
-                        {
-                            reader.Complete();
-                        }
+                        // Retrieve (or re-retrieve) the response PipeReader. The cleanup at the end of this finally
+                        // completes it unless a response was created, in which case the response owns it.
+                        frameReader = await responseCompletionSource.Task.ConfigureAwait(false);
                     }
                     catch
                     {
@@ -469,7 +462,11 @@ internal sealed class IceProtocolConnection : IProtocolConnection
                     DecrementDispatchInvocationCount();
                 }
 
-                frameReader?.Complete();
+                if (!responseCreated)
+                {
+                    frameReader?.Complete();
+                }
+                // else the response owns the PipeReader
             }
         }
     }
--- a/src/IceRpc/Internal/IceProtocolConnection.cs
+++ b/src/IceRpc/Internal/IceProtocolConnection.cs
@@ -464,8 +464,7 @@ internal sealed class IceProtocolConnection : IProtocolConnection
 
                 if (!responseCreated)
                 {
-                    Debug.Assert(frameReader is not null);
-                    frameReader.Complete();
+                    frameReader?.Complete();
                 }
                 // else the response owns the PipeReader
             }
--- a/src/IceRpc/Internal/IceProtocolConnection.cs
+++ b/src/IceRpc/Internal/IceProtocolConnection.cs
@@ -315,6 +315,7 @@ internal sealed class IceProtocolConnection : IProtocolConnection
                 CancellationTokenSource.CreateLinkedTokenSource(_disposedCts.Token, cancellationToken);
 
             PipeReader? frameReader = null;
+            bool responseCreated = false;
             TaskCompletionSource<PipeReader>? responseCompletionSource = null;
             int requestId = 0;
 
@@ -419,6 +420,7 @@ internal sealed class IceProtocolConnection : IProtocolConnection
                 };
 
                 frameReader = null; // response now owns frameReader
+                responseCreated = true;
                 return response;
             }
             catch (OperationCanceledException)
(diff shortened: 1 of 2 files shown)


src/IceRpc/Internal/IceProtocolConnection.cs
```diff
@@ -440,7 +442,15 @@ await _duplexConnectionWriter.WriteAsync(payloadBuffer, _disposedCts.Token)
...
-                        _ = await responseCompletionSource.Task.ConfigureAwait(false);
+                        PipeReader reader = await responseCompletionSource.Task.ConfigureAwait(false);
+
+                        // The read frames loop set the response frame reader (and transferred its ownership) but this
+                        // invocation was canceled before retrieving it: complete the reader here, no other code
+                        // completes it.
+                        if (!responseCreated && frameReader is null)
+                        {
+                            reader.Complete();
+                        }
```
bernardnormier: Replacing my earlier `frameReader ??= reader` comment with a fuller simplification. Now that `responseCreated` exists, it's the single source of truth for "the response owns the reader" — so we no longer need to null out `frameReader`, and this whole branch collapses to a plain (re-)retrieve. Three edits that go together (not split into separate suggestions on purpose: applying any one alone double-completes or leaks):

**1. Don't null `frameReader` after creating the response (~L422):**
```csharp
                // frameReader = null;  <- remove this line
                responseCreated = true; // response now owns frameReader
```

**2. This `finally` branch becomes just a (re-)retrieve:**
```csharp
                        // Retrieve (or re-retrieve) the response PipeReader. The cleanup at the end of this finally
                        // completes it unless a response was created, in which case the response owns it.
                        frameReader = await responseCompletionSource.Task.ConfigureAwait(false);
```

(the above remains in the try/catch to observe/ignore any exception)

**3. Guard the final completion with the flag (~L472):**
```csharp
                if (!responseCreated)
                {
                    frameReader?.Complete();
                }
                // else the response owns the PipeReader
```

Walks correctly through every state: success (re-retrieves the response's reader, guard skips it), cancel/orphan race (frameReader gets the unclaimed reader, guard completes it), retrieved-then-threw (re-fetches the same object, completed once), normal cancel / no reply (frameReader stays null, no-op), faulted source (await throws, observed by the `catch`).

pepone: 🤖 **Claude (Opus 4.8):** Applied all three edits together in 94fcbf3. `responseCreated` is now the single source of truth for ownership: removed the `frameReader = null` after creating the response, collapsed the inner `finally` branch to a plain (re-)retrieve `frameReader = await responseCompletionSource.Task`, and guarded the final completion with `if (!responseCreated) frameReader?.Complete();`. I traced all paths (success, cancel/orphan race, retrieved-then-threw, normal cancel/no reply, faulted source) — the reader is transferred to the response or completed exactly once, never leaked or double-completed. Build passes and all 49 IceProtocolConnection tests are green. A codex cross-check independently confirmed the same path analysis.

tests/IceRpc.Tests/IceProtocolConnectionTests.cs
```diff
@@ -868,6 +868,56 @@ public async Task Receiving_a_frame_larger_than_max_ice_frame_size_aborts_the_co
             Throws.InstanceOf<IceRpcException>().With.Property("IceRpcError").EqualTo(IceRpcError.ConnectionAborted));
     }
 
+    /// <summary>Verifies that an ice frame with a frame size smaller than the prologue size is rejected as invalid
+    /// data, aborting the connection. See icerpc/icerpc-csharp-audit#19.</summary>
+    [Test]
+    public async Task Receiving_a_frame_smaller_than_the_prologue_size_aborts_the_connection()
```
pepone, line 874:
```suggestion
    /// <summary>Verifies that an ice frame with a frame size smaller than the prologue size is rejected as invalid
    /// data, aborting the connection.</summary>
    [Test]
    public async Task Receiving_a_frame_smaller_than_the_prologue_size_aborts_the_connection()
```

src/IceRpc/Internal/IceProtocolConnection.cs
```diff
@@ -459,7 +462,11 @@ await _duplexConnectionWriter.WriteAsync(payloadBuffer, _disposedCts.Token)
                     DecrementDispatchInvocationCount();
                 }
 
-                frameReader?.Complete();
+                if (!responseCreated)
+                {
+                    frameReader?.Complete();
+                }
```
pepone, line 468:
```suggestion
                if (!responseCreated)
                {
                    Debug.Assert(frameReader is not null);
                    frameReader.Complete();
                }
```

bernardnormier: This is not correct. The frameReader can be null when responseCreated is false. For example, a oneway request (we're in the finally block.)

src/IceRpc/Internal/IceProtocolConnection.cs
```diff
@@ -459,7 +462,12 @@ await _duplexConnectionWriter.WriteAsync(payloadBuffer, _disposedCts.Token)
                     DecrementDispatchInvocationCount();
                 }
 
-                frameReader?.Complete();
+                if (!responseCreated)
+                {
+                    Debug.Assert(frameReader is not null);
+                    frameReader.Complete();
```
pepone:
```suggestion
                    frameReader?.Complete();
```

bernardnormier (approved):

PR #4769, opened 2026-06-12
Guard SslAuthenticationOptions ShallowClone against new .NET properties

pepone:
Fixes icerpc/icerpc-csharp-audit#20.

`SslAuthenticationOptionsExtensions.ShallowClone` copies each `SslClientAuthenticationOptions` / `SslServerAuthenticationOptions` property by hand and sits directly on the security-configuration path (`TcpClientTransport`, `QuicClientTransport`, `Server`). The list is complete against the current .NET property set, but a security-relevant property added by a future .NET release and set by the user would be silently discarded before the handshake — a silent weakening of the TLS configuration on a runtime upgrade.

Per the audit recommendation, this adds:

- Two reflection tests (`SslAuthenticationOptionsExtensionsTests`) asserting the settable public property set of each options class matches the list `ShallowClone` copies — they fail on the runtime upgrade that adds (or removes) a property, prompting the update.
- A `<remarks>` on the extension class explaining the constraint and pointing to the tests.

No behavior change; both tests pass against the current .NET 10 property set.

### Limitation

The tests run against the runtime this repository targets, so they catch the drift when IceRPC bumps its `TargetFramework` — not before. .NET servicing updates (10.0.x) don't add public API, so the property set is stable within .NET 10; but an application targeting a newer .NET major can run the `net10.0`-built IceRpc package on that newer runtime, where `Ssl*AuthenticationOptions` has new properties. If such an application sets a new property, the net10-compiled `ShallowClone` still drops it and these tests can't detect that. The exposure window is "new .NET major released → IceRPC ships a TFM-bumped package", and only for users who set a brand-new TLS property in that window.

### Alternatives considered

1. **Hand-copy + tripwire tests (this PR):** zero runtime cost, trimming/AOT-friendly; accepts the limitation above.
2. **Reflection-based copy in `ShallowClone`:** enumerate the settable public properties at runtime and copy them all. Future-proof across runtime upgrades with no recompile; costs a little reflection per connection creation (negligible next to a TLS handshake) and would break if assembly trimming is ever enabled (it isn't today).
3. **Eliminate the clone:** the call sites clone only to set `TargetHost` / `ApplicationProtocols` without mutating the user's instance; restructuring them to not need a copy removes the problem entirely, at the cost of a larger change.

### DIFF
--- a/src/IceRpc/Transports/Internal/SslAuthenticationOptionsExtensions.cs
+++ b/src/IceRpc/Transports/Internal/SslAuthenticationOptionsExtensions.cs
@@ -7,9 +7,9 @@ namespace IceRpc.Transports.Internal;
 /// <summary>Provides extension methods for <see cref="SslClientAuthenticationOptions" /> and <see
 /// cref="SslServerAuthenticationOptions" />.</summary>
 /// <remarks>The clone methods copy each property by hand and sit on the security-configuration path: a property
-/// added by a future .NET release and not copied here would be silently dropped from the user's TLS configuration.
-/// The SslAuthenticationOptionsExtensionsTests reflection tests fail when the .NET property set changes; when they
-/// do, add the new property here and to the test's property list.</remarks>
+/// added by a future .NET release is not copied and is silently set to its initial value in the user's
+/// authentication options. The SslAuthenticationOptionsExtensionsTests reflection tests fail when the .NET
+/// property set changes; when they do, add the new property here and to the test's property list.</remarks>
 internal static class SslAuthenticationOptionsExtensions
 {
     /// <summary>Makes a shallow copy of an SSL client authentication options.</summary>
--- a/src/IceRpc/Transports/Internal/SslAuthenticationOptionsExtensions.cs
+++ b/src/IceRpc/Transports/Internal/SslAuthenticationOptionsExtensions.cs
@@ -6,6 +6,10 @@ namespace IceRpc.Transports.Internal;
 
 /// <summary>Provides extension methods for <see cref="SslClientAuthenticationOptions" /> and <see
 /// cref="SslServerAuthenticationOptions" />.</summary>
+/// <remarks>The clone methods copy each property by hand and sit on the security-configuration path: a property
+/// added by a future .NET release and not copied here would be silently dropped from the user's TLS configuration.
+/// The SslAuthenticationOptionsExtensionsTests reflection tests fail when the .NET property set changes; when they
+/// do, add the new property here and to the test's property list.</remarks>
 internal static class SslAuthenticationOptionsExtensions
 {
     /// <summary>Makes a shallow copy of an SSL client authentication options.</summary>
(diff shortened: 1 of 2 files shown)


tests/IceRpc.Tests/Transports/SslAuthenticationOptionsExtensionsTests.cs
```diff
@@ -0,0 +1,68 @@
...
+        Assert.That(GetSettablePropertyNames(typeof(SslClientAuthenticationOptions)), Is.EquivalentTo(_clientProperties));
+
+    [Test]
+    public void ShallowClone_copies_all_ssl_server_authentication_options_properties() =>
+        Assert.That(GetSettablePropertyNames(typeof(SslServerAuthenticationOptions)), Is.EquivalentTo(_serverProperties));
+
+    private static IEnumerable<string> GetSettablePropertyNames(Type type) =>
+        type.GetProperties(BindingFlags.Public | BindingFlags.Instance)
+            .Where(property => property.CanWrite)
+            .Select(property => property.Name);
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Correct — applied. The guard now filters on `property.SetMethod?.IsPublic is true`, so a future public-getter/non-public-setter property (which neither `ShallowClone` nor the application could set) won't trip it.

src/IceRpc/Transports/Internal/SslAuthenticationOptionsExtensions.cs
```diff
@@ -6,6 +6,10 @@ namespace IceRpc.Transports.Internal;
 
 /// <summary>Provides extension methods for <see cref="SslClientAuthenticationOptions" /> and <see
 /// cref="SslServerAuthenticationOptions" />.</summary>
+/// <remarks>The clone methods copy each property by hand and sit on the security-configuration path: a property
+/// added by a future .NET release and not copied here would be silently dropped from the user's TLS configuration.
```
bernardnormier: I think it should be:
"a property added by a future .NET release is not copied and is silently set to its initial value in the user's authentication options"

tests/IceRpc.Tests/Transports/SslAuthenticationOptionsExtensionsTests.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using NUnit.Framework;
+using System.Net.Security;
+using System.Reflection;
+
+namespace IceRpc.Tests.Transports;
+
+/// <summary>Guards SslAuthenticationOptionsExtensions.ShallowClone against new properties added to the .NET
+/// Ssl*AuthenticationOptions classes: ShallowClone copies each property by hand, so a property added by a future
+/// .NET release would be silently dropped from the user's TLS configuration. When one of these tests fails, add the
```
bernardnormier: I think "is silently reset in the authentication options" would be more correct. No would and no drop.

bernardnormier (approved):

PR #4836, opened 2026-08-19
Reject encoding a fully-sliced class instance with the Compact class format

bernardnormier:
When a class instance decoded with an activator that knows none of its type IDs (a fully-sliced `UnknownIceClass`) is re-encoded using the Compact class format — the default — `EncodeInstance` writes the class instance marker and then nothing: the preserved slices are only re-encoded in Sliced format, and `UnknownIceClass.EncodeCore` is empty. The peer then misparses whatever bytes follow as the instance's first slice header, typically failing with `InvalidDataException: Cannot skip a class slice with no type ID`, and the corruption desynchronizes the rest of the encapsulation.

This PR makes the encoder reject this locally instead of emitting a corrupt payload:

- `EncodeInstance` now throws `NotSupportedException` when encoding an `UnknownIceClass` with a class format other than Sliced. Partially-known instances are unaffected: they still drop their preserved slices in Compact format and encode their known slices, as intended.
- The now-unreachable throw in `EncodeUnknownSlices` (its only call site requires the Sliced format) is replaced by a `Debug.Assert`, and the stale comment above it is corrected.
- Adds a regression test.

## What's Changed entry

Area: Ice codec

- Encoding a fully-sliced class instance with the Compact class format now fails with `NotSupportedException` instead of producing a corrupt payload.

### DIFF
--- a/src/IceRpc/Ice/Codec/IceEncoder.Class.cs
+++ b/src/IceRpc/Ice/Codec/IceEncoder.Class.cs
@@ -166,10 +166,18 @@ public ref partial struct IceEncoder
             _classContext.Current.InstanceType = InstanceType.Class;
             _classContext.Current.FirstSlice = true;
 
-            if (v.UnknownSlices.Count > 0 && _classContext.ClassFormat == ClassFormat.Sliced)
+            if (v.UnknownSlices.Count > 0)
             {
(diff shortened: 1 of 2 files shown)


externl (approved):
Looks good.

InsertCreativityHere (approved):
Looks good to me!

PR #4837, opened 2026-08-19
Reject invalid string type IDs when decoding Ice class slices

bernardnormier:
The Ice decoder stored a peer-supplied string type ID verbatim, without validation. This matters for preserved slices: `SliceInfo.TypeId` holds either a type ID or a compact ID without recording which one, and re-encoding a preserved slice parses `TypeId` as a compact ID unless it starts with `::`. A peer could therefore send a string type ID such as `"9999999999"` or `"-5"`; re-encoding the preserved slices then escaped as `OverflowException` (from `int.Parse`) or `ArgumentException` (from `EncodeSize`) instead of `InvalidDataException` — surfacing as `StatusCode.InternalError` for what is actually bad peer data.

This PR validates at the trust boundary: `DecodeTypeId` now throws `InvalidDataException` when a string type ID does not start with `::`. A valid Ice type ID always starts with `::`, and slices with compact IDs use their own type-ID kind on the wire — so after this check, a preserved slice's `TypeId` can always be re-encoded.

On the encoder side, the `catch (FormatException)` around the compact-ID parse is now unreachable — `SliceInfo` is only created by the decoder — so it's removed rather than left to mislabel a hypothetical internal bug as invalid peer data.

Adds a regression test.

## What's Changed entry

None — only changes how invalid data received from a peer is reported.

### DIFF
--- a/src/IceRpc/Ice/Codec/IceDecoder.Class.cs
+++ b/src/IceRpc/Ice/Codec/IceDecoder.Class.cs
@@ -380,6 +380,14 @@ public ref partial struct IceDecoder
             case TypeIdKind.String:
                 string typeId = DecodeString();
 
+                // A valid type ID always starts with "::". This validation is essential for preserved slices:
+                // SliceInfo.TypeId holds either a type ID or a compact ID without recording which one, and the
+                // re-encoding of a preserved slice parses TypeId as a compact ID unless it starts with "::".
+                if (!typeId.StartsWith("::", StringComparison.Ordinal))
+                {
+                    throw new InvalidDataException($"Received invalid type ID {typeId}.");
+                }
+
                 // The typeIds of slices in indirection tables can be decoded several times: when we skip the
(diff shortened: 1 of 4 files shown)


externl (approved):
Looks good.

externl:
The `int.Parse` justification in `IceEncoder.Class.cs` reaches into decoder behavior to explain an encoder invariant. `SliceInfo.TypeId`'s doc comment ("Gets the Ice type ID or compact ID for this slice") is where that belongs — say it's either `::`-prefixed or a decimal compact ID, and both sides can rely on it without the encoder narrating the decoder.

InsertCreativityHere (approved):
Looks good to me!

bernardnormier:
Good call — moved it to `SliceInfo.TypeId`, which now documents that a type ID starts with `::` while a compact ID is a non-negative integer in decimal notation. The encoder comment is back to the original one-liner, and the decoder just points at the doc for the contract. 6406f7561

PR #4838, opened 2026-08-19
Add PR template with the What's Changed entry convention

bernardnormier:
Each release's GitHub release notes carry a hand-written "What's Changed" section (see the v0.5.2 release). Writing it by revisiting every merged PR at release time doesn't scale, so this PR establishes a convention: every PR description ends with a `## What's Changed entry` section that records, at fix time, either the release-notes entry (with its area) or an explicit `None — <reason>`. At release time, the entries are collected from the milestone's merged PRs and only need a light editorial pass.

- `.github/pull_request_template.md` (new) carries the section skeleton and the rules in an HTML comment: entries describe observable behavior from the user's perspective, and fixes that only change the handling of invalid or malicious peer data get no entry.
- `AGENTS.md` gets a "Pull requests" section requiring the section and pointing at the template, since the template is not injected automatically for PRs created from the command line.

See #4836 and #4837 for the first two PRs using the convention.

## What's Changed entry

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,23 @@
+<!-- Describe the change and why it's needed. For a bug fix, describe the bug: when it occurs and what a user
+observes. -->
+
+## What's Changed entry
+
+<!-- This section feeds the "What's Changed" section of the GitHub release notes; it's collected from merged PRs at
+release time. Keep it, in one of two forms.
+
+For a user-facing change, specify the area (usually an area label name, such as "Core", "Slic" or "Ice codec") and
+write one bullet per user-visible change:
+
+Area: <area>
+
+- <entry>
+
+Write each entry from the user's perspective: the observable symptom or new behavior, in the documentation's
+vocabulary — not the internal mechanism. A fix that only changes how invalid or malicious data received from a peer
+is handled is not user-facing.
+
+For a change with no user-facing impact, explain briefly why:
+
+None — <short reason>.
+-->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,20 +4,27 @@ observes. -->
 ## What's Changed entry
 
 <!-- This section feeds the "What's Changed" section of the GitHub release notes; it's collected from merged PRs at
-release time. Keep it, in one of two forms.
+release time. It should be in one of two forms:
 
-For a user-facing change, specify the area (usually an area label name, such as "Core", "Slic" or "Ice codec") and
+1) For a user-facing change, specify the area (usually an area label name, such as "Core", "Slic" or "Ice codec") and
 write one bullet per user-visible change:
 
+```
 Area: <area>
 
 - <entry>
+```
+
+Most PRs have a single entry. A PR with several user-visible changes gets one bullet each, under one Area block
+per area.
 
 Write each entry from the user's perspective: the observable symptom or new behavior, in the documentation's
 vocabulary — not the internal mechanism. A fix that only changes how invalid or malicious data received from a peer
 is handled is not user-facing.
 
-For a change with no user-facing impact, explain briefly why:
+2) For a change with no user-facing impact, explain briefly why:
 
+```
 None — <short reason>.
+```
 -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,7 +20,7 @@ per area.
 
 Write each entry from the user's perspective: the observable symptom or new behavior, in the documentation's
 vocabulary — not the internal mechanism. A fix that only changes how invalid or malicious data received from a peer
-is handled is not user-facing.
+is handled is not generally user-facing.
 
 2) For a change with no user-facing impact, explain briefly why:
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -36,6 +36,12 @@ These conventions apply to all AI coding assistants (Copilot, Claude Code, etc.)
 - Run all tests: `dotnet test`
 - Run a single test project: `dotnet test tests/<Project>/<Project>.csproj --filter "FullyQualifiedName~<TestClass>"`
 
+## Pull requests
+
+- Every PR description ends with a `## What's Changed entry` section. Read
+  [.github/PULL_REQUEST_TEMPLATE.md](.github/PULL_REQUEST_TEMPLATE.md) for the format and the rules before writing
+  it — the template is not injected automatically when a PR is created from the command line.
+
 ## Dismissed audit patterns
 
 This section captures the reasoning behind `ai-audit` findings that have been closed as "not planned". Before opening a


.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
...
+write one bullet per user-visible change:
+
+Area: <area>
+
+- <entry>
+
+Write each entry from the user's perspective: the observable symptom or new behavior, in the documentation's
+vocabulary — not the internal mechanism. A fix that only changes how invalid or malicious data received from a peer
+is handled is not user-facing.
+
```
externl, line 24: I would just drop the example. I would say that if it can cause a crash and is and an important part of a patch release then it certainly warrants a changelog entry. 

```suggestion
vocabulary — not the internal mechanism.
```

externl (dismissed):

externl:
One note on the exclusion rule: "A fix that only changes how invalid or malicious data received from a peer is handled is not user-facing" rules out exactly what v0.5.2 led with — its first section is *"Hardening against malformed requests and responses"*, ending with *"If you expose an IceRPC server to untrusted clients (or talk to untrusted servers), you should upgrade to this release."*

.github/pull_request_template.md
InsertCreativityHere, line 1: The name of this file seems incorrect. Looks like a mix-up between the file vs directory naming. See section 3:
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

---

For a single template (what we're doing) it should be a file named `pull_request_template.md`.
For multiple templates, it should be a directory named `PULL_REQUEST_TEMPLATE/`.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
+<!-- Describe the change and why it's needed. For a bug fix, describe the bug: when it occurs and what a user
+observes. -->
+
+## What's Changed entry
+
+<!-- This section feeds the "What's Changed" section of the GitHub release notes; it's collected from merged PRs at
+release time. Keep it, in one of two forms.
+
+For a user-facing change, specify the area (usually an area label name, such as "Core", "Slic" or "Ice codec") and
```
InsertCreativityHere: We state that it should be in 1 of 2 forms, but the different forms are not obviously delineated.
Easily improved with:
```suggestion
1) For a user-facing change, specify the area (usually an area label name, such as "Core", "Slic" or "Ice codec") and
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+
+Area: <area>
+
+- <entry>
+
+Write each entry from the user's perspective: the observable symptom or new behavior, in the documentation's
+vocabulary — not the internal mechanism. A fix that only changes how invalid or malicious data received from a peer
+is handled is not user-facing.
+
+For a change with no user-facing impact, explain briefly why:
```
InsertCreativityHere:
```suggestion
2) For a change with no user-facing impact, explain briefly why:
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
...
+<!-- This section feeds the "What's Changed" section of the GitHub release notes; it's collected from merged PRs at
+release time. Keep it, in one of two forms.
+
+For a user-facing change, specify the area (usually an area label name, such as "Core", "Slic" or "Ice codec") and
+write one bullet per user-visible change:
+
+Area: <area>
+
+- <entry>
```
InsertCreativityHere, line 15: Also, could be made cleaner with:
```suggestion
'''
Area: <area>

- <entry>
- 
'''
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
...
+
+Write each entry from the user's perspective: the observable symptom or new behavior, in the documentation's
+vocabulary — not the internal mechanism. A fix that only changes how invalid or malicious data received from a peer
+is handled is not user-facing.
+
+For a change with no user-facing impact, explain briefly why:
+
+None — <short reason>.
```
InsertCreativityHere, line 28:
```suggestion
'''
None — <short reason>.
'''
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
+<!-- Describe the change and why it's needed. For a bug fix, describe the bug: when it occurs and what a user
+observes. -->
+
+## What's Changed entry
+
+<!-- This section feeds the "What's Changed" section of the GitHub release notes; it's collected from merged PRs at
+release time. Keep it, in one of two forms.
```
InsertCreativityHere: I don't understand the current wording. What is there to "keep" here?
I feel like writing a simple declarative instructions is simpler:
```suggestion
release time. It should be in one of two forms:
```

InsertCreativityHere (dismissed):
I like the concept! Much better to be making these notes at the time we make the change.

But I think there are a handful of simple improvements that could be made to the template.

bernardnormier:
I want to keep that rule, so I left the sentence in.

v0.5.2 is not the model I want to follow here. 

On the substance: for a fix whose only observable effect is on invalid or malicious peer data, an entry tells users nothing they can act on, and a release note full of them buries the changes that do affect their code. That is why #4837, #4839 and #4842 all say `None`.

The crash case you raise is real, but I think the vehicle is a security advisory, not a What's Changed bullet: an advisory reaches people who need to upgrade, carries a severity, and does not require the release notes to describe the malformed input. The release notes stay a list of what changed for a user writing code against IceRPC.

bernardnormier:
All applied, thanks — renamed to `pull_request_template.md`, numbered the two forms, and fenced both examples.

Your second `-` was not a typo on your part, it was a gap in mine: the template only showed one bullet in one area, which reads as a limit. A PR can have several entries in one area, and entries in more than one area. The example still shows the common case (most PRs have a single entry), with the rest stated just below it:

```
Most PRs have a single entry. A PR with several user-visible changes gets one bullet each, under one Area block
per area.
```

I kept the sentence about invalid/malicious peer data — see my reply to Joe above for the reasoning.

.github/pull_request_template.md
```diff
@@ -0,0 +1,30 @@
...
+
+- <entry>
+```
+
+Most PRs have a single entry. A PR with several user-visible changes gets one bullet each, under one Area block
+per area.
+
+Write each entry from the user's perspective: the observable symptom or new behavior, in the documentation's
+vocabulary — not the internal mechanism. A fix that only changes how invalid or malicious data received from a peer
+is handled is not user-facing.
```
externl:
```suggestion
is handled is not generally user-facing.
```

externl (approved):

.github/pull_request_template.md
InsertCreativityHere: Technically more correct...

InsertCreativityHere (approved):
Looks good to me!

PR #4839, opened 2026-08-19
Validate boolean values when decoding custom-mapped bool sequences

bernardnormier:
Decoding a plain `Sequence<bool>` validates each element (`checkElement: SliceDecoder.CheckBoolValue`), rejecting bytes that are neither 0 nor 1. But when a `Sequence<bool>` field uses a `cs::type` custom collection, the generated code took the fixed-size memcpy path without the validation callback — so a noncanonical byte like `2` landed directly in the decoded `bool`s. Such a `bool` breaks ordinary boolean reasoning in .NET (`b` and `!b` can both behave unexpectedly), from peer-controlled input.

The generator's custom-collection fixed-size branch now emits the same `CheckBoolValue` validation as the plain path.

Adds tests for the custom-mapped bool sequence (valid and invalid bytes), mirroring the existing plain `Sequence<bool>` tests.

## What's Changed entry

None — only changes how invalid data received from a peer is handled.

### DIFF
--- a/src/ZeroC.Slice.Generator/ITypeExtensions.cs
+++ b/src/ZeroC.Slice.Generator/ITypeExtensions.cs
@@ -94,7 +94,9 @@ internal static class ITypeExtensions
                 {
                     // cs::type on fixed-size primitives: wrap the fixed-size decode in the factory.
                     string csElemType = elemType.ToTypeString(currentNamespace);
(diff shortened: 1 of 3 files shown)


externl (approved):
Looks good.

InsertCreativityHere (approved):
Looks good to me!

PR #4516, opened 2026-04-15
Limit size of Slic control frame bodies during decoding

bernardnormier:
This PR caps the size of the Slic control frame bodies to 16,383 bytes.

PR #4842, opened 2026-08-20
Enforce the control frame body size limit on all non-data Slic frames

bernardnormier:
The Slic control frame body size limit (`MaxControlFrameBodySize`, 16,383 bytes) was applied with the predicate `header.FrameType < FrameType.Stream`, which covers `Initialize` through `Pong` only. Three frame types sort after `Stream` in the enum: `StreamReadsClosed`, `StreamWindowUpdate` and `StreamWritesClosed`. The first and last are harmless — their handler rejects any non-empty body — but `StreamWindowUpdate` is read through `ReadFrameBodyAsync`, which only rejects a size of 0 and then reads the declared number of bytes into a pipe with no pause writer threshold. A peer could therefore declare a body of up to `int.MaxValue` for a frame whose real body is a couple of bytes, and make the receiver buffer that many bytes before decoding.

The limit now applies to every frame type except `Stream` and `StreamLast`, whose size is bounded by the stream's flow control window. This is the exemption set #4516 intended when it introduced the limit.

Extends the existing oversized-body test with a `StreamWindowUpdate` case.

## What's Changed entry

None — internal hardening; no legitimate or accidental control frame is oversized.

### DIFF
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -1495,8 +1495,10 @@ internal class SlicConnection : IMultiplexedConnection
                 throw new InvalidDataException("The frame size can't be larger than int.MaxValue.", exception);
             }
 
-            // Reject oversized control frame bodies before any buffering occurs.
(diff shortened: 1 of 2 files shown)


externl (approved):

InsertCreativityHere (approved):
Looks good to me!

PR #4843, opened 2026-08-20
Abort the response stream when sending the payload continuation fails

### LINKED ISSUE #4789: [Audit-High] Response payload-continuation failure closes the stream gracefully, truncating the response

## Summary

In `SendResponseAsync`, `flushResult` is seeded with `isCanceled: true`, with an explicit comment saying it exists so that a throw from the *first* `CopyFromAsync` yields `CompleteOutput(false)`. It is never re-armed before the *second* copy. Entry into the second copy requires `!flushResult.IsCanceled`, so a throw there leaves the stale **successful** result in place and the `finally` runs `CompleteOutput(success: true)` → `SlicPipeWriter.Complete(null)` → `CloseWrites(graceful: true)` → `FrameType.StreamLast`.

The peer therefore reads a truncated response payload and reports success.

Exceptions out of the second copy are ordinary, not contract violations: `PipeWriterExtensions.CopyFromAsync` deliberately lets `TryRead`/`ReadAsync` exceptions escape, and `AsyncEnumerablePipeReader` propagates whatever the application's `MoveNextAsync` throws — the exact shape generated for server-streaming Slice and Protobuf operations.

The request-side twin uses `bool success = false` outside the `try` and is correct, so this is an asymmetry rather than a deliberate design choice.

## Trigger / impact

Reproduced over Slic/TCP: a dispatcher returns `Payload` = `"HEAD"` and `PayloadContinuation` = a reader yielding `"ABCD"` then throwing. Client output: `status=Ok`, `payload='HEADABCD'`, and **the payload read completes without error**.

The contrast run — the same throwing reader installed as `Payload`, so the sentinel applies — fails correctly with `IceRpcException: … 'TruncatedData'`. The status code is `Ok` in both runs (the header flushes first), so the payload-read outcome is the only client-visible difference.

Codex added two corrections worth carrying into the fix:

- The abortive `CompleteOutput` further down the method is **ineffective**, because the transport writer has already been completed by the `finally`.
- Since the response header is already on the wire, the correct failure the client should observe is transport-level `IceRpcError.TruncatedData` — not `StatusCode.TruncatedPayload`, which is request-side.

## Source references

- [src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1096-L1097](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1096-L1097) — the sentinel and its comment
- [src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1115-L1122](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1115-L1122) — second `CopyFromAsync`, guarded by `!flushResult.IsCanceled`
- [src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1124-L1129](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1124-L1129) — `finally` reading the stale `flushResult`
- [src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1530-L1535](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L1530-L1535) — the request-side twin, which is correct
- [src/IceRpc/Internal/PipeWriterExtensions.cs#L65-L77](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/Internal/PipeWriterExtensions.cs#L65-L77) — `CopyFromAsync` letting reader exceptions escape

## Recommendation

Mirror the request side — `bool payloadWriterSuccess = false;` outside the `try`, set only after the last successful copy, and `payloadWriter.CompleteOutput(payloadWriterSuccess)` in the `finally`. (Re-arming the sentinel before the second copy also works but is easier to regress.)

**[`PayloadContinuation_completed_on_invalid_response_payload_continuation`](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs#L1523) drives this exact path but tolerates the buggy outcome** — its cleanup accepts either a `TruncatedData` failure or a clean completion — and must be tightened alongside the fix.

## Audit context

Finding 3 of [general-correctness-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-07-27.md); Codex verdict **CONFIRMED** at **High** ("silently converts an application failure into a successful, incomplete streaming response, directly violating response integrity").

Severity: **High**.

bernardnormier:
Fixes #4789.

`SendResponseAsync` seeded `flushResult` with `isCanceled: true` so that a throw from the first `CopyFromAsync` would complete the payload writer abortively. It never re-armed this sentinel before copying the payload continuation, so a throw from that second copy left the successful result of the first copy in place: the `finally` completed the payload writer gracefully, the stream ended with a `StreamLast` frame, and the client read the truncated payload as a complete, successful response.

Exceptions from that second copy are ordinary rather than contract violations — `CopyFromAsync` lets reader exceptions escape, and the reader is the application's async enumerable for a server-streaming Slice or Protobuf operation.

This PR mirrors the request side, which was already correct: a `payloadWriterSuccess` flag declared outside the `try` and set only after the last copy completes. Every non-throwing path keeps its current behavior, including the one where the first copy reports the peer is no longer reading.

`PayloadContinuation_completed_on_invalid_response_payload_continuation` drove this path but tolerated either outcome in its cleanup; it now asserts the client always observes `IceRpcError.TruncatedData`, whether the invocation itself fails or the failure surfaces when reading the response payload.

## What's Changed entry

Area: Core

- Fixed a bug where a response payload could be silently truncated: when a failure (such as an error reading
  the payload from the application) interrupted the streaming of this payload, the client received the
  truncated payload as if it was complete.

### DIFF
--- a/src/IceRpc/Internal/IceRpcProtocolConnection.cs
+++ b/src/IceRpc/Internal/IceRpcProtocolConnection.cs
@@ -1093,7 +1093,7 @@ internal sealed class IceRpcProtocolConnection : IProtocolConnection
 
                     PipeWriter payloadWriter = response.GetPayloadWriter(streamOutput);
 
-                    // Set to true only after the payload and its continuation are fully copied.
+                    // Remains false if a copy throws or is canceled.
                     bool payloadWriterSuccess = false;
 
                     try
--- a/src/IceRpc/Internal/IceRpcProtocolConnection.cs
+++ b/src/IceRpc/Internal/IceRpcProtocolConnection.cs
@@ -1093,8 +1093,8 @@ internal sealed class IceRpcProtocolConnection : IProtocolConnection
 
                     PipeWriter payloadWriter = response.GetPayloadWriter(streamOutput);
 
-                    // We give flushResult an initial "failed" value, in case the first CopyFromAsync throws.
-                    var flushResult = new FlushResult(isCanceled: true, isCompleted: false);
+                    // Set to true only after the payload and its continuation are fully copied.
+                    bool payloadWriterSuccess = false;
 
                     try
                     {
@@ -1106,7 +1106,7 @@ internal sealed class IceRpcProtocolConnection : IProtocolConnection
                         // ignored.
                         bool hasContinuation = response.PayloadContinuation is not null;
 
-                        flushResult = await payloadWriter.CopyFromAsync(
+                        FlushResult flushResult = await payloadWriter.CopyFromAsync(
                             response.Payload,
                             stream.WritesClosed,
                             endStream: !hasContinuation,
@@ -1120,10 +1120,12 @@ internal sealed class IceRpcProtocolConnection : IProtocolConnection
                                 endStream: true,
                                 _disposedCts.Token).ConfigureAwait(false);
                         }
+
+                        payloadWriterSuccess = !flushResult.IsCanceled;
                     }
                     finally
                     {
-                        payloadWriter.CompleteOutput(success: !flushResult.IsCanceled);
+                        payloadWriter.CompleteOutput(payloadWriterSuccess);
                         response.Payload.Complete();
                         response.PayloadContinuation?.Complete();
                     }
--- a/tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs
+++ b/tests/IceRpc.Tests/IceRpcProtocolConnectionTests.cs
@@ -1551,16 +1551,21 @@ public sealed class IceRpcProtocolConnectionTests
             async () => await taskExceptionObserver.DispatchFailedException,
             Is.InstanceOf<InvalidOperationException>());
 
-        // Cleanup
-        try
-        {
-            await responseTask;
-        }
-        catch (IceRpcException exception) when (exception.IceRpcError == IceRpcError.TruncatedData)
-        {
-            // Depending on the timing, the payload stream send failure might abort the invocation before the response
-            // is sent.
-        }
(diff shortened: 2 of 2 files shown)


InsertCreativityHere (approved):
Looks good to me!

externl (approved):
Looks good.

One nit: the comment at `IceRpcProtocolConnection.cs:1096` — "Set to true only after the payload and its continuation are fully copied" — doesn't quite match the code. When the first `CopyFromAsync` returns `IsCompleted` (the peer stopped reading), the continuation is skipped and `payloadWriterSuccess` still becomes true. Something like "remains false if a copy throws" states why the flag is there, and is harder to read as an invitation to simplify it to `= true`.

bernardnormier:
Reworded the comment to state why the flag exists instead of when it's set: "Remains false if a copy throws or is canceled."

PR #4844, opened 2026-08-20
Wait for detached connections when shutting down a client connection

### LINKED ISSUE #4793: [Audit-Low] ClientConnection.ShutdownAsync skips the detached-connection wait, and its trailing catch contradicts the documented exception contract

## Summary

Two related defects in the same `try`/`catch`.

**1 — the detached-connection wait is skipped.** The active connection's `ShutdownAsync` and the `_detachedConnectionsTcs` wait are two **sequential** awaits inside one `try`, followed by a `catch` that ignores other shutdown exceptions. Both protocol implementations rethrow non-`OperationCanceledException` `IceRpcException` from `ShutdownAsync`, so a peer abort during the active connection's shutdown jumps past the detached wait entirely, and `PerformShutdownAsync` reports success while detached connections are still shutting down.

**2 — the documented exception contract is violated.** The `ShutdownAsync` doc comment promises an `IceRpcException` on shutdown failure, but the trailing `catch { }` swallows everything, so the method returns normally on a failed shutdown.

## Trigger / impact

A peer-initiated shutdown or `InactivityTimeout` detaches C1 (`RemoveFromActiveAsync` sets `_activeConnection = null; _detachedConnectionCount++`); a later invocation attaches C2 and decrements the count to 1; `ShutdownAsync` sees the count is non-zero so `_detachedConnectionsTcs` stays pending; C2's shutdown throws `IceRpcException(ConnectionAborted)`; the generic catch swallows it and the detached wait never runs.

Bounded, hence Low: `DisposeAsync` cancels `_disposedCts` and then awaits `_detachedConnectionsTcs.Task` unconditionally, so nothing leaks and detached connections are still disposed by their own background task. The defect is that `ShutdownAsync` **misreports completion** in the window before `DisposeAsync` — meaningful only when the detached connection is blocked on something slow (client-side callback dispatch, pending invocations) while the active one fails fast.

Codex noted that not every graceful peer closure reaches this path — the icerpc implementation specially accepts `ConnectionClosedByPeer` — but an abort or any other rethrown shutdown failure suffices.

## Source references

- [src/IceRpc/ClientConnection.cs#L426-L432](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/ClientConnection.cs#L426) — the two sequential awaits
- [src/IceRpc/ClientConnection.cs#L449-L452](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/ClientConnection.cs#L449) — the swallowing catch
- [src/IceRpc/ClientConnection.cs#L387-L393](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/ClientConnection.cs#L387) — the doc comment the catch contradicts
- [src/IceRpc/ConnectionCache.cs#L296-L298](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/ConnectionCache.cs#L296) — the correct `Task.WhenAll` shape

## Recommendation

Mirror `ConnectionCache.PerformShutdownAsync` and `Server.PerformShutdownAsync` — a single `Task.WhenAll` over both. Their `cts.Token.ThrowIfCancellationRequested()` re-check exists to un-hide an OCE inside a `WhenAll` aggregate and comes along with the `WhenAll`; it is not an independent defect here.

Then either narrow the trailing catch so the documented `IceRpcException` actually propagates, or amend the doc comment to match the swallow-everything behavior.

## Audit context

Finding 9 of [general-correctness-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-07-27.md), merged with the never-filed doc-contract note from the "Additional Minor Notes" of [general-correctness-audit-2026-06-10.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-06-10.md) — both concern the same `catch`. Codex verdict **CONFIRMED** at **Low** for both parts.

Severity: **Low**.

bernardnormier:
Fixes #4793.

`ClientConnection.PerformShutdownAsync` awaited the active connection's shutdown and the detached-connections wait sequentially inside a single `try`, followed by a `catch` that ignored shutdown failures. Both protocol implementations rethrow a non-`OperationCanceledException` `IceRpcException` from `ShutdownAsync`, so a failing active connection jumped past the detached-connections wait and `ShutdownAsync` reported success while detached connections were still shutting down.

This PR mirrors `ConnectionCache.PerformShutdownAsync` and `Server.PerformShutdownAsync`: a single `Task.WhenAll` over both, in the same nested-try shape, so a shutdown failure is ignored without skipping the wait. The `OperationCanceledException` translation to `OperationAborted` / `TimeoutException` is unchanged.

The `ShutdownAsync` doc comment listed an `IceRpcException` "if the connection shutdown failed", which this method does not throw, and omitted the `OperationAborted` it does throw when the connection is disposed mid-shutdown. It now uses the same wording as its `ConnectionCache` and `Server` counterparts.

## What's Changed entry

None — internal: the shutdown could misreport completion in a window that disposal already covered.

### DIFF
--- a/src/IceRpc/ClientConnection.cs
+++ b/src/IceRpc/ClientConnection.cs
@@ -386,7 +386,8 @@ public sealed class ClientConnection : IInvoker, IAsyncDisposable
     /// <exception cref="ObjectDisposedException">Thrown if this connection is disposed.</exception>
     /// <remarks><para>The returned task can also complete with one of the following exceptions:</para>
     /// <list type="bullet">
-    /// <item><description><see cref="IceRpcException" /> if the connection shutdown failed.</description></item>
+    /// <item><description><see cref="IceRpcException" /> with error <see cref="IceRpcError.OperationAborted" /> if this
+    /// connection is disposed while being shut down.</description></item>
     /// <item><description><see cref="OperationCanceledException" /> if cancellation was requested through the
     /// cancellation token.</description></item>
     /// <item><description><see cref="TimeoutException" /> if this shutdown attempt or a previous attempt exceeded <see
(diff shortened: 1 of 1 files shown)


externl (approved):
Looks good.

One note, not blocking: no test covers the fix. `ClientConnectionTests` has `Shutdown_times_out_after_shutdown_timeout` and `Shutdown_can_be_canceled`, but nothing exercises an active connection whose shutdown fails while a detached connection is still shutting down — the exact regression. `ConnectionCache` has no analogue either, so this matches the codebase rather than being a new gap.

InsertCreativityHere (approved):
Looks good to me!

PR #4845, opened 2026-08-20
Consume the MoveNextAsync ValueTask only once

### LINKED ISSUE #4795: [Audit-Low] AsyncEnumerablePipeReader consumes the same ValueTask<bool> twice (Slice and Protobuf)

> ⚠️ **Filed for review despite an adjacent dismissal.** Codex confirmed the defect but judged that it **does not clear the "actual victim" bar** set when #4465 was closed as Not planned. See "Relationship to #4465" below and close this as Not planned if you agree.

## Summary

In the flush-threshold branch, `bool hasNext = moveNext.Result;` consumes the `ValueTask<bool>` (→ `IValueTaskSource.GetResult(_token)`), and then `result = hasNext ? moveNext.AsTask() : null;` consumes it a **second** time (→ `GetStatus(_token)` plus another `GetResult(_token)`). `ValueTask` is documented as single-consumption.

The non-threshold branch consumes once and is correct — this is a slip in the flush-threshold branch only. The Protobuf copy has the identical shape.

## Trigger / impact

Reproduced against the built `IceRpc.Slice.dll`: a hand-written allocation-free `IAsyncEnumerator<int>` that is also an `IValueTaskSource<bool>`, returning `new ValueTask<bool>(this, _core.Version)` and calling `Reset()` in `GetResult` — the documented pooled-source pattern. A C# iterator streams 80000 bytes fine; the pooled-source enumerator dies with `InvalidOperationException` (stack: `ManualResetValueTaskSourceCore.GetStatus` ← `ValueTask.GetTaskForValueTaskSource` ← `AsyncEnumerableExtensions.cs:185`).

Because the 16 KiB threshold is crossed inside the *first* `ReadAsync`, the whole payload fails before any bytes are delivered — deterministic for an affected enumerator, not intermittent.

The victim population is narrow: all compiler-generated `async IAsyncEnumerable<T>` iterators and Ix.NET/EF-style implementations are immune (their source is not `Reset` until the next `MoveNextAsync`), and nothing in this repo triggers it.

## Relationship to #4465

#4465 (a *different* defect in this same file) was closed as **Not planned** under [AGENTS.md](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/AGENTS.md) dismissal #10 — cost/benefit — with a standing note to reopen only with new evidence: an actual victim and measurable impact.

Codex was asked to adjudicate exactly that, and its answer was **negative**:

> "The hand-written pooled source is legal and the failure is deterministic, but it remains a synthetic victim. Repository test enumerators return inline `ValueTask`s… Without a real production implementation or user report, this does not clear #4465's 'actual victim' filing bar. The proposed fix is correct."

So: the contract violation is real, the fix is one line and correct, but the demonstrated victim is constructed rather than observed. Filing it so the call is yours rather than mine.

## Source references

- [src/IceRpc.Slice/Operations/AsyncEnumerableExtensions.cs#L180](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Slice/Operations/AsyncEnumerableExtensions.cs#L180) — first consumption
- [src/IceRpc.Slice/Operations/AsyncEnumerableExtensions.cs#L185](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Slice/Operations/AsyncEnumerableExtensions.cs#L185) — second consumption
- [src/IceRpc.Protobuf/RpcMethods/Internal/AsyncEnumerableExtensions.cs#L175](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Protobuf/RpcMethods/Internal/AsyncEnumerableExtensions.cs#L175) — the Protobuf mirror

## Recommendation

`result = hasNext ? Task.FromResult(true) : null;` — the value is already known. Apply to the Protobuf mirror.

## Related issues

- #4465 — Not planned; different defect, same file. See above.
- icerpc-csharp-audit#31 / June Finding 8 — a third defect in this same file (OCE swallowing), already fixed.
- `M-06` and `L-10` of the companion repository-wide audit — same file, different defects again.

## Audit context

Finding 11 of [general-correctness-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-07-27.md); Codex verdict **CONFIRMED** at **Low**, with the filing-bar caveat quoted above.

Severity: **Low**.

bernardnormier:
Fixes #4795.

In the stream-flush-threshold branch of `AsyncEnumerablePipeReader`, the `ValueTask<bool>` returned by `MoveNextAsync` was consumed twice: first by reading `moveNext.Result`, then by calling `moveNext.AsTask()`. A `ValueTask` may only be consumed once — awaiting it, reading `Result` or calling `AsTask()` are all consumptions — so this is plain incorrect, even though it is never hit in practice today:

- When the `ValueTask` wraps a result directly or wraps a `Task`, both consumptions are harmless reads.
- When it wraps an `IValueTaskSource<bool>`, the second consumption uses a token that the first one may have invalidated. A pooled source that resets itself in `GetResult` — the documented pattern for reusing a source — makes `AsTask()` throw `InvalidOperationException`.

Compiler-generated `async IAsyncEnumerable<T>` iterators fall in the benign category: their value task source is reset at the beginning of the next `MoveNextAsync`, not in `GetResult`, so the token is still valid for the second consumption. That is why no enumerator used in practice fails today.

In this branch `hasNext` is known to be `true`, so the fix is to produce the task from that value instead of consuming the `ValueTask` again. The resulting `Task<bool>` is the same. The Protobuf copy of this code gets the same fix.

## What's Changed entry

None — no supported enumerator implementation is affected.

### DIFF
--- a/src/IceRpc.Protobuf/RpcMethods/Internal/AsyncEnumerableExtensions.cs
+++ b/src/IceRpc.Protobuf/RpcMethods/Internal/AsyncEnumerableExtensions.cs
@@ -172,7 +172,7 @@ internal static class AsyncEnumerableExtensions
                         // If we reached the stream flush threshold, it's time to flush.
                         if (written >= _streamFlushThreshold)
                         {
-                            result = hasNext ? moveNext.AsTask() : null;
+                            result = hasNext ? Task.FromResult(true) : null;
                             keepEncoding = false;
                         }
                         else
--- a/src/IceRpc.Slice/Operations/AsyncEnumerableExtensions.cs
+++ b/src/IceRpc.Slice/Operations/AsyncEnumerableExtensions.cs
@@ -182,7 +182,7 @@ public static class AsyncEnumerableExtensions
                         // If we reached the stream flush threshold, it's time to flush.
                         if (encoder.EncodedByteCount - sizePlaceholder.Length >= _streamFlushThreshold)
                         {
-                            result = hasNext ? moveNext.AsTask() : null;
+                            result = hasNext ? Task.FromResult(true) : null;
                             keepEncoding = false;
                         }
                         else


InsertCreativityHere (approved):
Looks good to me!

externl (approved):

PR #4846, opened 2026-08-20
Reject invalid opaque server addresses when decoding a proxy

### LINKED ISSUE #4796: [Audit-Low] Opaque ice server addresses that decode successfully cannot be re-encoded

## Summary

`DecodeServerAddress` writes the raw **signed** short transport code into parameter `t`, and the raw encapsulation body into `v` — where `size` may legitimately be 0, yielding `v=""`.

`ParseOpaqueParams` rejects exactly those two shapes (`transportCode < 0`, and `bytes.Length == 0`; `-1` is additionally overloaded as the "t not seen" sentinel). `ServiceAddress.CheckParams` accepts both, and the decoder uses the unchecked constructors, so nothing in between blocks the round trip.

The invariant to restore: everything `DecodeServerAddress` produces must be re-encodable by `EncodeServerAddress`.

## Trigger / impact

Reproduced with a hand-built proxy buffer:

- transport code `-1` → `decoded OK: ice://opaque/hello?transport=opaque&v=qrvM&t=-1`, then `FormatException: The value for parameter 't' is out of range`
- code `99` with a 6-byte encapsulation → `decoded OK: ...&v&t=99`, then `FormatException: Missing 'v' parameter`

Code `-2` fails too, so the `-1` sentinel overload is a secondary smell rather than the cause. `FormatException` is otherwise reserved for user-supplied URIs; the encode surface throws `NotSupportedException`/`ArgumentException`.

Low rather than Medium: conforming Ice peers are unlikely to produce either shape (the bundled `Ice::EndpointType` definitions are 0–9, and real encapsulations are non-empty), and this repo ships no `LocatorRegistry` implementation, so decode-then-re-encode requires application code.

Codex flagged one over-reach in the original write-up: the assertion that *every* possible extension endpoint must have a non-empty body is **not** established by this code or by the official opaque-endpoint syntax. Negative registered endpoint types are nonconforming, but an empty body is not clearly illegal.

## Source references

- [src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs#L130](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs#L130) — signed transport code written to `t`
- [src/IceRpc.Ice/Operations/IceProxyIceEncoderExtensions.cs#L201](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Ice/Operations/IceProxyIceEncoderExtensions.cs#L201) and [#L230](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Ice/Operations/IceProxyIceEncoderExtensions.cs#L230) — the two rejections

## Recommendation

Track "parameter seen" with `bool` flags instead of overloading `-1`/empty, and accept the full signed-short range for `t` plus an empty `v`.

## Related issues

- #4751 — the same accept-on-parse / fail-on-serialize class (alt-server `$`/`&` round-trip is not injective), currently open on the Future milestone.

## Audit context

Finding 12 of [general-correctness-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-07-27.md); Codex verdict **CONFIRMED** at **Low** — "a genuine decode/encode asymmetry … not unreachable enough to dismiss as Not-a-bug".

Severity: **Low**.

bernardnormier:
Fixes #4796.

Decoding a proxy could produce an opaque server address that cannot be encoded back:

- The transport code was decoded as a signed short and written verbatim into the `t` parameter, but no transport code is negative and the encoder rejects a negative `t`.
- An encapsulation whose body is empty produced an empty `v` parameter, and the encoder rejects an opaque server address without a body.

An application that decodes such a proxy and re-encodes it — when forwarding it, for instance — got a `FormatException`, an exception otherwise reserved for server addresses built from a URI.

Both values are invalid per the wire format, so this PR rejects them where the invalid data is received: `DecodeServerAddress` now throws `InvalidDataException` for a negative transport code and for an opaque server address with an empty body. The encoder is unchanged.

## What's Changed entry

None — only affects proxies from a nonconforming peer.

### DIFF
--- a/src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs
+++ b/src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs
@@ -72,7 +72,12 @@ public static class IceProxyIceDecoderExtensions
         // With the Ice encoding, the ice server addresses are transport-specific, with a transport-specific encoding.
 
         ServerAddress? serverAddress = null;
-        var transportCode = (TransportCode)decoder.DecodeShort();
+        short transportCodeValue = decoder.DecodeShort();
+        if (transportCodeValue < 0)
+        {
+            throw new InvalidDataException($"Received invalid transport code: {transportCodeValue}.");
+        }
+        var transportCode = (TransportCode)transportCodeValue;
 
         int size = decoder.DecodeInt();
         if (size < 6)
@@ -117,6 +122,11 @@ public static class IceProxyIceDecoderExtensions
                         break;
 
                     default:
+                        if (size == 0)
+                        {
+                            throw new InvalidDataException("Received opaque server address with an empty body.");
+                        }
+
                         // Create a server address for transport opaque
                         ImmutableDictionary<string, string>.Builder builder =
                             ImmutableDictionary.CreateBuilder<string, string>();


InsertCreativityHere (approved):
Looks good to me!

externl (commented):
The two guards are right, but the invariant in the description isn't restored: `TransportCode.Uri` (code 0) at `IceProxyIceDecoderExtensions.cs:116` builds a `ServerAddress` straight from a wire-supplied URI, and `ServerAddress(Uri)` doesn't validate opaque params. A peer sending `ice://opaque/?transport=opaque&t=-1&v=qrvM` still decodes fine and still throws `FormatException` on re-encode — the exact #4796 shape, through the path the diff doesn't guard.

Also no test for either new guard; `ProxyTests.cs:46` already has the encode-then-corrupt-bytes harness for this decoder.

bernardnormier:
You're right that the URI path accepts this — but so does the public API: `new ServiceAddress(new Uri("ice://opaque?transport=opaque&t=-1&v=qrvM"))` constructs fine, and so does `ice://host?transport=tcp&t=abc`. Server address params are opaque strings, parsed only when something uses them; encoding a proxy you built by hand from either URI throws the same `FormatException`, with no wire input involved.

So the URI decode path is exactly as permissive as `ServiceAddress(Uri)`, and "a decoded proxy can always be re-encoded" overstates what the decoder alone can deliver — I reworded the description. What the two guards reject are forms that are invalid per the wire format itself: a negative transport code and an opaque server address with an empty body.

On tests: I'm leaving these guards untested — this is extremely unlikely and harmless invalid data, and the corrupt-bytes harness doesn't reach either guard without significant new encoding code.

externl (approved):

PR #4848, opened 2026-08-20
Skip the retry log scope when logging is disabled

### LINKED ISSUE #4803: [Audit-Low] RetryInterceptor's `_logger != NullLogger.Instance` fast path is dead when configured via UseRetry
> **AI-generated audit finding** — this issue was opened from an automated correctness audit performed by Claude (Anthropic), model **Opus 5** (`claude-opus-5`). This particular item was re-verified directly against `main` rather than through the Codex cross-review pass that covers the rest of the batch. It **has not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

`RetryInterceptor` guards its per-attempt logging scope with a reference comparison against `NullLogger.Instance`:

```csharp
_logger != NullLogger.Instance && attempt > 1 ? _logger.RetryScope(attempt, _maxAttempts) : null;
```

The intent is clearly "skip the scope allocation when there is no logger". But `UseRetry` creates a `Logger<T>` wrapper, which is never reference-equal to `NullLogger.Instance` — so for any pipeline configured through `UseRetry` the guard is always true and the fast path never fires. The result is a logging scope allocated on every retry attempt even when logging is effectively disabled.

## Trigger / impact

Allocation-only; no behavioral difference. This is cleanup-grade, filed for completeness rather than urgency.

## Source references

- [src/IceRpc.Retry/RetryInterceptor.cs#L200](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Retry/RetryInterceptor.cs#L200) — the reference comparison
- [src/IceRpc.Retry/RetryPipelineExtensions.cs#L36](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Retry/RetryPipelineExtensions.cs#L36) — `NullLoggerFactory.Instance` passed to the factory, which still yields a `Logger<T>` wrapper

## Recommendation

Either drop the guard entirely (the scope is cheap relative to a retry), or replace the reference check with `_logger.IsEnabled(LogLevel.Debug)` — whichever matches the logging convention used elsewhere in the codebase.

## Audit context

From the "Additional Minor Notes" section of [general-correctness-audit-2026-06-10.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-06-10.md), which was single-pass verified and never filed. Re-checked against `main` at `2de29310a` on 2026-07-28: still present.

Severity: **Low**.

bernardnormier:
Fixes #4803.

`CreateRetryLogScope` skips the logging scope when `_logger != NullLogger.Instance`, but this comparison never succeeded for a pipeline configured through `UseRetry`: `loggerFactory.CreateLogger<RetryInterceptor>()` wraps every factory — including `NullLoggerFactory` — in a `Logger<RetryInterceptor>`. `UseRetry` now passes `NullLogger.Instance` directly when the factory is a `NullLoggerFactory`, matching the `UseLocator` fix in #4878, which makes the check effective.

The builder variant of `UseRetry` resolves `ILogger<RetryInterceptor>` from the service provider and is unchanged: with DI, the logger is always a `Logger<RetryInterceptor>` wrapper.

## What's Changed entry

None — internal optimization; no observable behavior change.

### DIFF
--- a/src/IceRpc.Retry/RetryInterceptor.cs
+++ b/src/IceRpc.Retry/RetryInterceptor.cs
@@ -4,6 +4,7 @@ using IceRpc.Extensions.DependencyInjection;
 using IceRpc.Features;
 using IceRpc.Retry.Internal;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using System.Diagnostics;
 using System.Runtime.ExceptionServices;
 
@@ -196,5 +197,5 @@ public class RetryInterceptor : IInvoker
     }
 
     private IDisposable? CreateRetryLogScope(int attempt) =>
-        attempt > 1 ? _logger.RetryScope(attempt, _maxAttempts) : null;
+        _logger != NullLogger.Instance && attempt > 1 ? _logger.RetryScope(attempt, _maxAttempts) : null;
 }
--- a/src/IceRpc.Retry/RetryInterceptor.cs
+++ b/src/IceRpc.Retry/RetryInterceptor.cs
@@ -4,7 +4,6 @@ using IceRpc.Extensions.DependencyInjection;
 using IceRpc.Features;
 using IceRpc.Retry.Internal;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
 using System.Diagnostics;
 using System.Runtime.ExceptionServices;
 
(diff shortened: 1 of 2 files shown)


externl (dismissed):
Looks good.

- Out of scope, but the same idiom survives next door *with* a real cost: `LocatorInterceptor.cs:205,212,232` compare against `NullLogger.Instance` while `LocatorPipelineExtensions.cs:32` feeds them `loggerFactory.CreateLogger<LocatorInterceptor>()` — a `Logger<T>` wrapper even for `NullLoggerFactory.Instance` — so the `Log*Decorator` chain is always installed. Happy to file an issue for it if you'd like.

bernardnormier:
Good catch — filed #4858.

externl (approved):
Looks good.

pepone (approved):

Issue #4801, opened 2026-07-28
Encode WellKnownTypes::Uuid in RFC 9562 (network) byte order instead of .NET Guid mixed-endian


pepone:
> **AI-generated audit finding** — this issue was opened from an automated correctness audit performed by Claude (Anthropic), model **Opus 5** (`claude-opus-5`), and independently cross-verified with OpenAI Codex (`codex-cli 0.145.0`).

## Summary

`EncodeUuid` uses the default `Guid.TryWriteBytes`, which emits .NET's **mixed-endian** layout (first three fields little-endian, remaining 8 bytes in order). `DecodeUuid` is symmetric through the default `new Guid(span)`. That is *not* the network byte order (big-endian) specified by [RFC 9562, Section 4 (UUID Format)](https://www.rfc-editor.org/rfc/rfc9562#section-4) (which obsoletes RFC 4122): "each field is encoded with the most significant byte first (known as 'network byte order')".

The Slice definition of the type documents only the width, so the byte order is unspecified:

```slice
/// Represents a Universally unique identifier (UUID), encoded as a span of 16 bytes (128-bit).
```

## Impact

Encode and decode are symmetric, so UUIDs round-trip correctly and nothing breaks today. However, an implementation in another language would reasonably assume canonical RFC 9562 byte order — the first 8 bytes would come out swapped.

## Fix

Encode and decode `WellKnownTypes::Uuid` in RFC 9562 (network) byte order:

- `value.TryWriteBytes(span, bigEndian: true, out _)` in `EncodeUuid`
- `new Guid(span, bigEndian: true)` in `DecodeUuid`

and document the byte order in `Uuid.slice` and the encoder/decoder doc comments.

This changes the wire encoding of `Uuid`, so it can't ship in a patch release — targeted at 0.7.0.

## Source references

Pinned to commit `2de29310a82428560125c5f804127ea054f46769`:

- [src/ZeroC.Slice.Codec/UuidSliceEncoderExtensions.cs#L14-L23](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/ZeroC.Slice.Codec/UuidSliceEncoderExtensions.cs#L14) — default `TryWriteBytes`
[...]

PR #4849, opened 2026-08-20
Document the little-endian host assumption

### LINKED ISSUE #4804: [Audit-Low] Slice codec assumes a little-endian host, with no endianness guards
> **AI-generated audit finding** — this issue was opened from an automated correctness audit performed by Claude (Anthropic), model **Opus 5** (`claude-opus-5`). This particular item was re-verified directly against `main` rather than through the Codex cross-review pass that covers the rest of the batch. It **has not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

The Slice codec writes and reads multi-byte primitives using the host's native byte order throughout, with no `BinaryPrimitives` calls and no endianness guards. On a big-endian platform (s390x, or a big-endian MIPS/PowerPC build) the encoded wire format would not match the Slice specification, making such a host wire-incompatible with every other peer.

This is informational: .NET's supported platforms are effectively all little-endian today, so nothing is broken in practice. It is filed so the assumption is recorded and can be either documented as a deliberate constraint or closed with the rest of the endianness question.

## Recommendation

Pick one deliberately:

- Document the little-endian requirement as a supported-platform constraint, and optionally add a static check at startup so a big-endian host fails loudly rather than silently producing wrong bytes; **or**
- Route primitive encode/decode through `BinaryPrimitives.*LittleEndian` so the codec is correct regardless of host byte order.

## Related issues

- The `Uuid` byte-order issue filed alongside this one is a *different* question — that one concerns **field layout** (which is a deliberate, cross-language convention and merely underdocumented), whereas this one concerns **host byte order**.

## Audit context

From the "Additional Minor Notes" section of [general-correctness-audit-2026-06-10.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-06-10.md), which was single-pass verified and never filed. Recorded there as informational.

Severity: **Low** (informational).

bernardnormier:
Fixes #4804.

The Slice and Ice encodings always encode multi-byte primitives in little-endian byte order. This implementation performs no byte-order conversion: it assumes the host is little-endian, so host order and wire order coincide and the codecs can copy arrays of fixed-size primitives directly to and from the wire.

This PR records that assumption as a supported-platform constraint in AGENTS.md's dismissed audit patterns, so future audits don't refile it or propose endianness guards / `BinaryPrimitives` conversions that would forfeit the copy fast paths for platforms .NET doesn't support.

## What's Changed entry

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -123,3 +123,11 @@ is no buffer to overflow, so the "zip bomb" framing (amplification of a small in
 does not apply. Bounding still happens, just at the consumer's decoder: Slice / Protobuf reject malformed bytes
 (`InvalidDataException`) and enforce per-decoder collection-allocation budgets. Don't propose a size cap inside the
 decompressor; the limit belongs to the decoder that materializes structured data from the stream. (#4507.)
+
+### 12. Little-endian host assumption
+
+The Slice and Ice encodings always encode multi-byte primitives in little-endian byte order. This C# implementation
+performs no byte-order conversion — it assumes the host is little-endian, so host order and wire order coincide, and
+the fast paths copy arrays of fixed-size primitives directly to and from the wire. A little-endian host is a
+supported-platform constraint, consistent with the platforms .NET itself supports. Don't file findings about missing
+endianness guards or propose routing these codecs through `BinaryPrimitives`. (#4804.)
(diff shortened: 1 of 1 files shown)


InsertCreativityHere (approved):
All sounds good to me!

externl (approved):
Looks good.

One nit: "Don't file findings about missing endianness guards" reads more broadly than intended. #4801 (`WellKnownTypes::Uuid` in RFC 9562 order) is open and accepted, and `DecodeUuid` is exactly an endianness-shaped finding — `new Guid(span)` takes .NET's mixed-endian layout. Narrowing it to *host* byte order in the Slice/Ice primitive codecs would keep a future agent from dismissing #4801-shaped work.

bernardnormier:
Good point — the pattern is now scoped to host byte order, with an explicit carve-out for types whose specified wire format is big-endian (`WellKnownTypes::Uuid`, RFC 9562 — #4801).

PR #4850, opened 2026-08-20
Make Set throw on a read-only feature collection

### LINKED ISSUE #4806: [Audit-Medium] Generic mutation bypasses read-only feature collections, and can poison FeatureCollection.Empty

## Summary

`ReadOnlyFeatureCollectionDecorator` throws from its indexer setter, but `Set<TFeature>` delegates straight to the mutable decoratee with no `IsReadOnly` check:

```csharp
public object? this[Type key]
{
    get => _decoratee[key];
    set => throw new InvalidOperationException("Cannot update a read-only feature collection.");
}

public void Set<TFeature>(TFeature? feature) => _decoratee.Set(feature);   // <-- no guard
```

## Impact

`FeatureCollection.Empty` is a **static singleton** (`new FeatureCollection().AsReadOnly()`), so `FeatureCollection.Empty.Set(feature)` mutates process-wide shared state through an interface that advertises `IsReadOnly => true`.

## Source references

- [src/IceRpc/Features/Internal/ReadOnlyFeatureCollectionDecorator.cs#L32](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/Features/Internal/ReadOnlyFeatureCollectionDecorator.cs#L32) — the unguarded `Set<TFeature>`
- [src/IceRpc/Features/FeatureCollection.cs#L12](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/Features/FeatureCollection.cs#L12) — `Empty` as a static singleton

## Recommendation

Make every mutation path throw when `IsReadOnly` is true. Add tests covering both the indexer and the generic setter, including `FeatureCollection.Empty`.

## Audit context

M-03 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md); confirmed by the [Claude verification pass](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-claude-review-2026-07-28.md). One-line fix.

Severity: **Medium**.

bernardnormier:
Fixes #4806.

`ReadOnlyFeatureCollectionDecorator` threw from its indexer setter but let `Set<TFeature>` delegate to the mutable decoratee, so the generic setter bypassed the read-only guard. Since `FeatureCollection.Empty` is a read-only decorator over a static singleton — and `IncomingRequest.Features` defaults to it — an application calling `request.Features.Set(feature)` instead of `request.Features = request.Features.With(feature)` silently mutated process-wide shared state, corrupting the features of unrelated requests.

`Set<TFeature>` now throws the same `InvalidOperationException` as the indexer setter.

## What's Changed entry

Area: Core

- Fixed a bug where setting a feature in a read-only feature collection updated this collection instead of
  throwing `InvalidOperationException`. In particular, setting a feature in `FeatureCollection.Empty` corrupted
  this shared feature collection.

### DIFF
--- a/src/IceRpc/Features/IFeatureCollection.cs
+++ b/src/IceRpc/Features/IFeatureCollection.cs
@@ -16,6 +16,8 @@ public interface IFeatureCollection : IEnumerable<KeyValuePair<Type, object>>
     /// <summary>Gets or sets a feature. Setting null removes the feature.</summary>
     /// <param name="key">The feature key.</param>
     /// <returns>The requested feature.</returns>
+    /// <exception cref="InvalidOperationException">Thrown by the setter when this feature collection is read-only.
+    /// </exception>
     object? this[Type key] { get; set; }
 
     /// <summary>Gets the requested feature. If the feature is not set, returns <see langword="null" />.</summary>
@@ -26,5 +28,6 @@ public interface IFeatureCollection : IEnumerable<KeyValuePair<Type, object>>
     /// <summary>Sets a new feature. Setting null removes the feature.</summary>
     /// <typeparam name="TFeature">The feature key.</typeparam>
     /// <param name="feature">The feature value.</param>
+    /// <exception cref="InvalidOperationException">Thrown when this feature collection is read-only.</exception>
     void Set<TFeature>(TFeature? feature);
 }
--- a/src/IceRpc/Features/Internal/ReadOnlyFeatureCollectionDecorator.cs
+++ b/src/IceRpc/Features/Internal/ReadOnlyFeatureCollectionDecorator.cs
@@ -30,7 +30,8 @@ internal class ReadOnlyFeatureCollectionDecorator : IFeatureCollection
     public TFeature? Get<TFeature>() => _decoratee.Get<TFeature>();
 
     /// <inheritdoc />
-    public void Set<TFeature>(TFeature? feature) => _decoratee.Set(feature);
+    public void Set<TFeature>(TFeature? feature) =>
+        throw new InvalidOperationException("Cannot update a read-only feature collection.");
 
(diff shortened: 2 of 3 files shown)


externl (approved):
Looks good.

- The throw isn't documented on the interface. Neither `IFeatureCollection.Set<TFeature>` nor `this[Type]` carries an `<exception cref="InvalidOperationException">`, and the `IsReadOnly` doc says only "read-only or read-write" — but this is now the contract every implementer has to honor on both mutation paths. `ReadOnlyFeatureCollectionDecorator.Set` also keeps `<inheritdoc />`, so its inherited text still reads "Sets a new feature. Setting null removes the feature"; fixing the interface doc resolves both.
- No coverage for `Set<T>(null)` on a read-only collection. The remove path goes through the same method, and it's the form the one internal caller uses (`LocatorInterceptor.cs:78`, guarded by `IsReadOnly`).
- Nit: `FeatureCollection.Empty.Set("foo")` in the new test — if this guard ever regresses, that line both fails and poisons the process-wide singleton for every other test in the assembly.

InsertCreativityHere (approved):
Looks good to me!

bernardnormier:
All three addressed:
- `IFeatureCollection` now documents the `InvalidOperationException` on both mutation paths, so the decorator's `<inheritdoc />` picks up the correct contract.
- Added the `Set<string>(null)` assertion for the remove path.
- Dropped the `FeatureCollection.Empty` assertion — `Empty` is `new FeatureCollection().AsReadOnly()`, the same decorator the test already constructs, so the line added the singleton-poisoning risk without adding coverage.

PR #4851, opened 2026-08-20
Ship the Ice Slice files in the IceRpc.Ice package

### LINKED ISSUE #4819: [Audit-Medium] IceRpc.Locator advertises IDL content absent from its package

## Summary

`IceRpc.Locator.props` is packed to `buildTransitive/` and adds a Slice directory relative to itself:

```xml
<SliceDirectory Include="$(MSBuildThisFileDirectory)../slice" />
```

But the package's content items only pack `LICENSE`, `README.md`, the icon, and the props file itself. The `.ice` files referenced by `SliceCompile` are build inputs, not package content — so the advertised `slice/` directory does not exist in the produced nupkg.

## Source references

- [src/IceRpc.Locator/IceRpc.Locator.props](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Locator/IceRpc.Locator.props) — the transitive directory entry
- [src/IceRpc.Locator/IceRpc.Locator.csproj#L26-L31](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Locator/IceRpc.Locator.csproj#L26) — package contents, with no slice content

## Recommendation

Pack the required Locator IDLs at the advertised path, or remove/correct the transitive directory entry. Add a package-consumer test that resolves and compiles the packaged IDL.

## Audit context

M-20 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md); confirmed by the [Claude verification pass](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-claude-review-2026-07-28.md).

Severity: **Medium**.

bernardnormier:
Fixes #4819.

The IceRpc.Locator package shipped a `buildTransitive/IceRpc.Locator.props` that added `$(MSBuildThisFileDirectory)../slice` as a `SliceDirectory` — but the package contains no `slice` directory. This props is a leftover from 0.5.x, when this package compiled `.slice` files packed via `SliceFile Pack="true".

The `.ice` files now ship with the IceRpc.Ice package instead: `Identity.ice`, `Locator.ice`, `LocatorRegistry.ice` and `Process.ice`, in the package's `slice/Ice` directory. All `Ice/*.ice` files ship in this one package regardless of which assembly compiles them: the Slice convention is to include them as `Ice/Locator.ice` with `-I <dir>/slice`, and the files include each other with `#include "Identity.ice"` (unchanged from upstream), which only resolves when they share a directory. Splitting them across packages would require consumers to add a `-I` on a `slice/Ice` subdirectory, which is never done.

A project that includes `Ice/Process.ice` with only IceRpc.Ice referenced gets a compile error on the generated code until it references IceRpc.Locator, which provides the generated types — rare, and the fix is a package reference.

The set excludes the wire-level enums (`OperationMode.ice`, `ReplyStatus.ice`) and `Object.ice`, which are implementation details.

## What's Changed entry

Area: icerpc+ice integration

- The IceRpc.Ice package now ships Identity.ice, Locator.ice, LocatorRegistry.ice and Process.ice in its slice
  directory.

### DIFF
--- a/src/IceRpc.Ice/IceRpc.Ice.csproj
+++ b/src/IceRpc.Ice/IceRpc.Ice.csproj
@@ -36,5 +36,10 @@
     <Content Include="../../LICENSE" Pack="true" PackagePath="/" />
     <Content Include="README.md" Pack="true" PackagePath="/" />
     <None Include="../../build/icerpc-icon.png" Pack="true" PackagePath="/" />
+    <Content
+      Include="../../slice/Ice/Identity.ice;../../slice/Ice/Locator.ice;../../slice/Ice/LocatorRegistry.ice;../../slice/Ice/Process.ice"
+      Pack="true"
+      PackagePath="slice/Ice"
+    />
   </ItemGroup>
 </Project>
--- a/src/IceRpc.Locator/IceRpc.Locator.csproj
+++ b/src/IceRpc.Locator/IceRpc.Locator.csproj
@@ -27,6 +27,5 @@
     <Content Include="../../LICENSE" Pack="true" PackagePath="/" />
     <Content Include="README.md" Pack="true" PackagePath="/" />
     <None Include="../../build/icerpc-icon.png" Pack="true" PackagePath="/" />
-    <Content Include="IceRpc.Locator.props" Pack="true" PackagePath="buildTransitive" />
   </ItemGroup>
 </Project>
--- a/src/IceRpc.Locator/IceRpc.Locator.props
+++ b/src/IceRpc.Locator/IceRpc.Locator.props
@@ -1,7 +1,0 @@
-<?xml version="1.0" encoding="utf-8" ?>
-<!-- Copyright (c) ZeroC, Inc. -->
-<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <ItemGroup>
-    <SliceDirectory Include="$(MSBuildThisFileDirectory)../slice" />
-  </ItemGroup>
-</Project>


pepone (approved):

pepone:
A quoted include does not require these files to be colocated. `#include "Identity.ice"` also resolves when the directory containing `Identity.ice` is passed to `slice2cs` with `-I`.

Keeping the files split preserves package ownership: `Identity.ice` belongs in `IceRpc.Ice`, while `Locator.ice`, `LocatorRegistry.ice`, and `Process.ice` belong in `IceRpc.Locator`, which contains the C# types generated from them.

With the proposed layout, a consumer can obtain and include `Process.ice` from `IceRpc.Ice` without referencing `IceRpc.Locator`. Code generated from a Slice definition that uses `Ice::Process` references types such as `IceRpc.Ice.ProcessProxy`, but those types are provided by `IceRpc.Locator.dll`, so that setup does not compile.

bernardnormier:
Right, a quoted include also resolves through `-I` — but only if the `-I` points at the directory that contains `Identity.ice`, i.e. `<IceRpc.Ice>/slice/Ice`. The convention is `-I <dir>/slice` with `#include "Ice/Locator.ice"`, never a `-I` on a subdirectory, and the files keep upstream's `#include "Identity.ice"`. So all `Ice/*.ice` files must ship together, and IceRpc.Ice is the package for them regardless of which assembly compiles each file — the generated types for Locator, LocatorRegistry and Process stay in IceRpc.Locator.dll. Updated the description to say this instead of the incorrect "only resolves when colocated" claim.

pepone (approved):
The packaging looks correct. One thing not clear is how users would include these files, we previously have the transitive build props to automate this.

bernardnormier:
The transitive props automation was `.slice`-only: `SliceDirectory` is an IceRpc.Slice.Tools item (still used by the IceRpc and ZeroC.Slice.Codec packages), and the deleted `IceRpc.Locator.props` was its 0.5.x leftover — after the `.ice` migration it pointed at a directory the package no longer shipped and fed an item ZeroC.Ice.Slice.Tools never reads. Today a consumer of the packaged `.ice` files must set `IncludeDirectories` by hand (`GeneratePathProperty` + `$(PkgIceRpc_Ice)/slice`). To restore the automation, ZeroC.Ice.Slice.Tools needs a `SliceDirectory`-like item that packages can contribute to from a `buildTransitive` props — filed zeroc-ice/ice#6677.

Issue #4852, opened 2026-08-20
Add a run smoke test for the project templates in CI


bernardnormier:
The test-templates CI action generates and builds all 12 project templates but never runs any of them, so a missing runtime file goes undetected. #4820 is an example: the non-DI templates did not copy their `certs` directory to the output directory, and every template still built fine.

Add a run smoke step: for at least one server/client template pair, build both projects, start the server binary from its output directory, run the client against it, and assert the invocation succeeds. Running from the output directory (not `dotnet run` from the project directory) is the point — it's the launch mode that catches missing output files.

PR #4854, opened 2026-08-20
Copy the certs directory to the output directory in the non-DI templates

### LINKED ISSUE #4820: [Audit-Medium] Six non-DI templates do not copy required certificates

## Summary

The non-DI template programs load files under `certs/`, but their project files omit the copy rule that the DI variants have. Verification found a clean split across all twelve templates:

| Template | certs copy rule in csproj |
|---|---|
| IceRpc-Ice-Client, IceRpc-Ice-Server | ❌ |
| IceRpc-Slice-Client, IceRpc-Slice-Server | ❌ |
| IceRpc-Protobuf-Client, IceRpc-Protobuf-Server | ❌ |
| IceRpc-Ice-DI-Client, IceRpc-Ice-DI-Server | ✅ |
| IceRpc-Slice-DI-Client, IceRpc-Slice-DI-Server | ✅ |
| IceRpc-Protobuf-DI-Client, IceRpc-Protobuf-DI-Server | ✅ |

All six non-DI templates reference `certs` from their program but have no copy rule; all six DI templates have it.

## Impact

A generated non-DI project produces no certificate under `bin`, and starting it from the output directory fails with `DirectoryNotFoundException`.

## Source references

- [src/IceRpc.Templates/Templates](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Templates/Templates)

## Recommendation

Apply the same output/publish content metadata to every template variant, and add generated-template smoke tests that launch from the output directory.

## Audit context

M-21 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md); confirmed by the [Claude verification pass](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-claude-review-2026-07-28.md).

Severity: **Medium**.

bernardnormier:
Fixes #4820.

The six non-DI templates load their certificates from `certs/` but their project files had no rule copying this directory to the build output, so an application created from these templates built fine and failed on startup when launched from its output directory. The six DI templates already have the copy rule; this PR adds the same rule to the non-DI templates.

No smoke test — the templates are exercised by the test-templates CI action, which builds but does not run them; #4852 tracks adding a run smoke test.

## What's Changed entry

Area: Templates

- Fixed a bug in the non-DI templates: you could build but not run the applications created with these
  templates because the certificates were missing.

### DIFF
--- a/src/IceRpc.Templates/Templates/IceRpc-Ice-Client/IceRpc-Ice-Client.csproj
+++ b/src/IceRpc.Templates/Templates/IceRpc-Ice-Client/IceRpc-Ice-Client.csproj
@@ -30,6 +30,7 @@
     <GlobalAnalyzerConfigFiles Include="$(MSBuildThisFileDirectory)CodeAnalysis.globalconfig" />
   </ItemGroup>
   <ItemGroup>
+    <!-- Copy the test certificates to the output directory. -->
     <None Update="certs\**">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
--- a/src/IceRpc.Templates/Templates/IceRpc-Ice-Client/IceRpc-Ice-Client.csproj
+++ b/src/IceRpc.Templates/Templates/IceRpc-Ice-Client/IceRpc-Ice-Client.csproj
@@ -29,4 +29,9 @@
     </PackageReference>
     <GlobalAnalyzerConfigFiles Include="$(MSBuildThisFileDirectory)CodeAnalysis.globalconfig" />
   </ItemGroup>
+  <ItemGroup>
+    <None Update="certs\**">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
 </Project>
--- a/src/IceRpc.Templates/Templates/IceRpc-Ice-DI-Client/IceRpc-Ice-DI-Client.csproj
+++ b/src/IceRpc.Templates/Templates/IceRpc-Ice-DI-Client/IceRpc-Ice-DI-Client.csproj
@@ -34,6 +34,7 @@
     <None Update="appsettings.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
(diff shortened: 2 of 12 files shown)


src/IceRpc.Templates/Templates/IceRpc-Ice-Client/IceRpc-Ice-Client.csproj
```diff
@@ -29,4 +29,9 @@
     </PackageReference>
     <GlobalAnalyzerConfigFiles Include="$(MSBuildThisFileDirectory)CodeAnalysis.globalconfig" />
   </ItemGroup>
+  <ItemGroup>
+    <None Update="certs\**">
```
pepone, line 34: Maybe we should add a comment here and in other templates, about copy the test certificates to the output directory.

Do we need a warning about don't doing that in production? Copying certificates doesn't seems such a good advice.

pepone (approved):

src/IceRpc.Templates/Templates/IceRpc-Ice-Client/IceRpc-Ice-Client.csproj
bernardnormier: Added the comment to all 12 templates (the DI ones had the same rule without a comment). On the production warning: each template's `certs/README.md` already says these test certificates must not be used in production, and the comment calls them test certificates; I'd rather not repeat the warning in the project file.

pepone (approved):

src/IceRpc.Templates/Templates/IceRpc-Ice-Client/IceRpc-Ice-Client.csproj
pepone: >  I'd rather not repeat the warning in the project file.

I was thinking about the general practice about copying certificates around, but less assume a bit of common sense. It should be obvious copying secret files is not a good idea in general.

PR #4855, opened 2026-08-20
Add AddIceRpcClientConnection overloads with a server address

### LINKED ISSUE #4822: [Audit-Medium] Seven package READMEs call a nonexistent dependency-injection overload

## Summary

Seven published package READMEs contain:

```csharp
.AddIceRpcClientConnection(new Uri("icerpc://localhost"))
```

The only public extension is **parameterless**:

```csharp
public static IServiceCollection AddIceRpcClientConnection(this IServiceCollection services)
```

Affected: Compressor, Deadline, Extensions.DependencyInjection, Logger, Metrics, RequestContext, Telemetry.

## Source references

- [src/IceRpc.Extensions.DependencyInjection/ClientConnectionServiceCollectionExtensions.cs#L38](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Extensions.DependencyInjection/ClientConnectionServiceCollectionExtensions.cs#L38) — the actual signature

## Recommendation

Update the snippets to configure the server address through the supported options API, and compile published README snippets in CI so this cannot regress.

## Audit context

M-24 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md); confirmed by the [Claude verification pass](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-claude-review-2026-07-28.md).

Severity: **Medium**.

bernardnormier:
Fixes #4822.

Seven package READMEs show `services.AddIceRpcClientConnection(new Uri("icerpc://localhost"))`, but the only overload was parameterless — the snippets never compiled. Rather than making seven README snippets more ceremonial, this PR adds the overloads the snippets assume, mirroring the `ClientConnection` constructors: an `AddIceRpcClientConnection(ServerAddress)` overload that sets `ClientConnectionOptions.ServerAddress` through the options pipeline, and an `AddIceRpcClientConnection(Uri)` convenience overload that delegates to it. The READMEs are correct as written.

Also reworks the doc comments of all three overloads to say what they register: a `ClientConnection` singleton, also registered as the `IInvoker` singleton.

## What's Changed entry

Area: Dependency injection

- Added AddIceRpcClientConnection overloads that accept the client connection's server address or server
  address URI.

### DIFF
--- a/src/IceRpc.Extensions.DependencyInjection/ClientConnectionServiceCollectionExtensions.cs
+++ b/src/IceRpc.Extensions.DependencyInjection/ClientConnectionServiceCollectionExtensions.cs
@@ -11,7 +11,7 @@ using System.Net.Quic;
 
 namespace IceRpc.Extensions.DependencyInjection;
 
-/// <summary>Provides an extension method for <see cref="IServiceCollection" /> to add a client connection.</summary>
+/// <summary>Provides extension methods for <see cref="IServiceCollection" /> to add a client connection.</summary>
 public static class ClientConnectionServiceCollectionExtensions
 {
     /// <summary>Adds a <see cref="ClientConnection" /> singleton that connects to the specified server address to this
--- a/src/IceRpc.Extensions.DependencyInjection/ClientConnectionServiceCollectionExtensions.cs
+++ b/src/IceRpc.Extensions.DependencyInjection/ClientConnectionServiceCollectionExtensions.cs
@@ -14,8 +14,34 @@ namespace IceRpc.Extensions.DependencyInjection;
 /// <summary>Provides an extension method for <see cref="IServiceCollection" /> to add a client connection.</summary>
 public static class ClientConnectionServiceCollectionExtensions
 {
-    /// <summary>Adds a <see cref="ClientConnection" /> and an <see cref="IInvoker" /> to this service collection.
-    /// </summary>
+    /// <summary>Adds a <see cref="ClientConnection" /> singleton that connects to the specified server address to this
+    /// service collection; this singleton is also registered as the <see cref="IInvoker" /> singleton.</summary>
+    /// <param name="services">The service collection to add services to.</param>
+    /// <param name="serverAddress">The server address of the client connection.</param>
+    /// <returns>The service collection.</returns>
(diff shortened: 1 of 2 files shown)


externl (approved):
Looks good.

- No named-options overload, so two `AddIceRpcClientConnection(address)` calls silently collapse: both `Configure` callbacks target the unnamed `ClientConnectionOptions` so the second address wins, while `AddSingleton` registers two `ClientConnection`s that both point at it. `AddIceRpcServer` avoids exactly this with its `optionsName` overloads; there's no client-connection equivalent.
- The test's `Throws.Nothing` only proves *an* address was set — `ClientConnection` throws when `ServerAddress` is null, so any address passes. Asserting `provider.GetRequiredService<IOptions<ClientConnectionOptions>>().Value.ServerAddress` would pin the actual value. The `ServerAddress` overload is untested.

src/IceRpc.Extensions.DependencyInjection/ClientConnectionServiceCollectionExtensions.cs
```diff
@@ -14,8 +14,34 @@ namespace IceRpc.Extensions.DependencyInjection;
 /// <summary>Provides an extension method for <see cref="IServiceCollection" /> to add a client connection.</summary>
```
InsertCreativityHere: This comment is out-dated now that we provide multiple methods instead of just one.

InsertCreativityHere (approved):
Looks good to me!

bernardnormier:
- The test now pins `ClientConnectionOptions.ServerAddress` to the exact address passed in. It covers the `Uri` overload, which delegates to the `ServerAddress` one.
- The named-options gap deserves its own design pass (named options plus how the `IInvoker` registration disambiguates) — filed #4859.

src/IceRpc.Extensions.DependencyInjection/ClientConnectionServiceCollectionExtensions.cs
bernardnormier: Fixed.

PR #4862, opened 2026-08-21
Validate the limits of IceFeature and SliceFeature

### LINKED ISSUE #4808: [Audit-Medium] Ice and Slice feature limits allow arithmetic overflow and unusable configurations

## Summary

The feature constructors compute the collection-allocation budget with unchecked `int` arithmetic:

- `IceFeature`: `8 * maxPayloadSize`
- `SliceFeature`: `8 * maxSegmentSize`

Large accepted values wrap negative — `int.MaxValue` yields `-8`. Ice additionally accepts `int.MaxValue` while the read path computes `maxSize + 1`, which wraps to `int.MinValue`. `MaxDepth == 0` is accepted even though it produces an unusable decoder.

## Source references

- [src/IceRpc.Ice/Features/IceFeature.cs#L64](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Ice/Features/IceFeature.cs#L64) and [#L81](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Ice/Features/IceFeature.cs#L81)
- [src/IceRpc.Slice/Features/SliceFeature.cs#L49](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Slice/Features/SliceFeature.cs#L49) and [#L62](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Slice/Features/SliceFeature.cs#L62)
- [src/IceRpc.Ice/Operations/Internal/PipeReaderExtensions.cs](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Ice/Operations/Internal/PipeReaderExtensions.cs) — the `maxSize + 1` read path

## Recommendation

Validate public option ranges against all downstream arithmetic. Use checked or widened arithmetic and reject configurations the decoder/read APIs cannot represent.

This is a different site from the already-fixed decoder-accounting overflows (icerpc-csharp-audit#5 and #7) — those covered allocation accounting; this is the public feature constructor.

## Audit context

M-05 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md); confirmed by the [Claude verification pass](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-claude-review-2026-07-28.md).

Severity: **Medium**.

bernardnormier:
Fixes #4808.

The `IceFeature` and `SliceFeature` constructors accepted limits their consumers can't handle:

- The derived collection-allocation budget (`8 * maxPayloadSize` / `8 * maxSegmentSize`) wrapped negative for sizes above `int.MaxValue / 8`.
- `IceFeature` accepted a max payload size of `int.MaxValue` while the Ice payload reader computes `maxSize + 1`, and accepted a max depth of `0` that `IceDecoder` rejects later, at decode time.
- Negative values other than `-1` were silently treated as "use the default", while the doc-comments only allow `-1`.

The constructors now throw `ArgumentOutOfRangeException` for all of these, per the documented contract:

- `maxCollectionAllocation`: `0`, a positive value, or `-1`.
- `maxSegmentSize`: a positive value or `-1` — the smallest Slice segment is 1 byte (the tag end marker).
- `maxPayloadSize`: `0`, a positive value, or `-1` — the smallest Ice payload is empty (a void operation); also rejects `int.MaxValue` (the Ice read path computes `maxSize + 1`).
- Both: at most `int.MaxValue / 8` (256 MB) when `maxCollectionAllocation` is `-1` (the derived `8 ×` budget must not overflow).
- `maxDepth`: a positive value or `-1`.

The Slice segment reader's asserts required `maxSize < int.MaxValue` without an arithmetic reason (the segment size decode uses a checked cast and a plain comparison), so they are relaxed to `maxSize > 0`; the Ice reader's asserts are relaxed to `maxSize >= 0` since the empty payload is a valid Ice payload.

## What's Changed entry

None — only rejects invalid configurations that no reasonable program uses.

### DIFF
--- a/src/IceRpc.Ice/Features/IIceFeature.cs
+++ b/src/IceRpc.Ice/Features/IIceFeature.cs
@@ -24,18 +24,20 @@ public interface IIceFeature
     IceEncodeOptions? EncodeOptions { get; }
 
     /// <summary>Gets the maximum collection allocation when decoding a payload, in bytes.</summary>
-    /// <value>The maximum collection allocation. Must be greater than or equal to <c>0</c>.</value>
+    /// <value>The maximum collection allocation.</value>
+    /// <remarks>Implementations must return a value greater than or equal to <c>0</c>.</remarks>
     int MaxCollectionAllocation { get; }
 
     /// <summary>Gets the maximum depth when decoding a class recursively.</summary>
-    /// <value>The maximum depth. Must be greater than <c>0</c>.</value>
+    /// <value>The maximum depth.</value>
+    /// <remarks>Implementations must return a value greater than <c>0</c>.</remarks>
     int MaxDepth { get; }
 
     /// <summary>Gets the maximum size of an Ice-encoded payload, in bytes. An Ice-encoded payload corresponds to the
     /// encoded arguments of an operation, or the encoded return values of an operation.</summary>
-    /// <value>The maximum size of an Ice-encoded payload, in bytes. Must be greater than or equal to <c>0</c> and less
-    /// than <see cref="int.MaxValue" />.</value>
+    /// <value>The maximum size of an Ice-encoded payload, in bytes.</value>
     /// <remarks>The payload size does not include the size of any header for this payload, such as the encapsulation
-    /// header with the ice protocol.</remarks>
+    /// header with the ice protocol. Implementations must return a value greater than or equal to <c>0</c> and less
+    /// than <see cref="int.MaxValue" />.</remarks>
     int MaxPayloadSize { get; }
 }
--- a/src/IceRpc.Ice/Features/IIceFeature.cs
+++ b/src/IceRpc.Ice/Features/IIceFeature.cs
@@ -33,8 +33,8 @@ public interface IIceFeature
 
     /// <summary>Gets the maximum size of an Ice-encoded payload, in bytes. An Ice-encoded payload corresponds to the
     /// encoded arguments of an operation, or the encoded return values of an operation.</summary>
-    /// <value>The maximum size of an Ice-encoded payload, in bytes. Must be greater than <c>0</c> and less than
-    /// <see cref="int.MaxValue" />.</value>
+    /// <value>The maximum size of an Ice-encoded payload, in bytes. Must be greater than or equal to <c>0</c> and less
+    /// than <see cref="int.MaxValue" />.</value>
     /// <remarks>The payload size does not include the size of any header for this payload, such as the encapsulation
     /// header with the ice protocol.</remarks>
     int MaxPayloadSize { get; }
--- a/src/IceRpc.Ice/Features/IIceFeature.cs
+++ b/src/IceRpc.Ice/Features/IIceFeature.cs
@@ -24,16 +24,17 @@ public interface IIceFeature
     IceEncodeOptions? EncodeOptions { get; }
 
     /// <summary>Gets the maximum collection allocation when decoding a payload, in bytes.</summary>
-    /// <value>The maximum collection allocation.</value>
+    /// <value>The maximum collection allocation. Must be greater than or equal to <c>0</c>.</value>
     int MaxCollectionAllocation { get; }
 
     /// <summary>Gets the maximum depth when decoding a class recursively.</summary>
-    /// <value>The maximum depth.</value>
+    /// <value>The maximum depth. Must be greater than <c>0</c>.</value>
     int MaxDepth { get; }
 
     /// <summary>Gets the maximum size of an Ice-encoded payload, in bytes. An Ice-encoded payload corresponds to the
     /// encoded arguments of an operation, or the encoded return values of an operation.</summary>
-    /// <value>The maximum size of an Ice-encoded payload, in bytes.</value>
+    /// <value>The maximum size of an Ice-encoded payload, in bytes. Must be greater than <c>0</c> and less than
+    /// <see cref="int.MaxValue" />.</value>
     /// <remarks>The payload size does not include the size of any header for this payload, such as the encapsulation
     /// header with the ice protocol.</remarks>
     int MaxPayloadSize { get; }
--- a/src/IceRpc.Ice/Features/IceFeature.cs
+++ b/src/IceRpc.Ice/Features/IceFeature.cs
@@ -21,17 +21,15 @@ public sealed class IceFeature : IIceFeature
     public IceEncodeOptions? EncodeOptions { get; }
 
     /// <inheritdoc/>
+    /// <value>The maximum collection allocation. Defaults to 8 times <see cref="MaxPayloadSize" />.</value>
     public int MaxCollectionAllocation { get; }
 
-    /// <summary>Gets the maximum depth when decoding a class recursively.</summary>
+    /// <inheritdoc/>
     /// <value>The maximum depth. Defaults to <c>100</c>.</value>
     public int MaxDepth { get; }
 
-    /// <summary>Gets the maximum size of an Ice-encoded payload, in bytes. An Ice-encoded payload corresponds to the
-    /// encoded arguments of an operation, or the encoded return values of an operation.</summary>
+    /// <inheritdoc/>
     /// <value>The maximum size of an Ice-encoded payload, in bytes. Defaults to <c>1</c> MB.</value>
-    /// <remarks>The payload size does not include the size of any header for this payload, such as the encapsulation
-    /// header with the ice protocol.</remarks>
     public int MaxPayloadSize { get; }
 
     /// <summary>Constructs an Ice feature.</summary>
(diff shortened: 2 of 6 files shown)


externl (commented):
Two things:

- **The guards validate the parameters, not the resulting feature.** When `defaultFeature` is a custom `IIceFeature`, `MaxPayloadSize` and `MaxDepth` come straight from it (`IceFeature.cs:96,98`) and skip every new check — so `new IceFeature(defaultFeature: mine)` still yields `MaxPayloadSize == int.MaxValue` / `MaxDepth == 0`, the exact states this PR sets out to make impossible. Same at `SliceFeature.cs:74`. Checking the resolved values after the `??`/ternary resolution would close it.

- **The `maxSize + 1` overflow is still reachable at the read site.** `IncomingFrameExtensions.cs:29,72` passes `feature.MaxPayloadSize` from whatever `IIceFeature` is in the request features, so a custom implementation returning `int.MaxValue` trips the `Debug.Assert` and passes `int.MinValue` to `ReadAtLeastAsync` in release. Clamping in `ReadFullPayloadAsync` — `(int)Math.Min((long)maxSize + 1, int.MaxValue)` — fixes it at the source rather than at one producer of `maxSize`.

bernardnormier:
I addressed both points by documenting the valid range of each limit in `IIceFeature` / `ISliceFeature`: implementations — including a custom `defaultFeature` — are trusted to honor this contract, like all other users of these interfaces trust it. The constructor checks only validate the arguments passed to `IceFeature` / `SliceFeature` itself.

Same reasoning for the `maxSize + 1` site: `MaxPayloadSize` is documented to be less than `int.MaxValue`, and the `Debug.Assert` catches a non-conforming implementation. Clamping there would silently paper over a bug in the feature implementation.

externl (dismissed):
Looks good!

src/IceRpc.Ice/Features/IceFeature.cs
```diff
@@ -46,6 +46,11 @@ public sealed class IceFeature : IIceFeature
     /// <param name="baseProxy">The base proxy, used when decoding service addresses into proxies.</param>
     /// <param name="defaultFeature">A feature that provides default values for all parameters. <see langword="null" />
     /// is equivalent to <see cref="Default" />.</param>
+    /// <exception cref="ArgumentOutOfRangeException">Thrown when <paramref name="maxCollectionAllocation" /> is a
+    /// negative value other than <c>-1</c>, when <paramref name="maxDepth" /> is <c>0</c> or a negative value other
+    /// than <c>-1</c>, when <paramref name="maxPayloadSize" /> is <c>0</c>, <see cref="int.MaxValue" /> or a negative
```
InsertCreativityHere: Missing an oxford comma here. And given the length of the list, this is one time it would actually be helpful for clarity!
```suggestion
    /// than <c>-1</c>, when <paramref name="maxPayloadSize" /> is <c>0</c>, <see cref="int.MaxValue" />, or a negative
```

src/IceRpc.Ice/Features/IceFeature.cs
InsertCreativityHere: We reject `maxDepth == 0` now, so it feels counter to explicitly check against `0` here:
```suggestion
        MaxDepth = maxDepth > 0 ? maxDepth : defaultFeature.MaxDepth;
```

src/IceRpc.Slice/Features/SliceFeature.cs
InsertCreativityHere: Same here. We reject 0, so no point checking for `=0`:

```suggestion
        MaxSegmentSize = maxSegmentSize > 0 ? maxSegmentSize : defaultFeature.MaxSegmentSize;
```

src/IceRpc.Ice/Features/IceFeature.cs
InsertCreativityHere, line 96: Same here, `maxPayloadSize` can never be 0, and if we _somehow_ made it to this line with `0`, this logic would violate the contract of `IIceFeature` by setting `MaxPayloadSize` to `0`.
```suggestion
        MaxPayloadSize = maxPayloadSize > 0 ? maxPayloadSize : defaultFeature.MaxPayloadSize;
```

InsertCreativityHere, line 92: Here too. `>` is sufficient over `>=`.
```suggestion
            (maxPayloadSize > 0 ? 8 * maxPayloadSize : defaultFeature.MaxCollectionAllocation);
```

src/IceRpc.Slice/Features/SliceFeature.cs
InsertCreativityHere: Same:
```suggestion
            (maxSegmentSize > 0 ? 8 * maxSegmentSize : defaultFeature.MaxCollectionAllocation);
```

InsertCreativityHere (dismissed):
Looks good to me. Just one "nit" as we're calling it now.

Most of the assignments check `xxx >= 0`, despite the fact that `0` is an explicitly invalid value for them now. So, we should really just be using `xxx > 0` instead.

src/IceRpc.Ice/Features/IceFeature.cs
bernardnormier: This list is now down to two items — `maxPayloadSize` accepts `0` since the empty payload is a valid Ice payload (see the other threads) — so the comma question went away.

src/IceRpc.Ice/Features/IceFeature.cs
bernardnormier: We decided to allow `maxPayloadSize = 0`: it's the true floor of the encoding — a void operation's payload is empty. So this check stays `>= 0`.

bernardnormier: Same as above — `0` is valid, so `>= 0`.

externl (approved):
Looks good.

One nit: the `MaxPayloadSize` `<value>` doc in `IIceFeature.cs` still says "Must be greater than `0`" — now that 0 is valid it should read "greater than or equal to `0`".

InsertCreativityHere (dismissed):
Looks good to me!

Maybe there's a separate issue open for this, but we still have the same old pattern present in our protobuf features. It'd be good to also update that in the same way we updated these.

src/IceRpc.Ice/Features/IIceFeature.cs
```diff
@@ -24,16 +24,17 @@ public interface IIceFeature
     IceEncodeOptions? EncodeOptions { get; }
 
     /// <summary>Gets the maximum collection allocation when decoding a payload, in bytes.</summary>
-    /// <value>The maximum collection allocation.</value>
+    /// <value>The maximum collection allocation. Must be greater than or equal to <c>0</c>.</value>
     int MaxCollectionAllocation { get; }
 
     /// <summary>Gets the maximum depth when decoding a class recursively.</summary>
-    /// <value>The maximum depth.</value>
+    /// <value>The maximum depth. Must be greater than <c>0</c>.</value>
```
InsertCreativityHere: The doc-comments for `MaxDepth` and `MaxPayloadSize` got updated here in `IIceFeature`, but not in `IceFeature`, which is arguably the more important place since that's the concrete types users will often use. Note that we aren't using `inheritdoc` there for these two properties.

This might also be true for the doc-comments in `ISliceFeature` vs `SliceFeature` as well.

src/IceRpc.Ice/Features/IceFeature.cs
```diff
@@ -55,6 +60,31 @@ public IceFeature(
         IIceProxy? baseProxy = null,
         IIceFeature? defaultFeature = null)
     {
+        if (maxCollectionAllocation < -1)
```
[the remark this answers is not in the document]
bernardnormier: Not adding constructor validation tests in this PR.

src/IceRpc.Slice/Features/SliceFeature.cs
```diff
@@ -34,21 +34,44 @@ public sealed class SliceFeature : ISliceFeature
...
+    /// other than <c>-1</c>, or when <paramref name="maxSegmentSize" /> is greater than <c>int.MaxValue / 8</c>
+    /// (256 MB) while <paramref name="maxCollectionAllocation" /> is <c>-1</c>.</exception>
     public SliceFeature(
         SliceEncodeOptions? encodeOptions = null,
         int maxCollectionAllocation = -1,
         int maxSegmentSize = -1,
         ISliceProxy? baseProxy = null,
         ISliceFeature? defaultFeature = null)
     {
+        if (maxCollectionAllocation < -1)
```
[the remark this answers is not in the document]
bernardnormier: Not adding constructor validation tests in this PR.

bernardnormier:
Filed #4887 for the same pattern in ProtobufFeature.

src/IceRpc.Ice/Features/IIceFeature.cs
bernardnormier: Reworked in fd9ffd327: the interfaces now document the consumer-facing value in <value> and the implementation contract as an "Implementations must return ..." remark; IceFeature / SliceFeature use inheritdoc with a <value> override that adds only the default value (locally defined tags take precedence over inherited ones). Same treatment in both features.

InsertCreativityHere (approved):
Looks good! Ship it!

Issue #4813, opened 2026-07-28
[Audit-Medium] Generators do not safely emit all legal C# namespace and type names


pepone:
> **AI-generated audit finding** — from a repository-wide audit performed by OpenAI Codex (`codex-cli`), independently verified against source by Claude (Anthropic), model **Opus 5** (`claude-opus-5`). It **has not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

Two related name-emission failures.

**1 — namespace segments are not escaped.** A valid C# namespace built from a keyword is emitted verbatim and does not parse. Reproduced with an isolated probe:

```slice
[cs::identifier("namespace")]
module Keyword
```

produces, in **both** generated files:

```csharp
namespace namespace;
```

```
error CS1001: Identifier expected
error CS1514: { expected
```

The `@` escaping is never applied.

**2 — framework references are not globally qualified.** Generated files emit bare `using IceRpc.Slice;` rather than `global::`-qualified references. A sibling namespace such as `Company.IceRpc`, or a Protobuf message named `IceRpc`, captures name lookup and produces missing-type errors.

## Source references

- [src/IceRpc.ServiceGenerator/Internal/SymbolExtensions.cs](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.ServiceGenerator/Internal/SymbolExtensions.cs)
- [src/IceRpc.ServiceGenerator/Internal/Emitter.cs](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.ServiceGenerator/Internal/Emitter.cs)
- [src/IceRpc.Protobuf.Generator/ClientGenerator.cs](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Protobuf.Generator/ClientGenerator.cs)
- [src/IceRpc.Protobuf.Generator/ServiceGenerator.cs](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Protobuf.Generator/ServiceGenerator.cs)

## Recommendation

Use Roslyn-aware identifier escaping for namespace segments, and globally qualify framework/runtime references emitted by every generator family.

## Audit context

[...]

PR #4864, opened 2026-08-24
Document the cs::identifier pass-through and the reserved namespaces

bernardnormier:
Records the disposition of #4813 in AGENTS.md's dismissed audit patterns:

- Pattern 1 (trusted IDL inputs) now says explicitly that `cs::identifier` emits exactly the string the user specifies: the generators do not validate, `@`-escape, or otherwise fix a "bad" identifier such as a C# keyword.
- New pattern 14: `IceRpc` and `ZeroC` are reserved namespaces — a project that consumes generated code must not define its own types or namespaces with these names. Generated code relies on this reservation and references framework namespaces unqualified, so findings proposing `global::`-qualified emission across the generators are rejected.

## What's Changed entry

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,7 +59,7 @@ values (cs::type, cs::attribute, cs::identifier, csharp_namespace, deprecated me
 nested types, slicec symbol output) surface as C# compiler errors on the generated source, which is an acceptable
 failure mode. In particular, `cs::identifier` emits exactly the string the user specifies: the generators do not
 validate, `@`-escape, or otherwise fix a "bad" identifier such as a C# keyword.
-(See #4444, #4449, #4459, #4467, #4470, #4485, #4495, #4496, #4497, #4502, #4503; the cs::identifier variant is #4813.)
+(See #4444, #4449, #4459, #4467, #4470, #4485, #4495, #4496, #4497, #4502, #4503, #4813.)
 
 ### 2. Trusted toolchain downloads
 
@@ -155,4 +155,5 @@ them for the lifetime of the process. Don't file findings proposing weak-key cac
 namespaces with these names. Generated code relies on this reservation — it references framework namespaces with
 ordinary `using` directives and unqualified names, not `global::` qualification on every reference. Shadowing a
 reserved namespace breaks the compilation of the generated code; that's an error in the consuming project, not in the
-generators. Don't file findings proposing `global::`-qualified emission across the generators. (#4813.)
+generators. Don't file findings proposing to `global::`-qualify the generated references to these namespaces as a
+defense against shadowing. (#4813.)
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -57,8 +57,9 @@ For dismissals newer than this section, also consult:
 data. Code-injection framing does not apply: anyone who can author IDL in a project can already author C#. Malformed
 values (cs::type, cs::attribute, cs::identifier, csharp_namespace, deprecated message, attribute payloads, deeply
 nested types, slicec symbol output) surface as C# compiler errors on the generated source, which is an acceptable
-failure mode.
-(See #4444, #4449, #4459, #4467, #4470, #4485, #4495, #4496, #4497, #4502, #4503.)
+failure mode. In particular, `cs::identifier` emits exactly the string the user specifies: the generators do not
+validate, `@`-escape, or otherwise fix a "bad" identifier such as a C# keyword.
+(See #4444, #4449, #4459, #4467, #4470, #4485, #4495, #4496, #4497, #4502, #4503; the cs::identifier variant is #4813.)
```
InsertCreativityHere: Pretty small nit, but feels odd to give "the cs::identifier variant is ..." special treatment, when all the others are just a list of PR numbers. Should probably drop the extra words and just make `#4813` normal like all the rest.

AGENTS.md
```diff
@@ -147,3 +148,11 @@ and, through its recursive merge, for every marked assembly it references — in
...
 them for the lifetime of the process. Don't file findings proposing weak-key caching or unload-safety for this cache.
 (#4827.)
+
+### 14. IceRpc and ZeroC are reserved namespaces
+
+`IceRpc` and `ZeroC` are reserved namespaces: a project that consumes generated code must not define its own types or
+namespaces with these names. Generated code relies on this reservation — it references framework namespaces with
+ordinary `using` directives and unqualified names, not `global::` qualification on every reference. Shadowing a
+reserved namespace breaks the compilation of the generated code; that's an error in the consuming project, not in the
+generators. Don't file findings proposing `global::`-qualified emission across the generators. (#4813.)
```
InsertCreativityHere: The last sentence feels a too strong to me: "Never propose `global::`-qualified emission in a generator" is much broader than just "IceRpc and ZeroC are reserved" and I think could be misinterpreted.

For example, we actually emit `global::`-qualified code all the time in our generators. Which feels counter to what the sentence states.

InsertCreativityHere (approved):
Looks good, but I'd reword/remove the last sentence about `global::` for being too broad.

externl (approved):
Looks good.

AGENTS.md
bernardnormier: Done — folded `#4813` into the list.

AGENTS.md
bernardnormier: Agreed — reworded to target only the shadowing defense: "Don't file findings proposing to `global::`-qualify the generated references to these namespaces as a defense against shadowing."

Issue #4830, opened 2026-07-28
[Audit-Low] Example runtime and robustness defects (5 findings)


pepone:
> **AI-generated audit findings** — from a repository-wide audit performed by OpenAI Codex (`codex-cli`). ⚠️ Unlike the High/Medium findings from the same audit, **these Low-severity items were not individually re-verified** by the follow-up Claude pass — they are grouped here for triage rather than asserted as confirmed. Verify each before acting on it.

Five example defects from the same audit, grouped for triage.

| ID | Finding |
|---|---|
| L-18 | Slice and Protobuf Thermostat examples fire-and-forget `PublishAsync`; non-cancellation failures become unobserved and skip cancellation-source cleanup. Channel-writer removal also happens after the iterator loop rather than in `finally`, so early disposal can retain a monitor |
| L-20 | Both TcpFallback servers call `quicServer.Listen()` before establishing the TCP listener — on a platform without QUIC the first call throws and the fallback endpoint never starts, defeating the example's whole point |
| L-23 | Upload/Download `Earth.png` assets contain JPEG data, so extension, content type and documentation disagree |
| L-24 | Slice Retry does not validate its server-count option; zero or negative reaches `Enumerable.Range(1, serverInstances - 1)` and fails with an incidental argument exception instead of the example's own message |
| L-25 | CurrentWeather uses `ReadAtLeastAsync(256)` as though it imposes a maximum; overlength data can be accepted when the final buffer is already complete, and a legal boundary-sized value can be rejected when completion arrives separately. The Program comment also incorrectly calls the default listener TCP |

## Note

L-20 is the most user-visible: the fallback example does not fall back. Worth doing first.

See also the separately filed M-27 (detached reader completion) and M-28 (Telemetry example bypasses its pipeline), which are the Medium members of this family.

## Audit context

L-18, L-20, L-23, L-24, L-25 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md).

externl:
L-24's client half is fixed in #4868. The Retry server has the same gap: it accepts a negative server number, which builds `icerpc://[::0]:{10000 + number}/` and fails with a `UriFormatException` instead of the example's own message.

PR #4867, opened 2026-08-24
Always complete the pipe reader in the Upload and Download examples

bernardnormier:
Fixes #4825.

When the generated code gives a `PipeReader` to the application — the response stream on the client side, the stream argument on the dispatch side — the application owns this pipe reader and must complete it. The Download client never completed the reader it received, and the Upload server completed it only on the success path: a failed or canceled copy skipped it. Both examples now complete the reader in a `finally` block, since these examples are the pattern users copy for byte streaming.

## What's Changed entry

None — example fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4868, opened 2026-08-24
Fix example defects: image file type, retry validation, query length check

bernardnormier:
Addresses findings L-20, L-23, L-24, and L-25 of #4830; the L-18 fix (Thermostat publish cleanup) follows in a separate PR.

- **L-23**: the Upload and Download `Earth.png` assets contain JPEG data. Renamed the assets and the produced files to `.jpg` so extension and content agree; the bytes are unchanged.
- **L-24**: the Retry client now rejects a zero or negative server instances argument with its own message instead of failing later with an `ArgumentOutOfRangeException`.
- **L-25**: the CurrentWeather server read the query string with `ReadAtLeastAsync(MaxQueryLength)` and `!IsCompleted`, which neither enforces the maximum (an overlength query arriving in a single buffer was accepted) nor reliably accepts a boundary-length query. It now reads `MaxQueryLength + 1` bytes and checks the buffer length, calling `AdvanceTo` before rejecting. Also corrected the Program.cs comment that called the default listener TCP.
- **L-20 is not a defect**: the TcpFallback example demonstrates a *client* falling back to TCP when QUIC connection establishment fails (e.g. a middlebox blocks QUIC); the server's role is to accept both kinds of clients and it intentionally requires a platform with QUIC support.

## What's Changed entry

None — example fixes.


examples/slice/Download/Server/images/Earth.jpg
InsertCreativityHere, line 1: Can confirm, this actually is a JPEG... First 4 bytes are `FF D8 FF E0`; JPEG magic.

InsertCreativityHere (approved):
Looks good to me!

externl (commented):
- The `.gitignore` in `examples/slice/Upload/Server/uploads/` and `examples/slice/Download/Client/downloads/` still lists `*.png`, so running either example now leaves the generated `.jpg` untracked in `git status`.

externl (approved):
Looks good.

bernardnormier:
Done — both .gitignore files now list `*.jpg`. Also merged main to resolve the conflicts with #4867 (the try/finally now wraps the `.jpg` file creation).

PR #4869, opened 2026-08-24
Fix reading publication cleanup in the Thermostat examples

### LINKED ISSUE #4830: [Audit-Low] Example runtime and robustness defects (5 findings)

bernardnormier:
Fixes #4830 — finding L-18; the other findings of this issue are addressed by #4868.

In both Thermostat servers, `ThermoBridge` intentionally does not await `PublishAsync`, so `PublishAsync` must not let exceptions escape. It caught only `OperationCanceledException`: a device connection loss made the `await foreach` throw `IceRpcException`, which faulted the discarded task (an unobserved task exception) and skipped the `publishCts` cleanup. `PublishAsync` now also catches `IceRpcException` with a console message and performs the cleanup in a `finally` block.

In `MonitorAsync`, the removal of the per-client channel writer ran after the iterator loop, so it was skipped when the consumer disposed the enumerator while it was suspended at a `yield return` — retaining the writer (and publishing to it) until server shutdown. The removal now runs in a `finally` block.

## What's Changed entry

None — example fix.

### DIFF
--- a/examples/protobuf/Thermostat/Server/ThermoFacade.cs
+++ b/examples/protobuf/Thermostat/Server/ThermoFacade.cs
@@ -72,21 +72,28 @@ internal sealed partial class ThermoFacade : IThermostatService
             // We stop yielding new values when the server shuts down or the client disconnects or stops reading.
             using var cts = CancellationTokenSource.CreateLinkedTokenSource(_shutdownToken, cancellationToken);
 
-            while (!cts.IsCancellationRequested)
+            try
             {
-                Reading reading;
-                try
+                while (!cts.IsCancellationRequested)
                 {
-                    reading = await channel.Reader.ReadAsync(cts.Token);
+                    Reading reading;
+                    try
+                    {
+                        reading = await channel.Reader.ReadAsync(cts.Token);
+                    }
+                    catch
+                    {
+                        break; // while
+                    }
+                    yield return reading;
                 }
-                catch
-                {
-                    break; // while
-                }
-                yield return reading;
             }
-
-            RemoveChannelWriter(node);
+            finally
+            {
+                // The consumer can dispose the enumerator while it's suspended at a yield return; only the
+                // finally block runs then.
+                RemoveChannelWriter(node);
+            }
         }
     }
 
@@ -146,14 +153,22 @@ internal sealed partial class ThermoFacade : IThermostatService
         {
             // Expected on shutdown or when superseded by a newer publish task.
         }
-
-        lock (_mutex)
+        catch (IceRpcException exception)
         {
-            // Cleanup unless a new publish task has already disposed publishCts.
-            if (_publishCts == publishCts)
+            // Expected when the device connection is lost. This method's task is not awaited, so we must not let
+            // the exception escape.
+            Console.WriteLine($"Stopped publishing readings: {exception.Message}");
+        }
+        finally
+        {
+            lock (_mutex)
             {
-                publishCts.Dispose();
-                _publishCts = null;
+                // Cleanup unless a new publish task has already disposed publishCts.
+                if (_publishCts == publishCts)
+                {
+                    publishCts.Dispose();
+                    _publishCts = null;
+                }
             }
         }
     }
--- a/examples/protobuf/Thermostat/Server/ThermoFacade.cs
+++ b/examples/protobuf/Thermostat/Server/ThermoFacade.cs
@@ -61,7 +61,8 @@ internal sealed partial class ThermoFacade : IThermostatService
             // We stop yielding new values when the server shuts down or the client disconnects or stops reading.
             using var cts = CancellationTokenSource.CreateLinkedTokenSource(_shutdownToken, cancellationToken);
 
-            // Each enumeration gets its own bounded channel with a single element.
+            // Each enumeration gets its own bounded channel holding a single reading: when the client falls behind, the
+            // channel keeps only the latest reading and drops the older ones.
             var channel = Channel.CreateBounded<Reading>(
                 new BoundedChannelOptions(1)
                 {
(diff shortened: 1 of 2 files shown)


externl (commented):
- `AddChannelWriter` is called before the iterator is returned, but `RemoveChannelWriter` is in the iterator's `finally`, which doesn't run if the body never started. `AsyncEnumerablePipeReader` creates the enumerator in its constructor and `Complete()` disposes it without advancing — reachable when the client disconnects between the dispatch returning and the first read of the payload continuation, leaving the writer registered until shutdown. Moving `AddChannelWriter` inside `ReadAsync` would put both under the iterator's control; the `_latestReading` seed means the client still gets the current reading on first read. Same applies to the Protobuf copy.

bernardnormier:
Good catch — done in both copies: `AddChannelWriter` now runs at the top of the iterator body, so it's paired with the `RemoveChannelWriter` in the finally block: both run when the enumeration starts, or neither runs when the consumer disposes the enumerator without ever starting it. The channel creation moved into the iterator as well, since nothing else uses it.

externl (approved):
Looks good.

examples/protobuf/Thermostat/Server/ThermoFacade.cs
```diff
@@ -146,14 +155,22 @@ internal async Task PublishAsync(IAsyncStream<Reading> readings)
...
             // Expected on shutdown or when superseded by a newer publish task.
         }
-
-        lock (_mutex)
+        catch (IceRpcException exception)
+        {
+            // Expected when the device connection is lost. This method's task is not awaited, so we must not let
+            // the exception escape.
+            Console.WriteLine($"Stopped publishing readings: {exception.Message}");
+        }
```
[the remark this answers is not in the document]
bernardnormier: Right — the decode exceptions escaped the stated no-escape requirement. The second catch now catches all exceptions.

examples/slice/Thermostat/Server/ThermoFacade.cs
```diff
@@ -143,14 +152,22 @@ internal async Task PublishAsync(IAsyncStream<Reading> readings)
...
             // Expected on shutdown or when superseded by a newer publish task.
         }
-
-        lock (_mutex)
+        catch (IceRpcException exception)
+        {
+            // Expected when the device connection is lost. This method's task is not awaited, so we must not let
+            // the exception escape.
+            Console.WriteLine($"Stopped publishing readings: {exception.Message}");
+        }
```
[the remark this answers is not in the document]
bernardnormier: Same fix as the Protobuf copy: the second catch now catches all exceptions.

examples/protobuf/Thermostat/Server/ThermoFacade.cs
```diff
@@ -72,21 +61,41 @@ async IAsyncEnumerable<Reading> ReadAsync([EnumeratorCancellation] CancellationT
...
-                }
-                catch
+                    SingleReader = true,
+                    SingleWriter = true,
+                    FullMode = BoundedChannelFullMode.DropOldest
+                });
+
+            // We pair AddChannelWriter with RemoveChannelWriter in the finally block: both run when the enumeration
+            // starts, or neither runs when the consumer disposes the enumerator without ever starting it.
+            LinkedListNode<ChannelWriter<Reading>> node = AddChannelWriter(channel.Writer);
```
pepone, line 76: The "both run when the enumeration starts" is confusing, maybe is clear to write "// Register the writer when iteration starts and unregister it when iteration ends or is disposed."

examples/protobuf/Thermostat/Server/ThermoFacade.cs
```diff
@@ -72,21 +61,41 @@ async IAsyncEnumerable<Reading> ReadAsync([EnumeratorCancellation] CancellationT
             // We stop yielding new values when the server shuts down or the client disconnects or stops reading.
             using var cts = CancellationTokenSource.CreateLinkedTokenSource(_shutdownToken, cancellationToken);
 
-            while (!cts.IsCancellationRequested)
-            {
-                Reading reading;
-                try
+            // Each enumeration gets its own bounded channel with a single element.
```
pepone: One thing not obvious with the design and the comment, is that we discard old values when a client falls behind.

pepone (approved):

examples/protobuf/Thermostat/Server/ThermoFacade.cs
bernardnormier: Reworded as suggested (in both ThermoFacade files).

examples/protobuf/Thermostat/Server/ThermoFacade.cs
bernardnormier: Added to the channel comment: when the client falls behind, the channel keeps only the latest reading and drops the older ones.

Issue #4832, opened 2026-07-28
[Audit-Low] Core, transport and codec edge cases (12 findings)


pepone:
> **AI-generated audit findings** — from a repository-wide audit performed by OpenAI Codex (`codex-cli`). ⚠️ Unlike the High/Medium findings from the same audit, **these Low-severity items were not individually re-verified** by the follow-up Claude pass — they are grouped here for triage rather than asserted as confirmed. Verify each before acting on it.

Twelve narrow correctness edge cases from the same audit, grouped for triage. Several are inconsistent-failure-mode issues rather than wrong results.

| ID | Finding |
|---|---|
| L-01 | `ServerAddress.Host` accepts `[::1]`; URI serialization adds a second pair of brackets, producing `icerpc://[[::1]]` and a `UriFormatException` |
| L-02 | Ice replies with a matching request ID are handed to the invocation before the reply header is fully decoded; malformed remainder data surfaces `IndexOutOfRangeException` or raw `InvalidDataException` on the caller path |
| L-03 | Once an icerpc stream is allocated, header encoding or payload-writer setup failure completes the stream output but can bypass completion of the request payload and continuation readers |
| L-04 | `IDuplexConnection.WriteAsync` documents `ArgumentException` for an empty buffer; `TcpConnection.WriteAsync` performs the empty socket/TLS write without validation |
| L-05 | Canceling a queued Coloc connect leaves the channel entry in place until an accept drains it, so a canceled full backlog can temporarily reject a valid connection |
| L-06 | `IMultiplexedStream` docs say a local stream has no ID and is not started until first write; QUIC exposes the ID and reports started from construction |
| L-07 | Client and server factories add one to `MaxIceRpcUnidirectionalStreams` without checked arithmetic, so `int.MaxValue` reaches the transport negative |
| L-08 | Coloc pause/resume validation depends on property assignment order; the pause threshold can be lowered below an existing resume threshold, failing only later inside `PipeOptions` |
| L-09 | The Ice decoder can read known fields beyond the declared slice body and compute a negative advance at the indirection table, yielding `ArgumentOutOfRangeException` instead of `InvalidDataException` |
| L-10 | Slice and Protobuf `ToPipeReader` call `GetAsyncEnumerator` in their constructors, so generated argument evaluation can create the reader before a null invoker or unsupported Ice continuation is rejected, leaving the enumerator unreachable |
[...]

PR #4871, opened 2026-08-25
Complete the request payloads when invoking with a null invoker

bernardnormier:
Addresses finding L-10 of #4832.

The generated proxy code creates the request payload — and, for a streamed Slice operation, the payload continuation whose construction starts the application's async enumerator — before calling `InvokeOperationAsync`. When the proxy has a null invoker, `InvokeOperationAsync` threw before any `OutgoingRequest` existed, so nothing completed these readers and the enumerator was unreachable. The Slice and Ice `InvokeOperationAsync` overloads now complete the payloads before throwing.

## What's Changed entry

None — resource cleanup on an error path caused by an application bug.


externl (approved):
Looks good.

PR #4873, opened 2026-08-25
Throw IceRpcException when an ice response header is invalid

bernardnormier:
Addresses finding L-02 of #4832.

With the ice protocol, the response header is decoded on the invocation code path, after the read frames loop hands over the reply frame. A malformed header made this decoding throw a raw `InvalidDataException` — or an `IndexOutOfRangeException` when the frame ended right after the request ID — directly to the application.

`InvokeAsync` now wraps response header decoding failures in an `IceRpcException` with error `IceRpcError.IceRpcError`, like the icerpc protocol connection does, and `DecodeResponseHeader` rejects an empty header with an `InvalidDataException`. The failure remains scoped to the invocation.

## What's Changed entry

None — improves the error thrown on malformed input.


externl (approved):
Looks good.

PR #4875, opened 2026-08-25
Store IPv6 hosts without brackets in ServerAddress

bernardnormier:
Addresses finding L-01 of #4832.

`Uri.CheckHostName` accepts both `::1` and `[::1]` as IPv6 hosts, but everything downstream of `ServerAddress.Host` assumes the unbracketed form: the URI serialization brackets any host containing `:`, so a bracketed host produced `icerpc://[[::1]]` and `ToUri()` threw `UriFormatException`. The `ServerAddress(Uri)` constructor was unaffected — `Uri.IdnHost` strips the brackets.

- The `Host` init setter now stores a bracketed IPv6 address without its brackets, like the `Uri` constructor does.
- `DecodeTcpServerAddressBody` (the other host validation relying on `Uri.CheckHostName`) now rejects a bracketed host: the host name in an Ice tcp/ssl server address is never bracketed, even when it's an IPv6 address.

## What's Changed entry

Area: Core

- Fixed a bug where setting `ServerAddress.Host` to a bracketed IPv6 address such as `[::1]` produced a server address whose `ToUri` throws `UriFormatException`. The brackets are now stripped on assignment.

### DIFF
--- a/src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs
+++ b/src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs
@@ -294,7 +294,8 @@ public static class IceProxyIceDecoderExtensions
     {
         var body = new TcpServerAddressBody(ref decoder);
 
-        if (Uri.CheckHostName(body.Host) == UriHostNameType.Unknown)
+        // The host name in an Ice tcp/ssl server address is never bracketed, even when it's an IPv6 address.
+        if (body.Host.StartsWith('[') || Uri.CheckHostName(body.Host) == UriHostNameType.Unknown)
         {
             throw new InvalidDataException($"Received service address with invalid host '{body.Host}'.");
         }
--- a/src/IceRpc/ServerAddress.cs
+++ b/src/IceRpc/ServerAddress.cs
@@ -32,7 +32,9 @@ public readonly record struct ServerAddress
             {
                 throw new ArgumentException($"Cannot set {nameof(Host)} to '{value}'.", nameof(value));
             }
-            _host = value;
+            // Brackets are only valid around an IPv6 address; store the address without them, like the Uri
+            // constructor does.
+            _host = value.StartsWith('[') ? value[1..^1] : value;
             OriginalUri = null; // new host invalidates OriginalUri
         }
     }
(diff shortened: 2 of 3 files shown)


externl (approved):
Looks good.

src/IceRpc/ServerAddress.cs
```diff
@@ -32,7 +32,9 @@ public string Host
             {
                 throw new ArgumentException($"Cannot set {nameof(Host)} to '{value}'.", nameof(value));
             }
-            _host = value;
+            // Brackets are only valid around an IPv6 address; store the address without them, like the Uri
+            // constructor does.
```
InsertCreativityHere: Can we add a note about mismatched brackets being rejected by the `if` check above or something?

I went down a rabbit hole wondering how this was safe for something for `[::1`, before learning that apparently `CheckHostName` enforces bracket pairing. But this behavior is not-at-all obvious here.

src/IceRpc/ServerAddress.cs
InsertCreativityHere, line 24: Should probably add a `<remarks>` about this bracket stripping.
Otherwise pretty surprising for a simple `string` property `init` to rewrite the string you pass in.

InsertCreativityHere (approved):
Looks good!

My Claude complained about the lack of testing `toUri` (the actual regression) and a lack of tests in general. I personally am not so convinced, so just making a note.

src/IceRpc/ServerAddress.cs
bernardnormier: Done - the comment now notes that CheckHostName returns Unknown for any bracketed value other than a well-formed bracketed IPv6 address, mismatched brackets included.

src/IceRpc/ServerAddress.cs
bernardnormier: Done.

src/IceRpc.Ice/Operations/IceProxyIceDecoderExtensions.cs
```diff
@@ -294,7 +294,8 @@ private static ServerAddress DecodeTcpServerAddressBody(this ref IceDecoder deco
     {
         var body = new TcpServerAddressBody(ref decoder);
 
-        if (Uri.CheckHostName(body.Host) == UriHostNameType.Unknown)
+        // The host name in an Ice tcp/ssl server address is never bracketed, even when it's an IPv6 address.
+        if (body.Host.StartsWith('[') || Uri.CheckHostName(body.Host) == UriHostNameType.Unknown)
```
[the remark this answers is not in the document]
bernardnormier: Not adding a test for this malformed-input rejection branch: it can only be reached with hand-corrupted wire bytes, and the only observable improvement it would lock in is the exception type.

PR #4876, opened 2026-08-25
Align transport implementations with their documented contracts

bernardnormier:
Addresses findings L-04, L-06 and L-07 of #4832.

- L-04: `IDuplexConnection.WriteAsync` documents `ArgumentException` for an empty buffer, and Coloc implements it, but `TcpConnection.WriteAsync` performed the empty socket/TLS write. It now throws like `TcpConnection.ReadAsync` already does for an empty read. A new `Write_with_empty_buffer_fails` conformance test covers all duplex transports — the read counterpart existed, the write one didn't.
- L-06: the `IMultiplexedStream.Id` and `IsStarted` docs stated that a local stream has no ID and is not started until the first write. That describes Slic; QUIC streams are started with an ID from construction. The docs now make the late start implementation-dependent.
- L-07: the client and server connection factories add one stream to `MaxIceRpcUnidirectionalStreams` for the icerpc protocol control stream; with `int.MaxValue` (previously a valid value) the addition wrapped negative. `MaxIceRpcBidirectionalStreams` and `MaxIceRpcUnidirectionalStreams` now reject values greater than `65,535` — the quic transport caps inbound stream counts at this value and Slic doesn't need more — and the control-stream addition saturates at `65,535`. #4880 tracks switching these properties to `ushort` in 0.7.

## What's Changed entry

Area: Core

- `ConnectionOptions.MaxIceRpcBidirectionalStreams` and `ConnectionOptions.MaxIceRpcUnidirectionalStreams` now reject values greater than `65,535`.

### DIFF
--- a/src/IceRpc/ConnectionOptions.cs
+++ b/src/IceRpc/ConnectionOptions.cs
@@ -82,14 +82,14 @@ public record class ConnectionOptions
     /// accepted on an icerpc connection. When this limit is reached, the peer is not allowed to open any new
     /// bidirectional stream. Since an bidirectional stream is opened for each two-way invocation, the sending of the
     /// two-way invocation will be delayed until another two-way invocation's stream completes.</summary>
-    /// <value>The maximum number of bidirectional streams. It can't be less than <c>1</c>. Defaults to <c>100</c>.
-    /// </value>
+    /// <value>The maximum number of bidirectional streams. It can't be less than <c>1</c> or greater than
+    /// <c>65,535</c>. Defaults to <c>100</c>.</value>
     public int MaxIceRpcBidirectionalStreams
     {
         get => _maxIceRpcBidirectionalStreams;
-        set => _maxIceRpcBidirectionalStreams = value > 0 ? value :
+        set => _maxIceRpcBidirectionalStreams = value is > 0 and <= ushort.MaxValue ? value :
             throw new ArgumentException(
-                $"{nameof(MaxIceRpcBidirectionalStreams)} can't be less than 1",
+                $"{nameof(MaxIceRpcBidirectionalStreams)} must be between 1 and {ushort.MaxValue}",
                 nameof(value));
     }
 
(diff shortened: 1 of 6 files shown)


externl (dismissed):
Looks good.

One note: the `int.MaxValue` saturation holds for Slic but not QUIC — `QuicConnectionOptions` caps inbound stream counts at `ushort.MaxValue`, so `MaxIceRpcUnidirectionalStreams = int.MaxValue` on the quic transport still throws at first connect. Worth correcting the description or noting the QUIC cap in the property doc.

bernardnormier:
Right — verified: System.Net.Quic accepts up to `65,535` inbound streams per direction and throws above that (msquic's stream counts are 16-bit). Since Slic doesn't need more than `65,535` either, these options now reject larger values outright, and the control-stream `+ 1` saturates at `65,535` — no transport-specific note needed in the transport-neutral `ConnectionOptions`. #4880 tracks switching the properties to `ushort` in 0.7.

externl (approved):
Looks good.

src/IceRpc/ConnectionOptions.cs
```diff
@@ -82,14 +82,14 @@ public int MaxIceFrameSize
     /// accepted on an icerpc connection. When this limit is reached, the peer is not allowed to open any new
     /// bidirectional stream. Since an bidirectional stream is opened for each two-way invocation, the sending of the
     /// two-way invocation will be delayed until another two-way invocation's stream completes.</summary>
-    /// <value>The maximum number of bidirectional streams. It can't be less than <c>1</c>. Defaults to <c>100</c>.
-    /// </value>
+    /// <value>The maximum number of bidirectional streams. It can't be less than <c>1</c> or greater than
+    /// <c>65,535</c>. Defaults to <c>100</c>.</value>
```
InsertCreativityHere, line 86: Probably for a separate issue if we decide to fix, but from a user-point-of-view, the real cap is actually 65534. If they set 65535, we can't honor it since we always need 1 stream (the `+1` for the control-stream). I don't think anything would break, but it is just a little weird that there's actually no difference between setting `65534` and `65535` here.

Also if we capped it like this, then we could drop the `int.Min`, since `65534 + 1` always fits within a `ushort`.

InsertCreativityHere (approved):

src/IceRpc/ConnectionOptions.cs
bernardnormier: See #4880.

And no, we should not cap it at 65534. If you set your max to `ushort.MaxValue`, you're not saying "I want exactly ushort.MaxValue streams", you're saying "I want as many streams as you can provide". So not adding an extra unidirectional stream in this case is completely fine.

PR #4877, opened 2026-08-25
Emit field attributes on variant enum record parameters

bernardnormier:
Addresses finding L-11 of #4832.

The attribute validator accepts `cs::attribute` on variant enum fields, but `VariantEnumGenerator` emitted bare `type name` positional parameters, silently dropping the attributes. The generator now emits `cs::attribute` and `deprecated` attributes with the `property:` target, which applies them to the property the record generates for the parameter — the same place struct fields put theirs.

Also fixes `BuildParameterList` receiving `""` instead of the current namespace, which made entity-typed fields fully qualified (`global::...`) even for types in the same namespace.

The remaining piece of L-11 — the `@param` doc comments of variant fields — is not fixable in this repository: slicec never fills the `entityInfo.comment` of variant fields. This is the same converter gap as the `@returns` loss tracked by L-12; both are now filed upstream as icerpc/slicec#809.

## What's Changed entry

Area: Slice codec

- The Slice compiler now applies `cs::attribute` and `deprecated` attributes on variant enum fields to the properties it generates for these fields.

### DIFF
--- a/tests/ZeroC.Slice.Generator.Tests/VariantEnumTests.slice
+++ b/tests/ZeroC.Slice.Generator.Tests/VariantEnumTests.slice
@@ -44,3 +44,7 @@ compact enum CompactShape {
     Triangle(side1: uint32, side2: uint32, side3: uint32)
     Oval(name: string?, major: uint32, minor: uint32)
 }
+
+enum ShapeWithFieldAttributes {
+    Circle([cs::attribute("System.ComponentModel.Description(\"The radius\")")] radius: uint32)
+}
--- a/src/ZeroC.Slice.Generator/VariantEnumGenerator.cs
+++ b/src/ZeroC.Slice.Generator/VariantEnumGenerator.cs
@@ -134,7 +134,7 @@ internal static class VariantEnumGenerator
 
         // Build parameter list for the record constructor.
         string nameWithParams = variant.Fields.Count > 0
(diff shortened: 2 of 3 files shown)


externl (approved):
Looks good.

tests/ZeroC.Slice.Generator.Tests/VariantEnumTests.slice
```diff
@@ -44,3 +44,7 @@ compact enum CompactShape {
     Triangle(side1: uint32, side2: uint32, side3: uint32)
     Oval(name: string?, major: uint32, minor: uint32)
 }
+
+enum ShapeWithFieldAttributes {
+    Circle([cs::attribute("System.ComponentModel.Description(\"The radius\")")] radius: uint32)
+}
```
InsertCreativityHere, line 50: Since there is different and special handling for it here, can we also add a test for the `deprecated` attribute here too?

InsertCreativityHere (approved):
Looks good to me!

tests/ZeroC.Slice.Generator.Tests/VariantEnumTests.slice
bernardnormier: It's very similar, that's why I didn't add an extra test.

PR #4883, opened 2026-08-26
Send the bundled StreamReadsClosed frame only with the last stream frame

### LINKED ISSUE #4798: [Audit-Low] Bundled StreamReadsClosed frame is emitted once per chunking-loop iteration instead of once with the last frame

## Summary

`writeReadsClosedFrame` is a loop-invariant parameter, but the emit site sits inside the `do { … } while (!source1.IsEmpty || !source2.IsEmpty);` chunking loop — unlike the sibling `if (lastStreamFrame) { stream.WroteLastStreamFrame(); }`, where `lastStreamFrame` is recomputed per chunk and is only true on the final iteration.

The whole point of the deferred flag, documented in `SlicStream.CloseReads`, is "to send both frames with a single write". The fire-and-forget path gets it right.

## Trigger / impact

A dispatcher reads the request payload to completion → `SlicPipeReader.ProcessReadResult` → `CloseReads(graceful: true)` → `_closeReadsOnWritesClosure = true`; the dispatcher then writes a 100 KB response with `endStream: true` → with the default `_maxStreamFrameSize = 32768` the loop runs 4× → **4 `StreamReadsClosed` frames instead of 1**.

The impact is genuinely small, and two details of the original claim were wrong:

- Each `StreamReadsClosed` is written **after** its iteration's chunk in the same flush, so it precedes only the remaining chunks, not the payload.
- A peer receiving one for a released stream is a **silent no-op** — the dispatch rejects only unknown stream IDs, and a released stream falls into the `TryGetValue` miss branch.

Duplicates are fully idempotent (`TrySetFlag`), so there is no double semaphore release and no protocol error. Codex confirmed all of this. Net: N−1 redundant 3-byte frames plus extra flush work, and a writes-closure notification reaching the peer earlier than designed.

## Source references

- [src/IceRpc/Transports/Slic/Internal/SlicConnection.cs#L985-L988](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs#L985-L988) — emit site inside the chunking loop
- [src/IceRpc/Transports/Slic/Internal/SlicConnection.cs#L967](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs#L967) — the correctly-guarded sibling
- [src/IceRpc/Transports/Slic/Internal/SlicStream.cs#L138-L142](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc/Transports/Slic/Internal/SlicStream.cs#L138-L142) — the "single write" intent

## Recommendation

`if (writeReadsClosedFrame && lastStreamFrame)` — `endStream` is always true whenever the flag is true, so the last iteration always qualifies.

Worth adding a test that writes more than `PeerMaxStreamFrameSize` with `endStream: true` on a stream whose reads were gracefully closed, asserting exactly one frame.

## Audit context

Finding 16 of [general-correctness-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-07-27.md); Codex verdict **CONFIRMED** at **Low**.

Severity: **Low**.

bernardnormier:
Fixes #4798.

When a stream's reads are gracefully closed before its writes, the sending of the `StreamReadsClosed` frame is deferred so that it can be bundled with the `StreamLast` (or `StreamWritesClosed`) frame in a single write on the duplex connection. In `WriteStreamDataFrameAsync`, however, the emit site sat inside the chunking loop with a loop-invariant flag: data larger than the peer's maximum stream frame size produced one `StreamReadsClosed` frame per chunk instead of one. The duplicates were harmless — the receiver's handling is idempotent — but redundant.

The fix guards the emit with `lastStreamFrame`, so the frame is written exactly once, bundled with the `StreamLast` frame. This is safe because `writeReadsClosedFrame` can only be true when `endStream` is true, so the last iteration always qualifies.

### Why delaying the frame to the last chunk is fine

The fix delays when the peer receives the `StreamReadsClosed` frame (previously the first duplicate went out with the first chunk). This matters for neither of the two reads-closure cases:

- When the application stops reading before consuming all the data ("I don't want more data"), `CloseReads(graceful: false)` sends the `StreamReadsClosed` frame immediately, as its own write. This is the case where delivery timing matters — it's what lets the peer abort a large write mid-flight — and it doesn't use the bundling path at all.
- The deferred/bundled path is only armed by `CloseReads(graceful: true)`, which is called only when the application consumed all the stream data including the end-of-stream — meaning the peer already closed its writes on the stream. At that point the notification is pure bookkeeping: the peer has nothing left to write, and it releases the stream only once its own reads complete, which necessarily happens after the `StreamLast` frame anyway.

The new test writes data spanning multiple stream frames on a stream whose reads were gracefully closed, and asserts the exact frame sequence received by the peer: `Stream`, `Stream`, `StreamLast`, `StreamReadsClosed`.

## What's Changed entry

None — the duplicate frames were harmless and not visible to applications.

### DIFF
--- a/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicConnection.cs
@@ -865,7 +865,7 @@ internal class SlicConnection : IMultiplexedConnection
     /// <param name="endStream"><see langword="true" /> to write a <see cref="FrameType.StreamLast" /> frame and
     /// <see langword="false" /> to write a <see cref="FrameType.Stream" /> frame.</param>
     /// <param name="writeReadsClosedFrame"><see langword="true" /> if a <see cref="FrameType.StreamReadsClosed" />
-    /// frame should be written after the stream frame.</param>
+    /// frame should be written after the last stream frame.</param>
     /// <param name="cancellationToken">A cancellation token that receives the cancellation requests.</param>
     /// <remarks>This method is called by streams and might be called on a closed connection. The connection might
     /// also be closed concurrently while it's in progress.</remarks>
@@ -982,7 +982,7 @@ internal class SlicConnection : IMultiplexedConnection
                         _duplexConnectionWriter.Write(sendSource2);
                     }
 
-                    if (writeReadsClosedFrame)
+                    if (writeReadsClosedFrame && lastStreamFrame)
                     {
                         WriteFrame(FrameType.StreamReadsClosed, stream.Id, encode: null);
                     }
--- a/tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
+++ b/tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
@@ -928,6 +928,69 @@ public class SlicTransportTests
         acceptedStream.Input.Complete();
     }
 
+    /// <summary>Verifies that the StreamReadsClosed frame bundled with the closure of writes is sent once, after the
+    /// last stream frame, when the data is written in multiple stream frames.</summary>
+    [Test]
+    public async Task Bundled_reads_closed_frame_is_sent_once_after_the_last_stream_frame()
+    {
+        // Arrange
+        await using ServiceProvider provider = new ServiceCollection()
+            .AddSlicTest()
+            .BuildServiceProvider(validateScopes: true);
+
+        var duplexClientTransport = provider.GetRequiredService<IDuplexClientTransport>();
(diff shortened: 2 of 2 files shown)


externl (approved):
Looks good.

pepone (approved):

Issue #4829, opened 2026-07-28
[Audit-Low] Documentation accuracy sweep (12 findings)


pepone:
> **AI-generated audit findings** — from a repository-wide audit performed by OpenAI Codex (`codex-cli`). ⚠️ Unlike the High/Medium findings from the same audit, **these Low-severity items were not individually re-verified** by the follow-up Claude pass — they are grouped here for triage rather than asserted as confirmed. Verify each before acting on it.

Twelve documentation defects from the same audit, grouped because they share one fix workflow: correct the text, then add validation so it cannot drift again. Each is independently actionable — close individually or as a batch.

| ID | Finding |
|---|---|
| L-19 | TcpFallback READMEs link to `../Secure`, which no longer exists |
| L-21 | Slice and Protobuf parent indexes describe their Stream examples as client-to-server; the code streams server-to-client |
| L-22 | Default-QUIC examples omit platform/runtime prerequisites even though the implementation throws when QUIC is unavailable |
| L-31 | Ice codec API docs describe the leading slice-size bytes and tagged-value encoding incorrectly (`SliceInfo.Bytes`, IceEncoder tag docs) |
| L-32 | Several package README samples omit required namespaces/imports or setup statements |
| L-33 | Telemetry README does not state that activities are created only for the icerpc protocol |
| L-34 | BUILDING.md omits that template instantiation/runtime checks are separate from the source build |
| L-35 | `docfx/docfx.json` hardcodes a sitemap base URL inconsistent with the publication/link configuration |
| L-36 | Grouped public-API XML comment drift across transport stream state, codec parameters/returns, generated API descriptions, and example-facing configuration |
| L-37 | `src/IceRpc.Ice/README.md` retains unfinished/TBD placeholder content in a published package README |
| L-38 | `src/IceRpc.Protobuf.Tools/README.md` has stale runtime and bundled-tool provenance guidance |
| L-39 | `src/IceRpc.ServiceGenerator/README.md` under-describes its Protobuf service-generation role |

## Suggested approach

Fix the factual errors first (L-19, L-21, L-31, L-33, L-37, L-38, L-39), then the structural ones (L-32, L-34, L-35, L-36, L-22). Compiling published README and DocFX snippets in CI would prevent most of this class recurring — see also the separately filed M-24 and M-25, which are the compile-breaking members of the same family.

## Audit context

[...]

PR #4884, opened 2026-08-26
Fix documentation inaccuracies

bernardnormier:
This PR fixes the trivial documentation findings of #4829:

- **L-19**: The TcpFallback READMEs referenced the `Secure` example, removed by #4211. The instructions to observe the TCP fallback are now self-contained: comment out `quicServer.Listen()` in the Server program.
- **L-21**: The Slice and Protobuf example indexes described the Stream example as streaming from client to server; it streams from server to client.
- **L-31**: `SliceInfo.Bytes` was documented as including the leading slice size; it excludes it, along with the tag end marker when the slice has tagged fields. The `EncodeTagged` doc attached a false format constraint to the `tag` parameter (the accepted formats are F1/F2/F4/F8/Size/FSize/OptimizedVSize) and documented only `VSize` as throwing.
- **L-33**: The Telemetry README now states that activities are created only for requests that use the `icerpc` protocol, matching the interceptor/middleware API docs.
- **L-35**: The docfx sitemap `baseUrl` still pointed at `docs.icerpc.dev`; the API reference is published at `code.icerpc.dev/csharp/<branch>/api/` (leftover from #4674).
- **L-37**: The IceRpc.Ice package README shipped a literal `[docs]: TBD` link; it now points to the IceRPC for Ice users documentation.
- **L-38**: The Protobuf.Tools README claimed the generator requires .NET 8 (it targets .NET 10) and credited the Linux ARM64 `protoc` to the Protobuf release page (it comes from Google.Protobuf.Tools; only the macOS ARM64 binary is downloaded separately). The tool description now matches the Slice.Tools README wording.
- **L-39**: The ServiceGenerator README now lists the Protobuf integration alongside Slice and Ice.
- **L-36** (grouped XML comment drift) is otherwise declined as unactionable — its transport-stream-state claim names a member that doesn't exist — but the sweep surfaced three micro-nits fixed here: a trailing space that unbound the `compactId` param doc, a param doc referencing the old `slicedData` parameter name, and two "first 3 bits" comments that meant the low-order bits.

L-32 (README snippets), L-22 (QUIC prerequisites) and L-34 (template builds) are addressed in separate PRs.

## What's Changed entry

None — documentation fixes.

### DIFF
--- a/src/IceRpc.ServiceGenerator/README.md
+++ b/src/IceRpc.ServiceGenerator/README.md
@@ -5,7 +5,7 @@ decorated with the `[IceRpc.Service]` attribute.
 
 [Package][package] | [Source code][source]
 
-This package is a dependency of the [IceRPC + Slice][icerpc+slice], [IceRPC + Protobuf][icerpc+protobuf] and
+This package is a dependency of the [IceRPC + Slice][icerpc+slice], [IceRPC + Protobuf][icerpc+protobuf], and
 [IceRPC + Ice][icerpc+ice] integrations.
 
 [icerpc+ice]: https://www.nuget.org/packages/IceRpc.Ice
--- a/src/IceRpc.ServiceGenerator/README.md
+++ b/src/IceRpc.ServiceGenerator/README.md
@@ -5,9 +5,11 @@ decorated with the `[IceRpc.Service]` attribute.
 
 [Package][package] | [Source code][source]
 
-This package is a dependency of the [IceRPC + Slice][icerpc+slice] and [IceRPC + Ice][icerpc+ice] integrations.
+This package is a dependency of the [IceRPC + Slice][icerpc+slice], [IceRPC + Protobuf][icerpc+protobuf] and
+[IceRPC + Ice][icerpc+ice] integrations.
 
 [icerpc+ice]: https://www.nuget.org/packages/IceRpc.Ice
+[icerpc+protobuf]: https://www.nuget.org/packages/IceRpc.Protobuf
 [icerpc+slice]: https://www.nuget.org/packages/IceRpc.Slice
 [package]: https://www.nuget.org/packages/IceRpc.ServiceGenerator
 [source]: https://github.com/icerpc/icerpc-csharp/tree/main/src/IceRpc.ServiceGenerator
--- a/src/IceRpc/Ice/Codec/IceEncoder.cs
+++ b/src/IceRpc/Ice/Codec/IceEncoder.cs
@@ -171,7 +171,7 @@ public ref partial struct IceEncoder
     // Other methods
 
     /// <summary>Encodes a non-null tagged value. The number of bytes needed to encode the value is known before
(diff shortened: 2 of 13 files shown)


externl (approved):
Looks good.

src/IceRpc.ServiceGenerator/README.md
```diff
@@ -5,9 +5,11 @@ decorated with the `[IceRpc.Service]` attribute.
 
 [Package][package] | [Source code][source]
 
-This package is a dependency of the [IceRPC + Slice][icerpc+slice] and [IceRPC + Ice][icerpc+ice] integrations.
+This package is a dependency of the [IceRPC + Slice][icerpc+slice], [IceRPC + Protobuf][icerpc+protobuf] and
```
InsertCreativityHere:
```suggestion
This package is a dependency of the [IceRPC + Slice][icerpc+slice], [IceRPC + Protobuf][icerpc+protobuf], and
```

src/IceRpc/Ice/Codec/IceEncoder.cs
```diff
@@ -202,11 +202,13 @@ public void EncodeTagged<T>(int tag, int size, T v, EncodeAction<T> encodeAction
...
-    /// <param name="tag">The tag. Must be either FSize or OptimizedVSize.</param>
-    /// <param name="tagFormat">The tag format.</param>
+    /// <param name="tag">The tag.</param>
+    /// <param name="tagFormat">The tag format. Must not be <see cref="TagFormat.VSize" /> or
+    /// <see cref="TagFormat.Class" />.</param>
     /// <param name="v">The value to encode.</param>
     /// <param name="encodeAction">The delegate that encodes the value after the tag header.</param>
-    /// <exception cref="ArgumentException">Thrown if <paramref name="tagFormat" /> is VSize.</exception>
+    /// <exception cref="ArgumentException">Thrown when <paramref name="tagFormat" /> is not a tag format supported by
+    /// this method.</exception>
```
InsertCreativityHere, line 211: `Thrown when` is the wording used in the C# example for `<exception>` tags.
But, 97% of the time in the time in this repo we use `Thrown if`.

If `Thrown when` is intentional here, we should open an issue to update all the other `Thrown if` to match it, instead of having it inconsistent.

InsertCreativityHere (approved):
Looks good to me.

src/IceRpc.ServiceGenerator/README.md
bernardnormier: Applied, thanks.

src/IceRpc/Ice/Codec/IceEncoder.cs
bernardnormier: "Thrown when" is intentional — we follow Microsoft's conventions whenever possible. The repo is actually more mixed than 97/3: 47 "Thrown when" vs 73 "Thrown if" in src. Filed #4894 to standardize on "Thrown when" throughout.

PR #4885, opened 2026-08-26
Make README code samples self-contained

bernardnormier:
This PR fixes the L-32 finding of #4829: several package README code samples were missing the using directives needed to compile as shown.

- `IceRpc.Telemetry`: all four samples use `ActivitySource` without `using System.Diagnostics;`, and the DI samples use `AddSingleton` without `using Microsoft.Extensions.DependencyInjection;`. Also fixes a comment that called the telemetry interceptor the "deadline interceptor", and drops an unused `using VisitorCenter;`.
- `IceRpc.Locator`: `LocatorProxy` requires `using IceRpc.Ice;`.
- `IceRpc.Transports.Coloc`: `SlicServerTransport`/`SlicClientTransport` require `using IceRpc.Transports.Slic;`.
- `IceRpc.Compressor`: the DI samples use `AddSingleton` without `using Microsoft.Extensions.DependencyInjection;` (using directives re-sorted alphabetically while at it), and the server sample now explains what `Chatbot` is.
- `IceRpc.Protobuf`: the server sample uses `IFeatureCollection` without `using IceRpc.Features;` (the Slice and Ice READMEs already have it).
- `IceRpc`, `IceRpc.Slice`, `IceRpc.Protobuf`, `IceRpc.Ice`: `await CancelKeyPressed;` now carries a comment pointing to the `examples/common` helper that provides it, like the existing pointer for `CreateServerAuthenticationOptions`.

## What's Changed entry

None — documentation fixes.


src/IceRpc.Ice/README.md
```diff
@@ -65,6 +65,8 @@ await using var server = new Server(new Chatbot(), new Uri("ice://[::0]"));
 server.Listen();
 
 // Wait until the console receives a Ctrl+C.
+// examples/common/Program.CancelKeyPressed.cs in the icerpc-csharp repo provides
+// the CancelKeyPressed helper.
```
externl, line 69: Should we just say in "this repo"?

externl (approved):

src/IceRpc.Compressor/README.md
```diff
@@ -110,8 +112,9 @@ host.Run();
 using IceRpc;
 using IceRpc.Extensions.DependencyInjection;
 using IceRpc.Slice;
```
InsertCreativityHere: Do we need `IceRpc.Slice` here? This doesn't seem to have any Slice-specific code in it.
And if we don't need it, having it could mislead uses to thinking `Compressor` is Slice-only, or relies on Slice in some way.

src/IceRpc.Ice/README.md
bernardnormier: I'd rather keep "the icerpc-csharp repo" (here and in the other READMEs): these are also the NuGet package READMEs, and on nuget.org "this repo" has no clear referent.

src/IceRpc.Ice/README.md
```diff
@@ -65,6 +65,8 @@ await using var server = new Server(new Chatbot(), new Uri("ice://[::0]"));
 server.Listen();
 
 // Wait until the console receives a Ctrl+C.
+// examples/common/Program.CancelKeyPressed.cs in the icerpc-csharp repo provides
```
InsertCreativityHere, line 68: I'm a fan of quoting off the file name, just to be clear about the start/end of it. Costs us nothing and improves readability. Same for the other spots we mention the file name now.
```suggestion
// 'examples/common/Program.CancelKeyPressed.cs' in the icerpc-csharp repo provides
```

InsertCreativityHere (approved):
Looks good to me!

A tangential thing I spotted while reviewing the other READMEs for consistency:
`src/IceRpc.Logger/README.md` has it's `using`s in the opposite order as these.
We have `using Microsoft...` before `using IceRpc;`, which I think is the wrong order.

src/IceRpc.Compressor/README.md
bernardnormier: Right — nothing in this snippet needs it (CreateSliceProxy, UseCompressor and Map<TService> all live in IceRpc.Extensions.DependencyInjection). Removed.

src/IceRpc.Ice/README.md
bernardnormier: I'd rather keep the paths unquoted: a full path like examples/common/Program.CancelKeyPressed.cs is self-delimiting (slashes + .cs extension), and the pre-existing path mentions in these READMEs are unquoted, so quoting only the new spots would be inconsistent.

bernardnormier:
Fixed the using order in the IceRpc.Logger README snippets.

PR #4886, opened 2026-08-26
Document the QUIC platform prerequisites

bernardnormier:
This PR fixes the L-22 finding of #4829: the examples that use QUIC — the default multiplexed transport, so every example that doesn't select another transport — didn't mention that running them requires QUIC to be set up on the system, even though the QUIC transport throws when it isn't.

Since each example is meant to be studied in isolation, every QUIC-using example README (31 in total: 17 slice, 12 protobuf, 2 json) now carries a short self-contained note: QUIC works out of the box on Windows, while Linux and macOS need extra setup steps, with a link to .NET's QUIC platform dependencies documentation. The platform specifics themselves stay in the Microsoft documentation.

The QUIC Transport section of the IceRpc package README is also reworked: it now makes clear that you can build a QUIC application on any platform — the QUIC transport ships in the main IceRpc assembly — and that only the runtime depends on the system's QUIC setup, instead of the vague "IceRPC has the same platform dependencies as System.Net.Quic".

## What's Changed entry

None — documentation fixes.

### DIFF
--- a/examples/protobuf/TcpFallback/README.md
+++ b/examples/protobuf/TcpFallback/README.md
@@ -9,6 +9,9 @@ The server program creates a server for QUIC and another server for TCP; both se
 The client program first attempts to establish a QUIC connection; if this connection establishment fails, it falls back
 to a TCP connection.
 
+This example uses QUIC, IceRPC's default multiplexed transport. On Linux and macOS, QUIC requires extra setup
+steps; see .NET's [QUIC platform dependencies][quic-platform].
+
 You can build the client and server applications with:
 
 ``` shell
@@ -30,3 +33,5 @@ dotnet run
 ```
 
 In order to see the fallback to TCP, run the TCP fallback client with the server from the [../Secure] example.
+
+[quic-platform]: https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/quic/quic-overview#platform-dependencies
--- a/examples/protobuf/Thermostat/README.md
+++ b/examples/protobuf/Thermostat/README.md
@@ -49,6 +49,9 @@ Server also calls a service on the device that implements Protobuf service `Ther
 
 ## Build and run
 
+This example uses QUIC, IceRPC's default multiplexed transport. On Linux and macOS, QUIC requires extra setup
+steps; see .NET's [QUIC platform dependencies][quic-platform].
+
 You can build the client, server and device with:
 
 ``` shell
@@ -86,3 +89,5 @@ set point to 70°F with:
 cd Client
 dotnet run set 70
 ```
+
+[quic-platform]: https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/quic/quic-overview#platform-dependencies
--- a/examples/protobuf/Thermostat/README.md
+++ b/examples/protobuf/Thermostat/README.md
@@ -52,7 +52,7 @@ Server also calls a service on the device that implements Protobuf service `Ther
 This example uses QUIC, IceRPC's default multiplexed transport. On Linux and macOS, QUIC requires extra setup
 steps; see .NET's [QUIC platform dependencies][quic-platform].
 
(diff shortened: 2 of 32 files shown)


examples/slice/TcpFallback/README.md
```diff
@@ -30,3 +33,5 @@ dotnet run
 ```
 
 In order to see the fallback to TCP, run the TCP fallback client with the server from the [../Secure] example.
```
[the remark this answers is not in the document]
bernardnormier: Fixed by #4884

examples/protobuf/TcpFallback/README.md
```diff
@@ -30,3 +33,5 @@ dotnet run
 ```
 
 In order to see the fallback to TCP, run the TCP fallback client with the server from the [../Secure] example.
```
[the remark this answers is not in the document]
bernardnormier: Fixed by #4884

externl (approved):
Looks good.

examples/protobuf/Thermostat/README.md
```diff
@@ -49,6 +49,9 @@ Server also calls a service on the device that implements Protobuf service `Ther
 
 ## Build and run
 
+This example uses QUIC, IceRPC's default multiplexed transport. On Linux and macOS, QUIC requires extra setup
+steps; see .NET's [QUIC platform dependencies][quic-platform].
+
 You can build the client, server and device with:
```
InsertCreativityHere: This would be more consistent with the wording we use in every other Protobuf example.
```suggestion
You can build the client, server, and device applications with:
```

examples/slice/Thermostat/README.md
```diff
@@ -48,6 +48,9 @@ Server also calls a service on the device that implements Slice interface `Therm
 
 ## Build and run
 
+This example uses QUIC, IceRPC's default multiplexed transport. On Linux and macOS, QUIC requires extra setup
+steps; see .NET's [QUIC platform dependencies][quic-platform].
+
 You can build the client, server and device with:
```
InsertCreativityHere: Same here:

```suggestion
You can build the client, server, and device applications with:
```

src/IceRpc/README.md
```diff
@@ -7,10 +7,13 @@ and package represent the base assembly and package for the [C# implementation o
...
 
-IceRPC's default transport is QUIC, a new UDP-based multiplexed transport used by HTTP/3 and other modern application
+IceRPC's default multiplexed transport is QUIC, a new UDP-based transport used by HTTP/3 and other modern application
 protocols.
 
-IceRPC has the same platform dependencies as `System.Net.Quic`. See .NET's [QUIC Platform dependencies][platform].
+The QUIC transport implementation is included in the main IceRpc assembly, so you can build an application that uses
+QUIC on any platform. At runtime, this implementation relies on `System.Net.Quic`, which works out of the box on
+Windows but requires extra setup steps on Linux and macOS, as documented in .NET's
+[QUIC Platform dependencies][platform].
```
InsertCreativityHere: In the examples, `platform` is lowercase. Which I think is correct. Not clear why `Platform` should be upper-cased here. This was pre-existing! But might as well fix it now.

```suggestion
[QUIC platform dependencies][platform].
```

InsertCreativityHere (approved):
Looks good to me!

examples/protobuf/Thermostat/README.md
bernardnormier: Applied, thanks.

examples/slice/Thermostat/README.md
bernardnormier: Applied, thanks.

src/IceRpc/README.md
bernardnormier: Applied — this was the only remaining uppercase occurrence; all 32 READMEs now use the same wording.

PR #4888, opened 2026-08-26
Validate the limit of ProtobufFeature

### LINKED ISSUE #4887: ProtobufFeature doesn't validate its limits
`ProtobufFeature` has the pattern that #4862 fixed for `IceFeature` and `SliceFeature`:

- The constructor treats any negative `maxMessageLength` as "use the default value" instead of accepting only `-1`, and rejects nothing — for example, `new ProtobufFeature(maxMessageLength: int.MaxValue)` is accepted.
- `IProtobufFeature.MaxMessageLength` doesn't document a valid range, for the consumer or for implementations.

The fix should mirror #4862:

- Validate the constructor parameter and document the accepted range.
- Document the contract on `IProtobufFeature.MaxMessageLength` (implementations note) and the default on `ProtobufFeature.MaxMessageLength`.
- Review the `MaxMessageLength` read sites (`InvokerExtensions`, `IncomingRequestExtensions`, and the helpers they call) for `maxSize + 1`-style overflow with extreme values.

Noted by @InsertCreativityHere while reviewing #4862.

bernardnormier:
Fixes #4887.

This PR applies the #4862 model to `ProtobufFeature`:

- The constructor now rejects a `maxMessageLength` that is a negative value other than `-1` with an `ArgumentOutOfRangeException`. `0` remains valid: an all-defaults Protobuf message encodes to 0 bytes.
- `IProtobufFeature.MaxMessageLength` documents the implementation contract (a value greater than or equal to `0`), and `ProtobufFeature.MaxMessageLength` inherits these docs and documents its default value.

The `MaxMessageLength` read sites were reviewed for `maxSize + 1`-style overflow: there is none — `DecodeMessageLength` compares the wire length against `(uint)maxMessageLength` directly, and its safety comment already relies on the non-negative contract that the interface now documents. This is also why the valid range has no `int.MaxValue` exclusion, unlike `maxPayloadSize` in `IceFeature`.

## What's Changed entry

None — only rejects invalid configurations that no reasonable program uses.

### DIFF
--- a/src/IceRpc.Protobuf/Features/IProtobufFeature.cs
+++ b/src/IceRpc.Protobuf/Features/IProtobufFeature.cs
@@ -10,6 +10,7 @@ public interface IProtobufFeature
 {
     /// <summary>Gets the maximum length of an encoded Protobuf message, in bytes.</summary>
     /// <value>The maximum length of a Protobuf message, in bytes.</value>
+    /// <remarks>Implementations must return a value greater than or equal to <c>0</c>.</remarks>
     int MaxMessageLength { get; }
 
     /// <summary>Gets the options to use when encoding the payload of an outgoing response.</summary>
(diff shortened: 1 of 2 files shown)


externl (approved):

InsertCreativityHere (approved):
Looks good to me!

Issue #4831, opened 2026-07-28
[Audit-Low] Test infrastructure and coverage gaps (6 findings)


pepone:
> **AI-generated audit findings** — from a repository-wide audit performed by OpenAI Codex (`codex-cli`). ⚠️ Unlike the High/Medium findings from the same audit, **these Low-severity items were not individually re-verified** by the follow-up Claude pass — they are grouped here for triage rather than asserted as confirmed. Verify each before acting on it.

Six test-infrastructure defects from the same audit, grouped for triage.

| ID | Finding |
|---|---|
| L-13 | Four Protobuf continuation-completion cases decorate `request.Payload`, assign that decorator to `PayloadContinuation`, and overwrite the real generated streaming continuation — so they assert against the replacement, not the actual client stream |
| L-14 | `TestDispatcher` disposal uses `SemaphoreSlim.CurrentCount` to decide whether to release; the count is zero precisely when dispatches are waiting, so held tasks can stay blocked |
| L-15 | `build/Tests.runsettings` excludes `IceRpc.Tests.Common` from coverage but not `ZeroC.Tests.Common`, letting shared codec helpers inflate product coverage |
| L-16 | The `ReadOnlySequencePipeWriter.WriteAsync` conformance case checks aggregate byte count but not content, unlike the sibling cases |
| L-17 | `ProtocolBridgingTests.Forwarder` transfers response-field `ReadOnlySequence` values into a new response and disposes the owning request before serialization — latent because existing scenarios usually have no response fields |
| L-40 | 123 QUIC and 34 IPv6/default-QUIC tests were skipped for platform reasons; the skips are legitimate, but no required CI environment asserts these groups actually execute |

## Note

L-13 and L-14 are the ones that weaken real signal — a test asserting the wrong object, and a helper that can hang. L-40 is a CI-configuration decision rather than a code fix.

See also the separately filed M-22/M-23, which together make the unobserved-task-exception harness inert.

## Audit context

L-13 through L-17 and L-40 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md).

bernardnormier:
All six findings were verified and dispositioned:

| Finding | Disposition |
|---|---|
| L-13 | Fixed by #4889 — the tests also asserted completion with a lambda that could never fail; the assertions are now real. |
[...]

PR #4889, opened 2026-08-26
Fix the Protobuf payload completion tests

bernardnormier:
This PR fixes the L-13 finding of #4831.

The four client-streaming and bidi-streaming payload completion tests in `OperationTests` wrapped `request.Payload` — which is empty for these RPC shapes — and assigned the wrapper to `request.PayloadContinuation`, replacing the stream reader created by the generated code. As a result the client stream was never sent, the replaced reader was never completed, and the assertions observed the replacement. The tests now decorate `request.PayloadContinuation` and assert that the empty `request.Payload` invariant holds, and the two streaming success tests verify that all streamed messages reach the service.

In addition, all eight tests asserted completion with `Assert.That(() => payload.Completed, Throws.Nothing)` — a lambda that reads a `Task` property and can never throw, so completion was never actually verified. The success tests now assert `await payload.Completed` is null and the failure tests assert `payload.Completed.IsCompletedSuccessfully`, matching the idiom used in `InvokeOperationAsyncTests`.

Also removes an unused `message` local copy-pasted from the unary tests.

## What's Changed entry

None — test fixes.

### DIFF
--- a/tests/IceRpc.Protobuf.Tests/OperationTests.cs
+++ b/tests/IceRpc.Protobuf.Tests/OperationTests.cs
@@ -326,8 +326,8 @@ public partial class OperationTests
         await client.UnaryOpAsync(message);
 
         // Assert
-        Assert.That(() => requestPayload!.Completed, Throws.Nothing);
-        Assert.That(() => responsePayload!.Completed, Throws.Nothing);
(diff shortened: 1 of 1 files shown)


externl (approved):
Looks good.

pepone (approved):

PR #4891, opened 2026-08-26
Unblock held dispatches when disposing TestDispatcher

bernardnormier:
This PR fixes the L-14 finding of #4831.

`TestDispatcher.Dispose` released the `_hold` semaphore only when `SemaphoreSlim.CurrentCount` was positive. The count is positive only when no dispatch is waiting on the semaphore, so the release could never unblock a waiting dispatch. `Dispose` now unconditionally releases `_holdDispatchCount` permits — the maximum number of dispatches that can be waiting.

No current test hangs because of this: every held dispatch waits with a connection-scoped cancellation token, and the tests dispose the connection before the dispatcher. The fix protects future tests that hold a dispatch on a token that isn't canceled first.

Also documents the `_holdDispatchCount` and `_dispatchCount` fields.

## What's Changed entry

None — test fixes.

### DIFF
--- a/tests/IceRpc.Tests.Common/TestDispatcher.cs
+++ b/tests/IceRpc.Tests.Common/TestDispatcher.cs
@@ -27,7 +27,12 @@ public sealed class TestDispatcher : IDispatcher, IDisposable
         new(TaskCreationOptions.RunContinuationsAsynchronously);
 
     private readonly byte[]? _responsePayload;
+
+    // Each of the first _holdDispatchCount dispatches waits on the _hold semaphore; it's also the maximum number of
+    // dispatches that can be waiting on this semaphore at any point.
     private readonly int _holdDispatchCount;
+
+    // The number of dispatches started so far.
     private int _dispatchCount;
 
     /// <inheritdoc/>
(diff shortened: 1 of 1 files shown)


tests/IceRpc.Tests.Common/TestDispatcher.cs
```diff
@@ -72,9 +77,10 @@ public async ValueTask<OutgoingResponse> DispatchAsync(IncomingRequest request,
     /// <inheritdoc/>
     public void Dispose()
     {
-        if (_hold.CurrentCount > 0)
+        if (_holdDispatchCount > 0)
         {
-            _hold.Release(_hold.CurrentCount);
+            // Release enough permits to unblock any dispatch still waiting on the semaphore.
+            _hold.Release(_holdDispatchCount);
```
InsertCreativityHere, line 83: This is either very minor or very major, but I think we're violating the contract of `Dispose` here.

My understanding is that it should always be safe to call `Dispose` multiple times, and it should behave idempotently. Here though, when I call it the first time, everything works, and `_hold` is `Dispose`d, When I call it a 2nd time, `_holdDispatchCount` is unchanged, and then `_hold.Release` will throw right?

To be totally correct, we should probably set `_holdDispatchCount` to `0` here. To avoid double-releasing.

Does this actually matter for us? Probably not.

InsertCreativityHere (approved):
Looks good to me, just one point worth bringing up about the `Dispose`.

tests/IceRpc.Tests.Common/TestDispatcher.cs
bernardnormier: Right, a second Dispose would throw. Leaving as-is for now: nothing disposes a TestDispatcher twice, and I'd rather not add just-in-case code. We'll fix it if that's ever needed.

externl (approved):
Looks good.

PR #4892, opened 2026-08-26
Assert content in the multiplexed stream write conformance tests

bernardnormier:
This PR fixes the L-16 finding of #4831.

The `Stream_write` conformance test verified only the aggregate byte count of the received data, and its source buffers were all zeros — so even a content assertion could not have detected reordered or duplicated segments. The buffers now carry distinct byte patterns and the test compares the full received content against the written data, like the sibling conformance tests do.

Also asserts the received byte's value in `Stream_write_empty_buffer_is_noop`, which previously checked only the length.

## What's Changed entry

None — test fixes.


externl (approved):
Looks good.

One note: the mod-256 ramp makes every 4096-byte segment within a buffer byte-identical, so a 256-aligned segment swap or duplicate still passes. An aperiodic fill — e.g. `(byte)(seed + i + (i >> 8))` — would close that.

pepone (approved):

bernardnormier:
Applied @externl's note: the fill is now `(byte)(seed + i + (i >> 8))`, aperiodic within a buffer, so a 4096-aligned segment swap changes the content too.

PR #4893, opened 2026-08-26
Exclude ZeroC.Tests.Common from code coverage

bernardnormier:
This PR fixes the L-15 finding of #4831: the coverage configuration excluded `IceRpc.Tests.Common` but not its `ZeroC.Tests.Common` sibling, so the shared test helpers showed up as a covered package in the coverage reports.

## What's Changed entry

None — test configuration fix.


externl (approved):
Looks good.

PR #4895, opened 2026-08-27
Use "Thrown when" in exception doc comments

### LINKED ISSUE #4894: Use "Thrown when" in `<exception>` doc comments
Our `<exception>` doc comments are split between `Thrown when ...` (47 occurrences in src) and `Thrown if ...` (73 in src, 7 in tests). Microsoft's convention for `<exception>` documentation is "Thrown when ...".

Replace "Thrown if" with "Thrown when" throughout.

bernardnormier:
Fixes #4894.

This PR replaces "Thrown if" with "Thrown when" in all `<exception>` doc comments, following Microsoft's convention. Sentences that continued with "or if ..." now use "or when ..." to match.

It also fixes "shutdown" used as a verb — in doc comments, code comments, one exception message, and one log message: the verb is "shut down"; "shutdown" is the noun.

The `slice/IceRpc/Internal/IceRpcDefinitions.slice` doc comment has the same verb misuse but is vendored from icerpc-slice, so it's not touched here.

## What's Changed entry

None.

### DIFF
--- a/src/IceRpc/ConnectionCache.cs
+++ b/src/IceRpc/ConnectionCache.cs
@@ -138,12 +138,12 @@ public sealed class ConnectionCache : IInvoker, IAsyncDisposable
     /// <param name="request">The outgoing request being sent.</param>
     /// <param name="cancellationToken">A cancellation token that receives the cancellation requests.</param>
     /// <returns>The corresponding <see cref="IncomingResponse" />.</returns>
-    /// <exception cref="InvalidOperationException">Thrown if no <see cref="IServerAddressFeature" /> feature is set and
-    /// the request's service address has no server addresses.</exception>
-    /// <exception cref="IceRpcException">Thrown with <see cref="IceRpcError.InvocationRefused" /> if the connection
-    /// cache is shutdown, or with <see cref="IceRpcError.NoConnection" /> if the request's
+    /// <exception cref="InvalidOperationException">Thrown when no <see cref="IServerAddressFeature" /> feature is set
+    /// and the request's service address has no server addresses.</exception>
+    /// <exception cref="IceRpcException">Thrown with <see cref="IceRpcError.InvocationRefused" /> when the connection
+    /// cache is shut down, or with <see cref="IceRpcError.NoConnection" /> when the request's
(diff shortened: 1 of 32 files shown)


src/IceRpc/ConnectionCache.cs
```diff
@@ -138,12 +138,12 @@ await Task.WhenAll(
...
     /// <param name="cancellationToken">A cancellation token that receives the cancellation requests.</param>
     /// <returns>The corresponding <see cref="IncomingResponse" />.</returns>
-    /// <exception cref="InvalidOperationException">Thrown if no <see cref="IServerAddressFeature" /> feature is set and
-    /// the request's service address has no server addresses.</exception>
-    /// <exception cref="IceRpcException">Thrown with <see cref="IceRpcError.InvocationRefused" /> if the connection
-    /// cache is shutdown, or with <see cref="IceRpcError.NoConnection" /> if the request's
+    /// <exception cref="InvalidOperationException">Thrown when no <see cref="IServerAddressFeature" /> feature is set
+    /// and the request's service address has no server addresses.</exception>
+    /// <exception cref="IceRpcException">Thrown with <see cref="IceRpcError.InvocationRefused" /> when the connection
+    /// cache is shut down, or with <see cref="IceRpcError.NoConnection" /> when the request's
```
InsertCreativityHere, line 144: This `shut?down` might be incorrect? I can't 100% tell.

But from my understanding, we don't "shut down" the connection cache, we're checking if it's in the state of being "shutdown". Either way. It is not critical.

InsertCreativityHere (approved):
Checked by hand, and by number, this is all of them fixed!
There were previous 71 `thrown if`, and 19 `thrown when`. There are now 90 `thrown when`.

There's also 2 `thrown with` and ` thrown by` which are already fine, and left alone.

src/IceRpc/ConnectionCache.cs
bernardnormier: "Shut down" is correct here: the state is expressed with the past participle — "the connection cache is shut down" parallels "the store is closed". "Shutdown" (one word) is the noun/attributive form ("this shutdown timeout"), never an adjective. Same distinction as everywhere else in this PR ("is already shut down or shutting down").

Issue #4896, opened 2026-08-27
Variant enum case field types can be shadowed by the case record names


bernardnormier:
The C# variant enum mapping nests one record class per case inside the union record. Since #4877, case parameter types that live in the same namespace as the enum are emitted unqualified. Inside the nested scope, the case record names shadow namespace-level type names, so a field type whose name matches any case name binds to the wrong type.

For example:

```slice
module Example

struct Interface { name: string }

enum Symbol {
    Interface(v: Interface)
}
```

generates:

```csharp
internal abstract partial record class Symbol
{
    partial record class Interface(Interface V) : Symbol
```

where the parameter type `Interface` binds to `Symbol.Interface` itself, producing `error CS8910: The primary constructor conflicts with the synthesized copy constructor.` A collision with a *different* case name silently binds to that sibling case record instead.

Observed by running the `Regenerate` target on `src/ZeroC.Slice.Symbols`: `slice/Compiler/CodeGenerator.slice` wraps each symbol type in a `Symbol` union whose case names match the wrapped types, and the regenerated `CodeGenerator.cs` no longer compiles. The checked-in `Compiler` files must stay on the pre-#4877 output until this is fixed.

Fix direction: fully qualify case parameter types inside the union body — unqualified same-namespace names are inherently unsafe in that nested scope.

PR #4897, opened 2026-08-27
Make generated code compile without implicit usings

### LINKED ISSUE #4791: [Audit-Medium] Generated code depends on ImplicitUsings being enabled in the consumer project

## Summary

The generators emit `global::`-qualified names essentially everywhere, and the fixed preamble usings include neither `System` nor `System.Linq`. Two spots break the pattern, so generated code only compiles when the **consumer** project has implicit usings switched on.

The SDK adds those namespaces only when `ImplicitUsings` is explicitly `true`/`enable` — omitting the property means off.

### 6a — `Span<byte>`

Emitted unqualified in every `Request.Encode{Op}` / `Response.Encode{Op}` helper, i.e. for any operation with at least one non-streamed parameter or return value. Compiling the checked-in `tests/IceRpc.Slice.Generator.Tests/generated/ModuleIdentifierTests1.IceRpc.cs` with `<ImplicitUsings>disable</ImplicitUsings>` — or with the property simply **omitted** — fails with `error CS0246: The type or namespace name 'Span<>' could not be found`.

This is the broad one: it affects essentially all generated interface code.

### 6b — `.Count()`

The tagged fixed-size-collection branch emits `int count_ = {param}.Count();` on an `IList<T>`/`IDictionary<K,V>`, requiring `Enumerable.Count`. A project with `struct S { tag(1) values: Sequence<int32>? tag(2) dict: Dictionary<int32,int32>? }` and implicit usings disabled fails with `error CS1955: Non-invocable member 'ICollection<int>.Count' cannot be used like a method.` Flipping to `enable` builds clean, isolating the cause.

This branch has **no generated-file coverage**. One detail of the original write-up was wrong and Codex corrected it: `OperationTests.slice` *does* contain a tagged `Sequence<int32>?`, so the corpus is not limited to `Sequence<uint8>?` as claimed. But operation fields route through a separate `ReadOnlyMemory` path that never reaches `.Count()`, and the only struct case has a 1-byte element type which `GetCollectionElementSize` excludes — so the conclusion holds: **no checked-in generated file contains `.Count()`**.

Codex found no other generator-owned unqualified `System` type. Unqualified `List`/`HashSet` occurrences originate from trusted `cs::type` text, which is a different (and accepted) category.

## Trigger / impact

Any consumer project that omits or disables `ImplicitUsings` — which is the SDK default — deterministically fails to compile generated code. Enabling implicit usings is a workaround, not a generator invariant.

Both defects are invisible in-tree because [`build/Common.props`](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/build/Common.props#L13) sets `ImplicitUsings=true` repo-wide, and every template and example enables it.

## Source references

- [src/IceRpc.Slice.Generator/EncodeHelper.cs#L20](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Slice.Generator/EncodeHelper.cs#L20) — unqualified `Span<byte>`
- [src/ZeroC.Slice.Generator/FieldExtensions.cs#L78](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/ZeroC.Slice.Generator/FieldExtensions.cs#L78) — `.Count()` on an `ICollection<T>`
- [src/IceRpc.Slice.Generator/Program.cs#L19](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Slice.Generator/Program.cs#L19) and [src/ZeroC.Slice.Generator/Program.cs#L21](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/ZeroC.Slice.Generator/Program.cs#L21) — the fixed preamble usings

## Recommendation

- Emit `global::System.Span<byte>`.
- For 6b use `{varName}.Count` on the already-bound pattern variable, which fixes the binding *and* drops a redundant nullable dereference.

Adding usings to the preamble also works but is less robust — a Slice module can map to a namespace that shadows `System`.

Worth adding a generator test project with `<ImplicitUsings>disable</ImplicitUsings>` so this cannot regress.

## Audit context

Finding 6 of [general-correctness-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-07-27.md); Codex verdict **CONFIRMED** at **Medium**, with the corpus-coverage detail corrected as noted above.

Severity: **Medium**.

bernardnormier:
Fixes #4791.

The Slice generators emit fully qualified names for BCL types — except in two spots that relied on the consuming project enabling `ImplicitUsings`: the `Span<byte>` local emitted in every `Request.Encode*`/`Response.Encode*` helper, and the `.Count()` call emitted for tagged fixed-size collection fields. A consuming project without implicit usings — the SDK default — failed to compile the generated code.

Both emissions are now fully qualified, like everything else the generators emit: `global::System.Span<byte>`, and `global::System.Linq.Enumerable.Count(...)` in its static form — which also works when a `cs::type` attribute maps a sequence to an array, and drops a redundant nullable dereference by counting the non-null pattern variable.

To keep this from regressing, the Slice, Ice and Protobuf generator test projects (including the `IceRpc.Ice.Generator.Base.Tests.ReferencedAssemblies` projects) now build with `<ImplicitUsings>disable</ImplicitUsings>`, and their handwritten test files declare their usings explicitly. This also verifies that the `slice2cs` and `protoc-gen-icerpc-csharp` outputs compile without implicit usings — both already qualify all their references.

The new `Sequence<int32>?` and `Dictionary<int32, int32>?` tagged fields in `TaggedTests.slice` cover the previously untested fixed-size tagged collection encoding.

`cs::type` text is passed through verbatim: in a project that doesn't enable implicit usings, qualify the C# type names you write in `cs::type` — the test corpora now do (`cs::type("global::System.Collections.Generic.List<bool>")`).

The checked-in `src/ZeroC.Slice.Symbols/Compiler` files are deliberately not regenerated here: regenerating them is blocked by #4896.

## What's Changed entry

The Slice compiler now generates code that compiles in projects that don't enable `ImplicitUsings`.

### DIFF
--- a/src/IceRpc.Slice.Generator/Program.cs
+++ b/src/IceRpc.Slice.Generator/Program.cs
@@ -16,6 +16,14 @@ await GeneratorDriver.RunAsync(
         ? CodeBlock.FromBlocks([ProxyGenerator.Generate(interfaceDef), ServiceGenerator.Generate(interfaceDef)])
         : null,
     mapOutputPath: path => Path.ChangeExtension(Path.GetFileName(path), ".IceRpc.cs"),
-    usings: ["IceRpc.Slice", "IceRpc.Slice.Operations", "ZeroC.Slice.Codec"]).ConfigureAwait(false);
+    usings:
+    [
+        "IceRpc.Slice",
+        "IceRpc.Slice.Operations",
+        "System",
+        "System.Collections.Generic",
+        "System.Linq",
+        "ZeroC.Slice.Codec"
+    ]).ConfigureAwait(false);
 
 return 0;
--- a/src/IceRpc.Slice.Generator/Program.cs
+++ b/src/IceRpc.Slice.Generator/Program.cs
@@ -16,14 +16,6 @@ await GeneratorDriver.RunAsync(
         ? CodeBlock.FromBlocks([ProxyGenerator.Generate(interfaceDef), ServiceGenerator.Generate(interfaceDef)])
         : null,
     mapOutputPath: path => Path.ChangeExtension(Path.GetFileName(path), ".IceRpc.cs"),
-    usings:
-    [
-        "IceRpc.Slice",
-        "IceRpc.Slice.Operations",
-        "System",
-        "System.Collections.Generic",
-        "System.Linq",
-        "ZeroC.Slice.Codec"
-    ]).ConfigureAwait(false);
+    usings: ["IceRpc.Slice", "IceRpc.Slice.Operations", "ZeroC.Slice.Codec"]).ConfigureAwait(false);
 
 return 0;
--- a/src/IceRpc.Slice.Generator/EncodeHelper.cs
+++ b/src/IceRpc.Slice.Generator/EncodeHelper.cs
@@ -17,7 +17,7 @@ internal static class EncodeHelper
                 {{encodeOptionsName}}?.PipeOptions ?? SliceEncodeOptions.Default.PipeOptions);
             var encoder_ = new SliceEncoder(pipe_.Writer);
 
-            Span<byte> sizePlaceholder_ = encoder_.GetPlaceholderSpan(4);
+            global::System.Span<byte> sizePlaceholder_ = encoder_.GetPlaceholderSpan(4);
             int startPos_ = encoder_.EncodedByteCount;
 
             {{encodeBody}}
--- a/src/ZeroC.Slice.Generator/FieldExtensions.cs
+++ b/src/ZeroC.Slice.Generator/FieldExtensions.cs
@@ -75,7 +75,7 @@ internal static class FieldExtensions
                 $$"""
                 if ({{param}} is {{csType}} {{varName}})
                 {
-                    int count_ = {{param}}.Count();
+                    int count_ = global::System.Linq.Enumerable.Count({{varName}});
                     {{encoderName}}.EncodeTagged({{tag}}, size: {{sizeExpr}}, {{varName}}, {{encodeLambda}});
                 }
                 """).ToString();
--- a/src/ZeroC.Slice.Generator/Program.cs
+++ b/src/ZeroC.Slice.Generator/Program.cs
@@ -18,6 +18,6 @@ await GeneratorDriver.RunAsync(
         _ => null,
     },
     mapOutputPath: path => Path.ChangeExtension(Path.GetFileName(path), ".cs"),
-    usings: ["ZeroC.Slice.Codec"]).ConfigureAwait(false);
+    usings: ["System.Collections.Generic", "System.Linq", "ZeroC.Slice.Codec"]).ConfigureAwait(false);
 
 return 0;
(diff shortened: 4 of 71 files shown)


src/IceRpc.Slice.Generator/Program.cs
```diff
@@ -16,6 +16,14 @@ await GeneratorDriver.RunAsync(
         ? CodeBlock.FromBlocks([ProxyGenerator.Generate(interfaceDef), ServiceGenerator.Generate(interfaceDef)])
         : null,
     mapOutputPath: path => Path.ChangeExtension(Path.GetFileName(path), ".IceRpc.cs"),
-    usings: ["IceRpc.Slice", "IceRpc.Slice.Operations", "ZeroC.Slice.Codec"]).ConfigureAwait(false);
+    usings:
+    [
+        "IceRpc.Slice",
+        "IceRpc.Slice.Operations",
+        "System",
+        "System.Collections.Generic",
```
InsertCreativityHere: This PR claims that we emit unqualified collection types, but as far as I can tell, that's simply incorrect. Every `IList`, `Dictionary`, `IEnumerable`, etc. that we emit, is always fully qualified. Adding this `using` doesn't fix any bug.

I do see a use-case in making the `cs::type` attribute slightly more ergonomic.
I'm not completely against that, but I definitely think it should be in a different PR, since "making `cs::type` more friendly" is a completely unrelated to the implicit-usings.

InsertCreativityHere (commented):
I think all the changes to the tests are awesome and good and we should keep them. But I'm not sure about the changes to the 2 generators. Our original philosophy was to always qualify our own type-emissions, instead of injecting `using`s, and it's clear that in 2 places, we weren't careful enough.

That philosophy isn't set in stone of course, but this PR seems like a weak reason to deviate. The bugs in question could be fixed by just slightly tweaking 2 lines of code in our generators, with 0 side-effects to any users.

Adding 2 or 3 new `using`s to every generated file (possibly breaking some user's Slice in a patch release)... it just feels like a worse solution to the problem here.

bernardnormier:
> Adding 2 or 3 new `using`s to every generated file (possibly breaking some user's Slice in a patch release)... -t just feels like a worse solution to the problem here.

It appears you misunderstood the issue.

Before: we relied on the default implicit usings. You had to specify these default implicit usings in your project to get the generated code to build successfully

After (with this PR): we no longer depend on the default implicit usings because we generate them (really: those we may need) in the generated .cs files.

src/IceRpc.Slice.Generator/Program.cs
bernardnormier: You're right — I checked: the generated corpora contain 541 fully qualified System.Collections.Generic references and zero unqualified generator-emitted collection types. The PR now takes the direction your comment points to: no preamble usings at all. The two emissions that relied on implicit usings are now fully qualified like everything else (global::System.Span<byte>, and global::System.Linq.Enumerable.Count(...) in its static form), and the corpora's cs::type texts qualify their collection types since those projects now disable implicit usings — with implicit usings enabled, cs::type("List<T>") keeps working as before.

externl (approved):
Looks good.

InsertCreativityHere (approved):
Looks good to me!

pepone (approved):

Issue #4898, opened 2026-08-27
Improve the runtime test coverage of variant enums


bernardnormier:
The variant enum tests in `ZeroC.Slice.Generator.Tests` only exercise fields with builtin types (`uint32`, `uint16?`, `string`) and a tagged basic enum. No test encodes or decodes a variant enum with a struct-typed field:

- plain struct field
- optional struct field (bit sequence path)
- tagged struct field

The `ShadowedSymbol` enum in `VariantEnumTests.slice` (added by the fix for #4896) declares all three shapes, so the generated code compiles — but nothing round-trips them at runtime. Sequence- and dictionary-typed variant fields are similarly uncovered.

PR #4899, opened 2026-08-27
Fix shadowing of type names by variant case record names

### LINKED ISSUE #4896: Variant enum case field types can be shadowed by the case record names

bernardnormier:
Fixes #4896.

The variant enum mapping nests one case record per case inside the union record, so within the union body the case record names shadow same-namespace type names. The case parameter list and the case `Encode` method now emit fully qualified type names (empty current namespace).

The parameter list regressed in #4877; the `Encode` method exposure predates it — a tagged (or otherwise lambda-encoded) field whose type name matches a case name produced code that binds the type name to the sibling case record (`error CS8121`). It went unnoticed because no variant enum had such a field.

The new `ShadowedSymbol` enum in `VariantEnumTests.slice` covers both: a self-colliding case (`MyStruct(v: MyStruct)`), and colliding optional and tagged fields. Compiling the generated code is the regression test; #4898 tracks runtime coverage for struct-typed variant fields, a pre-existing test gap that is independent of shadowing.

The second commit regenerates the checked-in `src/ZeroC.Slice.Symbols/Compiler` files with the fixed generator: only the version stamps change, confirming the fix restores the exact output these files were generated with.

## What's Changed entry

Area: Slice codec

- Fixed the generated code for variant enums whose case names match the type names used by their fields.


externl (approved):
Looks good.

src/ZeroC.Slice.Generator/VariantEnumGenerator.cs
```diff
@@ -132,9 +132,12 @@ private static CodeBlock GenerateVariantRecord(
     {
         string variantName = variant.Name;
 
+        // Inside the union record, the case record names shadow the type names of the enclosing namespace. We set
+        // currentNamespace to "" to generate fully qualified type names for the parameter list and the encode method.
```
InsertCreativityHere: Big fan of adding a comment so we don't accidentally remove this again!
A slight tweak though, since this isn't something that always happens. In reality it should be rare.
```suggestion
        // Inside the union record, the case record names can shadow type names from the enclosing namespace. We set
        // currentNamespace to "" to generate fully qualified type names for the parameter list and the encode method.
```

InsertCreativityHere (approved):
Looks good to me!

src/ZeroC.Slice.Generator/VariantEnumGenerator.cs
bernardnormier: Applied, thanks.

PR #4900, opened 2026-08-27
Improve the runtime test coverage of variant enums

### LINKED ISSUE #4898: Improve the runtime test coverage of variant enums

bernardnormier:
Fixes #4898.

The new `VariantFieldKinds` enum in `VariantEnumTests.slice` covers the variant field shapes that had no runtime coverage: a plain struct field, an optional struct field (bit sequence path), a tagged struct field, a sequence field and a dictionary field. Each case gets an encode/decode round-trip test.

The sequence and dictionary cases map to `IList<int>` / `IDictionary<int, string>` record parameters, which the generated record equality compares by reference — those tests assert on the decoded case's property instead.

## What's Changed entry

None.

### DIFF
--- a/tests/ZeroC.Slice.Generator.Tests/VariantEnumTests.cs
+++ b/tests/ZeroC.Slice.Generator.Tests/VariantEnumTests.cs
@@ -168,6 +168,110 @@ public class VariantEnumTests
     }
 
     [Test]
+    public void Decode_variant_enum_with_struct_field()
+    {
+        // Arrange
+        var buffer = new MemoryBufferWriter(new byte[256]);
+        var encoder = new SliceEncoder(buffer);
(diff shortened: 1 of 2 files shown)


externl (approved):
Looks good.

InsertCreativityHere (approved):
Looks good to me!

PR #4901, opened 2026-08-28
Match protoc's name mapping in the Protobuf generator and build tasks

### LINKED ISSUE #4792: [Audit-Medium] Predicted Protobuf output filename diverges from protoc's own C# filename

## Summary

`ToPascalCase` capitalizes only when the previous character is not a letter-or-digit. protoc's own `UnderscoresToPascalCase` additionally sets `cap_next_letter` after every **digit** (and treats non-ASCII as a separator).

The two diverge whenever a `.proto` basename contains a digit immediately followed by a lowercase ASCII letter. `UpToDateCheckTask` and `OutputFileNamesTask` use `ToPascalCase` to predict what `--csharp_out` will write, and the targets add `%(OutputDir)/%(OutputFileName).cs` to `Compile` — so the build references a file protoc never wrote.

Codex probed the repo's own pinned protoc 35.1 directly and confirmed the divergence on every listed name:

| `.proto` | protoc writes | task predicts |
|---|---|---|
| `v2ray.proto` | `V2Ray.cs` | `V2ray.cs` |
| `utf8string.proto` | `Utf8String.cs` | `Utf8string.cs` |
| `sha256hash.proto` | `Sha256Hash.cs` | `Sha256hash.cs` |
| `oauth2client.proto` | `Oauth2Client.cs` | `Oauth2client.cs` |
| `s3object.proto` | `S3Object.cs` | `S3object.cs` |

The IceRpc plug-in shares `Converter.cs`, so `V2ray.IceRpc.cs` is self-consistent — only the protoc-side prediction is wrong.

## Trigger / impact

On a case-sensitive volume a clean `dotnet build` fails: `CSC : error CS2001: Source file '.../generated/V2Ray.cs' could not be found.`

Two effects are **permanent** on case-sensitive filesystems:

1. `File.Exists(csharpOutput)` is never true, so `UpToDate` is always false and **protoc re-runs on every build**.
2. `ProtoClean` deletes the phantom name, orphaning protoc's real output.

CS2001 itself fires on the **first (clean) build only** in the default layout — on rebuild, `Exclude="@(Compile->'%(FullPath)')"` matches case-insensitively against the default-glob-included `V2Ray.cs` and suppresses the bogus item. It persists on every build when `EnableDefaultCompileItems=false` or when `OutputDir` is outside the project. Codex verified this accidental-recovery path in the MSBuild source.

## Why this is not the already-dismissed PascalCase issue

#4446, #4452 and #4473 were all closed as **Not planned**, and [AGENTS.md](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/AGENTS.md) dismissals #3 and #4 record the rationale: identifier-mapping collisions are rare, and output-filename collisions fail at build time by design.

This is a **different defect**, and Codex was asked specifically to adjudicate that question. Its conclusion:

> "This is distinct from #4446/#4452: those concern two inputs collapsing onto one mapped name; here one input's predicted built-in output disagrees with protoc's actual output."

There is no collision here — a single `.proto` file, on its own, produces a build that references a nonexistent file and an up-to-date check that never succeeds.

## Source references

- [src/IceRpc.Protobuf.Tools/Converter.cs#L20](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Protobuf.Tools/Converter.cs#L20) — the capitalization predicate
- [src/IceRpc.Protobuf.Tools/UpToDateCheckTask.cs#L43](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Protobuf.Tools/UpToDateCheckTask.cs#L43) and [src/IceRpc.Protobuf.Tools/OutputFileNamesTask.cs#L33](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Protobuf.Tools/OutputFileNamesTask.cs#L33) — the prediction sites

## Recommendation

Make the filename converter match protoc's `UnderscoresToPascalCase` (capitalize after digits, restrict to ASCII) — or, better, stop predicting entirely and derive the generated-file list from the `.d` file protoc already writes.

## Related issues

- #4446, #4452, #4473 — all **Not planned**; the collision class, distinct from this as explained above.
- `M-12` of the companion repository-wide audit concerns the generated C# **namespace**, not the filename.

## Audit context

Finding 7 of [general-correctness-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-07-27.md); Codex verdict **CONFIRMED** at **Medium** — "a valid single input can fail every clean build on common case-sensitive systems, while incremental checks and Clean remain permanently incorrect".

Severity: **Medium**.

### LINKED ISSUE #4812: [Audit-Medium] Protobuf package fallback disagrees with protoc's generated C# namespace

## Summary

When `csharp_namespace` is absent, IceRPC uses `descriptor.Package` **verbatim**:

```csharp
else
{
    return descriptor.Package;
}
```

protoc's C# generator instead PascalCases it. For `package lower.snake_case`, protoc generates `Lower.SnakeCase` while IceRPC generates `lower.snake_case`, so the IceRPC service code cannot resolve the request and response types.

## Source references

- [src/IceRpc.Protobuf.Generator/FileDescriptorExtensions.cs#L16-L19](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Protobuf.Generator/FileDescriptorExtensions.cs#L16)

## Recommendation

Implement the same fallback namespace mapping as Google's C# generator, and add compile tests for lowercase, underscore, empty, and explicitly overridden package namespaces.

## Related

Distinct from #4792, which is the same class of protoc-divergence but concerns the generated **filename** rather than the namespace.

## Audit context

M-12 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md); confirmed by the [Claude verification pass](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-claude-review-2026-07-28.md).

Severity: **Medium**.

bernardnormier:
Fixes #4812.
Fixes #4792.

`Converter.ToPascalCase` capitalized a letter only after a non-letter-or-digit character, while protoc's `UnderscoresToCamelCase` also capitalizes after every digit and drops non-ASCII characters. As a result, for a file like `sha256hash.proto`, the build tasks predicted `Sha256hash.cs` while protoc writes `Sha256Hash.cs` — on a case-sensitive file system, a clean build fails with CS2001, the up-to-date check never succeeds, and `ProtoClean` deletes the wrong file (#4792). Separately, when a file has no `csharp_namespace` option, the generator used the Protobuf package verbatim as the C# namespace, while protoc maps `package lower.snake_case` to `Lower.SnakeCase` — so the generated client/service code could not resolve the message types protoc generates (#4812).

The converter now replicates protoc's algorithm and is renamed to make the contract explicit — Google.Protobuf exposes no public API for this mapping, so it's a re-implementation with `<remarks>` pinning it to its protoc counterparts:

- `ToProtocPascalCase` matches `UnderscoresToCamelCase`: used for the generated file names (build tasks + generator) and the service/rpc identifier mapping.
- `ToCsharpNamespace` matches `GetFileNamespace` (periods preserved): used by the package fallback in `GetCsharpNamespace`.

The new `sha256hash.proto` test file (digit-followed-by-letter name, `package lower.snake_case`, no `csharp_namespace` option) exercises the whole chain at build time: protoc writes `Sha256Hash.cs`, the tasks predict the same name, and `Sha256Hash.IceRpc.cs` declares `namespace Lower.SnakeCase`, matching protoc. The new test invokes an operation through the generated client and service. The converter output was also verified against protoc 3.35.1 for the probe names from #4792 (`v2ray` → `V2Ray`, `utf8string` → `Utf8String`, `oauth2client` → `Oauth2Client`, `s3object` → `S3Object`).

## What's Changed entry

Area: Protobuf

- The Protobuf build tasks and `protoc-gen-icerpc-csharp` now map `.proto` file names and package names to C# exactly like `protoc`. In particular, the generated code for a file without the `csharp_namespace` option now uses the same namespace as the code protoc generates, and a generated file name capitalizes letters that follow digits (`sha256hash.proto` → `Sha256Hash.IceRpc.cs`). Service and rpc names are mapped the same way (`auth2service` → `Auth2ServiceClient`).


src/IceRpc.Protobuf.Tools/Converter.cs
```diff
@@ -6,19 +6,48 @@ namespace IceRpc.CaseConverter.Internal;
...
+            else
             {
-                previousChar = c;
-                continue;
+                capitalizeNext = true;
+                if (c == '.' && preservePeriods)
+                {
+                    sb.Append('.');
+                }
             }
```
InsertCreativityHere: This can be done in 4 lines of code, which I also think better express the intent than char comparisons:
```suggestion
            if (char.IsAsciiLetterOrDigit(c) || (c == '.' && preservePeriods))
            {
                sb.Append(capitalizeNext ? char.ToUpperInvariant(c) : c);
            }
            capitalizeNext = !char.IsAsciiLetter(c);
```
Just to be sure, I had my Claude ensure these are truly identical implementations.

InsertCreativityHere (approved):
Assuming the description of Protobuf's behavior is correct: looks good to me!

But I think the implementation of `Convert` could be alot cleaner.

src/IceRpc.Protobuf.Tools/Converter.cs
```diff
@@ -6,19 +6,48 @@ namespace IceRpc.CaseConverter.Internal;
...
     {
         var sb = new StringBuilder(value.Length);
-        char previousChar = '_';
+        bool capitalizeNext = true;
         foreach (char c in value)
         {
-            if (!char.IsLetterOrDigit(c))
+            if (c is >= 'a' and <= 'z')
+            {
+                sb.Append(capitalizeNext ? (char)(c + ('A' - 'a')) : c);
```
pepone: Can we keep using ToUpperInvariant?


```suggestion
                sb.Append(capitalizeNext ? char.ToUpperInvariant(c) : c);
```

pepone (approved):

src/IceRpc.Protobuf.Tools/Converter.cs
bernardnormier: Applied, thanks — it covers pepone's `ToUpperInvariant` point too.

src/IceRpc.Protobuf.Tools/Converter.cs
bernardnormier: Done via Austin's rewrite above.

PR #4904, opened 2026-08-28
Point multi-server applications to AddIceRpcConnectionCache

### LINKED ISSUE #4859: Cannot register multiple client connections with AddIceRpcClientConnection
`AddIceRpcClientConnection` always configures the unnamed `ClientConnectionOptions`. Calling it twice with different server addresses silently collapses: both `Configure` callbacks target the same unnamed options (the second address wins) while two `ClientConnection` singletons are registered, both pointing at it.

`AddIceRpcServer` supports the equivalent scenario with its `optionsName` overloads; there is no client connection counterpart. This needs a small design pass: named options, plus how the `IInvoker` registration disambiguates when there are several connections.

Noted by @externl while reviewing #4855.

bernardnormier:
Fixes #4859.

As discussed in #4859, the silent collapse when calling `AddIceRpcClientConnection` twice is standard DI behavior (additive `Configure` callbacks on the unnamed options, duplicate singleton registrations), and named client connections would be a poor fit — the multi-server scenario is what `ConnectionCache` is designed for.

The remarks of the three `AddIceRpcClientConnection` overloads now point applications that connect to multiple servers to `AddIceRpcConnectionCache`.

## What's Changed entry

None.


externl (approved):

InsertCreativityHere (approved):
Looks good to me!

Issue #4911, opened 2026-08-31
The compression trailer write can block indefinitely on flow control


bernardnormier:
When completing the compressor's payload writer without an exception, `CompressorPipeWriter.Complete` disposes the Brotli/Deflate stream, which writes the compression trailer (a few bytes) into the decoratee. This write goes through `PipeWriter.AsStream`'s synchronous `Write`, i.e. `decoratee.WriteAsync(trailer, CancellationToken.None)` awaited synchronously.

A peer that behaves properly unblocks this write quickly: it keeps reading (its decompressor needs the trailer to reach the end of the payload), or it abandons the payload by completing its input and the resulting stop-sending unblocks the write immediately. Connection shutdown, loss and disposal also unblock it, with an exception that completes the decoratee.

But we can't count on the peer to behave properly — it can be buggy or malicious. A peer that stops reading without completing its input, after the payload consumed the last of the send window, blocks the trailer write with no cancellation. `Complete` runs after the payload copy completes, so the invocation token is never consulted: the invocation deadline cannot rescue it, and the synchronous wait pins a thread pool thread. Only connection-level events (idle timeout, shutdown, dispose) end the wait — a peer that answers pings while stalling the stream can extend it indefinitely. The same stall during the payload copy is bounded by the invocation token; the graceful `Complete` is uniquely exposed.

The underlying limitation is in System.IO.Pipelines: neither `Complete` nor `CompleteAsync` takes a cancellation token, so a `PipeWriter` decorator that must write trailing bytes has no cancelable "finish" call to write them in. In particular, switching the payload writer completion from `Complete` to `CompleteAsync` does not fix this bug, independently of any other difficulty: the trailer write it performs still cannot be canceled.

## Fix options

[...]

PR #4908, opened 2026-08-31
Document the synchronous Complete convention

bernardnormier:
Adds a dismissed-audit pattern for the decision recorded in #965: IceRPC completes pipe readers and writers with `Complete`, never `CompleteAsync`, and treats `Complete` as a fast, non-blocking call. The pattern also records why this works (no `Complete()` with unflushed bytes; decorators and transports don't wait in `Complete`), why switching to `CompleteAsync` wouldn't help (no cancellation token), and the one known deviation: the compressor's trailer write, a real bug tracked by #4911.

The review of #4840 showed this decision was not captured anywhere an AI reviewer would find it.

## What's Changed entry

None — documentation for AI coding agents.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4912, opened 2026-08-31
Check for unflushed bytes in CompressorPipeWriter.Complete

bernardnormier:
`SlicPipeWriter.Complete` and `QuicPipeWriter.Complete` throw `InvalidOperationException` when called without an exception while the writer holds unflushed bytes. This PR adds the same check to `CompressorPipeWriter`, per the synchronous `Complete` convention (#4908): a graceful `Complete` must have no payload data left to flush — for the compressor, the only bytes `Complete` writes are the compression trailer. The trailer comment is reworded accordingly.

See #4911 for a related flow-control concern with the trailer write.

## What's Changed entry

None — enforces an existing convention; no behavior change for conforming payload writers.

### DIFF
--- a/src/IceRpc.Compressor/CompressorPipeWriter.cs
+++ b/src/IceRpc.Compressor/CompressorPipeWriter.cs
@@ -29,13 +29,19 @@ internal class CompressorPipeWriter : PipeWriter
     {
         if (!_isCompleted)
         {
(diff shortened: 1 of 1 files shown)


externl (approved):
Looks good!

InsertCreativityHere (approved):
Looks good to me!

pepone (approved):

Issue #4906, opened 2026-08-31
Quic tests fail with QUIC_STATUS_TLS_ERROR since libmsquic 2.6.1


externl:
Every Quic test on ubuntu-24.04 fails at connect with `System.Security.Authentication.AuthenticationException: Authentication failed: Status code: QUIC_STATUS_TLS_ERROR`, then the test host crashes and the run aborts. This also fails `DefaultTransportTests.Connect_with_default_multiplexed_transport` in IceRpc.Tests.

- CI installs libmsquic unpinned; the last green run (Aug 18) got 2.6.0, failing runs get 2.6.1 from packages.microsoft.com. A re-run of main's last green commit fails identically, so no repo change is involved: https://github.com/icerpc/icerpc-csharp/actions/runs/33187165565/job/99498846517
- The runner image also moved (ubuntu24/20260810.271 -> 20260823.283), so it may be msquic 2.6.1 against the image's OpenSSL.
- Workaround: We could pin `libmsquic=2.6.0` in the install step until 2.6.1 is understood. I think we should figure out show severe this is first though.

externl:
Root cause: msquic 2.6.1's hostname-validation change (microsoft/msquic#6275, cherry-pick of microsoft/msquic#6274) fails any handshake whose target host is an IP literal or null with `QUIC_STATUS_TLS_ERROR`, before .NET's `RemoteCertificateValidationCallback` runs — and every test here connects to `icerpc://127.0.0.1`. Reproduced on macOS with brew libmsquic 2.6.1 using a minimal QuicConnection probe: `TargetHost` of `127.0.0.1` or null fails; `localhost` — even a mismatched DNS name — completes and defers to the validation callback. The server cert's SAN includes `IP:127.0.0.1`, so this breaks even legitimate IP validation; reported upstream: microsoft/msquic#6288.

Note macOS CI passes only because the runner's Homebrew index still serves 2.5.9.

externl:
.NET fixed this issue in https://github.com/dotnet/runtime/issues/133096. Hopefully it lands in the next release of .NET 10.

externl:
The fix (which ended up being in .NET) should be included in the next release of .NET (https://github.com/microsoft/msquic/issues/6288#issuecomment-5616341444).

PR #4916, opened 2026-09-02
Clarify the IIceFeature, ISliceFeature and IProtobufFeature docs

### LINKED ISSUE #4905: Clarify the IIceFeature property documentation
The XML documentation for several `IIceFeature` properties does not fully describe their behavior:

- `BaseProxy` does not document what the base proxy contributes or what happens when it is `null`.
  - A proxy decoded from an incoming request receives `InvalidInvoker.Instance` by default.
  - A proxy decoded from an incoming response inherits the invoker and encode options of the proxy that sent the request.
- `MaxPayloadSize` does not state that it applies only when decoding incoming payloads. Its definition also omits Ice exception payloads.
- `MaxCollectionAllocation` does not explain that it is a cumulative allocation budget across decoded strings and collections, rather than a per-collection limit.

Update these doc comments so the public contract matches the existing implementation. This is a documentation-only change.

bernardnormier:
Fixes #4905

Documentation-only change to the three feature interfaces:

- **BaseProxy** (Ice, Slice): a decoded proxy inherits the invoker and encode options of the base proxy. When null, a proxy decoded from a request gets `InvalidInvoker.Instance`, while a proxy decoded from a response inherits from the proxy that sent the request. For Slice, also documents that relative service addresses are resolved against the base proxy's service address.
- **MaxPayloadSize** / **MaxSegmentSize** / **MaxMessageLength**: these limits apply only when decoding incoming payloads. The full payload, segment, or message is read into memory before decoding, and the read throws `InvalidDataException` when the limit is exceeded. The size excludes the encapsulation header, segment size prefix, or 5-byte message prefix respectively. `MaxPayloadSize` now also lists Ice exception payloads.
- **MaxCollectionAllocation** (Ice, Slice): a cumulative budget across every string, sequence, and dictionary decoded from one payload (Ice) or one segment (Slice), not a per-collection limit.

- **ConnectionOptions.MaxIceFrameSize**: documents that this property, not `MaxPayloadSize`, is the counterpart of the Ice property `Ice.MessageSizeMax`, and how the two differ.

### DIFF
--- a/src/IceRpc.Ice/Features/IIceFeature.cs
+++ b/src/IceRpc.Ice/Features/IIceFeature.cs
@@ -16,9 +16,9 @@ public interface IIceFeature
 
     /// <summary>Gets the base proxy used when decoding a service address into a proxy.</summary>
     /// <value>The base proxy. A decoded proxy inherits the invoker and encode options of the base proxy. When
-    /// <see langword="null" />, a proxy decoded from an incoming request receives
-    /// <see cref="InvalidInvoker.Instance" /> as its invoker, and a proxy decoded from an incoming response inherits
-    /// the invoker and encode options of the proxy that sent the request.</value>
+    /// <see langword="null" />, the proxy that sent the request serves as the base proxy for a proxy decoded from an
+    /// incoming response, while a proxy decoded from an incoming request receives
+    /// <see cref="InvalidInvoker.Instance" /> as its invoker.</value>
     IIceProxy? BaseProxy { get; }
 
     /// <summary>Gets the options to use when encoding the payload of an outgoing response.</summary>
@@ -44,15 +44,11 @@ public interface IIceFeature
     /// encoded arguments of an operation, the encoded return values of an operation, or the encoded Ice exception
     /// carried by a response with status code <see cref="StatusCode.ApplicationError" />.</summary>
     /// <value>The maximum size of an Ice-encoded payload, in bytes.</value>
-    /// <remarks><para>This limit applies only when decoding the payload of an incoming request or response: the
-    /// decoding throws <see cref="InvalidDataException" /> when the payload size carried by the request or response
-    /// exceeds this maximum, before decoding any byte of the payload. It does not restrict the size of the payloads
-    /// encoded by the application. The payload size does not include the size of any header for this payload, such
-    /// as the encapsulation header with the ice protocol. Implementations must return a value greater than or equal
-    /// to <c>0</c> and less than <see cref="int.MaxValue" />.</para>
-    /// <para>This property is the counterpart of the Ice property
-    /// <see href="https://docs.zeroc.com/ice/3.8/cpp/ice#Ice.MessageSizeMax">Ice.MessageSizeMax</see>, with two
-    /// differences: the maximum payload size is in bytes rather than kilobytes, and it excludes the protocol header
-    /// that Ice.MessageSizeMax includes.</para></remarks>
+    /// <remarks>This limit applies only when decoding the payload of an incoming request or response: the decoding
+    /// throws <see cref="InvalidDataException" /> when the payload size carried by the request or response exceeds
+    /// this maximum, before decoding any byte of the payload. It does not restrict the size of the payloads encoded by
+    /// the application. The payload size does not include the size of any header for this payload, such as the
+    /// encapsulation header with the ice protocol. Implementations must return a value greater than or equal to
+    /// <c>0</c> and less than <see cref="int.MaxValue" />.</remarks>
     int MaxPayloadSize { get; }
 }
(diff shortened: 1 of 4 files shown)


src/IceRpc.Ice/Features/IIceFeature.cs
```diff
@@ -34,10 +40,14 @@ public interface IIceFeature
...
     /// <value>The maximum size of an Ice-encoded payload, in bytes.</value>
-    /// <remarks>The payload size does not include the size of any header for this payload, such as the encapsulation
-    /// header with the ice protocol. Implementations must return a value greater than or equal to <c>0</c> and less
-    /// than <see cref="int.MaxValue" />.</remarks>
+    /// <remarks>This limit applies only when decoding the payload of an incoming request or response: the payload is
+    /// read in full into memory before it is decoded, and this read throws <see cref="InvalidDataException" /> when
+    /// the payload exceeds this size. It does not restrict the size of the payloads encoded by the application. The
+    /// payload size does not include the size of any header for this payload, such as the encapsulation header with
+    /// the ice protocol. Implementations must return a value greater than or equal to <c>0</c> and less than
+    /// <see cref="int.MaxValue" />.</remarks>
```
pepone: Maybe link to MaxCollectionAllocation and state this is on the wire size, not the decoded memory size.

src/IceRpc.Slice/Features/ISliceFeature.cs
```diff
@@ -19,13 +24,20 @@ public interface ISliceFeature
 
     /// <summary>Gets the maximum collection allocation when decoding a payload, in bytes.</summary>
     /// <value>The maximum collection allocation.</value>
-    /// <remarks>Implementations must return a value greater than or equal to <c>0</c>.</remarks>
+    /// <remarks>This value is a cumulative budget for the decoding of one Slice payload segment, not a limit on the
+    /// size of each collection: the decoder adds the estimated memory size of every string, sequence, and dictionary
+    /// it decodes, and throws <see cref="InvalidDataException" /> once this total exceeds the maximum collection
+    /// allocation. Implementations must return a value greater than or equal to <c>0</c>.</remarks>
```
pepone: The estimation is done before the actual decoding isn't it? maybe worth clarify.

externl (approved):
Looks good!

src/IceRpc.Ice/Features/IIceFeature.cs
bernardnormier: Reworded in fd4a66f02c6ae32a54a5a402816c7da5b624d24e to say what is checked: the payload size carried by the request or response, before any byte of the payload is decoded. The summary already pins the value to the size of the Ice-encoded payload. I also added a paragraph relating this property to Ice.MessageSizeMax, which Ice users will look for.

src/IceRpc.Slice/Features/ISliceFeature.cs
bernardnormier: Yes, the budget is charged from the encoded size or element count before the collection is decoded. Clarified in fd4a66f02c6ae32a54a5a402816c7da5b624d24e, in both ISliceFeature and IIceFeature.

pepone (commented):
- `IIceFeature.MaxPayloadSize`, Ice.MessageSizeMax paragraph: the counterpart of `Ice.MessageSizeMax` is `ConnectionOptions.MaxIceFrameSize` (connection-level, whole ice frame including the header, bytes, default 1 MB, closes the connection), not this property. With the defaults, an oversized ice payload trips `MaxIceFrameSize` before this limit is ever consulted, and `0` here rejects every non-empty payload instead of disabling the limit like `Ice.MessageSizeMax=0`. Either point Ice users to `MaxIceFrameSize`, or reframe the paragraph around those differences and drop "two differences".
- `ISliceFeature.BaseProxy`, null case for responses: the sender also becomes the base for relative service addresses, not just for the invoker and encode options. Suggest: "When `null`, the proxy that sent the request serves as the base proxy for a proxy decoded from an incoming response, while a proxy decoded from an incoming request receives `InvalidInvoker.Instance` as its invoker and keeps its relative service address as is." Same shape works for `IIceFeature.BaseProxy`.
- Milestone 0.6.1, to match #4905.
- Rebase on main to pick up #4920; the macOS red is #4906.

bernardnormier:
Thanks, all four points addressed in 4199c7ba48f70b17b14cf55331e61c62bd0c24dc:
- Removed the Ice.MessageSizeMax paragraph from `IIceFeature.MaxPayloadSize`. The counterpart note now lives on `ConnectionOptions.MaxIceFrameSize`, where it belongs.
- `BaseProxy` null case reworded as suggested, in both `ISliceFeature` and `IIceFeature`.
- Milestone 0.6.1 and the documentation label set.
- Rebased on main.

pepone (approved):

PR #4589, opened 2026-05-07
IceRpc.Slice.Tools: prune stale generated outputs across rebuilds

pepone:
The build now writes a manifest (obj/slicec.outputs.txt) listing every path produced by the latest Slicec run. Before each subsequent build (and before SlicecClean), any path in the previous manifest that the current build no longer produces is deleted, so:

- removing or renaming a .slice file no longer leaves orphan .cs files matching the SDK's default Compile glob,
- changing OutputDir prunes the old folder's outputs,
- toggling IceRpc=true->false prunes the old .IceRpc.cs.

Closes #4474.

PR #4923, opened 2026-09-05
Prune stale Protobuf outputs and tighten the protoc up-to-date check

### LINKED ISSUE #4817: [Audit-Medium] Protobuf incremental build accepts stale outputs and never prunes removed inputs

Two defects in the same incremental-build machinery, filed together because they share a fix surface.

## M-17 — removing or renaming a Proto input leaves active stale generated code

`ProtoClean` computes what to delete from the **current** item set:

```xml
<OutputFileNamesTask Sources="@(ProtoFile)"> ... </OutputFileNamesTask>
<Delete Files="@(_ProtoFile->'%(OutputDir)/%(OutputFilename).cs')" />
```

A removed `.proto` is no longer in `@(ProtoFile)`, so its outputs are never deleted. There is no output manifest or pruning step equivalent to the Slice tooling, and the default Compile glob keeps compiling the orphaned generated types.

## M-18 — incremental validity checks can accept stale or incomplete outputs

`UpToDateCheckTask`:

- compares dependencies against the **newest** output (`Math.Max` across `.d`, `.cs`, `.IceRpc.cs`) instead of the oldest, so a deliberately older `.IceRpc.cs` is masked by a newer `.cs`;
- calls `File.GetLastWriteTime` on dependency paths without existence checks — a **missing** dependency returns a year-1601 timestamp and therefore reads as up to date;
- includes no task/generator version, no generator options or metadata, and ignores generator binaries newer than all outputs.

## Source references

- [src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets#L124-L131](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets#L124) — `ProtoClean` over the current item set
- [src/IceRpc.Protobuf.Tools/UpToDateCheckTask.cs#L48-L62](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Protobuf.Tools/UpToDateCheckTask.cs#L48) — newest-output comparison and unchecked dependency timestamps

## Recommendation

Persist the previous generated-output set and remove outputs no longer produced before `CoreCompile`. Require every dependency and output to exist, compare the newest **input** against the **oldest output**, and fingerprint all generator versions/options that affect output.

## Audit context

M-17 and M-18 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md); both confirmed by the [Claude verification pass](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-claude-review-2026-07-28.md).

Severity: **Medium**.

pepone:
Fixes #4817

Builds on the item rename made in #4922.

### M-17: stale outputs of removed or renamed Proto files

`ProtoClean` only ever deleted the outputs of the *current* `ProtoFile` items, so a removed or renamed `.proto` left `Hello.cs` and `Hello.IceRpc.cs` behind for the SDK's default `Compile` glob to keep compiling. The targets now mirror the Slice tooling (#4589):

- A new `_ComputeProtoOutputs` target computes the absolute paths of the files the build produces for the current items: the two C# files, the dependency file and the options record described below. `ProtoCompile`, `ProtoClean` and the two targets below depend on it. It runs only the name computation, not the up-to-date check, for the reason noted in #4922: on Rebuild it runs before Clean.
- `_ProtoWriteOutputManifest` (after `ProtoCompile`) records those paths in `$(IntermediateOutputPath)protoc.outputs.txt`.
- `_ProtoPruneStaleOutputs` (before `ProtoCompile` and `ProtoClean`) deletes anything in the previous manifest that the current build no longer produces, and removes it from `Compile` first so csc doesn't fail with CS2001. It prunes nothing when `_ComputeProtoOutputs` was skipped because the task assembly is missing. When the last Proto file is removed it also deletes the manifest, which `_ProtoWriteOutputManifest` no longer replaces; otherwise every later build would treat the former output paths as its own and delete whatever appears there.
- `ProtoClean` deletes the manifest and the inputs cache along with the generated files.

### M-18: up-to-date check accepting stale outputs

`UpToDateCheckTask` now:

- compares the newest input against the **oldest** output rather than the newest, so an old `Hello.IceRpc.cs` is no longer masked by a fresh `Hello.cs`;
- treats a **missing** dependency as out of date, instead of reading the placeholder timestamp `File.GetLastWriteTime` returns for missing files. A deleted import now surfaces as a protoc error rather than a silently stale build;
- takes `AdditionalInputs`: the `protoc` binary, `IceRpc.Protobuf.Generator.dll` and `$(IntermediateOutputPath)protoc.inputs.cache`, so a rebuilt or upgraded generator regenerates the code;
- sees a fresh `protoc.inputs.cache` when the configuration shared by all Proto files changes: `ProtoCompile` writes the compiler and generator versions and the `ProtoSearchPath` entries to it with `WriteOnlyWhenDifferent`, before the check. The file's timestamp moves only when a value changes; every Proto file is then out of date, including those a failed run left untouched;
- sees a fresh per-file record when a Proto file's `AdditionalOptions` change: `ProtoCompile` writes them, one per line, to `<OutputDir>/<Name>.options` next to the dependency file, with the same `WriteOnlyWhenDifferent` rule, and the task lists the record among that file's inputs. Only that file is then out of date, and a change reverted after a failed build is still caught, since the record is rewritten again. The record is also one of the files `_ComputeProtoOutputs` lists, so the manifest, prune and clean cover it. Adding a Proto file doesn't regenerate the others.

The README note saying that changing `AdditionalOptions` requires a `dotnet clean` is replaced by a short description of the incremental behavior.

### Verification

One-file probe importing the source-build props/targets (macOS, .NET SDK 10.0.201), protoc runs per build:

| Scenario | Before | After |
|---|---|---|
| No change | 0 | 0 |
| `Hello.IceRpc.cs` older than the `.proto`, `Hello.cs` newer | 0 (stale) | 1 |
| Imported `other.proto` deleted | 0 (stale) | protoc error for the missing import |
| `AdditionalOptions` of one of two files changed / unchanged / reverted | 0 (stale) | 1 / 0 / 1, that file only |
| A file switches to the `AdditionalOptions` value the other file uses | 0 (stale) | 1 |
| One file's options changed and `protoc` fails on the other; the failure fixed | n/a | 2, then 1 (the failed file) |
| One file's options changed and `protoc` fails on the other; the change reverted and the failure fixed | n/a | 2, then 2 (the reverted file's code no longer carries the option) |
| `ProtoSearchPath` changed | 0 (stale) | 2 |
| Task assembly missing (source build, `--no-dependencies`) | task-load error | task-load error, generated code intact |
| `third.proto` added | 1 | 1 |
| `hello.proto` renamed to `greeter.proto` | 1, `Hello.*` left behind | 1, `Hello.*` deleted, no duplicate types |
| Last `.proto` removed | 0, outputs left behind | 0, `generated/` empty |
| Generator assembly touched | 0 (stale) | 1 |
| Fresh Rebuild / second Rebuild / Clean | 1 / 1 / outputs deleted | 1 / 1 / outputs, options records, manifest and inputs cache deleted |

From the branch, `IceRpc.Protobuf.BuildTelemetry` builds clean, the `IceRpc.Protobuf.Tests` suite passes (70 tests), and a second build of that test project runs protoc zero times. Upgrading from a previous version regenerates every Proto file once, since no options record exists yet. Not exercised: Windows, where the only platform-specific addition is the `.exe` extension on the protoc input path.

## What's Changed entry

Area: Protobuf

- Removing or renaming a `.proto` file, or changing its `OutputDir`, now deletes the C# files previously generated
  for it during the next build, instead of leaving them for the compiler to pick up.
- `protoc` now regenerates a Proto file's code when any of its generated files is older than the Proto file or one of
  its imports, when an import is missing, when `protoc` or the generator is updated, when the file's
  `AdditionalOptions` change, and when `ProtoSearchPath` changes. Running `dotnet clean` after changing
  `AdditionalOptions` is no longer needed.

### DIFF
--- a/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
+++ b/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
@@ -60,9 +60,94 @@
       Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)"
     />
   </ItemGroup>
-  <Target Name="ProtoCompile" BeforeTargets="CoreCompile" Condition="@(ProtoFile) != ''">
+  <!--
+    Computes the files that protoc produces for the current ProtoFile items, as absolute paths, and the fingerprint
+    of the configuration that produces them. ProtoCompile and the prune, manifest and clean targets share it, and
+    MSBuild runs it once per build.
+
+    It deliberately doesn't run the up-to-date check, and ProtoCompile keeps its own _ProtoFile item: Rebuild runs
+    Clean and Build in the same project instance, so an up-to-date check made before Clean would mark as current the
+    very outputs Clean is about to delete. A task <Output> also appends to an existing item, so sharing _ProtoFile
+    with ProtoCompile would leave it with two entries per ProtoFile.
+
+    The task assembly only exists in source builds once IceRpc.Protobuf.Tools has been built, which isn't the case
+    for a Clean that precedes any Build; the Exists check skips the computation in that case.
+  -->
+  <Target Name="_ComputeProtoOutputs"
+          Condition="@(ProtoFile) != '' And Exists('$(IceRpcProtobufToolsTaskAssembliesPath)IceRpc.Protobuf.Tools.dll')">
+    <OutputFileNamesTask Sources="@(ProtoFile)">
+      <Output ItemName="_ProtoNamedFile" TaskParameter="ComputedSources" />
+    </OutputFileNamesTask>
+    <ItemGroup>
+      <_ProtoOutput Include="$([MSBuild]::NormalizePath('%(_ProtoNamedFile.OutputDir)/%(_ProtoNamedFile.OutputFileName).cs'))" />
+      <_ProtoOutput Include="$([MSBuild]::NormalizePath('%(_ProtoNamedFile.OutputDir)/%(_ProtoNamedFile.OutputFileName).IceRpc.cs'))" />
+      <_ProtoOutput Include="$([MSBuild]::NormalizePath('%(_ProtoNamedFile.OutputDir)/%(_ProtoNamedFile.OutputFileName).d'))" />
+      <!-- The xxx.BuildTelemetry.txt is generated when both IceRpcBuildTelemetry and IceRpcBuildTelemetryDebug are true -->
+      <_ProtoOutput
+        Include="$([MSBuild]::NormalizePath('%(_ProtoNamedFile.OutputDir)/%(_ProtoNamedFile.OutputFileName).BuildTelemetry.txt'))"
+        Condition="'$(IceRpcBuildTelemetry)' == 'true' And '$(IceRpcBuildTelemetryDebug)' == 'true'" />
+    </ItemGroup>
+    <PropertyGroup>
+      <!--
+        Everything that changes the generated code without touching a Proto file: the compiler and generator
+        versions, the import search path and the per-file protoc options. A change regenerates every Proto file.
+      -->
+      <_ProtocFingerprint>protoc=$(ProtocBundledPluginVersion);icerpc-csharp=$(ProtocIceRpcPluginVersion);search-path=@(ProtoSearchPath->'%(FullPath)', ',');options=@(ProtoFile->'%(Identity)=%(AdditionalOptions)', ',')</_ProtocFingerprint>
+    </PropertyGroup>
+  </Target>
+
+  <!--
+    Delete generated files recorded in the previous build's manifest that the current build no longer produces.
+    Without this, a removed or renamed .proto file (or a ProtoFile whose OutputDir changed) leaves orphan .cs files
+    behind that the SDK's default Compile glob keeps picking up.
+
+    BeforeTargets covers both the ProtoFile-present case (ProtoCompile/ProtoClean) and the ProtoFile-empty case where
+    those targets are skipped by their conditions but stale files from a previous build still need pruning
+    (CoreCompile/Clean).
+  -->
+  <Target Name="_ProtoPruneStaleOutputs"
+          BeforeTargets="ProtoCompile;CoreCompile;ProtoClean;Clean"
+          DependsOnTargets="_ComputeProtoOutputs"
+          Condition="Exists('$(IntermediateOutputPath)protoc.outputs.txt')">
+    <ReadLinesFromFile File="$(IntermediateOutputPath)protoc.outputs.txt">
+      <Output TaskParameter="Lines" ItemName="_PreviousProtoOutput" />
+    </ReadLinesFromFile>
+    <ItemGroup>
+      <_StaleProtoOutput Include="@(_PreviousProtoOutput)" Exclude="@(_ProtoOutput)" />
+      <!--
+        The SDK's default Compile glob captured these paths during item evaluation, with project-relative
+        Identities. We materialize the relative form into its own item so that <Compile Remove> can match it
+        - inlining the MakeRelative call inside an item transform pattern doesn't evaluate the property
+        function as expected.
+      -->
+      <_StaleProtoOutputRelative
+        Include="$([MSBuild]::MakeRelative($(MSBuildProjectDirectory), %(_StaleProtoOutput.Identity)))"
+        Condition="'%(_StaleProtoOutput.Identity)' != ''" />
+      <!--
+        Remove now, otherwise csc still reports the stale files as inputs and fails with CS2001 once we
+        delete them on disk. We try both the absolute and relative form to cover Compile entries added by
+        either explicit ProjectReference users or the SDK glob.
+      -->
+      <Compile Remove="@(_StaleProtoOutput)" />
+      <Compile Remove="@(_StaleProtoOutputRelative)" />
+    </ItemGroup>
+    <Delete Files="@(_StaleProtoOutput)" />
+  </Target>
+
+  <Target Name="ProtoCompile" BeforeTargets="CoreCompile" DependsOnTargets="_ComputeProtoOutputs" Condition="@(ProtoFile) != ''">
+    <ItemGroup>
+      <!-- Tools whose update must regenerate the code, in addition to the inputs recorded in the dependency files. -->
+      <_ProtocInput Include="$(IceRpcProtocPath)$(IceRpcProtocPrefix)/protoc$(_ProtocExecutableExtension)" />
+      <_ProtocInput Include="$(IceRpcProtocGenPath)IceRpc.Protobuf.Generator.dll" />
+    </ItemGroup>
+
     <MakeDir Directories="%(ProtoFile.OutputDir)" />
-    <UpToDateCheckTask OutputDir="%(ProtoFile.OutputDir)" Sources="@(ProtoFile)">
+    <UpToDateCheckTask
+      OutputDir="%(ProtoFile.OutputDir)"
+      Sources="@(ProtoFile)"
+      AdditionalInputs="@(_ProtocInput)"
+      Fingerprint="$(_ProtocFingerprint)"
+      FingerprintFile="$(IntermediateOutputPath)protoc.fingerprint.txt">
       <Output ItemName="_ProtoFile" TaskParameter="ComputedSources" />
     </UpToDateCheckTask>
 
@@ -121,19 +206,36 @@
       />
     </ItemGroup>
   </Target>
+
   <!--
-    _ProtoCleanFile is deliberately distinct from the _ProtoFile item that ProtoCompile emits: a task <Output> appends
-    to an existing item, and Rebuild runs Clean and Build in the same project instance, so sharing the item would leave
-    ProtoCompile with two entries per ProtoFile, run protoc twice, and pass duplicate Compile items to csc.
+    Record this build's outputs and configuration so the next build can prune anything no longer produced and
+    detect a configuration change. As an AfterTargets of ProtoCompile this only runs once protoc succeeded, so a
+    failed run never records a fingerprint its outputs don't match. The fingerprint is escaped because
+    WriteLinesToFile would otherwise split it on the semicolons.
   -->
-  <Target Name="ProtoClean" BeforeTargets="Clean" Condition="Exists('$(IceRpcProtobufToolsTaskAssembliesPath)IceRpc.Protobuf.Tools.dll')">
-    <OutputFileNamesTask Sources="@(ProtoFile)">
-      <Output ItemName="_ProtoCleanFile" TaskParameter="ComputedSources" />
-    </OutputFileNamesTask>
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).cs')" />
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).IceRpc.cs')" />
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).d')" />
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).BuildTelemetry.txt')" />
+  <Target Name="_ProtoWriteOutputManifest"
+          AfterTargets="ProtoCompile"
+          DependsOnTargets="_ComputeProtoOutputs"
+          Condition="@(ProtoFile) != ''">
+    <MakeDir Directories="$(IntermediateOutputPath)" />
+    <WriteLinesToFile
+      File="$(IntermediateOutputPath)protoc.outputs.txt"
+      Lines="@(_ProtoOutput)"
+      Overwrite="true"
+      WriteOnlyWhenDifferent="true" />
+    <WriteLinesToFile
+      File="$(IntermediateOutputPath)protoc.fingerprint.txt"
+      Lines="$([MSBuild]::Escape($(_ProtocFingerprint)))"
+      Overwrite="true"
+      WriteOnlyWhenDifferent="true" />
+  </Target>
+
+  <Target Name="ProtoClean" BeforeTargets="Clean" DependsOnTargets="_ComputeProtoOutputs">
+    <Delete Files="@(_ProtoOutput)" />
+    <Delete Files="$(IntermediateOutputPath)protoc.outputs.txt"
+            Condition="Exists('$(IntermediateOutputPath)protoc.outputs.txt')" />
+    <Delete Files="$(IntermediateOutputPath)protoc.fingerprint.txt"
+            Condition="Exists('$(IntermediateOutputPath)protoc.fingerprint.txt')" />
   </Target>
 
   <!-- Package ProtoFile items -->
--- a/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
+++ b/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
@@ -121,14 +121,19 @@
       />
     </ItemGroup>
   </Target>
+  <!--
+    _ProtoCleanFile is deliberately distinct from the _ProtoFile item that ProtoCompile emits: a task <Output> appends
+    to an existing item, and Rebuild runs Clean and Build in the same project instance, so sharing the item would leave
+    ProtoCompile with two entries per ProtoFile, run protoc twice, and pass duplicate Compile items to csc.
+  -->
   <Target Name="ProtoClean" BeforeTargets="Clean" Condition="Exists('$(IceRpcProtobufToolsTaskAssembliesPath)IceRpc.Protobuf.Tools.dll')">
     <OutputFileNamesTask Sources="@(ProtoFile)">
-      <Output ItemName="_ProtoFile" TaskParameter="ComputedSources" />
+      <Output ItemName="_ProtoCleanFile" TaskParameter="ComputedSources" />
     </OutputFileNamesTask>
-    <Delete Files="@(_ProtoFile->'%(OutputDir)/%(OutputFilename).cs')" />
-    <Delete Files="@(_ProtoFile->'%(OutputDir)/%(OutputFilename).IceRpc.cs')" />
-    <Delete Files="@(_ProtoFile->'%(OutputDir)/%(OutputFilename).d')" />
-    <Delete Files="@(_ProtoFile->'%(OutputDir)/%(OutputFilename).BuildTelemetry.txt')" />
+    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).cs')" />
+    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).IceRpc.cs')" />
+    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).d')" />
+    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).BuildTelemetry.txt')" />
   </Target>
 
   <!-- Package ProtoFile items -->
--- a/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
+++ b/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
@@ -130,10 +130,10 @@
     <OutputFileNamesTask Sources="@(ProtoFile)">
       <Output ItemName="_ProtoCleanFile" TaskParameter="ComputedSources" />
     </OutputFileNamesTask>
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).cs')" />
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).IceRpc.cs')" />
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).d')" />
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFilename).BuildTelemetry.txt')" />
+    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFileName).cs')" />
+    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFileName).IceRpc.cs')" />
+    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFileName).d')" />
+    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFileName).BuildTelemetry.txt')" />
   </Target>
 
   <!-- Package ProtoFile items -->
--- a/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
+++ b/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
@@ -115,11 +115,6 @@
     </ItemGroup>
     <Delete Files="@(_StaleProtoOutput)" />
     <!--
-      With no ProtoFile left, _ProtoWriteOutputManifest do't run to replace the manifest; delete it here, or
-      every later build deletes whatever appears at the former output paths.
-    -->
-    <Delete Files="$(IntermediateOutputPath)protoc.outputs.txt" Condition="@(ProtoFile) == ''" />
-    <!--
       With no ProtoFile left, _ProtoWriteOutputManifest doesn't run to replace the manifest; delete it here, or
       every later build deletes whatever appears at the former output paths.
     -->
--- a/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
+++ b/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
@@ -80,38 +80,22 @@
         Include="$([MSBuild]::NormalizePath('%(_ProtoNamedFile.OutputDir)/%(_ProtoNamedFile.OutputFileName).BuildTelemetry.txt'))"
         Condition="'$(IceRpcBuildTelemetry)' == 'true' And '$(IceRpcBuildTelemetryDebug)' == 'true'" />
     </ItemGroup>
-    <PropertyGroup>
-      <!-- Anything that changes the generated code without touching a Proto file; a change regenerates them all. -->
-      <_ProtocFingerprint>protoc=$(ProtocBundledPluginVersion);icerpc-csharp=$(ProtocIceRpcPluginVersion);search-path=@(ProtoSearchPath->'%(FullPath)', ',');options=@(ProtoFile->'%(Identity)=%(AdditionalOptions)', ',')</_ProtocFingerprint>
-    </PropertyGroup>
   </Target>
 
-  <!--
-    BeforeTargets includes CoreCompile and Clean for the case where ProtoFile is empty: ProtoCompile and ProtoClean
-    are then skipped, but the outputs of a previous build still need pruning.
-  -->
+  <!-- Runs even when ProtoFile is empty: the outputs of a previous build still need pruning. -->
   <Target Name="_ProtoPruneStaleOutputs"
-          BeforeTargets="ProtoCompile;CoreCompile;ProtoClean;Clean"
+          BeforeTargets="ProtoCompile;ProtoClean"
           DependsOnTargets="_ComputeProtoOutputs"
           Condition="Exists('$(IntermediateOutputPath)protoc.outputs.txt')">
     <ReadLinesFromFile File="$(IntermediateOutputPath)protoc.outputs.txt">
       <Output TaskParameter="Lines" ItemName="_PreviousProtoOutput" />
     </ReadLinesFromFile>
     <ItemGroup>
-      <_StaleProtoOutput Include="@(_PreviousProtoOutput)" Exclude="@(_ProtoOutput)" />
-      <!--
-        The SDK's default Compile glob records project-relative identities, and <Compile Remove> matches identities
-        exactly. A MakeRelative call inlined in an item transform doesn't evaluate, hence the separate item.
-      -->
-      <_StaleProtoOutputRelative
-        Include="$([MSBuild]::MakeRelative($(MSBuildProjectDirectory), %(_StaleProtoOutput.Identity)))"
-        Condition="'%(_StaleProtoOutput.Identity)' != ''" />
-      <!--
-        Remove before deleting, or csc fails with CS2001 on the missing files. The absolute form covers explicit
-        Compile items, the relative form the SDK glob.
-      -->
-      <Compile Remove="@(_StaleProtoOutput)" />
-      <Compile Remove="@(_StaleProtoOutputRelative)" />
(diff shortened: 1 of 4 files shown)


src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -60,9 +60,94 @@
...
+    </ItemGroup>
+
     <MakeDir Directories="%(ProtoFile.OutputDir)" />
-    <UpToDateCheckTask OutputDir="%(ProtoFile.OutputDir)" Sources="@(ProtoFile)">
+    <UpToDateCheckTask
+      OutputDir="%(ProtoFile.OutputDir)"
+      Sources="@(ProtoFile)"
+      AdditionalInputs="@(_ProtocInput)"
+      Fingerprint="$(_ProtocFingerprint)"
+      FingerprintFile="$(IntermediateOutputPath)protoc.fingerprint.txt">
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Not reproducible. `WriteLinesToFile` writes each item's `ItemSpec`, which MSBuild hands to the task unescaped, so `protoc.fingerprint.txt` holds the literal semicolon-separated fingerprint on one line. `$([MSBuild]::Escape)` only stops the value from being split into several `Lines` items when the parameter is bound. Verified on this branch with `IceRpc.Protobuf.Tests`: the recorded file reads `protoc=36.0;icerpc-csharp=0.7.0-preview.1;search-path=;options=...`, a second unchanged build runs protoc 0 times with no "configuration changed" message, and changing the fingerprint (`-p:ProtocIceRpcPluginVersion=x`) then reverting it regenerates all 5 files each time.

src/IceRpc.Protobuf.Tools/UpToDateCheckTask.cs
```diff
@@ -31,42 +46,41 @@ public class UpToDateCheckTask : Microsoft.Build.Utilities.Task
...
         {
-            bool upToDate = true;
             string fileName = source.GetMetadata("FileName").ToProtocPascalCase();
             string dependOutput = Path.Combine(OutputDir, $"{fileName}.d");
-            string csharpOutput = Path.Combine(OutputDir, $"{fileName}.cs");
-            string icerpcOutput = Path.Combine(OutputDir, $"{fileName}.IceRpc.cs");
+            string[] outputs =
+            [
+                dependOutput,
+                Path.Combine(OutputDir, $"{fileName}.cs"),
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Declining this one. `fileName` is the item's MSBuild `FileName` metadata (the file name without extension, converted to PascalCase), so it is never a rooted path and `Path.Combine` cannot drop `OutputDir` here. The three calls are unchanged from `main`; the alert fires because the lines moved.

src/IceRpc.Protobuf.Tools/UpToDateCheckTask.cs
```diff
@@ -31,42 +46,41 @@ public class UpToDateCheckTask : Microsoft.Build.Utilities.Task
...
+                Path.Combine(OutputDir, $"{fileName}.IceRpc.cs"),
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Declining this one. `fileName` is the item's MSBuild `FileName` metadata (the file name without extension, converted to PascalCase), so it is never a rooted path and `Path.Combine` cannot drop `OutputDir` here. The three calls are unchanged from `main`; the alert fires because the lines moved.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -60,9 +60,76 @@
...
+  </Target>
+
+  <!--
+    BeforeTargets includes CoreCompile and Clean for the case where ProtoFile is empty: ProtoCompile and ProtoClean
+    are then skipped, but the outputs of a previous build still need pruning.
+  -->
+  <Target Name="_ProtoPruneStaleOutputs"
+          BeforeTargets="ProtoCompile;CoreCompile;ProtoClean;Clean"
+          DependsOnTargets="_ComputeProtoOutputs"
+          Condition="Exists('$(IntermediateOutputPath)protoc.outputs.txt')">
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Reproduced: a Debug build, a removed or renamed Proto file, then a Release build compiles the stale output until Debug is built again. The per-configuration/TargetFramework manifest deliberately mirrors IceRpc.Slice.Tools (#4589), so parallel multi-target inner builds do not share one state file. Building several configurations against the same shared `OutputDir` is a niche setup this PR does not handle. The suppressed comment on line 116 (manifest retained after the last Proto file is removed) is fixed as of 84d0d0a: `_ProtoPruneStaleOutputs` now deletes the manifest when there are no ProtoFile items.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -60,12 +60,94 @@
...
+      <_ProtoOutput Include="$([MSBuild]::NormalizePath('%(_ProtoNamedFile.OutputDir)/%(_ProtoNamedFile.OutputFileName).cs'))" />
+      <_ProtoOutput Include="$([MSBuild]::NormalizePath('%(_ProtoNamedFile.OutputDir)/%(_ProtoNamedFile.OutputFileName).IceRpc.cs'))" />
+      <_ProtoOutput Include="$([MSBuild]::NormalizePath('%(_ProtoNamedFile.OutputDir)/%(_ProtoNamedFile.OutputFileName).d'))" />
+      <_ProtoOutput
+        Include="$([MSBuild]::NormalizePath('%(_ProtoNamedFile.OutputDir)/%(_ProtoNamedFile.OutputFileName).BuildTelemetry.txt'))"
+        Condition="'$(IceRpcBuildTelemetry)' == 'true' And '$(IceRpcBuildTelemetryDebug)' == 'true'" />
+    </ItemGroup>
+    <PropertyGroup>
+      <!-- Anything that changes the generated code without touching a Proto file; a change regenerates them all. -->
+      <_ProtocFingerprint>protoc=$(ProtocBundledPluginVersion);icerpc-csharp=$(ProtocIceRpcPluginVersion);search-path=@(ProtoSearchPath->'%(FullPath)', ',');options=@(ProtoFile->'%(Identity)=%(AdditionalOptions)', ',')</_ProtocFingerprint>
```
bernardnormier: `options=@(ProtoFile->'%(Identity)=%(AdditionalOptions)', ',')` changes whenever a Proto file is added, removed or renamed, so each of those regenerates every Proto file in the project. On main, adding a file ran protoc once. The README's "Incremental builds" section describes the per-file behavior, not this.

Only the option values need to be in the fingerprint, not the file set: for example the distinct `%(AdditionalOptions)` values, or a per-file record next to the `.d` compared in the task.

Verified at this head: adding a Proto file to IceRpc.Protobuf.Tests logs "The protoc configuration changed since the previous build; all Proto files are out of date".

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -60,12 +60,94 @@
...
+      <_ProtocInput Include="$(IceRpcProtocGenPath)IceRpc.Protobuf.Generator.dll" />
+    </ItemGroup>
+
     <MakeDir Directories="%(ProtoFile.OutputDir)" />
-    <UpToDateCheckTask OutputDir="%(ProtoFile.OutputDir)" Sources="@(ProtoFile)">
+    <UpToDateCheckTask
+      OutputDir="%(ProtoFile.OutputDir)"
+      Sources="@(ProtoFile)"
+      AdditionalInputs="@(_ProtocInput)"
+      Fingerprint="$(_ProtocFingerprint)"
```
bernardnormier: See the review body: this parameter, `FingerprintFile`, the `FingerprintChanged` output, the Delete below and the second WriteLinesToFile in `_ProtoWriteOutputManifest` all go away if the fingerprint is a cache file listed in `@(_ProtocInput)`.

bernardnormier (commented):
The manifest and prune half looks right: it mirrors the Slice tooling, and the Rebuild, failed-protoc and last-file-removed scenarios all check out. My comments are mostly on the fingerprint half, which does more than it needs to, plus a few small things.

The fingerprint machinery (three task parameters, an output property, the conditional Delete of the fingerprint file, the escaped second WriteLinesToFile, and the AfterTargets ordering that keeps them consistent) can be replaced by the mechanism the task already has. Write the fingerprint values to `$(IntermediateOutputPath)protoc.inputs.cache` with `WriteOnlyWhenDifferent="true"` before the up-to-date check, and add that file to `@(_ProtocInput)`. The file's mtime moves only when a value changes, every output is then older than an input, and everything regenerates. The partial-failure case falls out too: a rewritten cache stays newer than the outputs protoc didn't get to. This is what the SDK does for csc with `CoreCompileInputs.cache`. It removes the fingerprint code from `UpToDateCheckTask`, the Delete step and the Escape call.

Follow-up, not for this PR: `IceRpc.Slice.Tools.targets` has the bug 84d0d0a fixes here. When the last `.slice` file is removed, `_SlicecWriteOutputManifest` doesn't run, `slicec.outputs.txt` survives, and every later build deletes whatever appears at the former output paths. Worth an issue.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -60,12 +60,94 @@
...
+      -->
+      <_StaleProtoOutputRelative
+        Include="$([MSBuild]::MakeRelative($(MSBuildProjectDirectory), %(_StaleProtoOutput.Identity)))"
+        Condition="'%(_StaleProtoOutput.Identity)' != ''" />
+      <!--
+        Remove before deleting, or csc fails with CS2001 on the missing files. The absolute form covers explicit
+        Compile items, the relative form the SDK glob.
+      -->
+      <Compile Remove="@(_StaleProtoOutput)" />
+      <Compile Remove="@(_StaleProtoOutputRelative)" />
```
bernardnormier: `MatchOnMetadata` matches the SDK glob's relative identities directly, so the relative item, its comment and the MakeRelative call are not needed. The Condition covers the case below.

```suggestion
      <_StaleProtoOutput Include="@(_PreviousProtoOutput)" Exclude="@(_ProtoOutput)"
                         Condition="'@(_ProtoOutput)' != '' Or '@(ProtoFile)' == ''" />
      <!-- Remove before deleting, or csc fails with CS2001 on the missing files. -->
      <Compile Remove="@(_StaleProtoOutput)" MatchOnMetadata="FullPath" MatchOnMetadataOptions="PathLike" />
```

Verified with SDK 10.0: this removes both the glob's `generated/Stale.cs` and an absolute Compile entry. It also fixes a real, if unlikely, failure: the unquoted `MakeRelative(..., %(Identity))` splits on a comma in the path and fails with MSB4186. The Slice copy has the same shape.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -60,12 +60,94 @@
...
+  </Target>
+
+  <!--
+    BeforeTargets includes CoreCompile and Clean for the case where ProtoFile is empty: ProtoCompile and ProtoClean
+    are then skipped, but the outputs of a previous build still need pruning.
+  -->
+  <Target Name="_ProtoPruneStaleOutputs"
+          BeforeTargets="ProtoCompile;CoreCompile;ProtoClean;Clean"
+          DependsOnTargets="_ComputeProtoOutputs"
+          Condition="Exists('$(IntermediateOutputPath)protoc.outputs.txt')">
```
bernardnormier, line 94: When `_ComputeProtoOutputs` is skipped by its Exists condition while ProtoFile is not empty (source build, `--no-dependencies` after the Tools bin was cleaned), `_ProtoOutput` is empty, so every recorded output is treated as stale and deleted, and only then does ProtoCompile fail to load the task. The old ProtoCompile failed the same way but left the generated code alone. The Condition on `_StaleProtoOutput` in the suggestion below guards it.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -60,12 +60,94 @@
...
+      <_ProtocFingerprint>protoc=$(ProtocBundledPluginVersion);icerpc-csharp=$(ProtocIceRpcPluginVersion);search-path=@(ProtoSearchPath->'%(FullPath)', ',');options=@(ProtoFile->'%(Identity)=%(AdditionalOptions)', ',')</_ProtocFingerprint>
+    </PropertyGroup>
+  </Target>
+
+  <!--
+    BeforeTargets includes CoreCompile and Clean for the case where ProtoFile is empty: ProtoCompile and ProtoClean
+    are then skipped, but the outputs of a previous build still need pruning.
+  -->
+  <Target Name="_ProtoPruneStaleOutputs"
+          BeforeTargets="ProtoCompile;CoreCompile;ProtoClean;Clean"
```
bernardnormier: BeforeTargets run before their target even when that target's Condition is false, and ProtoClean has no Condition at all now, so `CoreCompile` and `Clean` in the list are redundant and the comment's reason doesn't hold.

```suggestion
  <!-- Runs even when ProtoFile is empty: the outputs of a previous build still need pruning. -->
  <Target Name="_ProtoPruneStaleOutputs"
          BeforeTargets="ProtoCompile;ProtoClean"
```

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -60,12 +60,94 @@
...
+    </ItemGroup>
+    <Delete Files="@(_StaleProtoOutput)" />
+    <!--
+      With no ProtoFile left, _ProtoWriteOutputManifest doesn't run to replace the manifest; delete it here, or
+      every later build deletes whatever appears at the former output paths.
+    -->
+    <Delete Files="$(IntermediateOutputPath)protoc.outputs.txt" Condition="@(ProtoFile) == ''" />
+  </Target>
+
+  <Target Name="ProtoCompile" BeforeTargets="CoreCompile" DependsOnTargets="_ComputeProtoOutputs" Condition="@(ProtoFile) != ''">
```
bernardnormier: 129 columns.

```suggestion
  <Target Name="ProtoCompile"
          BeforeTargets="CoreCompile"
          DependsOnTargets="_ComputeProtoOutputs"
          Condition="@(ProtoFile) != ''">
```

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -121,18 +203,35 @@
...
-    </OutputFileNamesTask>
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFileName).cs')" />
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFileName).IceRpc.cs')" />
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFileName).d')" />
-    <Delete Files="@(_ProtoCleanFile->'%(OutputDir)/%(OutputFileName).BuildTelemetry.txt')" />
+  <Target Name="_ProtoWriteOutputManifest"
+          AfterTargets="ProtoCompile"
+          DependsOnTargets="_ComputeProtoOutputs"
+          Condition="@(ProtoFile) != ''">
+    <MakeDir Directories="$(IntermediateOutputPath)" />
```
bernardnormier: WriteLinesToFile creates the directory, and PrepareForBuild already created it. Drop the line.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -121,18 +203,35 @@
...
+      WriteOnlyWhenDifferent="true" />
+  </Target>
+
+  <Target Name="ProtoClean" BeforeTargets="Clean" DependsOnTargets="_ComputeProtoOutputs">
+    <Delete Files="@(_ProtoOutput)" />
+    <Delete Files="$(IntermediateOutputPath)protoc.outputs.txt"
+            Condition="Exists('$(IntermediateOutputPath)protoc.outputs.txt')" />
+    <Delete Files="$(IntermediateOutputPath)protoc.fingerprint.txt"
+            Condition="Exists('$(IntermediateOutputPath)protoc.fingerprint.txt')" />
   </Target>
```
bernardnormier, line 232: Delete skips missing files (the first Delete already relies on that), so the Exists guards are redundant and one task does it. Same for the Exists half of the condition on the Delete in ProtoCompile, moot if the fingerprint becomes a cache file.

```suggestion
  <Target Name="ProtoClean" BeforeTargets="Clean" DependsOnTargets="_ComputeProtoOutputs">
    <Delete Files="@(_ProtoOutput)" />
    <Delete Files="$(IntermediateOutputPath)protoc.outputs.txt;$(IntermediateOutputPath)protoc.fingerprint.txt" />
  </Target>
```

src/IceRpc.Protobuf.Tools/UpToDateCheckTask.cs
```diff
@@ -77,37 +96,71 @@ public override bool Execute()
...
+        string? missingOutput = outputs.FirstOrDefault(output => !File.Exists(output));
+        if (missingOutput is not null)
         {
-            var depends = new List<string>();
-            string dependContents = File.ReadAllText(dependOutput);
+            Log.LogMessage(MessageImportance.Low, $"'{source}' is out of date: output '{missingOutput}' is missing.");
+            return false;
+        }
+
+        // Every output must be newer than every input, so compare against the oldest output.
```
bernardnormier: The second clause narrates the `Min` below it.

```suggestion
        // Every output must be newer than every input.
```

src/IceRpc.Protobuf.Tools/README.md
```diff
@@ -75,9 +75,15 @@ once per Proto file.
...
 
-> [!NOTE]
-> Changing `AdditionalOptions` does not mark previously generated code as out of date. Run `dotnet clean` and then
-> build again to regenerate the code with the new options.
+## Incremental builds
+
+`protoc` runs only for the Proto files whose generated code is missing or out of date. A Proto file is out of date
+when the Proto file itself, one of the files it imports, `protoc` or the `protoc-gen-icerpc-csharp` generator is newer
+than one of its generated files. Changing `AdditionalOptions` or `ProtoSearchPath`, or upgrading this package,
+regenerates the code of all Proto files.
```
bernardnormier: protoc doesn't run on its own; the build runs it.

```suggestion
The build runs `protoc` only for the Proto files whose generated code is missing or out of date. A Proto file is
out of date when the Proto file itself, one of the files it imports, `protoc` or the `protoc-gen-icerpc-csharp`
generator is newer than one of its generated files. Changing `AdditionalOptions` or `ProtoSearchPath`, or upgrading
this package, regenerates the code of all Proto files.
```

src/IceRpc.Protobuf.Tools/README.md
```diff
@@ -75,9 +75,15 @@ once per Proto file.
...
+
+The generated code of a Proto file that is removed from the project, renamed, or given a different `OutputDir` is
+deleted during the next build.
```
bernardnormier: Three stacked qualifiers on a passive sentence; active voice reads better.

```suggestion
When you remove a Proto file from the project, rename it, or change its `OutputDir`, the next build deletes the code
previously generated for it.
```

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
pepone: 🤖 **Claude:** Done in e32378740. The cache records the distinct `AdditionalOptions` values (`@(ProtoFile->'options=%(AdditionalOptions)'->DistinctWithCase())`; with-case because `Distinct()` folds case and protoc options don't). Verified on the branch: adding a Proto file runs protoc once, removing it runs nothing and prunes its outputs. One limitation of the distinct set, for the record: a file switching to an option value another file already uses leaves the set unchanged and isn't regenerated. A per-file record next to the `.d` would close that at the cost of a manifest entry, clean/prune participation and a compare in the task; say if you'd rather have that.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
pepone: 🤖 **Claude:** Done in e32378740: `Fingerprint`, `FingerprintFile`, `FingerprintChanged`, the Delete and the second WriteLinesToFile are gone. `protoc.inputs.cache` is written in ProtoCompile before the check and listed in `@(_ProtocInput)`.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
pepone: 🤖 **Claude:** Applied in e32378740. Verified with SDK 10.0.201: renaming `hello.proto` to `greeter.proto` deletes `Hello.*`, removes the glob's Compile entry and the build succeeds.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
pepone: 🤖 **Claude:** Applied in e32378740. Verified: with the task assembly path pointing at a missing directory and `--no-dependencies`, `_ProtoPruneStaleOutputs` runs, deletes nothing, and the build fails on the task load with the generated code intact. ProtoClean's manifest Delete got the same condition: in a source build `dotnet clean` cleans the Tools project first, and deleting the manifest there would leave a removed Proto file's outputs unpruned by the next build.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
pepone: 🤖 **Claude:** Applied in e32378740. Verified: with the last Proto file removed, the target still prunes the outputs and deletes the manifest.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
pepone: 🤖 **Claude:** Done in e32378740.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
pepone: 🤖 **Claude:** Done in e32378740.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
pepone: 🤖 **Claude:** Done in e32378740; the Delete now covers `protoc.inputs.cache` instead of the fingerprint.

src/IceRpc.Protobuf.Tools/UpToDateCheckTask.cs
pepone: 🤖 **Claude:** Done in e32378740.

src/IceRpc.Protobuf.Tools/README.md
pepone: 🤖 **Claude:** Done in e32378740.

src/IceRpc.Protobuf.Tools/README.md
pepone: 🤖 **Claude:** Done in e32378740.

pepone:
🤖 **Claude:** Done in e32378740: the fingerprint machinery is replaced by `protoc.inputs.cache`, written with `WriteOnlyWhenDifferent` before the up-to-date check and listed in `@(_ProtocInput)`. Verified with a two-file probe against the branch (protoc runs per build): unchanged 0; options changed / unchanged / reverted 2 / 0 / 2; search path changed 2; Proto file added 1; a cache change with protoc failing on one file, then that file fixed, regenerates only that file; Rebuild twice 2 / 2; Clean deletes outputs, manifest and cache. `IceRpc.Protobuf.Tests` passes and a second build runs protoc 0 times. The Slice follow-up is #4930.

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
```diff
@@ -60,9 +63,87 @@
...
+      <_StaleProtoOutput Include="@(_PreviousProtoOutput)" Exclude="@(_ProtoOutput)"
+                         Condition="'@(_ProtoOutput)' != '' Or '@(ProtoFile)' == ''" />
+      <!-- Remove before deleting, or csc fails with CS2001 on the missing files. -->
+      <Compile Remove="@(_StaleProtoOutput)" MatchOnMetadata="FullPath" MatchOnMetadataOptions="PathLike" />
+    </ItemGroup>
+    <Delete Files="@(_StaleProtoOutput)" />
+    <!--
+      With no ProtoFile left, _ProtoWriteOutputManifest doesn't run to replace the manifest; delete it here, or
+      every later build deletes whatever appears at the former output paths.
+    -->
```
bernardnormier: The second clause describes the bug; the first is the why.

```suggestion
    <!-- With no ProtoFile left, _ProtoWriteOutputManifest doesn't run to replace the manifest. -->
```

bernardnormier (commented):
All points from the previous round are addressed in e32378740: the fingerprint machinery is replaced by
`protoc.inputs.cache` listed in `@(_ProtocInput)`, the `_StaleProtoOutput` condition guards the skipped
`_ComputeProtoOutputs` case, `MatchOnMetadata` replaces the relative item, the redundant BeforeTargets, MakeDir and
Exists guards are gone, and the README wording is updated.

One request on the options thread, and two comments that describe the behavior this PR replaces; that belongs
in the PR description.

src/IceRpc.Protobuf.Tools/UpToDateCheckTask.cs
```diff
@@ -78,36 +69,69 @@ public override bool Execute()
...
-            // Strip everything before and including "Xxx.cs:" (the output target).
-            const string outputPrefix = ".cs:";
-            int i = dependContents.IndexOf(outputPrefix, StringComparison.CurrentCultureIgnoreCase);
-            if (i == -1 || i + outputPrefix.Length >= dependContents.Length)
+        foreach (string input in ProcessDependencies(dependOutput).Concat(additionalInputs))
+        {
+            if (!File.Exists(input))
             {
-                return depends;
+                // File.GetLastWriteTime returns a placeholder date for a missing file, older than any output.
```
bernardnormier: A missing input is out of date; the log line says so. The placeholder-date behavior is in the PR description.

```suggestion
```

src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.targets
bernardnormier: Yes to the per-file record. protoc runs once per Proto file, so the options are a per-file input, and the build
already keeps per-file state for every Proto file: its dependency file. Record the options next to it and the check
becomes exact instead of a set approximation. The cache file stays for the inputs that are global: the versions and
the search path.

It is smaller than it sounds:

- `_ComputeProtoOutputs` gets one more `_ProtoOutput` Include, `%(OutputFileName).options` (or a name you prefer).
  That one line covers the manifest, prune and clean.
- `ProtoCompile` gets a `WriteLinesToFile` after `ProtocTask`, batched on `%(_ProtoFile...)` with the same
  `UpToDate != 'true'` condition, `Lines="%(_ProtoFile.AdditionalOptions)"`. Running after protoc, a failed run
  leaves no fresh record.
- `UpToDateCheckTask` adds the record to `outputs` and compares its lines with the `AdditionalOptions` metadata it
  already gets on each source. Missing or different means out of date. No new parameter.

The `options=` line, the `DistinctWithCase` transform and its comment then leave the cache.

pepone: 🤖 **Claude:** Done in 033a19c88, with one change to the mechanism. The record is `<OutputDir>/<Name>.options`, one option per line, listed by `_ComputeProtoOutputs` so the manifest, prune and clean cover it, and the `options=` line and `DistinctWithCase` left the cache. But `ProtoCompile` writes it *before* the check with `WriteOnlyWhenDifferent`, and the task lists it among the file's inputs, the same rule as `protoc.inputs.cache` applied per file. Two reasons, both reproduced with the after-protoc version first:

- A record written after protoc isn't refreshed for a file whose sibling failed. Change A's options and break B: A regenerates, B fails, no record is written. Revert A and fix B: A's old record matches again and its outputs are newer than its inputs, so A keeps the code generated with the reverted options.
- Comparing the record with the metadata in the task means re-splitting the raw string, which diverges from MSBuild's own split for `%3B` and, on Unix, for a backslash in an option value. With the record as an input there is nothing to compare.

Verified with a two-file probe: one file's options changed / unchanged / reverted regenerates that file only (1 / 0 / 1); a file switching to the value the other uses, 1; a sibling failure with an options change regenerates only the failed file once fixed (1), and with the change reverted, the reverted file as well (2); third file added 1, removed 0 with `Third.options` pruned; task assembly missing, task-load error with the generated code intact; Clean removes the records. `IceRpc.Protobuf.Tests`: first build 5 (no records yet), second 0, 70 tests pass.

src/IceRpc.Protobuf.Tools/UpToDateCheckTask.cs
```diff
@@ -31,42 +36,28 @@ public class UpToDateCheckTask : Microsoft.Build.Utilities.Task
...
-            bool upToDate = true;
             string fileName = source.GetMetadata("FileName").ToProtocPascalCase();
             string dependOutput = Path.Combine(OutputDir, $"{fileName}.d");
-            string csharpOutput = Path.Combine(OutputDir, $"{fileName}.cs");
-            string icerpcOutput = Path.Combine(OutputDir, $"{fileName}.IceRpc.cs");
+            string[] outputs =
+            [
+                dependOutput,
+                Path.Combine(OutputDir, $"{fileName}.cs"),
+                Path.Combine(OutputDir, $"{fileName}.IceRpc.cs"),
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Declining, as for the earlier alert on the same lines: `fileName` is the item's `FileName` metadata converted to PascalCase, never a rooted path, so `Path.Combine` cannot drop `OutputDir` here. The new `.options` call has the same shape.

bernardnormier (approved):

PR #4926, opened 2026-09-08
Generate variant field and return value doc comments

### LINKED ISSUE #4832: [Audit-Low] Core, transport and codec edge cases (12 findings)

bernardnormier:
Closes #4832 by fixing the two remaining findings, the `@param` half of L-11 and L-12. Both depend on the doc comments
that slicec 0.5 now forwards for variant fields and return fields (icerpc/slicec#809).

- `VariantEnumGenerator` emits a `<param>` tag on the variant's record for each field documented with `@param` on the
  variant.
- `ProxyGenerator` and `ServiceGenerator` emit a `<returns>` tag from the operation's `@returns` tags: the description
  of a single return value, or a bulleted list with one item per documented element of a tuple return. Operations with
  `cs::encodedReturn` keep no `<returns>` tag on the service side, since their return value is a `PipeReader` the Slice
  doc comment doesn't describe.

The new tests read the generated members from the test assembly's XML documentation file.

## What's Changed entry

Area: Slice codec

- The Slice compiler now generates a `<param>` doc comment for each field of a variant enum documented with `@param`.
- The Slice compiler now generates a `<returns>` doc comment for operations documented with `@returns`. For an
  operation that returns a tuple, the comment lists each documented return value.

### DIFF
--- a/src/IceRpc.Slice.Generator/ServiceGenerator.cs
+++ b/src/IceRpc.Slice.Generator/ServiceGenerator.cs
@@ -310,9 +310,14 @@ internal static class ServiceGenerator
             op.CancellationTokenParamName,
             docComment: "A cancellation token that receives the cancellation requests.");
 
-        if (op.NonStreamedReturns.Count == 0 && !op.HasStreamedReturn)
+        if (op.ReturnType.Count == 0)
         {
             operationBuilder.AddComment("returns", "A value task that completes when this implementation completes.");
+        }
+        else if (!op.Attributes.HasAttribute(CSAttributes.CSEncodedReturn) &&
+            op.GetReturnsDocComment(currentNamespace) is string returns)
+        {
+            operationBuilder.AddComment("returns", returns);
         }
 
         // Build SliceOperation attribute with optional named parameters
--- a/src/ZeroC.Slice.Generator/BuilderExtensions.cs
+++ b/src/ZeroC.Slice.Generator/BuilderExtensions.cs
@@ -21,6 +21,20 @@ internal static class BuilderExtensions
         return builder;
     }
 
+    /// <summary>Adds a <c>&lt;param&gt;</c> tag for each field that has a doc comment.</summary>
+    internal static T AddDocCommentParams<T>(this T builder, IEnumerable<Field> fields, string currentNamespace)
+        where T : ICommentBuilder<T>
+    {
+        foreach (Field field in fields)
+        {
+            if (DocCommentFormatter.FormatOverview(field.Comment, currentNamespace) is string overview)
+            {
+                builder.AddComment("param", "name", field.Name, overview);
+            }
+        }
+        return builder;
+    }
+
     /// <summary>Adds the <c>&lt;seealso&gt;</c> tags from a <see cref="Comment"/> to a builder.</summary>
     internal static T AddDocCommentSeeAlso<T>(this T builder, Comment? comment, string currentNamespace)
         where T : ICommentBuilder<T>
--- a/src/ZeroC.Slice.Generator/BuilderExtensions.cs
+++ b/src/ZeroC.Slice.Generator/BuilderExtensions.cs
@@ -21,20 +21,6 @@ internal static class BuilderExtensions
         return builder;
     }
 
-    /// <summary>Adds a <c>&lt;param&gt;</c> tag for each field that has a doc comment.</summary>
-    internal static T AddDocCommentParams<T>(this T builder, IEnumerable<Field> fields, string currentNamespace)
-        where T : ICommentBuilder<T>
-    {
-        foreach (Field field in fields)
-        {
-            if (DocCommentFormatter.FormatOverview(field.Comment, currentNamespace) is string overview)
-            {
-                builder.AddComment("param", "name", field.Name, overview);
-            }
-        }
-        return builder;
-    }
-
     /// <summary>Adds the <c>&lt;seealso&gt;</c> tags from a <see cref="Comment"/> to a builder.</summary>
     internal static T AddDocCommentSeeAlso<T>(this T builder, Comment? comment, string currentNamespace)
         where T : ICommentBuilder<T>
--- a/src/IceRpc.Slice.Generator/OperationExtensions.cs
+++ b/src/IceRpc.Slice.Generator/OperationExtensions.cs
@@ -144,6 +144,27 @@ internal static class OperationExtensions
             return fn.Build();
         }
 
+        /// <summary>Returns the <c>&lt;returns&gt;</c> doc comment for the operation: the doc comment of its single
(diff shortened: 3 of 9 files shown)


src/ZeroC.Slice.Generator/BuilderExtensions.cs
```diff
@@ -21,6 +21,20 @@ internal static T AddDocCommentSummary<T>(this T builder, Comment? comment, stri
         return builder;
     }
 
+    /// <summary>Adds a <c>&lt;param&gt;</c> tag for each field that has a doc comment.</summary>
```
InsertCreativityHere: Tangentially related: we already have an `Addfield` method on `ContainedBuilder`... but it appears to be dead code? If it's not useful for this method to call, we should delete this dead method.

src/IceRpc.Slice.Generator/ServiceGenerator.cs
```diff
@@ -310,10 +310,15 @@ private static CodeBlock BuildServiceOperationDeclaration(Operation op, string c
...
-        if (op.NonStreamedReturns.Count == 0 && !op.HasStreamedReturn)
+        if (op.ReturnType.Count == 0)
         {
             operationBuilder.AddComment("returns", "A value task that completes when this implementation completes.");
         }
+        else if (!op.Attributes.HasAttribute(CSAttributes.CSEncodedReturn) &&
+            op.GetReturnsDocComment(currentNamespace) is string returns)
+        {
+            operationBuilder.AddComment("returns", returns);
+        }
```
InsertCreativityHere, line 321: Can be in a separate PR, but it would be good to still generate an `<returns>` even for `CSEncodedReturn`. Even if we just say something about a value task that completes with a pipe-reader holding the encoded...".

Note that before this PR, we would still generate one. It's only after this PR where we don't anymore.

InsertCreativityHere (approved):
Didn't study the test too closely, but everything looks good to me!

pepone (approved):
Approving. Two non-blocking observations, candidates for a follow-up rather than changes to this PR:

- A variant field renamed with `cs::identifier("@Other")` and documented with `@param` yields `<param name="@Other">`, which fails with CS1572: the XML name has to be `Other`, without the `@`. Operation parameters already behave the same way on main, so this is not a regression; stripping a leading `@` when emitting the `<param name>` value would fix both sites.
- For a `cs::encodedReturn` operation that also has a streamed return, the service-side tuple keeps the stream element under its Slice name and type, so its `@returns` text still applies there. The service declaration could emit it alongside a `Payload` item instead of dropping the whole `<returns>`.

src/IceRpc.Slice.Generator/ServiceGenerator.cs
bernardnormier: Before this PR there was no `<returns>` on encodedReturn service operations either: main only emitted one for operations without a return value, which `cs::encodedReturn` rejects. Agreed it should get one. Filed #4927 to emit a fixed description of the payload element, and the `@returns` text of the stream element when the operation also streams.

bernardnormier:
@pepone thanks for the two observations. Both are now tracked as follow-ups: #4928 for the `@` in the `param` name of
a verbatim `cs::identifier`, and #4927 for the `<returns>` of `cs::encodedReturn` service operations, including the
stream element of the tuple.

src/ZeroC.Slice.Generator/BuilderExtensions.cs
bernardnormier: Good catch. `AddField` was the primary-constructor-parameter adder the variant records should have used all along. It's now `AddPrimaryConstructorParameter` and `VariantEnumGenerator` goes through it, which also removed the separate `<param>` helper.

InsertCreativityHere (approved):
Looks good to me.

pepone (approved):

PR #4931, opened 2026-09-10
Extract helpers for optional encoding in the Slice generators

### LINKED ISSUE #4870: Extract a helper for the optional encode value expression in the Slice generators
The choice between `(value ?? default!)`, `value!.Value`, and `value!` when encoding an optional is written out at five separate encode sites, so a new encode site has to reproduce it by hand.

- `src/ZeroC.Slice.Generator/ITypeExtensions.cs:222-224` — sequence of optionals; also hand-rolls the value-type test that `TypeRef.IsValueType` already provides
- `src/ZeroC.Slice.Generator/ITypeExtensions.cs:243-245` — `Result` success and failure types
- `src/ZeroC.Slice.Generator/TypeRefExtensions.cs:48-50` — dictionary values and the general `TypeRef` lambda
- `src/ZeroC.Slice.Generator/FieldExtensions.cs:212-214` — optional fields and parameters
- `src/IceRpc.Slice.Generator/OperationExtensions.cs:356-358` — stream of optionals

A single helper on `TypeRef` returning the value expression would collapse all five.

bernardnormier:
Fixes #4870.

The choice between `(value ?? default!)`, `value!.Value`, and `value!` when encoding an optional was written out at
five encode sites. This PR adds `TypeRef.UnwrapNonNullOptional(param)` and uses it everywhere.

Two further collapses fell out of it:

- The sequence-of-optionals site had its own lambda builder that, once it used `TypeRef.IsValueType` instead of a
  hand-rolled value-type test, was identical to `GetEncodeLambda`. `EncodeSequence` now mirrors `EncodeDictionary`:
  one lambda call plus a method-name switch.
- The `Result` success/failure lambda and the stream-of-optionals lambda were the same bool-prefixed lambda in two
  projects. Both now use the new `TypeRef.GetEncodeLambdaWithNullMarker`.

Generated code is unchanged except that the optional field/parameter site now emits `x!.Value` and `x!` inside its
null check, like the other sites, and multi-line encode bodies in a stream-of-optionals lambda are now indented.

## What's Changed

None — generator internals only.

### DIFF
--- a/src/IceRpc.Slice.Generator/OperationExtensions.cs
+++ b/src/IceRpc.Slice.Generator/OperationExtensions.cs
@@ -125,7 +125,7 @@ internal static class OperationExtensions
                 bool useSegments;
                 if (streamField.DataTypeIsOptional)
                 {
-                    encodeLambda = GetStreamOfOptionalEncodeLambda(streamField, currentNamespace);
+                    encodeLambda = streamField.DataType.GetEncodeLambdaWithNullMarker(currentNamespace);
                     useSegments = true;
                 }
                 else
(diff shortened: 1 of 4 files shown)


externl (approved):
Looks good to me!

pepone (approved):

PR #4932, opened 2026-09-10
Generate a returns doc comment for cs::encodedReturn service operations

### LINKED ISSUE #4927: Generate a returns doc comment for cs::encodedReturn service operations
The service-side declaration of a `cs::encodedReturn` operation has no `<returns>` doc comment. Its return value is a
`PipeReader` holding the encoded return value, or a tuple of that reader and the streamed return, so the Slice
`@returns` tags don't describe it directly.

The generator should emit a fixed description for the payload element, and for an operation with a streamed return, a
tuple list with that description plus the stream element's `@returns` text, since the stream keeps its Slice name and
type in the tuple.

Follow-up to #4926.

bernardnormier:
Fixes #4927.

The service-side declaration of a `cs::encodedReturn` operation now gets a `<returns>` doc comment. Its return value
is a `PipeReader` holding the encoded return value, or a tuple of that reader and the streamed return, so the Slice
`@returns` tags don't describe it directly:

- Without a streamed return, the comment is the fixed text "The encoded return value."
- With a streamed return, the comment is a tuple list with the `Payload` element and that fixed text, followed by the
  stream element with its `@returns` text, or "The streamed return value." when it is not documented.

The "A tuple containing" list formatting is now shared with the existing `<returns>` comment for tuple returns.

Two documented `cs::encodedReturn` operations were added to `DocumentationTests.slice`, and `DocumentationTests.cs`
checks their generated comments in the XML documentation file. Generated output for the streamed case:

```csharp
/// <returns>
/// A tuple containing:
/// <list type="bullet">
/// <item><term>Payload</term><description>The encoded return value.</description></item>
/// <item><term>Entries</term><description>The log entries.</description></item>
/// </list>
/// </returns>
[SliceOperation("streamLog", EncodedReturn = true)]
global::System.Threading.Tasks.ValueTask<(global::System.IO.Pipelines.PipeReader Payload, global::System.Collections.Generic.IAsyncEnumerable<string> Entries)> StreamLogAsync(
```

Follow-up to #4926.

## What's Changed

Area: Slice codec

- The Slice compiler now generates a `<returns>` doc comment for the service-side declaration of operations with the
  `cs::encodedReturn` attribute.

### DIFF
--- a/src/IceRpc.Slice.Generator/OperationExtensions.cs
+++ b/src/IceRpc.Slice.Generator/OperationExtensions.cs
@@ -172,14 +172,16 @@ internal static class OperationExtensions
 
             if (op.StreamedReturn is Field streamReturn)
             {
-                var items = new List<(string, string)> { (op.EncodedReturnPayloadName, payloadDescription) };
-                if (DocCommentFormatter.FormatOverview(streamReturn.Comment, currentNamespace) is string overview)
-                {
-                    items.Add((streamReturn.Name, overview));
-                }
-                return TupleReturnsDocComment(items);
+                string streamDescription =
+                    DocCommentFormatter.FormatOverview(streamReturn.Comment, currentNamespace) ??
+                    "The streamed return value.";
+                return TupleReturnsDocComment(
+                    [(op.EncodedReturnPayloadName, payloadDescription), (streamReturn.Name, streamDescription)]);
             }
-            return payloadDescription;
+            else
+            {
+                return payloadDescription;
+            }
         }
 
         private static string TupleReturnsDocComment(IEnumerable<(string Name, string Description)> items)
--- a/src/IceRpc.Slice.Generator/OperationExtensions.cs
+++ b/src/IceRpc.Slice.Generator/OperationExtensions.cs
@@ -157,12 +157,36 @@ internal static class OperationExtensions
             var items = op.ReturnType
                 .Select(r => (r.Name, Overview: DocCommentFormatter.FormatOverview(r.Comment, currentNamespace)))
                 .Where(item => item.Overview is not null)
-                .Select(item => $"<item><term>{item.Name}</term><description>{item.Overview}</description></item>")
+                .Select(item => (item.Name, item.Overview!))
                 .ToList();
 
-            return items.Count > 0
-                ? $"A tuple containing:\n<list type=\"bullet\">\n{string.Join("\n", items)}\n</list>"
-                : null;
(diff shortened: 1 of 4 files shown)


src/IceRpc.Slice.Generator/OperationExtensions.cs
```diff
@@ -157,12 +157,36 @@ internal CodeBlock BuildEncodeStreamMethod(
...
+            const string payloadDescription = "The encoded return value.";
+
+            if (op.StreamedReturn is Field streamReturn)
+            {
+                var items = new List<(string, string)> { (op.EncodedReturnPayloadName, payloadDescription) };
+                if (DocCommentFormatter.FormatOverview(streamReturn.Comment, currentNamespace) is string overview)
+                {
+                    items.Add((streamReturn.Name, overview));
+                }
+                return TupleReturnsDocComment(items);
```
InsertCreativityHere, line 180: In the case where the user hasn't documented their `stream` parameter, this will generate a doc-comment describing a tuple of one element, which is better than the current state, but feels incomplete.

Maybe we could add a fallback for when `FormatOverview` returns null, so we always generate 2 items in the list?

InsertCreativityHere (approved):
Looks good, but I also agree, the testing here seems lackluster. Just making sure the doc-comment doesn't emit any warnings could let many kinds of failures through silently.

pepone (approved):

src/IceRpc.Slice.Generator/OperationExtensions.cs
bernardnormier: Done in 678b20f: the stream item is always listed, with "The streamed return value." when it has no `@returns`.

tests/IceRpc.Slice.Generator.Tests/DocumentationTests.slice
```diff
@@ -52,6 +52,17 @@ interface MySessionManager {
...
+    /// Retrieves the log of the given user.
+    /// @param user: The user.
+    /// @returns: The log entries.
+    [cs::encodedReturn] getLog(user: string) -> Sequence<string>
+
+    /// Streams the log of the given user.
+    /// @param user: The user.
+    /// @returns count: The number of entries.
+    /// @returns entries: The log entries.
+    [cs::encodedReturn] streamLog(user: string) -> (count: int32, entries: stream string)
```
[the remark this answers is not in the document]
bernardnormier: Added XML documentation tests for both operations in 678b20f.

InsertCreativityHere (approved):
Looks good to me!

PR #4934, opened 2026-09-10
Fix the generated code for a verbatim cs::identifier on a parameter

### LINKED ISSUE #4928: A verbatim cs::identifier on a documented parameter produces an invalid param doc comment
A parameter or variant field mapped with `cs::identifier("@event")` and documented with `@param` generates
`<param name="@event">`, which the C# compiler rejects with CS1572: the XML name of a verbatim identifier is `event`,
without the `@`. This affects operation parameters on main and, since #4926, variant fields.

`FunctionBuilder.AddParameter` and `ContainerBuilder.AddPrimaryConstructorParameter` should strip a leading `@` from
the name they put in the `param` tag.

bernardnormier:
Fixes #4928.

A parameter or variant field mapped with `cs::identifier("@event")` and documented with `@param` generated
`<param name="@event">`, which the C# compiler rejects with CS1572. `FunctionBuilder.AddParameter` and
`ContainerBuilder.AddPrimaryConstructorParameter` now strip the leading `@` from the name in the `param` tag.

The test definitions exposed a second bug with the same trigger: the generators name the local variables that hold
decoded parameters and return values by prefixing the parameter name, which turned `@event` into `sliceP_@event`, a
syntax error. An operation with a verbatim parameter identifier did not compile at all. A new `DecodedVariableName`
property on `Entity` builds that name without the `@`, and the nine sites in the proxy, service, and operation
generators use it.

The test Slice files gain an operation with a documented `@event` parameter and a variant enum with a documented
`@event` field. Warnings are errors in the test builds, so these definitions fail to compile without the fix. Generated
output with it:

```csharp
/// <param name="event">The event name.</param>
...
var sliceP_event = decoder.DecodeString();
return (sliceP_user, sliceP_event);

/// <param name="event">The event name.</param>
partial record class Created(string @event) : SubscriptionChange
```

## What's Changed

Area: Slice codec

- Fixed the code generated for an operation parameter or variant field mapped to a verbatim C# identifier with
  `cs::identifier`. The generated doc comment and decoding code did not compile.

### DIFF
--- a/src/ZeroC.Slice.Generator/EntityExtensions.cs
+++ b/src/ZeroC.Slice.Generator/EntityExtensions.cs
@@ -32,10 +32,6 @@ internal static class EntityExtensions
         internal string ParameterName => entity.Attributes.FindAttribute(CSAttributes.CSIdentifier) is Attribute attr ?
             attr.Args[0] : entity.Identifier.ToCamelCase();
 
-        /// <summary>Gets the name of the local variable that holds the decoded value of this parameter or return
-        /// field.</summary>
-        internal string DecodedVariableName => $"sliceP_{entity.ParameterName.TrimStart('@')}";
-
         /// <summary>Gets the name of the encoder or decoder extensions class. The returned name is fully qualified when
         /// the entity is in a different namespace.</summary>
         internal string ExtensionsClass(string currentNamespace, bool decoder)
--- a/src/ZeroC.Slice.Generator/EntityExtensions.cs
+++ b/src/ZeroC.Slice.Generator/EntityExtensions.cs
@@ -32,6 +32,10 @@ internal static class EntityExtensions
         internal string ParameterName => entity.Attributes.FindAttribute(CSAttributes.CSIdentifier) is Attribute attr ?
             attr.Args[0] : entity.Identifier.ToCamelCase();
 
+        /// <summary>Gets the name of the local variable that holds the decoded value of this parameter or return
+        /// field.</summary>
+        internal string DecodedVariableName => $"sliceP_{entity.ParameterName.TrimStart('@')}";
+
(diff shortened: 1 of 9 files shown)


src/ZeroC.Slice.Generator/EntityExtensions.cs
```diff
@@ -32,6 +32,10 @@ internal static class EntityExtensions
         internal string ParameterName => entity.Attributes.FindAttribute(CSAttributes.CSIdentifier) is Attribute attr ?
             attr.Args[0] : entity.Identifier.ToCamelCase();
 
+        /// <summary>Gets the name of the local variable that holds the decoded value of this parameter or return
+        /// field.</summary>
+        internal string DecodedVariableName => $"sliceP_{entity.ParameterName.TrimStart('@')}";
```
InsertCreativityHere, line 37: We have a `FieldExtensions.cs` file where I think this would be better placed.
We only call this method on fields, and calling it on anything else would be a bug I believe.

InsertCreativityHere (approved):
Looks good to me!

My Claude also found this minor issue. Not sure if we care about:
- The `@` is still kept when making return-type lists. This compiles since it's not a real value passed to a tag, but might be desirable to remove them anyways? For example, `GetReturnsDocComment` would output `<term>@event</term>` for an identifier `@event`.

pepone (approved):

src/ZeroC.Slice.Generator/EntityExtensions.cs
bernardnormier: Moved to FieldExtensions in 91afe32.

bernardnormier:
Also stripped the `@` from the `<term>` in the returns doc comment in 91afe32, since the tuple element is really named `event`.

InsertCreativityHere (approved):
Looks good to me!

PR #4940, opened 2026-09-11
Extract toolchain archives into a staging folder and gate on the executable

### LINKED ISSUE #4800: [Audit-Low] Interrupted toolchain extraction leaves a partial directory that is never re-extracted

> ⚠️ **Filed for review despite being borderline.** Codex confirmed the mechanism but judged it "technically real, but below the filing bar absent an actual incident". Close as Not planned if you agree — recorded here so the reasoning is on file rather than rediscovered.

## Summary

`ExtractTask` calls `ZipFile.ExtractToDirectory(src, dest)` straight into the final location — no temp-then-rename, no try/finally, no cleanup — while the MSBuild guard is `Condition="!Exists(<dest dir>)"`. `ExtractToDirectory` creates the directory before extracting entries, so a partial extraction is permanently skipped thereafter.

Nothing heals it: `SlicecClean`/`ProtoClean` clean consumer-generated files, not `obj/slicec-*`, so `dotnet clean` does not help. The following unconditional `<Copy SourceFiles=".../%(SlicecPlatform.Exe)">` then fails MSB3030 or copies a truncated binary, with no hint that the fix is `rm -rf src/IceRpc.Slice.Tools/obj`.

## Trigger / impact

A single Ctrl-C usually will **not** do it — MSBuild requests graceful cancellation and `ExtractTask` is not `ICancelableTask`, so a ~2 MB extraction completes. Realistic triggers are a hard kill/SIGKILL/OOM, a disk-full `IOException`, or power loss.

Codex refined the symptom analysis: the protoc archive is multi-entry, but `bin/protoc` is its *first* file payload, so that does not materially widen the missing-executable window. **A truncated executable silently copied to `obj/tools` is the likelier symptom in both cases**, producing a cryptic exec failure rather than a clean MSB3030.

"Permanently" means until the directory is manually removed. Limited to local development trees; CI uses fresh checkouts, and shipped runtime behavior is unaffected.

Worth noting that MSBuild's own `DownloadFile` task already deletes partial downloads, so this is a gap rather than an accepted repo-wide pattern.

## Source references

- [build/ExtractTask.cs#L32](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/build/ExtractTask.cs#L32) — extraction straight into the final directory
- [src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.csproj#L110](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.csproj#L110) — the `!Exists(<dir>)` guard
- [src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj#L94-L97](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj#L94) — same shape

## Recommendation

Extract to a sibling temp directory and `Move` on success (or delete the destination in a `catch` before rethrowing), and gate the `Condition` on the expected executable rather than on the directory.

## Related issues

- #4450 and #4471 (both **Not planned**) concern *integrity hashes* on toolchain downloads — AGENTS.md dismissal #2. This finding is about partial-extraction cleanup, not provenance, so the dismissal does not directly apply. Flagging it since it is the same code area.

## Audit context

Finding 18 of [general-correctness-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/general-correctness-audit-2026-07-27.md); Codex verdict **CONFIRMED** at **Low**, with the filing-bar caveat quoted above.

Severity: **Low**.

pepone:
`ExtractTask` unzipped the slicec and protoc archives straight into their final `obj/` folder, and the extraction step
was skipped whenever that folder existed. A build killed mid-extraction (hard kill, disk full, power loss) therefore
left a partial folder that was never re-extracted: the following `Copy` failed with MSB3030, or copied a truncated
executable, until the developer removed `obj/` by hand.

This PR makes the extraction all-or-nothing and self-healing:

- `ExtractTask` extracts into a sibling `<folder>.extracting` staging folder and moves it into place once every archive
  is extracted. The staging folder is removed on failure and replaced on the next run.
- The slicec and protoc extraction steps are now skipped only when the extracted executable exists, so a folder left
  without its executable is re-extracted on the next build instead of failing the following `Copy`.

Verified locally on macOS: a zip with a corrupt entry no longer leaves a destination folder behind, Unix executable
permissions are still preserved, a slicec or protoc folder whose executable was deleted heals on rebuild, and the
full solution builds. The task only uses netstandard 2.0 APIs, which is what `RoslynCodeTaskFactory` compiles against.

Fixes #4800

## What's Changed entry

None — build-system change to how this repository downloads its own toolchain; no user-facing impact.

### DIFF
--- a/build/ExtractTask.cs
+++ b/build/ExtractTask.cs
@@ -11,10 +11,13 @@ using Microsoft.Build.Utilities;
 /// <summary>A custom MSBuild task that extracts .zip archive files to a destination folder. This tasks uses
 /// <see cref="ZipFile.ExtractToDirectory(string, string)"/> to extract the files, which ensures that Unix permissions
 /// are correctly restored. The MSBuild Unzip task does not restore Unix permissions.</summary>
+/// <remarks>The archives are extracted into a sibling staging folder that replaces the destination folder only once
+/// all of them are extracted, so that an interrupted build does not leave a partially extracted destination folder
+/// behind.</remarks>
 public class ExtractTask : Task
 {
-    /// <summary>Gets or sets a <see cref="ITaskItem"/> with a destination folder path to unzip the files to.
-    /// </summary>
+    /// <summary>Gets or sets a <see cref="ITaskItem"/> with a destination folder path to unzip the files to. An
+    /// existing destination folder is replaced.</summary>
     [Required]
     public ITaskItem DestinationFolder { get; set; }
 
@@ -27,10 +30,32 @@ public class ExtractTask : Task
     /// <returns>Returns whether or not the execution completed successfully.</returns>
     public override bool Execute()
     {
-        foreach (ITaskItem sourceFile in SourceFiles)
+        string destinationFolder =
+            DestinationFolder.ItemSpec.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+        string stagingFolder = destinationFolder + ".extracting";
+
+        DeleteFolder(stagingFolder);
+        try
         {
-            ZipFile.ExtractToDirectory(sourceFile.ItemSpec, DestinationFolder.ItemSpec);
+            foreach (ITaskItem sourceFile in SourceFiles)
+            {
+                ZipFile.ExtractToDirectory(sourceFile.ItemSpec, stagingFolder);
+            }
+            DeleteFolder(destinationFolder);
+            Directory.Move(stagingFolder, destinationFolder);
+        }
+        finally
+        {
+            DeleteFolder(stagingFolder);
         }
         return true;
+    }
+
+    private static void DeleteFolder(string path)
+    {
+        if (Directory.Exists(path))
+        {
+            Directory.Delete(path, recursive: true);
+        }
     }
 }
--- a/build/ExtractTask.cs
+++ b/build/ExtractTask.cs
@@ -11,9 +11,8 @@ using Microsoft.Build.Utilities;
 /// <summary>A custom MSBuild task that extracts .zip archive files to a destination folder. This tasks uses
 /// <see cref="ZipFile.ExtractToDirectory(string, string)"/> to extract the files, which ensures that Unix permissions
 /// are correctly restored. The MSBuild Unzip task does not restore Unix permissions.</summary>
-/// <remarks>The archives are extracted into a sibling staging folder that replaces the destination folder only once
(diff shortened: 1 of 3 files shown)


pepone:
🤖 **Claude:** On the "Orphaned `.extracting` folders can remain after an interrupted extraction" note (suppressed on both csproj guards):

The premise that the destination can still contain the executable while a stale staging folder exists does not hold. `ExtractTask` only runs when the executable is absent from the destination, nothing in the task writes that executable before `Directory.Move`, and the move is a single directory rename. So after a kill at any point inside the task the executable is still missing, the next build's condition is true, the task runs again, and its first statement removes the stale staging folder before re-extracting.

Verified with a build-level check that reproduces the exact state described (a complete `slicec` in `obj/slicec-<version>-<platform>.extracting`, destination without `slicec`): the next `dotnet build src/IceRpc.Slice.Tools` re-extracted the executable and left no staging folder behind. A Codex review of this commit reached the same conclusion: extraction runs only when the destination executable is absent, so the orphan-folder concern does not apply.

The only leftover still possible is a staging folder for a toolchain version that was bumped between the kill and the next build. It accumulates the same way the per-version zips and extracted folders already do in `obj/`.

build/ExtractTask.cs
```diff
@@ -11,10 +11,13 @@
 /// <summary>A custom MSBuild task that extracts .zip archive files to a destination folder. This tasks uses
 /// <see cref="ZipFile.ExtractToDirectory(string, string)"/> to extract the files, which ensures that Unix permissions
 /// are correctly restored. The MSBuild Unzip task does not restore Unix permissions.</summary>
+/// <remarks>The archives are extracted into a sibling staging folder that replaces the destination folder only once
+/// all of them are extracted, so that an interrupted build does not leave a partially extracted destination folder
+/// behind.</remarks>
```
bernardnormier: State the guarantee rather than the failure it prevents.

```suggestion
/// <remarks>The destination folder is replaced only once every archive is extracted into a sibling staging folder,
/// so it is either complete or absent.</remarks>
```

bernardnormier (approved):
One doc comment nit.

externl (approved):

PR #4943, opened 2026-09-11
Record the RPC outcome on telemetry activities

### LINKED ISSUE #4807: [Audit-Medium] Telemetry activities do not record failed RPC outcomes

## Summary

`TelemetryInterceptor` and `TelemetryMiddleware` both start and tag an `Activity`, then `return await _next.InvokeAsync(...)` directly. Neither sets an error status or failure classification for a non-OK response, an exception thrown by invocation/dispatch, or a cancellation outcome.

## Impact

Traces present failed RPCs as successful or unclassified, weakening alerting and latency/error analysis.

Note this is distinct from the logger-framing dismissal recorded in [AGENTS.md](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/AGENTS.md) (#4431/#4432/#4433): that concerns how log *messages* frame equivalent wire outcomes. An OpenTelemetry activity status is a structured result field, not log framing.

## Source references

- [src/IceRpc.Telemetry/TelemetryInterceptor.cs#L38-L56](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Telemetry/TelemetryInterceptor.cs#L38)
- [src/IceRpc.Telemetry/TelemetryMiddleware.cs](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Telemetry/TelemetryMiddleware.cs)

## Recommendation

Record the RPC result before disposing the activity: set status and error attributes for non-OK responses and caught exceptions, then rethrow without changing RPC behavior.

## Audit context

M-04 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md); confirmed by the [Claude verification pass](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-claude-review-2026-07-28.md).

Severity: **Medium**.

pepone:
Fixes #4807

This PR builds on #4956, whose two commits are cherry-picked here; #4956 should be merged first.

The telemetry interceptor and middleware start an `Activity` per icerpc request, but the activity ended with its status
left `Unset` whatever the outcome: a response with a status code other than `Ok`, an exception thrown by the invocation
or the dispatch, and a cancellation all looked the same as a successful call in the exported traces, and no `error.type`
attribute was recorded. Trace backends render `Unset` as success, so failed RPCs did not show up as errors.

Both components now record the outcome before the activity stops, following the OpenTelemetry semantic conventions for
RPC spans and the "Recording errors" guidance. The rules differ by span kind, as they do for HTTP client and server
spans:

- The interceptor (client span) tags every response with `rpc.status_code`. A status code other than `Ok` sets the
  activity status to `Error`, with the response error message as the status description, and tags `error.type` with
  the status code name. An exception sets the activity status to `Error` with the exception message as the status
  description and tags `error.type` with the exception's fully qualified type name; no status code is recorded since
  there is no response.
- The middleware (server span) tags the activity with the `rpc.status_code` of the dispatch outcome: the status code of
  the returned response or, for an exception thrown by the dispatch, the status code of the failure response given by
  `DispatchException.FromException`. A status code that reports a problem with the request (`ApplicationError`,
  `NotFound`, `InvalidData`, `TruncatedPayload`, `Unauthorized`) leaves the activity status unset, like a 4xx on an
  HTTP server span, whether the dispatch returns it or throws it. Any other failure status code sets the activity
  status to `Error` and tags `error.type` with the status code name, whether the dispatch returns the response or
  throws. The status description is the error message of the response the caller receives.
- A cancellation by the token passed to `InvokeAsync` or `DispatchAsync` is the caller's decision, not a failure: the
  activity status stays unset, no `error.type` is set, and the `icerpc.canceled` tag is set to `true`. The middleware
  records no status code in this case, since the connection sends no response. A cancellation by another token is an
  error like any other exception.
- A successful response leaves the status `Unset` and sets no `error.type` on both sides.
- `StatusCode` is an unchecked enum, so `error.type` maps an undefined status code to `_OTHER` to keep its cardinality
  bounded. `rpc.status_code` keeps the exact value.

The RPC behavior is unchanged: the same response is returned and the same exception propagates.

The error message the caller receives is composed inside the `OutgoingResponse` constructor, so the middleware could
not get it. `DispatchException` gains a public `ErrorMessage` property that gives it: the exception's `Message`,
followed by the type and the message of its `InnerException` when there is one. Both protocol connections now build
the failure response with `new OutgoingResponse(request, dispatchException.StatusCode, dispatchException.ErrorMessage)`
and no exception argument, and the constructor and the new property share one copy of the format string in
`IceRpc.Internal.DispatchErrorMessage`. The wire messages don't change, so the protocol tests are unchanged.

New tests capture the activity from an `ActivityListener` when it stops and assert the status, the description, and
the tags for a failure response, for an exception, for a cancellation, and for a successful response, on both the
interceptor and the middleware. The middleware tests cover the status codes on both sides of the server-error line,
for returned responses and for thrown exceptions.

## What's Changed entry

Area: Telemetry

- The telemetry interceptor and middleware now record the outcome of each invocation and dispatch on the activity.
  Every activity that has a response carries an `rpc.status_code` tag. On the interceptor, a status code other than
  `Ok` or an exception sets the activity status to `Error` with an `error.type` tag. On the middleware, a status code
  that reports a server failure sets the activity status to `Error`, while a status code that reports a problem with
  the request, such as `NotFound` or `Unauthorized`, leaves it unset. A cancellation by the caller's cancellation token
  sets the `icerpc.canceled` tag instead of marking the activity as failed.

Area: Core

- `DispatchException` has a new `ErrorMessage` property that gives the error message of the failure response that
  carries this exception.


src/IceRpc.Telemetry/Internal/ActivityExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    /// <summary>Records the status code of a response. A status code other than <see cref="StatusCode.Ok" /> marks
+    /// the activity as failed.</summary>
+    /// <param name="activity">The activity.</param>
+    /// <param name="statusCode">The status code of the response.</param>
+    /// <param name="errorMessage">The error message of the response.</param>
+    internal static void RecordStatusCode(this Activity activity, StatusCode statusCode, string? errorMessage)
+    {
+        string statusCodeName = statusCode.ToString();
+        activity.SetTag("rpc.response.status_code", statusCodeName);
```
externl: `rpc.response.status_code` is already deprecated upstream: https://github.com/open-telemetry/semantic-conventions/pull/3921 renamed it to `rpc.status_code`, and the registry on `main` lists the old name as replaced. It is unreleased as of v1.44.0, but I'd ship the new name rather than one the next release retires. The remarks, the README and the tests carry the old name too.

src/IceRpc.Telemetry/Internal/ActivityExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        if (statusCode != StatusCode.Ok)
```
externl: HTTP draws this line by span kind: a client span records 4xx and 5xx as errors, which matches what the interceptor does, but a server span records only 5xx and unhandled exceptions. .NET follows that in `HttpClient`, in ASP.NET Core's built-in tracing and in the OpenTelemetry contrib instrumentation. Would it make sense for the middleware to do the same and leave the codes that report the caller's problem, such as `NotFound`, `Unauthorized` and `InvalidData`, unset?

bernardnormier (commented):
Agreed with Joe on both points.

`RecordStatusCode` and `RecordException` are inconsistent: the first records the status code tag, the second
doesn't. Give `RecordException` a `StatusCode?` parameter, with null meaning no status code tag. The interceptor
passes null: an exception there means no response. The middleware passes the status code the caller will get, which
the protocol connection decides after the middleware returns (`ConvertToInternalError`, the mapping of
`InvalidDataException` and friends). Don't reproduce that mapping in the middleware: fold the two halves of the
conversion (the type switch in `IceRpcProtocolConnection` and `ToOutgoingResponse`) into one internal helper in
IceRpc, call it from both places, and add `InternalsVisibleTo("IceRpc.Telemetry")`. Internal because the mapping is
icerpc-specific: the ice connection maps only `InvalidDataException`. Please file an issue on whether the two
protocols should share one mapping, exposed publicly for any middleware that reports outcomes.

Please delete the new paragraph in the README: the README is a high-level description, and the tags are a detail
that belongs in the class remarks.

The new remarks sentence in both classes is hard to parse: four facts in one sentence with a nested parenthesis.
One sentence per fact, for example on the interceptor:

    The activity records the outcome of the invocation. When the invocation returns a response, the
    `rpc.status_code` tag holds its status code. When the status code is not `Ok` or the invocation throws an
    exception, the activity status is `Error` and the `error.type` tag identifies the failure.

Two test nits, take them or leave them: `Assert.That(outcome.Tags, Does.ContainKey("x").WithValue("y"))` replaces
each ContainsKey/indexer pair, and `Assert.CatchAsync` replaces the try/catch that captures the thrown exception.

src/IceRpc.Telemetry/Internal/ActivityExtensions.cs
pepone: 🤖 **Claude:** Done in d18069092: the tag is now `rpc.status_code` in the code, the remarks and the tests. Checked against the semantic-conventions registry on `main`, which lists `rpc.response.status_code` as renamed to `rpc.status_code`. The README paragraph is gone, per Bernard's comment.

src/IceRpc.Telemetry/Internal/ActivityExtensions.cs
pepone: 🤖 **Claude:** Done in d18069092. The middleware now leaves the activity status unset for `ApplicationError`, `NotFound`, `InvalidData`, `TruncatedPayload` and `Unauthorized`, while still recording `rpc.status_code` for them. `NotImplemented`, `Unavailable`, `InternalError`, `DeadlineExceeded`, `NotSupported` and unknown values set `Error`, mirroring the gRPC server list (UNIMPLEMENTED, UNAVAILABLE, INTERNAL, DEADLINE_EXCEEDED, UNKNOWN). `TruncatedPayload` sits on the request side because its documentation says it is usually caused by the invoker; `ApplicationError` because it is an application-defined outcome rather than a server failure. The middleware remarks list both sets. The interceptor is unchanged: every status code other than `Ok` is an error, as for HTTP client spans.

pepone:
🤖 **Claude:** Applied in d18069092:

- `RecordException` takes a `StatusCode?`. The interceptor passes null; the middleware passes the status code the
  icerpc connection derives from the exception. The type switch in `IceRpcProtocolConnection` and
  `DispatchException.ToOutgoingResponse` are folded into `IceRpc.Internal.ExceptionExtensions` (`ToStatusCode` and
  `ToOutgoingResponse`): the connection calls the fold, the middleware calls `ToStatusCode`, and IceRpc adds
  `InternalsVisibleTo("IceRpc.Telemetry")`. The ice connection keeps its own mapping and still calls
  `ToOutgoingResponse` on the `DispatchException` it builds. The protocol connection tests pass unchanged.
- A dispatch canceled by the middleware's cancellation token records no status code, mirroring the connection's
  `OperationCanceledException` filter: it sends no response in that case.
- Filed #4954 on sharing the mapping between the two protocols and exposing it publicly.
- README paragraph removed. Remarks rewritten one sentence per fact, using your interceptor text.
- Tests use `Does.ContainKey().WithValue()` and `Assert.CatchAsync`.

pepone:
🤖 **Claude:** Two policy questions on the middleware that d18069092 leaves as reviewed, for @externl and
@bernardnormier:

1. **Exceptions vs. equivalent responses.** Every exception thrown by the dispatch sets the activity status to
   `Error`, per "5xx and unhandled exceptions". In IceRPC, exceptions also map to request-problem codes:
   `throw new DispatchException(StatusCode.Unauthorized)`, or an `InvalidDataException` while decoding the payload,
   produces the same `Unauthorized` / `InvalidData` response as returning it, yet the thrown form is `Error` and the
   returned form is unset. If the derived status code should decide in both cases, the change is one line in the
   middleware's exception path, `isError = statusCode is null || IsServerError(statusCode)`, which keeps every
   exception mapped to `InternalError` as an error.

2. **Cancellation.** A dispatch canceled by the middleware's cancellation token is recorded as `Error` with
   `error.type = System.OperationCanceledException` and no status code, matching how the PR treats cancellation on the
   interceptor (as `HttpClient` does). On the server side this cancellation means the connection is shutting down or
   the peer went away, not a server failure, and the HTTP conventions (Development status) say a caller-requested
   cancellation should not be an error. Should the middleware leave the status unset in that case, and should the
   interceptor do the same when the token passed to `InvokeAsync` is the one that fired?

bernardnormier (commented):
The `DispatchExceptionExtensions` to `ExceptionExtensions` change makes the conversion harder to follow: the
ConvertToInternalError rule now lives in both `ToStatusCode` and `ToOutgoingResponse`. Rather than an internal
icerpc-only helper plus `InternalsVisibleTo`, let's fix #4954 first: one public method that gives the status code
of the response for an exception thrown by a dispatch, for both protocols. Then this PR builds on it and the
`IceRpc` changes here go away. The status description stays the exception's message; the caller's composed error
message adds nothing that isn't already on the activity.

`ActivityExtensions` isn't pulling its weight either: please remove it and inline the tag and status calls in the
interceptor and the middleware.

On your two questions:

1. Yes, the derived status code decides in both paths. Throwing a `DispatchException` is how a service returns a
   failure in IceRPC, so a thrown `Unauthorized` and a returned `Unauthorized` must record the same outcome. A
   dispatch exception with a status code then goes through the same code as a response.
2. A cancellation isn't an error when the `OperationCanceledException` carries the token passed to `InvokeAsync`
   or `DispatchAsync`, since that's the caller's own decision: leave the status unset, don't set `error.type`, and
   record it explicitly with an `icerpc.canceled` tag set to true. The HTTP client conventions say the same about
   the status for a cancellation the instrumentation can recognize as the caller's. Any other exception, a
   cancellation by another token included, stays an error.

pepone:
🤖 **Claude:** Applied in 2ad1d9ea4. It sits on top of the two #4956 commits, which are cherry-picked into this branch
(ea425cae0, 8a5c341e9), so #4956 should be merged first.

- The `IceRpc` changes are gone: `src/IceRpc` now matches #4956. The middleware derives the status code of an exception
  with `DispatchException.FromException`. The status description stays the exception's message.
- `ActivityExtensions` is removed. The tag and status calls are inline in the interceptor and the middleware.
- The derived status code decides in both middleware paths. A thrown `DispatchException` with `Unauthorized`, an
  `InvalidDataException` or a truncated payload leaves the status unset, like the equivalent response. For a server
  error status code, a dispatch exception that `FromException` returns unchanged sets `error.type` to its status code
  name, as a response does. Any other exception, including a `DispatchException` with `ConvertToInternalError`, sets
  `error.type` to the exception's type name, so the status code, the type and the message are all on the activity.
- A cancellation that carries the token passed to `InvokeAsync` or `DispatchAsync` sets `icerpc.canceled` to true and
  sets neither the status nor `error.type`. The middleware records no `rpc.status_code` for it. The filter also
  requires that token to be canceled: without that check, an `OperationCanceledException` created without a token
  would pass for the caller's cancellation when the caller passes `CancellationToken.None`. Any other cancellation
  stays an error.

src/IceRpc.Telemetry/Internal/ActivityExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+        this Activity activity,
+        StatusCode statusCode,
+        string? errorMessage,
+        bool isError)
+    {
+        string statusCodeName = statusCode.ToString();
+        activity.SetTag("rpc.status_code", statusCodeName);
+        if (isError)
+        {
+            activity.SetTag("error.type", statusCodeName);
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Done in 2ad1d9ea4. `error.type` holds the status code name only for a defined `StatusCode` and `_OTHER`
otherwise, on both the interceptor and the middleware. `rpc.status_code` keeps the exact value. The `(StatusCode)42`
tests now expect `42` and `_OTHER`.

src/IceRpc.Telemetry/TelemetryMiddleware.cs
```diff
@@ -43,7 +54,39 @@ public async ValueTask<OutgoingResponse> DispatchAsync(IncomingRequest request,
...
+                activity.SetTag("rpc.status_code", dispatchException.StatusCode.ToString());
+                if (IsServerError(dispatchException.StatusCode))
+                {
+                    // Like a returned response, a dispatch exception that FromException returns unchanged identifies
+                    // the failure by its status code. Any other exception is identified by its type.
+                    activity.SetTag(
+                        "error.type",
+                        ReferenceEquals(dispatchException, exception) ?
+                            dispatchException.StatusCode.ToErrorType() : exception.GetType().FullName);
+                    activity.SetStatus(ActivityStatusCode.Error, exception.Message);
```
bernardnormier: Same rule as the response path: the status code identifies the failure, and the description is the error message the
caller receives. The `ReferenceEquals` branch and its comment go away.

```suggestion
                    activity.SetTag("error.type", dispatchException.StatusCode.ToErrorType());
                    activity.SetStatus(ActivityStatusCode.Error, dispatchException.ErrorMessage);
```

The server-error exception test cases then expect the status code name as the error type on every row, and the
remarks and the PR description drop the sentence about other exceptions being identified by their type.

bernardnormier (commented):
Everything from my last review is addressed: the `IceRpc` changes are gone and `src/IceRpc` matches #4956, the
middleware derives the status code with `DispatchException.FromException`, `ActivityExtensions` is inlined, the
derived status code decides in both middleware paths, and a cancellation by the caller's token sets
`icerpc.canceled` with the status left unset. I checked that last one against the real pipeline: the protocol
connection and `ClientConnection` rethrow a caller cancellation with `cancellationToken.ThrowIfCancellationRequested()`,
so the exception the interceptor sees carries the caller's token.

One more change, on the middleware's exception path (inline comment). The `ReferenceEquals` branch makes that block
hard to follow, and the status description should be the error message the caller receives, not the exception's
message. That message is composed inside the `OutgoingResponse` constructor today, so the middleware can't get it.
Please add a public computed property to `DispatchException`:

```csharp
/// <summary>Gets the error message of the failure response that carries this exception.</summary>
public string ErrorMessage =>
    InnerException is null ?
        Message :
        $"{Message} The failure was caused by an exception of type '{InnerException.GetType()}' with message: {InnerException.Message}";
```

Both connections then create the response with `new OutgoingResponse(request, dispatchException.StatusCode,
dispatchException.ErrorMessage)` and no exception argument, and the `OutgoingResponse` constructor delegates to a
shared internal helper rather than keeping a second copy of the format string. The wire messages don't change, so the
protocol tests stay as they are. Here or in a small PR first, your call.

src/IceRpc.Telemetry/TelemetryMiddleware.cs
pepone: 🤖 **Claude:** Done in 32fa9191a. The `ReferenceEquals` branch and its comment are gone: `error.type` is
`dispatchException.StatusCode.ToErrorType()` and the status description is `dispatchException.ErrorMessage`, the same
rule as the response path.

The server-error exception test cases now expect the status code name on every row — the `ConvertToInternalError` row
goes from `IceRpc.DispatchException` to `InternalError` and the `NotSupportedException` row from
`System.NotSupportedException` to `NotSupported` — and each row carries the error message the caller receives.

pepone:
🤖 **Claude:** Applied in 32fa9191a, in this PR rather than a separate one.

- `DispatchException` has the public `ErrorMessage` property with the body you gave. `OutgoingResponse`'s constructor
  and the new property share one copy of the format string, in `IceRpc.Internal.DispatchErrorMessage.Compose`. Both
  connections build the failure response with `new OutgoingResponse(request, dispatchException.StatusCode,
  dispatchException.ErrorMessage)` and no exception argument. The wire messages don't change and the protocol
  connection tests pass unchanged.
- The middleware's exception path is the two lines from your suggestion; the `ReferenceEquals` branch and its comment
  are gone, and the server-error exception test cases expect the status code name on every row.
- The PR description drops the sentence about other exceptions being identified by their type and gains an
  `Area: Core` entry for the new property. The class remarks needed no change: neither mentions exception type names,
  they say only that `error.type` identifies the failure.

Local run: the telemetry tests pass (42), and `IceRpc.Tests` passes except four `DefaultTransportTests` QUIC cases
that fail identically without this commit (`QUIC_STATUS_TLS_ERROR` on connect) — a local QUIC setup problem, not a
regression.

bernardnormier (approved):
Looks good.

src/IceRpc.Telemetry/TelemetryMiddleware.cs
```diff
@@ -43,7 +54,34 @@ public async ValueTask<OutgoingResponse> DispatchAsync(IncomingRequest request,
...
+                throw;
+            }
+            catch (Exception exception)
+            {
+                DispatchException dispatchException = DispatchException.FromException(exception);
+                activity.SetTag("rpc.status_code", dispatchException.StatusCode.ToString());
+                if (IsServerError(dispatchException.StatusCode))
+                {
+                    activity.SetTag("error.type", dispatchException.StatusCode.ToErrorType());
+                    activity.SetStatus(ActivityStatusCode.Error, dispatchException.ErrorMessage);
```
externl, line 81: The recording-errors guidance asks for the exception message as the status description and advises against repeating the status code in it: https://opentelemetry.io/docs/specs/semconv/general/recording-errors/#recording-errors-on-spans. The composed message opens with the status code, so `exception.Message` would sit closer to the spec here. It would also let `DispatchException.ErrorMessage`, `DispatchErrorMessage`, and the `OutgoingResponse` change go, keeping this PR inside the telemetry package.

src/IceRpc.Telemetry/Internal/StatusCodeExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+
+/// <summary>Provides an extension method for <see cref="StatusCode" /> to identify a failure in the
+/// <c>error.type</c> tag of an activity.</summary>
+internal static class StatusCodeExtensions
+{
+    /// <summary>Gets the <c>error.type</c> of a failure status code. <see cref="StatusCode" /> is an unchecked enum, so
+    /// an undefined value maps to <c>_OTHER</c> to keep the cardinality of <c>error.type</c> bounded.</summary>
+    /// <param name="statusCode">The failure status code.</param>
+    /// <returns>The name of <paramref name="statusCode" /> when it is defined; otherwise, <c>_OTHER</c>.</returns>
+    internal static string ToErrorType(this StatusCode statusCode) =>
```
externl, line 13: Does `_OTHER` earn its own class? `rpc.status_code` already keeps the raw value, and the low-cardinality requirement on `error.type` exists for metric dimensions, which these activities don't feed. `ToString()` on both tags would do the job, and `StatusCodeExtensions` and the two `(StatusCode)42` rows could go with it.

src/IceRpc.Telemetry/TelemetryMiddleware.cs
pepone: @bernardnormier do you agree? I ask because adding ErrorMessage was Bernard's suggestion from the previous review.

src/IceRpc.Telemetry/Internal/StatusCodeExtensions.cs
pepone: I’d keep `_OTHER` here. The OpenTelemetry RPC span conventions recommend predictable, low-cardinality `error.type` values for spans too, so the guidance isn’t limited to metric dimensions. `StatusCode` is an unchecked  `varuint62` enum; unknown values could produce many distinct error types. Grouping them under `_OTHER` makes that tag useful for querying error classes, while `rpc.status_code` still preserves the exact value for diagnosis. The helper and unknown-code test cases seem worth keeping to make that policy explicit.

src/IceRpc.Telemetry/TelemetryMiddleware.cs
bernardnormier: No, I don't agree with @externl. 

We should record what we actually send with the response (as per your latest updates). The status code is occasionally "duplicated" in the message (when the error message is not supplied); not a big deal in my view.

externl (approved):

PR #4945, opened 2026-09-11
Add ConnectionCacheOptions.MaxConnections

pepone:
`ConnectionCache` created a new connection for every server address it had not connected to yet, with no bound on the total number of connections it manages. An application that routes invocations through a cache over a large or externally supplied address set (a long alt-server list, a locator result) could therefore open an unbounded number of outbound connections.

`ConnectionCacheOptions` now has a `MaxConnections` property that mirrors `ServerOptions.MaxConnections` (validated `>= 0`, defaults to `0` meaning unlimited). The cache counts what the server counts: active connections plus connections being established, shut down or disposed. Once the limit is reached, an invocation that requires a new connection fails with `IceRpcException(IceRpcError.LimitExceeded)`. Before failing, the cache still walks the request's server addresses and uses an active or pending connection to any of them, so the limit never prevents reuse. `LimitExceeded` is not retried by the retry interceptor.

### Tests

- `Invocation_fails_when_max_connections_is_reached`: with `MaxConnections = 1` and one active connection, an invocation to a second server fails with `LimitExceeded` while an invocation to the first server still succeeds.
- `Pending_connection_counts_toward_max_connections`: a connection whose transport connect is held counts toward the limit; an invocation to a second server fails with `LimitExceeded` and the held invocation completes once released.
- `Use_active_connection_to_alt_server_when_max_connections_is_reached`: with `PreferExistingConnection = false` and the limit reached, an invocation whose main address has no connection is sent over the active connection to its alt server instead of failing.
- `ConnectionCacheOptionsTests.MaxConnections_rejects_negative_values`, and `LimitExceeded` added to the retry interceptor's non-retryable exceptions.

All three cache tests fail without the `ConnectionCache` change (no exception thrown, connect timeout, and the request going to the main server respectively).

## What's Changed entry

Area: Core

- Added `ConnectionCacheOptions.MaxConnections`, the maximum number of connections a `ConnectionCache` manages (defaults to 0, meaning unlimited). Once the limit is reached, an invocation that requires a new connection fails with `IceRpcError.LimitExceeded`.

### DIFF
--- a/src/IceRpc/ConnectionCache.cs
+++ b/src/IceRpc/ConnectionCache.cs
@@ -13,7 +13,9 @@ namespace IceRpc;
 /// <summary>Represents an invoker that routes outgoing requests to connections it manages.</summary>
 /// <remarks><para>The connection cache routes requests based on the request's <see cref="IServerAddressFeature" />
 /// feature or the server addresses of the request's target service.</para>
-/// <para>The connection cache keeps at most one active connection per server address.</para></remarks>
+/// <para>The connection cache keeps at most one active connection per server address.
+/// <see cref="ConnectionCacheOptions.MaxConnections" /> bounds the total number of connections it manages.</para>
+/// </remarks>
 public sealed class ConnectionCache : IInvoker, IAsyncDisposable
 {
     // Connected connections.
@@ -36,6 +38,8 @@ public sealed class ConnectionCache : IInvoker, IAsyncDisposable
     private readonly CancellationTokenSource _disposedCts = new();
 
     private Task? _disposeTask;
+
+    private readonly int _maxConnections;
 
     private readonly Lock _mutex = new();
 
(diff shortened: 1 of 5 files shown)


bernardnormier (commented):
A new `ConnectionCacheOptions` property needs a use case, and there is no clear one here. A client connects only to
addresses it trusts or has validated, and the number of connections is the number of servers it chose to talk to.

Even taking the description's scenario as real (a large or externally supplied address set), it's not clear that a
cap set through a property, enforced by failing the invocation, is what such an application wants. It may prefer to
know how many connections the cache holds and decide for itself, which a count exposed by `ConnectionCache` would
give it without a new failure mode on the invocation path. That would be a separate discussion with a use case
behind it.

Let's close this PR, and close the finding as by design.

Issue #4833, opened 2026-07-28
[Audit-Low] Build tooling and developer environment (5 findings)


pepone:
> **AI-generated audit findings** — from a repository-wide audit performed by OpenAI Codex (`codex-cli`). ⚠️ Unlike the High/Medium findings from the same audit, **these Low-severity items were not individually re-verified** by the follow-up Claude pass — they are grouped here for triage rather than asserted as confirmed. Verify each before acting on it.

Five build/tooling defects from the same audit, grouped for triage.

| ID | Finding |
|---|---|
| L-26 | Direct packing of a tooling project can produce an incomplete package — the tooling globs assume dependent generator/telemetry output already exists from a whole-solution build |
| L-27 | Slice telemetry always reports the IceRPC C# generator even when `IceRpc` metadata disables it, and omits custom generator metadata; the list should be built from the actual invocation |
| L-28 | `whitespace.yml` uses the mutable `zeroc-ice/github-actions@main` while other external actions are commit-pinned |
| L-29 | `.prettierrc` requires `@prettier/plugin-xml`, but the repository has no package manifest or lockfile installing it |
| L-30 | The devcontainer Dockerfile puts literal `~/.cargo/bin` in `ENV PATH`; executable lookup does not expand tildes inside PATH entries, so non-login processes can fail to find Cargo |

## Note

L-28 is a one-line supply-chain hardening fix and consistent with how every other external action in the repository is already pinned — worth doing regardless of the rest.

L-26 is also filed separately as a Medium (it was verified), since an isolated pack silently omits the plug-ins rather than failing.

## Audit context

L-26 through L-30 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md).

pepone:
All the items in this issue have been resolved or dismissed.

| ID | Resolution |
|---|---|
| L-26 | Fixed by #4924 (filed as the verified Medium #4824). The `IceRpc.Protobuf.Tools` / `IceRpc.Slice.Tools` plug-in globs now run in a pack-time target — as the protoc and slicec binaries already did — so an isolated `dotnet pack` picks up what its project references just built, and the target now *fails* the pack, naming the project, the configuration and the build command, when a bundled plug-in is missing instead of silently producing an incomplete package. |
[...]

PR #4946, opened 2026-09-11
Fix Slice build telemetry

pepone:
Slice build telemetry reported the IceRPC C# generator for every Slice file, including files with `IceRpc="false"`, because the generator list was part of the project-level `_BuildTelemetryOptions` property. List the generators per `_SliceFile` item instead: the IceRPC generator entry is added in the `IceRpc` metadata block, and the build-telemetry generator's options are assembled from the item's metadata. Only the bundled generators are reported, and `SlicecTask` is unchanged.

Verified with a project containing one Slice file with `IceRpc="true"` and one with `IceRpc="false"`: the second now reports `slicec-csharp-generator` and `slicec-build-telemetry` only, and the number of `SlicecTask` invocations is unchanged since `Generators` already varied with `IceRpc`.

Addresses L-27 in #4833.

## What's Changed entry

None — build telemetry data correction.


src/IceRpc.Slice.Tools/SlicecTask.cs
```diff
@@ -54,7 +61,26 @@ protected override string GenerateCommandLineCommands()
...
+            {
+                // Report the executable name only; paths and plug-in options can contain private information.
+                string name = Path.GetFileName(generator.Split(',', 2)[0].Replace('\\', '/'));
+                if (Path.GetExtension(name) is ".sh" or ".bat" or ".exe")
+                {
+                    name = Path.GetFileNameWithoutExtension(name);
+                }
+                string version = name is
+                    "slicec-csharp-generator" or "slicec-icerpc-csharp-generator" or "slicec-build-telemetry" ?
+                    IceRpcSliceToolsVersion : "unknown";
```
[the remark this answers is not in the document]
pepone: Addressed in 9827f8c66. MSBuild now passes the configured bundled-generator paths explicitly, and SlicecTask compares complete generator entries against these paths instead of their basenames. A custom entry such as /private/tools/slicec-csharp-generator.sh therefore reports version unknown. Added regression cases for custom scripts sharing each bundled generator name on Unix and Windows; all 13 tests pass. The MSBuild invocation check also verifies a custom/bundled filename collision.

src/IceRpc.Slice.Tools/SlicecTask.cs
```diff
@@ -54,7 +64,25 @@ protected override string GenerateCommandLineCommands()
     {
         var builder = new CommandLineBuilder(false);
 
-        foreach (string generator in Generators)
+        IEnumerable<string> generators = Generators;
+        if (BuildTelemetryGenerator.Length > 0)
+        {
+            string generatorOptions = string.Concat(Generators.Append(BuildTelemetryGenerator).Select(generator =>
```
externl, line 70: As far as I'm aware this is currently not happening.

- We never pass paths to the build telemetry server
- We don't pass anything with an extension
- There are no unknowns. 
- We control what's passed into this task and no one else will be calling it.

If anything I think maybe something incorrect here should just assert or throw.

src/IceRpc.Slice.Tools/SlicecTask.cs
```diff
@@ -16,11 +16,21 @@ public class SlicecTask : ToolTask
     /// <summary>Additional options to pass to the <c>slicec</c> compiler.</summary>
     public string[] AdditionalOptions { get; set; } = [];
 
+    /// <summary>The build-telemetry generator script and its options, or an empty string to disable telemetry.
+    /// </summary>
+    public string BuildTelemetryGenerator { get; set; } = "";
+
+    /// <summary>The bundled generator scripts whose versions are known.</summary>
+    public string[] BundledGenerators { get; set; } = [];
```
externl, line 24: Shouldn't we always know the version. Its our tooling.

BUILDING.md
```diff
@@ -36,6 +37,15 @@ dotnet build
 
 This command builds all the tools, sources and tests with the default configuration (debug).
 
+## Formatting XML files
+
+To use the repository's Prettier configuration, install Node.js and run `npm ci` from the repository root. This installs
+Prettier and its XML plugin for the VS Code Prettier extension and the command line:
+
+```shell
+npx prettier --write path/to/project.csproj
```
externl, line 46: Doesn't `npx` download the package on demand?.  We should also enforce in ci. Maybe we can add an npm script `rpm run ...`

src/IceRpc.Slice.Tools/SlicecTask.cs
pepone: The task support custom generators.

src/IceRpc.Slice.Tools/SlicecTask.cs
pepone: This is because the generators supported by the task is not fixed, you can configure your own generators.

externl: Shouldn't we require a version?

src/IceRpc.Slice.Tools/SlicecTask.cs
externl: Right, but they should pass valid things like a version. I'm fine to strip path, as what we pass should only be generator name anyway. How does this currently work with our wrapper scripts?

bernardnormier (commented):
Please keep only L-27 in this PR and move the other fixes to separate PRs.

On L-27 itself, the first half is real: `_BuildTelemetryOptions` is a property, so every Slice file reports the
IceRPC generator whatever its `IceRpc` metadata says. The second half is not a defect. Custom generators are the
user's tooling, we can't know their version, and reporting their names collects more private information, not
less. The fix for the real half is a few lines in the targets, with no change to `SlicecTask`, no new task
parameters and no new test project. Drop the three `generator=` entries from `_BuildTelemetryOptions`, add a
metadata in the `IceRpc` block:

    <_SliceFile Condition="'%(IceRpc)' == 'true'">
      ...
      <BuildTelemetryGenerators>,generator=slicec-icerpc-csharp-generator:$(IceRpcSliceToolsVersion)</BuildTelemetryGenerators>
    </_SliceFile>

and list the generators per item in the telemetry entry:

    <Generators>%(_SliceFile.Generators);$(_IceRpcSliceBuildTelemetryPath)slicec-build-telemetry$(_SliceGeneratorScriptExtension)$(_BuildTelemetryOptions),generator=slicec-csharp-generator:$(IceRpcSliceToolsVersion)%(_SliceFile.BuildTelemetryGenerators),generator=slicec-build-telemetry:$(IceRpcSliceToolsVersion)</Generators>

`Generators` already varies with `IceRpc`, so the batches of the `SlicecTask` invocation are unchanged. That also
settles Joe's two threads on the task: no `unknown` version and no path handling, because only our generators are
reported.

tests/IceRpc.Slice.Tools.Tests/SlicecTaskTests.cs
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) ZeroC, Inc.
```
bernardnormier: Please drop the new test project. Build telemetry exists for us to collect data about slicec compilations; it's
not a product feature that needs unit tests. And with the targets-only fix above there is no task change to test.

pepone:
🤖 **Claude:** Applied the targets-only fix in 68195ba04: the generator entries moved from `_BuildTelemetryOptions` into per-item metadata, and the `SlicecTask` change, the README change and the test project are reverted. Verified with a two-file project (`IceRpc` true and false): the `IceRpc="false"` batch now reports `slicec-csharp-generator` and `slicec-build-telemetry` only, still with two `SlicecTask` invocations. L-28, L-29 and L-30 moved to #4951, #4952 and #4953.

tests/IceRpc.Slice.Tools.Tests/SlicecTaskTests.cs
pepone: 🤖 **Claude:** Dropped in 68195ba04.

src/IceRpc.Slice.Tools/SlicecTask.cs
pepone: 🤖 **Claude:** Superseded by 68195ba04, which reverts `SlicecTask` to main per Bernard's review. Only the bundled generators are reported, with their names and `$(IceRpcSliceToolsVersion)` written in the targets. The wrapper scripts (`slicec-*-generator.sh`/`.bat`) are still passed to slicec as `--generator` paths but never appear in the telemetry.

src/IceRpc.Slice.Tools/SlicecTask.cs
pepone: 🤖 **Claude:** Superseded by 68195ba04: no `unknown` version remains, every reported generator carries `$(IceRpcSliceToolsVersion)`.

BUILDING.md
pepone: 🤖 **Claude:** This change moved to #4952. On npx: after `npm ci`, `npx prettier` runs the copy in `node_modules`; without it npx fetches the latest `prettier` package alone and the `@prettier/plugin-xml` plugin named in `.prettierrc` is missing. On CI enforcement: with the locked versions, 32 of the 209 XML-style files in the repository currently fail `prettier --check`, so it needs a formatting pass first. An npm script is easy to add; let's settle both in #4952.

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets
```diff
@@ -73,7 +73,7 @@
   <Target Name="_ComputeSliceFiles" Condition="@(SliceFile) != ''">
     <PropertyGroup>
       <!-- Build the per-generator options suffix for the build-telemetry generator. -->
```
bernardnormier: The generator entries moved to the items below.

```suggestion
      <!-- Options of the build-telemetry generator; the generator entries are added per Slice file below. -->
```

bernardnormier (approved):
Please drop "and developer tooling" from the title now that only L-27 is left, and replace the What's Changed entry
by `None — build telemetry data correction`.

src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.targets
```diff
@@ -87,9 +87,10 @@
       <_SliceFile Condition="'%(IceRpc)' == 'true'">
         <Generators>%(_SliceFile.Generators);$(_IceRpcSliceGeneratorPath)slicec-icerpc-csharp-generator$(_SliceGeneratorScriptExtension)</Generators>
         <GeneratedPath>%(_SliceFile.GeneratedPath);$([MSBuild]::NormalizePath('%(OutputDir)/%(FileName).IceRpc.cs'))</GeneratedPath>
+        <BuildTelemetryGenerators>,generator=slicec-icerpc-csharp-generator:$(IceRpcSliceToolsVersion)</BuildTelemetryGenerators>
       </_SliceFile>
       <_SliceFile Condition="'$(IceRpcBuildTelemetry)' == 'true'">
-        <Generators>%(_SliceFile.Generators);$(_IceRpcSliceBuildTelemetryPath)slicec-build-telemetry$(_SliceGeneratorScriptExtension)$(_BuildTelemetryOptions)</Generators>
+        <Generators>%(_SliceFile.Generators);$(_IceRpcSliceBuildTelemetryPath)slicec-build-telemetry$(_SliceGeneratorScriptExtension)$(_BuildTelemetryOptions),generator=slicec-csharp-generator:$(IceRpcSliceToolsVersion)%(_SliceFile.BuildTelemetryGenerators),generator=slicec-build-telemetry:$(IceRpcSliceToolsVersion)</Generators>
```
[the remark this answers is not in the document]
bernardnormier: On purpose: only our generators are reported. See the review.

Issue #4788, opened 2026-07-28
[Audit-Medium] Slic local stream releases its stream-count permit before the peer has consumed the data


pepone:
> **AI-generated audit finding** — this issue was opened from an automated correctness audit performed by Claude (Anthropic), model **Opus 5** (`claude-opus-5`), and independently cross-verified with OpenAI Codex (`codex-cli 0.145.0`). It **has not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

`CloseWrites` gates the "send a closing frame" path on `IsStarted && !WritesClosed && !_writesClosePending`, but `WriteLastStreamFrameAsync` has already set `_writesClosePending = true` for any `WriteAsync(..., endStream: true)`. On the **success** path that flag is never rolled back — the `catch` added by #4765 rolls it back only when the write fails before the frame was queued.

A subsequent `Output.Complete()` therefore falls into the frame-less `else` branch and calls `TrySetWritesClosed()` immediately, contradicting the invariant `WroteLastStreamFrame` documents ("For local streams, writes will be closed only once the peer sends the StreamReadsClosed frame") and that the comment in `CloseWrites` restates. `TrySetState` then reaches `SlicConnection.ReleaseStream`, which releases the bidirectional/unidirectional semaphore while the peer still counts the stream as open.

## Trigger / impact

**Every one-way icerpc invocation is the default victim**: `CreateStreamAsync(bidirectional: false)` produces a stream whose reads are already closed at construction, so `Output.Complete()` after the endStream write frees the permit with zero peer involvement.

Reproduced end-to-end over Slic/TCP (not merely at the transport API): with `MaxIceRpcUnidirectionalStreams = 2` and a blocking dispatcher, 12 one-way `InvokeAsync` calls all returned in ~0 ms with no throttling; the server then logged `The maximum unidirectional stream count 3 was reached.` and tore the connection down, silently losing the in-flight one-way requests (the client transparently reconnected). The same sequence with a non-endStream write correctly blocks, isolating `_writesClosePending` as the cause.

Bidirectional streams are affected more weakly — Codex noted that a premature `WritesClosed` does not release the permit until local reads also close, but the permit can still be released on the response `StreamLast` before the peer's `StreamReadsClosed` acknowledgment.

## Source references

Pinned to commit `2de29310a82428560125c5f804127ea054f46769`:

[...]

PR #4947, opened 2026-09-11
Send the deferred Slic StreamReadsClosed frame when the peer closes writes first

pepone:
A client that cancels an invocation while the server is still sending the response can leak a bidirectional stream slot on its connection. Each such cancellation leaves one stream that never closes, and once the connection's bidirectional stream limit is reached, every new invocation on it blocks until the connection is closed.

The cause is on the server side of the stream. When the server has consumed the whole request, it defers its `StreamReadsClosed` frame so that it can be bundled with the response's `StreamLast` frame in a single write. When the client's own `StreamReadsClosed` frame (sent when it stops reading the response) closes the server's writes first, the `StreamLast` frame is never written and the deferred `StreamReadsClosed` frame was never sent: the client stream is left waiting for it and is never released.

This happens in three timings, each covered by a new test:

- the response's endStream write is parked (exhausted peer window or connection write semaphore) and is canceled by the client's frame;
- the client's frame arrives before the endStream write is attempted, so the write fails right away;
- no endStream write is attempted at all (`CopyFromAsync` returns as soon as writes are closed and the dispatch completes the output).

The rollback added in #4765 doesn't cover these cases because `_writesClosedTcs`, which it keys on, is also completed when the peer closes writes.

The leak shows up when the client's `StreamLast` frame is sent by the output completion rather than by an endStream write: raw `IMultiplexedStream` usage, or a plain `PipeWriter` payload decorator such as the compressor. After an endStream write, `CloseWrites` currently closes the local stream's writes as soon as the output is completed instead of waiting for the peer's `StreamReadsClosed` frame (#4788), which hides the leak for that shape; the tests therefore send the client's `StreamLast` frame through `Output.Complete()`.

### Fix

`ReceivedReadsClosedFrame` now sends the deferred `StreamReadsClosed` frame on its own when the peer closes writes while the frame is still owed, and clears the deferral. If a `StreamLast` write is already past cancellation and bundles the frame too, the peer receives it twice, which is harmless: a released stream is dropped by the connection and a live one handles it idempotently.

Two ordering details keep this race-free, without closing reads or writes under the stream mutex:

- `ReceivedReadsClosedFrame` closes writes before it captures the deferral under the mutex: a `CloseReads` that ran before left the deferral for the handler, and one that runs after sees writes closed and sends the frame itself.
- `CloseWrites` clears the deferral when it claims it, so `ReceivedReadsClosedFrame` can't send the frame before the stream is released; the bundled frame follows the closure of writes on the same thread.

### Tests

`Deferred_reads_closed_frame_is_sent_when_peer_closes_reads_during_end_stream_write`, `..._before_end_stream_write` and `..._without_end_stream_write` cover the three timings: with `MaxBidirectionalStreams = 1`, the client sends its request and completes the output, the server consumes the request and exhausts the client's window, the client closes its reads at the given point, and the test asserts that the client can create a new stream once the server output is completed. Without the fix, all three fail when their `[CancelAfter]` token is canceled (2 min, above CI's `--blame-hang-timeout`).

Full `IceRpc.Tests` suite passes (1051 passed / 0 failed), `dotnet format` clean.

## What's Changed entry

Area: Slic

- Fixed a stream leak that could eventually block all new invocations on a connection when invocations were canceled while the server was still sending their response.

### DIFF
--- a/src/IceRpc/Transports/Slic/Internal/SlicStream.cs
+++ b/src/IceRpc/Transports/Slic/Internal/SlicStream.cs
@@ -55,7 +55,7 @@ internal class SlicStream : IMultiplexedStream
     private readonly SlicConnection _connection;
     private ulong _id = ulong.MaxValue;
     private readonly SlicPipeReader? _inputPipeReader;
-    // This mutex protects _writesClosePending, _closeReadsOnWritesClosure and _wroteLastStreamFrame.
+    // This mutex protects _writesClosePending, _closeReadsOnWritesClosure.
     private readonly Lock _mutex = new();
     private readonly SlicPipeWriter? _outputPipeWriter;
     // FlagEnumExtensions operations are used to update the state. These operations are atomic and don't require mutex
@@ -63,9 +63,6 @@ internal class SlicStream : IMultiplexedStream
     private int _state;
     private readonly TaskCompletionSource _writesClosedTcs = new(TaskCreationOptions.RunContinuationsAsynchronously);
     private bool _writesClosePending;
-    // Set once the StreamLast frame is queued on the connection. Unlike WritesClosed, it is not set when the peer
-    // closes writes.
-    private bool _wroteLastStreamFrame;
 
     internal SlicStream(SlicConnection connection, bool isBidirectional, bool isRemote)
     {
@@ -133,7 +130,6 @@ internal class SlicStream : IMultiplexedStream
     internal void CloseReads(bool graceful)
     {
         bool writeReadsClosedFrame = false;
-        bool deferReadsClosedFrame = false;
 
         lock (_mutex)
         {
@@ -143,39 +139,25 @@ internal class SlicStream : IMultiplexedStream
                 // writes are closed, we don't send the StreamReadsClosed frame just yet. Instead, when writes are
                 // closed, CloseWrites will bundle the sending of the StreamReadsClosed with the sending of the
                 // StreamLast or StreamWritesClosed frame. This allows to send both frames with a single write on the
-                // duplex connection.
+                // duplex connection. If the peer closes its reads first, ReceivedReadsClosedFrame sends the frame on
+                // its own.
                 if (graceful &&
                     IsBidirectional &&
                     IsRemote &&
                     !_state.HasFlag(State.WritesClosed) &&
                     !_writesClosePending)
                 {
-                    deferReadsClosedFrame = true;
+                    // Reads are closed before the deferral is published, both under the mutex, so that the stream
+                    // count is decreased before the StreamReadsClosed frame can be sent and ReceivedReadsClosedFrame
+                    // can't close writes in between.
+                    TrySetReadsClosed();
+                    _closeReadsOnWritesClosure = true;
                 }
                 else if (!graceful || IsRemote)
                 {
                     // If forcefully closed because the input was completed before the data was fully read or if writes
                     // are already closed and the stream is a remote stream, we send the StreamReadsClosed frame to
                     // notify the peer that reads are closed.
-                    writeReadsClosedFrame = true;
-                }
-            }
-        }
-
-        if (deferReadsClosedFrame)
-        {
-            // Reads must be closed before the StreamReadsClosed frame can be sent, by CloseWrites or by a concurrent
-            // StreamLast write, so that the connection's stream count is decreased before the peer receives the frame.
-            TrySetReadsClosed();
-
-            lock (_mutex)
-            {
-                if (!_state.HasFlag(State.WritesClosed) && !_writesClosePending)
-                {
-                    _closeReadsOnWritesClosure = true;
-                }
-                else
-                {
                     writeReadsClosedFrame = true;
                 }
             }
@@ -228,7 +210,9 @@ internal class SlicStream : IMultiplexedStream
         {
             if (IsStarted && !_state.HasFlag(State.WritesClosed) && !_writesClosePending)
             {
+                // The frame sent below can't be canceled, so the deferred StreamReadsClosed frame is claimed here.
                 writeReadsClosedFrame = _closeReadsOnWritesClosure;
+                _closeReadsOnWritesClosure = false;
                 _writesClosePending = true;
                 writeWritesClosedFrame = true;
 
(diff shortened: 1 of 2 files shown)


pepone:
slic-closure-before-after-4947

bernardnormier (commented):
Please rewrite the PR description with the consequence first and the mechanism after, in plain terms: a client that
cancels an invocation while the server is still sending the response can leak a bidirectional stream slot, until
new invocations on the connection block. Then why: the server defers its `StreamReadsClosed` frame to bundle it
with the response's `StreamLast`, and that frame is lost when the client's own `StreamReadsClosed` closes the
server's writes first. The three timings, the earlier rollback and the affected shapes can follow, but the first
paragraph shouldn't name fields.

Have you considered sending the owed frame from `ReceivedReadsClosedFrame` instead? That's the event that strands
the deferred flag, and the fix fits in that one method:

```csharp
internal void ReceivedReadsClosedFrame()
{
    bool writeReadsClosedFrame;
    lock (_mutex)
    {
        // The deferred StreamReadsClosed frame can no longer ride along with the StreamLast frame.
        writeReadsClosedFrame = _closeReadsOnWritesClosure;
        _closeReadsOnWritesClosure = false;
    }

    TrySetWritesClosed();
    _outputPipeWriter?.CompleteWrites(exception: null);

    if (writeReadsClosedFrame)
    {
        try
        {
            WriteStreamFrame(FrameType.StreamReadsClosed, encode: null, writeReadsClosedFrame: false);
        }
        catch (IceRpcException)
        {
            // Ignore connection failures.
        }
    }
}
```

It covers the three timings: with no write or a write attempted after the frame, the flag is still set and the
frame goes out here, after the stream is released; with a parked write, the write is canceled by `CompleteWrites`
and the frame goes out here too, and if the write wins the race and gets queued the peer receives the frame twice,
which is harmless (a released stream is dropped by the connection, a live one handles it idempotently); with the
`StreamLast` frame already queued, the stream is released and the handler doesn't run. A concurrent `CloseWrites`
either captures the flag first and bundles it, or finds it cleared and sends a bare `StreamLast`. The frame write
is fire-and-forget, as the pong frame is from the same loop. `_wroteLastStreamFrame`, the mutex in
`WroteLastStreamFrame`, the second lock in `CloseReads` and the new branch in `CloseWrites` then go away; the
test stays as is.

The ordering fix in `CloseReads` is a separate, one-line change: close reads inside the lock before setting the
flag.

pepone:
🤖 **Claude:** Done in 35f2c7348.

- PR description rewritten with the consequence first; the mechanism, the three timings, the #4765 rollback and the affected shapes follow.
- Fix moved to `ReceivedReadsClosedFrame` as sketched: it captures and clears the deferral and sends the frame on its own after `CompleteWrites`. `_wroteLastStreamFrame`, the mutex in `WroteLastStreamFrame`, the second lock in `CloseReads` and the fallback branch in `CloseWrites` are gone; the test is unchanged.
- `CloseReads` closes reads inside the lock before setting the flag.

Two deviations from the sketch, both for ordering with the mutex `CloseReads` and `CloseWrites` already use:

- `TrySetWritesClosed()` runs inside the `ReceivedReadsClosedFrame` lock rather than after the capture. Outside it, the handler can capture the flag while it is still false, `CloseReads` then sees writes open and publishes the deferral, the handler closes writes without sending, the later endStream write fails on the canceled token (no rollback, the TCS is completed) and `CloseWrites` is blocked by `WritesClosed`: the frame is lost again.
- `CloseWrites` clears the flag when it captures it (its frame write is fire-and-forget and never canceled, so no duplicate is needed) and, for a remote stream, calls `TrySetWritesClosed()` inside its lock. Outside it, the handler can run between `CloseWrites` setting the writes-closed bit and `ReleaseStream`, send the frame, and the peer can start a replacement stream before this side's count is decremented.

`ReleaseStream` takes no connection lock and the TCS/semaphore continuations run asynchronously, so nothing waits under the mutex.

rect41

bernardnormier (commented):
Please don't call `TrySetReadsClosed` or `TrySetWritesClosed` inside the mutex. Nothing on main does, and
`TrySetState` runs `ReleaseStream` and completes the TCS, which makes the locking much harder to follow. The
ordering you want holds without it, because the state bits are monotonic:

- `ReceivedReadsClosedFrame`: call `TrySetWritesClosed()` before the lock, then capture and clear the flag under it.
  A `CloseReads` that ran before left the flag for the handler; one that runs after sees writes closed and sends
  the frame itself. The TCS is completed before the capture, so the rollback in the last-frame write can't undo it.
- `CloseWrites`: keep `TrySetWritesClosed()` where it was, outside the lock. Clearing the flag under the lock is
  what keeps the handler from sending after this capture; the frame write follows the closure on this thread.
- `CloseReads`: drop the added `TrySetReadsClosed()`. The gap between publishing the deferral and closing reads
  exists on main, is a few instructions wide, and isn't part of this leak. Leave it as it was.

On the test: the `[Values("during-write", "before-write", "without-write")]` parameter selects a code path, and the
body branches on it twice. Our other string parameters are data (a protocol name, an address) with one body for
every value, and the testing book we follow says a test is a linear sequence with no if statements. Please make
it three tests, one per timing, each linear: `..._during_end_stream_write`, `..._before_end_stream_write` and
`..._without_end_stream_write`. The shared arrange (connection with one bidirectional stream and a small window,
client request completed, server consumes it and exhausts the client's window) and the shared act and assert can
go to a private helper.

Two more on the test, at the assert:

- The comment describes the bug ("if the deferred StreamReadsClosed frame was lost this would wait forever"). The
  test verifies that the client stream is released, so say that: "Assert: the client stream is released, so a new
  stream can be created."
- Rather than `.AsTask().WaitAsync(TimeSpan.FromMinutes(2))`, put `[CancelAfter]` on the test, take a
  `CancellationToken cancellationToken` parameter and pass it to `CreateStreamAsync` instead of `default`. Same
  fallback, no wrapper, and the other awaits in the test can take the token too. The six `WaitAsync` fallbacks
  from #4765 and #4766 can follow in a separate PR.

pepone:
🤖 **Claude:** Done in b475849c2.

- `ReceivedReadsClosedFrame` calls `TrySetWritesClosed()` before the lock and only captures and clears the flag under it. `CloseWrites` calls `TrySetWritesClosed()` outside the lock again, as on main, and `CloseReads` no longer calls `TrySetReadsClosed()` under it, so nothing calls `TrySetState` under the mutex. Both races from my last reply came from the handler closing writes after its capture; closing writes first rules them out.
- The test is split into `..._during_end_stream_write`, `..._before_end_stream_write` and `..._without_end_stream_write`, each linear. `CreateSingleBidirectionalStreamServiceProvider` and `CreateStreamWithDeferredReadsClosedFrameAsync` hold the shared arrange, and `CompleteServerOutputAndAssertClientStreamReleasedAsync` the shared act and assert.
- The assert comment reads "Assert: the client stream is released, so a new stream can be created." `[CancelAfter(120_000)]` and a `CancellationToken` parameter replace the `WaitAsync` wrapper, and every await in the tests takes the token (`WritesClosed` through `WaitAsync(cancellationToken)`). Without the fix, all three tests fail when the token is canceled.

The PR description is updated to match. I've left the six `WaitAsync` fallbacks from #4765 and #4766 for a separate PR.

bernardnormier (approved):
Looks good

PR #4951, opened 2026-09-11
Pin the whitespace action to a commit

pepone:
`whitespace.yml` used the mutable `zeroc-ice/github-actions@main` while every other external action in the workflows is pinned to a commit. Pin it to the current head of `main`; the repository has no tags.

Addresses L-28 in #4833, split out of #4946.

## What's Changed entry

None — CI workflow change only.


externl:
I actually think it's fine to leave this one on main. It's an action that we own and control.

pepone:
> I actually think it's fine to leave this one on main. It's an action that we own and control.

The downside to use main is that any update automatically used by all repositories, let's keep main for now.

PR #4957, opened 2026-09-14
Apply the Protobuf targets cleanups to the Slice targets

### LINKED ISSUE #4948: Apply the Protobuf targets cleanups from #4923 to IceRpc.Slice.Tools.targets
Follow-up suggested in the review of #4930: `IceRpc.Slice.Tools.targets` should take the cleanups made to the Protobuf targets in #4923. The two files mirror each other, and the Protobuf side now has:

- `_SlicecPruneStaleOutputs` with `BeforeTargets="Slicec;SlicecClean"` only. A `BeforeTargets` target runs even when its target's `Condition` is false, and `SlicecClean` has no condition, so `CoreCompile` and `Clean` in the list are redundant.
- `<Compile Remove="@(_StaleSlicecOutput)" MatchOnMetadata="FullPath" MatchOnMetadataOptions="PathLike" />` instead of the `_StaleSlicecOutputRelative` item, its comment and the `MakeRelative` call. `MatchOnMetadata` matches the SDK glob's relative identities directly. The unquoted `MakeRelative(..., %(Identity))` also splits on a comma in the path and fails with MSB4186.
- No `MakeDir` in `_SlicecWriteOutputManifest`: `WriteLinesToFile` creates the directory, and `PrepareForBuild` already created it.
- No `Exists` guard on the manifest `Delete` in `SlicecClean`: `Delete` skips missing files.
- `WriteOnlyWhenDifferent="true"` on the manifest `WriteLinesToFile`.

Behavior to keep, verified on the Protobuf side: renaming a Slice file removes the glob's `Compile` entry and deletes the old outputs; removing the last Slice file still prunes and deletes the manifest; `Rebuild` and `Clean` unchanged.

pepone:
Fixes #4948

Applies the cleanups #4923 made to the Protobuf targets to `IceRpc.Slice.Tools.targets`, which mirrors them:

- `_SlicecPruneStaleOutputs` runs `BeforeTargets="Slicec;SlicecClean"`. A `BeforeTargets` target runs even when its
  target's `Condition` is false, and `SlicecClean` has no condition, so `CoreCompile` and `Clean` were redundant. The
  comment that gave them as the reason is replaced.
- The stale outputs are removed from `Compile` with `MatchOnMetadata="FullPath"`, which matches the SDK glob's
  project-relative identities directly, so the `_StaleSlicecOutputRelative` item, its comment and the `MakeRelative`
  call are gone. The unquoted `MakeRelative(..., %(Identity))` split on a comma in the path and failed with MSB4186.
  As the manifest was then never rewritten, every later build, `Clean` included, failed the same way until the
  manifest was deleted by hand.
- `_SlicecWriteOutputManifest` no longer runs `MakeDir`: `WriteLinesToFile` creates the directory, and
  `PrepareForBuild` already created it. It writes the manifest with `WriteOnlyWhenDifferent`.
- `SlicecClean` drops the `Exists` guard on the manifest `Delete`, which skips missing files.

### Verification

Three-file probe importing the source-build props/targets (Windows 11, .NET SDK 10.0.401). `a.slice` and `b.slice`
generate into `generated/`, which the SDK glob picks up with relative identities; `c.slice` generates outside the
project directory, picked up by an explicit `Compile` glob with absolute identities.

| Scenario | Before | After |
|---|---|---|
| Fresh build | outputs and manifest written | same |
| Unchanged build | manifest rewritten | manifest untouched |
| Rename `a.slice` to `alpha.slice` | `a.*` deleted, build succeeds | same |
| Remove `c.slice` (absolute `Compile` entry) | `c.*` deleted, build succeeds | same |
| Remove the last Slice files | outputs and manifest deleted | same |
| Then create `generated/alpha.cs` by hand, build | kept | kept |
| Add the Slice files back / Rebuild / Clean / Clean again | recorded / recorded / outputs and manifest deleted / succeeds | same |
| Remove all Slice files, Clean | outputs and manifest deleted | same |
| Project directory with a comma: rename `a.slice`, then every step above | MSB4186 on each build and Clean | same as without a comma |

`IceRpc.Slice.Generator.Tests` builds and passes (214 tests), and a second build leaves its manifest untouched.

## What's Changed entry

Area: Slice tools

- Renaming or removing a Slice file in a project whose path contains a comma no longer makes every later build fail
  with error MSB4186.


bernardnormier (approved):
Looks good.

PR #4958, opened 2026-09-14
Keep the Slic stream-count permit until the peer consumed the last stream frame

### LINKED ISSUE #4788: [Audit-Medium] Slic local stream releases its stream-count permit before the peer has consumed the data

pepone:
A burst of one-way requests on a Slic connection could make the server abort the connection with "The maximum unidirectional stream count N was reached", which lost every request still in flight on it. The client's own stream limit exists to prevent this and was not enforced: the client released a one-way stream's slot as soon as it had sent the request, before the server consumed it, so it could open more streams than the server allows.

The cause is on the sending side of the stream. When the last stream frame is sent by a write with endStream, as the icerpc protocol connection does when it copies a request or response payload, completing the output afterwards closed writes at once, in the branch of `CloseWrites` that has no frame to send. The stream was released, and its permit reused, before the peer's `StreamReadsClosed` frame acknowledged that the data was consumed. When the last frame is sent by the output completion instead, `CloseWrites` already waits for that frame before it closes a local stream's writes, so the two paths disagreed on when a local stream is released.

A unidirectional stream, whose reads are closed at construction, was released as soon as its output was completed. A bidirectional stream was released when the response's `StreamLast` frame was read, before the peer's `StreamReadsClosed` frame; since a server usually consumes the request before it responds, the count mismatch there needs a server that responds before it consumed the whole request.

### Fix

The no-frame branch of `CloseWrites` closes writes only for unstarted streams. A started remote stream reaches it only with writes already closed. A started local stream that queued its `StreamLast` frame closes writes when the peer's `StreamReadsClosed` frame is received, as the `StreamLast` frame written by `CloseWrites` itself already does.

This builds on #4947, now merged: without that fix, a server stream whose client closed reads first never sent its deferred `StreamReadsClosed` frame, and with this change the client stream would wait for it forever. That is the shape #4947's description notes as hidden by the early closure.

### Tests

- `Stream_creation_hangs_until_remote_data_is_consumed_after_end_stream_write` (conformance, bidirectional and unidirectional): the endStream variant of `Stream_creation_hangs_until_remote_data_is_consumed`. With a single allowed stream, the client writes its last frame with endStream and completes its output; a second `CreateStreamAsync` must not complete while the server hasn't consumed the data. It runs on Slic and QUIC. Without the fix, both cases fail on Slic.
- `Deferred_reads_closed_frame_is_sent_when_peer_closes_reads_after_its_end_stream_write`: the fourth timing for #4947, where the client's `StreamLast` frame is sent by an endStream write and the client closes its reads before the server closes its writes. It guards the dependency on #4947: without that fix, it fails when its `[CancelAfter]` token is canceled.

Slic, multiplexed conformance and protocol connection tests pass, `dotnet format` clean.

Fixes #4788.

## What's Changed entry

Area: Slic

- Fixed a bug where a burst of one-way requests could exceed the server's maximum unidirectional stream count and abort the connection. When this limit is reached, the client now waits for the server to consume an earlier request before sending the next one.


src/IceRpc/Transports/Slic/Internal/SlicStream.cs
```diff
@@ -265,7 +265,12 @@ internal void CloseWrites(bool graceful)
...
         else
         {
-            TrySetWritesClosed();
+            // A started local stream either already closed writes or queued a StreamLast frame; in the latter case,
+            // writes are closed once the peer's StreamReadsClosed frame is received (see WroteLastStreamFrame).
+            if (!IsStarted || IsRemote)
+            {
+                TrySetWritesClosed();
+            }
         }
```
bernardnormier: The `IsRemote` term is dead: a started remote stream reaches this branch only with writes already closed, since
`WroteLastStreamFrame` closes them when the `StreamLast` frame is queued and a failed endStream write rolls
`_writesClosePending` back. With the comment on the branch that acts, this becomes:

```suggestion
        else if (!IsStarted)
        {
            // The peer knows nothing about an unstarted stream, so its writes are closed right away.
            TrySetWritesClosed();
        }
        // Otherwise, the stream either already closed writes or, for a local stream, queued a StreamLast frame. Once
        // the peer's StreamReadsClosed frame is received (see ReceivedReadsClosedFrame), it closes writes and releases
        // its stream-count permit.
```

Verified with the Slic, multiplexed conformance and icerpc protocol connection tests.

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
```diff
@@ -2186,6 +2186,155 @@ public async Task Deferred_reads_closed_frame_is_sent_when_peer_closes_reads_wit
         await CompleteServerOutputAndAssertClientStreamReleasedAsync(sut, serverStream, cancellationToken);
     }
 
+    /// <summary>Verifies that a local unidirectional stream keeps its stream-count permit when its output is completed
+    /// after a write with endStream, until the peer consumed the data.</summary>
+    [Test]
+    [CancelAfter(120_000)]
+    public async Task Unidirectional_stream_keeps_its_permit_after_end_stream_write_until_the_peer_consumed_the_data(
```
bernardnormier: These two tests are the endStream variant of `Stream_creation_hangs_until_remote_data_is_consumed` in
`MultiplexedConnectionConformanceTests`, which the issue asked for. `ReadOnlySequencePipeWriter` is public and both
transports implement it, so the conformance test covers QUIC too and replaces both tests here. I ran it: it passes on
Slic and QUIC with this fix, and both cases fail on Slic without it.

```csharp
    [Test]
    public async Task Stream_creation_hangs_until_remote_data_is_consumed_after_end_stream_write(
        [Values(true, false)] bool bidirectional)
    {
        // Arrange
        IServiceCollection serviceCollection = CreateServiceCollection();
        serviceCollection.AddOptions<MultiplexedConnectionOptions>().Configure(
            options =>
            {
                options.MaxBidirectionalStreams = 1;
                options.MaxUnidirectionalStreams = 1;
            });

        await using ServiceProvider provider = serviceCollection.BuildServiceProvider(validateScopes: true);
        var sut = provider.GetRequiredService<ClientServerMultiplexedConnection>();
        await sut.AcceptAndConnectAsync();

        IMultiplexedStream clientStream1 = await sut.Client.CreateStreamAsync(bidirectional, default);
        await ((ReadOnlySequencePipeWriter)clientStream1.Output).WriteAsync(
            new ReadOnlySequence<byte>(_oneBytePayload),
            endStream: true,
            default);
        clientStream1.Output.Complete();

        IMultiplexedStream serverStream1 = await sut.Server.AcceptStreamAsync(default);
        if (bidirectional)
        {
            serverStream1.Output.Complete();
        }

        // Act/Assert
        Assert.That(
            async () =>
            {
                using var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(100));
                _ = await sut.Client.CreateStreamAsync(bidirectional, cts.Token);
            },
            Throws.InstanceOf<OperationCanceledException>());
    }
```

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
```diff
@@ -2186,6 +2186,155 @@ public async Task Deferred_reads_closed_frame_is_sent_when_peer_closes_reads_wit
...
+        nextStream.Output.Complete();
+        nextStream.Input.Complete();
+    }
+
+    /// <summary>Verifies that the StreamReadsClosed frame deferred by the graceful closure of reads on a remote
+    /// bidirectional stream is sent when the peer sent its StreamLast frame with an endStream write and closes its
+    /// reads before the remote stream closes its writes, so that the peer's stream is released.</summary>
+    [Test]
+    [CancelAfter(120_000)]
+    public async Task Deferred_reads_closed_frame_is_sent_when_peer_closes_reads_after_its_end_stream_write(
```
bernardnormier, line 2194: This test re-implements `CreateStreamWithDeferredReadsClosedFrameAsync`. A `clientEndStreamWrite` parameter on the
helper gives it the shape of its three siblings:

```csharp
        // Arrange
        await using ServiceProvider provider = CreateSingleBidirectionalStreamServiceProvider();
        var sut = provider.GetRequiredService<ClientServerMultiplexedConnection>();
        (IMultiplexedStream clientStream, IMultiplexedStream serverStream) =
            await CreateStreamWithDeferredReadsClosedFrameAsync(sut, cancellationToken, clientEndStreamWrite: true);

        clientStream.Input.Complete(new OperationCanceledException());
        await serverStream.WritesClosed.WaitAsync(cancellationToken);

        // Act/Assert
        await CompleteServerOutputAndAssertClientStreamReleasedAsync(sut, serverStream, cancellationToken);
```

with, in the helper:

```csharp
        if (clientEndStreamWrite)
        {
            _ = await ((ReadOnlySequencePipeWriter)clientStream.Output).WriteAsync(
                new ReadOnlySequence<byte>(new byte[1]),
                endStream: true,
                cancellationToken);
        }
        else
        {
            _ = await clientStream.Output.WriteAsync(new byte[1], cancellationToken);
        }
        clientStream.Output.Complete();
```

The helper's window-exhausting write is harmless here. Verified: the four deferred-frame tests pass with this shape.

bernardnormier (commented):
The fix is right. For a started local stream, the no-frame branch of `CloseWrites` is reached only when writes are
already closed or an endStream write queued the `StreamLast` frame, and in the second case `ReceivedReadsClosedFrame`
closes writes, as it already does for the `StreamLast` frame written by `CloseWrites`. A failed endStream write rolls
`_writesClosePending` back, so the output completion still takes the frame branch.

Two changes on the tests, inline. Please also copy the `slic` label and the 0.6.1 milestone from #4788.

The What's Changed entry describes the mechanism and leaves out the behavior a user sees: one-way invocations now
wait once the server's limit is reached. Suggested wording:

> Fixed a bug where a burst of one-way requests could exceed the server's maximum unidirectional stream count and
> abort the connection. When this limit is reached, the client now waits for the server to consume an earlier request
> before sending the next one.

One consequence to decide on, not a change request: a client with this fix now depends on the server's
`StreamReadsClosed` frame in the shape #4947 fixed. Against a server without #4947, an invocation canceled while the
server is sending its response leaks a bidirectional permit on the client, which the early closure used to hide.
That may deserve a line in the 0.6.1 release notes.

src/IceRpc/Transports/Slic/Internal/SlicStream.cs
pepone: Done in 5ec52a560, with your suggested branch and comment.

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
pepone: Done in 5ec52a560: both Slic tests are replaced by `Stream_creation_hangs_until_remote_data_is_consumed_after_end_stream_write` in `MultiplexedConnectionConformanceTests`. I confirmed both cases fail on Slic without the fix. QUIC fails here with a local TLS error on every test, so CI is the QUIC check.

tests/IceRpc.Tests/Transports/Slic/SlicTransportTests.cs
pepone: Done in 5ec52a560: `CreateStreamWithDeferredReadsClosedFrameAsync` takes `clientEndStreamWrite`, and the test uses your shape.

pepone:
Thanks. I added the `slic` label and the 0.6.1 milestone, and used your What's Changed wording in the description. I'm leaving the older-server case out of the 0.6.1 release notes.

bernardnormier (approved):
Looks good.

PR #4942, opened 2026-09-11
Pack the IceRpc.Protobuf.Tools task assembly from bin instead of obj

pepone:
Fixes #4814

The release workflow signs the assemblies matching `src/<Project>/bin/Release/net*/<Project>.dll` and then runs `dotnet pack --no-build`. `IceRpc.Protobuf.Tools` packed its MSBuild task assembly from `$(IntermediateOutputPath)`, so `tasks/IceRpc.Protobuf.Tools.dll` in the published package was the unsigned obj copy. This packs it from `$(OutputPath)` instead, as `IceRpc.Slice.Tools` already does.

Verified locally: after a Release build, appending bytes to the bin copy (to stand in for signing) and running `dotnet pack --no-build` produced a package whose `tasks/` dll matched the obj copy before this change and the bin copy after it.

The unsigned first-party dependency copies in the Tools packages' `tools/` folder are tracked in #4949.

## What's Changed entry

None — packaging fix, the task assembly is now signed like the other assemblies in the package.

Issue #4814, opened 2026-07-28
[Audit-Medium] Stable release signing omits the packed MSBuild task assembly


pepone:
> **AI-generated audit finding** — from a repository-wide audit performed by OpenAI Codex (`codex-cli`), independently verified against source by Claude (Anthropic), model **Opus 5** (`claude-opus-5`). It **has not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

The signing catalog collects, per project directory, files matching `*\bin\Release\net*\$name.dll`. But `IceRpc.Protobuf.Tools` packs its MSBuild task assembly from `obj`, not `bin`:

```xml
<None Include="$(IntermediateOutputPath)/$(AssemblyName).dll" Pack="true" PackagePath="tasks/" Visible="false" />
```

The signed `bin` copy and the packed `obj` copy are different files, so the assembly shipped in the package is **unsigned**.

## Correction to the finding as originally written

The audit also claimed the Tools packages' copied internal dependencies are unsigned. Verification shows they are **not** affected: `IceRpc.Protobuf.Generator.dll` and `IceRpc.Protobuf.BuildTelemetry.dll` each live under their own project directory and match `$name.dll`, so the catalog does pick them up. The confirmed hole is the `obj`-packed task assembly.

## Source references

- [.github/workflows/build-packages.yml#L46-L56](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/.github/workflows/build-packages.yml#L46) — the signing catalog glob
- [src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj#L57](https://github.com/icerpc/icerpc-csharp/blob/2de29310a82428560125c5f804127ea054f46769/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj#L57) — the `obj`-sourced pack item

## Recommendation

Produce a final package-content manifest before signing, sign every first-party managed binary from the exact location that will be packed, and verify package contents before publication.

## Audit context

M-14 of [repository-wide-audit-2026-07-27.md](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-2026-07-27.md); confirmed and **narrowed** by the [Claude verification pass](https://github.com/icerpc/icerpc-csharp-audit/blob/main/repository-wide-audit-claude-review-2026-07-28.md).

Severity: **Medium**.

PR #4959, opened 2026-09-14
Sign the first-party dependency copies packed into the Tools packages

### LINKED ISSUE #4949: [Audit-Medium] Stable release signing omits the first-party dependency copies packed into the Tools packages
> Found while fixing #4814 (PR #4942). The gap was spotted by Claude (Anthropic, Fable 5.1) with an independent OpenAI Codex cross-check, then verified by emulating the signing catalog's glob against a local Release build. It has not been triaged by a human for severity or priority; the severity below mirrors #4814, which is the same class of problem.

## Summary

The stable release workflow signs, for each project directory under `src`, only the assembly named after that directory (`*\bin\Release\net*\<Project>.dll`), and then runs `dotnet pack --no-build`:

- [.github/workflows/build-packages.yml#L46-L56](https://github.com/icerpc/icerpc-csharp/blob/ce468b3e5b5a7e2b3d88c52710f371c226c502bd/.github/workflows/build-packages.yml#L46-L56)

`IceRpc.Protobuf.Tools` and `IceRpc.Slice.Tools` build their `tools/` payload by globbing the generator and build-telemetry projects' `bin/Release/net10.0/*` folders:

- [src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj#L65-L70](https://github.com/icerpc/icerpc-csharp/blob/ce468b3e5b5a7e2b3d88c52710f371c226c502bd/src/IceRpc.Protobuf.Tools/IceRpc.Protobuf.Tools.csproj#L65-L70)
- [src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.csproj#L80-L89](https://github.com/icerpc/icerpc-csharp/blob/ce468b3e5b5a7e2b3d88c52710f371c226c502bd/src/IceRpc.Slice.Tools/IceRpc.Slice.Tools.csproj#L80-L89)

Those folders hold copy-local copies of first-party assemblies that belong to other projects, and the catalog never lists such copies. Each published Tools package therefore ships unsigned copies of:

| Package | Unsigned first-party copies in `tools/` | Copied from |
|---|---|---|
| `IceRpc.Protobuf.Tools` | `IceRpc.dll`, `IceRpc.Protobuf.dll`, `ZeroC.Slice.Codec.dll` | `IceRpc.Protobuf.BuildTelemetry/bin` |
| `IceRpc.Slice.Tools` | `IceRpc.dll`, `IceRpc.Slice.dll`, `ZeroC.Slice.Codec.dll`, `ZeroC.Slice.Symbols.dll` | `ZeroC.Slice.Generator/bin`, `IceRpc.Slice.Generator/bin`, `IceRpc.Slice.BuildTelemetry/bin` |

The generator and build-telemetry assemblies themselves (`IceRpc.Protobuf.Generator.dll`, `IceRpc.Slice.BuildTelemetry.dll`, ...) are signed: they live under their own project directory and match the catalog glob. The same assemblies shipped in their own packages (`IceRpc.dll` in the `IceRpc` package, and so on) are signed too. Only the copies packed into `tools/` are not.

#4814 narrowed the original finding to the MSBuild task assembly and stated that the dependency copies were not affected. That statement was wrong; #4942 fixes the task assembly only.

The same globs exist on `0.6.x`.

## Reproduction

After `dotnet build --configuration Release`, emulate the catalog:

```sh
for d in src/*/; do n=$(basename "$d"); find "$d" -path "*/bin/Release/net*/$n.dll"; done
```

and list the first-party assemblies in the globbed bin folders:

```sh
ls src/IceRpc.Protobuf.BuildTelemetry/bin/Release/net10.0 src/IceRpc.Slice.BuildTelemetry/bin/Release/net10.0 \
   src/ZeroC.Slice.Generator/bin/Release/net10.0 src/IceRpc.Slice.Generator/bin/Release/net10.0 | grep -E '^(IceRpc|ZeroC)'
```

The catalog lists each first-party assembly once, at its own project's `bin`. The copies in the four folders above are absent from it, and they are the files `dotnet pack` puts into `tools/`.

## Possible fixes

- Widen the catalog to every first-party assembly under `src/*/bin/Release/net*/` (`IceRpc*.dll` and `ZeroC.*.dll`), so the copy-local copies are signed in place. This is the smallest change.
- Alternatively, stop globbing bin folders for `tools/` and pack the payload from an explicit list of signed paths, which also stops third-party copies from drifting in unnoticed.

Severity: **Medium**.

pepone:
Fixes #4949

The release workflow signed, for each project under `src`, only the assembly named after the project in that project's own `bin` folder. `IceRpc.Protobuf.Tools` and `IceRpc.Slice.Tools` pack the `bin` folders of their generators and build telemetry plug-ins into `tools/`, and those folders also hold copy-local copies of other first-party assemblies. The catalog never listed these copies, so the published Tools packages shipped them unsigned. The 0.6.0 packages on nuget.org have 7 of them (`Get-AuthenticodeSignature` reports `NotSigned`):

- `IceRpc.Slice.Tools`: `tools/IceRpc.dll`, `tools/IceRpc.Slice.dll`, `tools/ZeroC.Slice.Codec.dll`, `tools/ZeroC.Slice.Symbols.dll`
- `IceRpc.Protobuf.Tools`: `tools/IceRpc.dll`, `tools/IceRpc.Protobuf.dll`, `tools/ZeroC.Slice.Codec.dll`

This PR searches all of `src` for each project's assembly, so every copy under a `bin/Release/net*` folder is signed where it is. That is the smaller of the two fixes proposed in the issue, and it also covers any copies the Tools payloads pick up later. The catalog grows from 24 to 67 files. The 36 new entries that no package includes cost some extra signatures. For scale, signing took 51s in the 0.6.0 release job, which finished in 3m20s against a 10-minute timeout.

Verified locally the same way as #4942: after a Release build, I ran each version of the "Collect Files to Sign" script as written in the workflow, appended a marker to every catalog file to stand in for signing, ran `dotnet pack --no-build`, and checked every first-party dll in the resulting packages:

- before: 23 marked, 7 unmarked (the 7 copies above)
- after: 30 marked, 0 unmarked

The "Collect Files to Sign" step is the same on `0.6.x`, so this commit cherry-picks cleanly.

## What's Changed entry

None — release packaging fix: the first-party assemblies bundled in the Tools packages are now signed like the rest of the package.


.github/workflows/build-packages.yml
```diff
@@ -46,10 +46,12 @@ jobs:
...
+            # For each project in src, we sign the main assembly named after the project, in its own bin folder and in
+            # the bin folders of the projects that copy it: the Tools packages pack the bin folders of their generators
+            # and build telemetry plug-ins, copy-local dependencies such as IceRpc.dll included.
             $assemblies = Get-ChildItem -Path src -Directory | ForEach-Object {
               $name = $_.Name
-              Get-ChildItem -Path $_.FullName -Filter "$name.dll" -Recurse |
+              Get-ChildItem -Path src -Filter "$name.dll" -Recurse |
               Where-Object { $_.FullName -like "*\bin\Release\net*\$name.dll" } |
               ForEach-Object { Resolve-Path -Relative $_.FullName }
             }
```
bernardnormier, line 57: With the search rooted at `src`, the outer loop only supplies the name, so `src` is walked once per project. One walk with the names
as a set gives the same catalog (68 entries on my build, no difference with `Compare-Object`) in a fraction of the
time.

```suggestion
            # Sign each project's assembly in every bin folder where it's found, not just its own. This is needed
            # because each Tools package ships the whole bin folder of its generator and build telemetry projects, and
            # these folders contain copies of IceRpc.dll and other first-party assemblies.
            $names = (Get-ChildItem -Path src -Directory).Name
            $assemblies = Get-ChildItem -Path src -Filter *.dll -Recurse -File |
              Where-Object { $names -contains $_.BaseName -and $_.DirectoryName -like "*\bin\Release\net*" } |
              ForEach-Object { Resolve-Path -Relative $_.FullName }
```

bernardnormier (commented):
The fix is right. I reproduced the catalog on a local Release build: the wider search adds exactly 43 entries, all
copy-local copies of first-party assemblies, including the 7 that `dotnet pack` puts into the Tools packages, and it
picks up no reference assembly.

Please drop the `ai-audit` label: it stays on the issue, not on the PR.

Two things this PR doesn't cover, for separate issues:

- `slicec.exe` in `IceRpc.Slice.Tools` is unsigned too: it has no certificate table, and the slicec release workflow
  has no signing step.
- A post-pack step that fails the job when a first-party binary in a produced package isn't signed would have caught
  both #4814 and #4949.

pepone:
for `slicec.exe` I addressed that in the repository that produces it, see https://github.com/icerpc/slicec/pull/825

.github/workflows/build-packages.yml
pepone: Applied in e51715089, thanks.

bernardnormier (approved):

PR #4963, opened 2026-09-15
Fix verbatim identifiers in generated service dispatchers

### LINKED ISSUE #4939: Rosyln Source Generators Can't Properly Handle Verbatim Identifiers
Found while thinking about #4934...

In general, we allow users to use verbatim identifiers (via `cs::identifier`) as an escape hatch when the default mapping for a Slice identifier would be reserved in C#. For example, if they have an operation with a parameter named `event` (which is a keyword in C#), they could avoid needing to alter their Slice definition by annotating the parameter with `[cs::identifier("@event")]`. This is well-supported (modulo a couple bugs) and a documented use-case.

However this setup breaks down when a user goes to actually implement the service in the normal way (by writing a `partial class` and marking it with `[Service]`). This uses a Rosyln source generator (in this case `IceRpc.ServiceGenerator`), which loads names from the semantic model. This semantic model _does not include '@'_, since these are purely lexical artifacts.

----

For a concrete example, take:
```
    subscribe(user: string, [cs::identifier("@event")] event: string)
```
The Slice compiler for C# gracefully handles the verbatim modifier, and so we will generate this perfectly okay code:
```
internal static ValueTask<(string User, string @event)> DecodeSubscribeAsync(
    CancellationToken cancellationToken) => ...
```
But when a user implements this service using the `[Service]` attribute, which lacks awareness of `@`, it will generate:
```
(args, features, cancellationToken) =>
    ((IProbeDispatchService)this).SubscribeAsync(args.User, args.event, features, cancellationToken),
```
Note that here `args.event` lacks any escaping, and will fail to compile, since `event` is a keyword.

pepone:
A service marked with `[Service]` can fail to compile when an operation parameter or return value uses a verbatim identifier such as `cs::identifier("@event")`. Roslyn exposes the name as `event`, and the service generator emits invalid accesses such as `args.event`.

`cs::identifier` gives the exact C# identifier and the Slice generator escapes nothing, so `@event` is the only mapping that keeps a Slice parameter or return member named like a C# keyword. The `[Service]` generator was the one place that didn't honor it. Supporting it keeps the name as written, in metadata and docs included, and code on an older C# version keeps compiling against the same Slice definitions. The benefit is modest, since callers using named arguments or tuple element names write `@event` too.

Escape reserved keywords when emitting tuple accesses and qualified type and namespace names. Contextual keywords are valid identifiers in every position the generator emits a name, so they stay as is. Keep the generated source file names free of `@`, which Roslyn rejects in source hint names.

Fixes #4939.

## Validation

- Service-generator tests: 12 passed, including 5 new compilation cases covering request/return tuples, streamed returns, and keyword type and namespace names.
- Slice-generator tests: 215 passed, including an end-to-end invocation with escaped parameter and return identifiers.
- `git diff --check` passed.

## What's Changed entry

Area: Slice tools

- Fixed service implementations marked with `[Service]` failing to compile when generated dispatch code references verbatim C# identifiers, including parameters and return values renamed with `cs::identifier`.


tests/IceRpc.ServiceGenerator.Tests/IdentifierTests.cs
```diff
@@ -0,0 +1,214 @@
...
+            "Contracts",
+            [CSharpSyntaxTree.ParseText(definitions)],
+            references,
+            new CSharpCompilationOptions(
+                OutputKind.DynamicallyLinkedLibrary,
+                nullableContextOptions: NullableContextOptions.Enable));
+
+        Assert.That(compilation.GetDiagnostics().Where(d => d.Severity == DiagnosticSeverity.Error), Is.Empty);
+
+        if (referenceAssembly)
```
bernardnormier: A symbol name never carries `@`, whether the symbol comes from source or from metadata, so both values of
`referenceAssembly` reach `GetEscapedName` with the same input. Dropping the axis removes this block, the second
compilation and the parameter on the three tests.

src/IceRpc.ServiceGenerator/Internal/SymbolExtensions.cs
```diff
@@ -21,6 +22,13 @@ internal static class SymbolExtensions
         return null;
     }
 
+    /// <summary>Gets the symbol name escaped for use as a C# identifier.</summary>
+    /// <remarks>Roslyn symbol names omit the <c>@</c> prefix used by verbatim identifiers.</remarks>
+    internal static string GetEscapedName(this ISymbol symbol) =>
+        SyntaxFacts.GetKeywordKind(symbol.Name) != SyntaxKind.None ||
+        SyntaxFacts.GetContextualKeywordKind(symbol.Name) != SyntaxKind.None ?
+            $"@{symbol.Name}" : symbol.Name;
```
bernardnormier: A contextual keyword is a valid identifier in every position the generator uses a name: after a `.` in a member
access or a qualified name. That includes `await`, since the generated `DispatchAsync` and its lambdas are not async.
Only the reserved keywords need the `@`, which is also where Roslyn's own `EscapeKeywordIdentifiers` stops.

```suggestion
    internal static string GetEscapedName(this ISymbol symbol) =>
        SyntaxFacts.GetKeywordKind(symbol.Name) != SyntaxKind.None ? $"@{symbol.Name}" : symbol.Name;
```

src/IceRpc.ServiceGenerator/Internal/ProtobufServiceMethod.cs
```diff
@@ -52,7 +52,7 @@ internal ProtobufServiceMethod(
         OperationName = (string)items[0].Value!;
 
         _interfaceName = method.ContainingType.GetFullName();
-        _methodName = method.Name;
+        _methodName = method.GetEscapedName();
```
bernardnormier: The Protobuf generator names every method `{PascalCase}Async` and Protobuf has no way to rename it, so this name
can't be a keyword; only a hand-written interface can produce one. Please keep `method.Name` and, in
`Qualified_names_and_protobuf_method_names_are_escaped`, give the method an ordinary name and drop the method part
of the test name. The namespace and the nested types are the real case.

```suggestion
        _methodName = method.Name;
```

bernardnormier (commented):
The fix is correct. Every place the generator turns a symbol name into code now goes through the escaping: tuple
accesses, qualified type and namespace names, the Protobuf method name, and the hint name has the `@` stripped. I
checked the generated code for a plain `event` parameter and for `@event` through `cs::identifier`.

The description should say why `@` in `cs::identifier` is worth supporting, since rejecting it would be the smaller
fix. As I see it: `cs::identifier` gives the exact C# identifier, and the generator escapes nothing, so `@event` is
the only mapping that keeps a Slice parameter or return member named like a C# keyword. The `[Service]` generator was
the one place that didn't honor it. The benefit is modest, since callers using named arguments or tuple element names
write `@event` too, but the name stays as it is, in metadata and docs included, and code on an older C# version keeps
compiling against the same Slice definitions.

tests/IceRpc.ServiceGenerator.Tests/IdentifierTests.cs
pepone: Removed the `referenceAssembly` axis from all three tests and simplified the helper to a single compilation.

src/IceRpc.ServiceGenerator/Internal/SymbolExtensions.cs
pepone: Updated `GetEscapedName` to escape only reserved keywords using `SyntaxFacts.GetKeywordKind`.

src/IceRpc.ServiceGenerator/Internal/ProtobufServiceMethod.cs
pepone: Restored `method.Name`, changed the test method to `EchoAsync`, and renamed the test to `Qualified_names_are_escaped`.

bernardnormier (approved):
Looks good.

PR #4964, opened 2026-09-15
Handle verbatim identifiers in generated name collision checks

### LINKED ISSUE #4936: The `*ParamName` methods in `OperationExtensions` are Incomplete
We have a handful of methods which exist to escape certain special identifiers that we use in generated code.
For example, take `FeaturesParamName`:
```
        /// <summary>Gets the escaped name for the injected "features" parameter, appending "_" if any operation
        /// parameter uses the name "features".</summary>
        internal string FeaturesParamName =>
            op.Parameters.Any(p => p.ParameterName == "features") ? "features_" : "features";
```
Since this is a direct match against `ParameterName`, it won't catch _escaped_ field names.
For example, if I had a parameter that I remapped with `cs::identifier("@features")`, this would defeat the check and we would end up with a method with both `features` and `@features`, which isn't allowed.

----

This is a real gap in these functions checks, but it might not be one worth worrying about.
Since on the face of it, there is no good reason to ever remap an identifier to `@features`, since `features` is not reserved in C#. Opening for completeness, but closing as not-planned may very well be the right path here.

pepone:
Renaming a Slice parameter with `cs::identifier("@features")` generates both `@features` and an injected `features` parameter, which C# rejects as duplicate names. The same problem affects `cancellationToken`, request and response `encodeOptions`, and the synthetic `Payload` field of an encoded return with a stream.

Ignore the leading `@` when comparing names in these five collision checks. Keep emitting the user-specified identifier exactly and append the existing `_` suffix to the injected name when a collision is found.

Fixes #4936.

## Validation

- Added five end-to-end regression tests covering each collision check; the generated declarations failed to compile before the fix.
- All 219 tests in `IceRpc.Slice.Generator.Tests` pass.
- `git diff --check` passes.

## What's Changed entry

Area: Slice codec

- Fixed duplicate names in generated C# when `cs::identifier` uses verbatim spellings of injected parameter names (`@features`, `@cancellationToken`, or `@encodeOptions`) or the `@Payload` streamed return field with `cs::encodedReturn`.


InsertCreativityHere (approved):
I agree with the fix, but I don't think it's important enough to warrant 141 lines of additional tests.

Note that we already have tests for these special names without a '@', (in `OperationTests.slice` (which we should consider moving these tests too as well)), and for those, we do a compile-only test, making sure our service implementation compiles.

I think that would be sufficient here, and I don't see the value in actually calling these operations.

bernardnormier (commented):
I don't think we should fix this. None of `features`, `cancellationToken`, `encodeOptions` and `Payload` is a C#
keyword or is likely to become one, and `@features` is the same identifier as `features`, so there is no reason to ever write `cs::identifier("@features")`. This is a made-up scenario, and #4936 itself suggests closing it as not planned. I'd close both.

PR #4965, opened 2026-09-15
Verify Authenticode signatures in packed NuGet binaries

### LINKED ISSUE #4961: Fail the release packaging job when a first-party binary in a package is unsigned
`build-packages.yml` decides what to sign before packing: "Collect Files to Sign" builds a catalog from the `bin`
folders, and `dotnet pack --no-build` then packs whatever it finds. Nothing checks that the catalog covers every
first-party binary a package includes, so a package that ships a file the catalog missed is published unsigned, and
nobody notices until someone runs `Get-AuthenticodeSignature` on the nuget.org package. This has happened twice:

- #4814: the MSBuild task assembly packed into `tasks/` was unsigned.
- #4949: the copies of `IceRpc.dll` and other first-party assemblies packed into the Tools packages' `tools/` were
  unsigned.

## Proposal

When `authenticode_sign` is set, add a step after "Pack IceRPC" (and after "Pack IceRPC Templates", if the templates
package ever includes binaries) that:

1. extracts each produced `.nupkg`,
2. finds every first-party binary in it, and
3. fails the job if `Get-AuthenticodeSignature` doesn't report `Valid` for any of them, listing the files.

Third-party files keep their own signatures (or none), so the step must tell first-party binaries apart. For example,
the Tools packages include `protoc`, `Google.Protobuf.dll`, and MSBuild dependencies. One option is to use the same
rule as the catalog: a `.dll` whose base name is a project name under `src`. That rule can't catch a binary the catalog
misses because of its name, though, so it might be better to check every `.dll` and `.exe` in the packages against an
explicit list of third-party files. Once icerpc/slicec#825 is released, `slicec.exe` in `IceRpc.Slice.Tools` can be
checked too.

Suggested by @bernardnormier in the review of #4959.

pepone:
The release workflow can publish unsigned first-party binaries when the signing catalog misses a file that `dotnet pack` includes, such as an MSBuild task assembly or a dependency copied into a Tools package.

When `authenticode_sign` is enabled, inspect every produced `.nupkg` after both packing steps and require a valid Authenticode signature on every DLL and EXE except an explicit list of bundled third-party files. Temporarily exempt `slicec.exe` until the bundled compiler includes icerpc/slicec#825. Report each invalid signature with its package and entry path, then fail before artifact upload.

Fixes #4961.

Validation: six temporary PowerShell fixture tests passed, covering valid binaries, multiple invalid signatures across packages, exceptions, exception path matching, missing packages, and corrupt archives. Workflow placement, the signing condition, extraction cleanup, and `git diff --check` also passed. Signature results were simulated on Linux; real Authenticode validation still requires Windows CI.

## What's Changed entry

None — this adds a release packaging check with no user-facing API or runtime change.


.github/workflows/build-packages.yml
```diff
@@ -95,6 +95,61 @@ jobs:
         working-directory: src/IceRpc.Templates
         run: dotnet pack --configuration Release --output ../../
 
+      - name: Verify packaged Authenticode signatures
+        if: github.repository == 'icerpc/icerpc-csharp' && inputs.authenticode_sign
+        shell: pwsh
+        run: |
+          $exceptions = @(
+            # Third-party binaries retain their upstream signatures (or lack of signatures).
```
bernardnormier: I would prefer to list "known exceptions" and check that all other binaries have a signature.

For example, if we expect Google.Protobuf.dll to have a signature, and it looses this signature after some update, the job will fail, which is fine.

.github/workflows/build-packages.yml
```diff
@@ -95,6 +95,74 @@ jobs:
...
+            # Remove when the bundled slicec includes https://github.com/icerpc/slicec/issues/825.
+            'tools/windows-x64/slicec.exe'
+          )
+          # Match publisher subjects so certificate rotation does not require updating thumbprints.
+          $zeroCSubject = 'CN=ZeroC, O=ZeroC, L=Jupiter, S=Florida, C=US'
+          $microsoftSubject = 'CN=.NET, O=Microsoft Corporation, L=Redmond, S=Washington, C=US'
+          $thirdPartySubjects = @{
+            'tools/Microsoft.Extensions.DependencyInjection.Abstractions.dll' = $microsoftSubject
+            'tools/Microsoft.Extensions.Logging.Abstractions.dll' = $microsoftSubject
+          }
```
bernardnormier: The goal here is to check we didn't forget to sign anything, or that an unsigned DLL or EXE didn't somehow get
through. Who signed a binary, and how, doesn't matter for that, so a simpler check that everything outside the
exceptions is Authenticode-signed is enough.

```suggestion
```

.github/workflows/build-packages.yml
```diff
@@ -95,6 +95,74 @@ jobs:
...
+                Where-Object { $_.Extension -in '.dll', '.exe' }
+              foreach ($binary in $binaries) {
+                $relativePath = [System.IO.Path]::GetRelativePath($extractPath, $binary.FullName).Replace('\', '/')
+                if ($relativePath -in $exceptions) {
+                  continue
+                }
+                $expectedSubject = $zeroCSubject
+                if ($thirdPartySubjects.ContainsKey($relativePath)) {
+                  $expectedSubject = $thirdPartySubjects[$relativePath]
+                }
```
bernardnormier:
```suggestion
```

.github/workflows/build-packages.yml
```diff
@@ -95,6 +95,74 @@ jobs:
...
+                $signature = Get-AuthenticodeSignature -LiteralPath $binary.FullName
+                if ($signature.Status -ne 'Valid') {
+                  $failures.Add("$($package.Name): $relativePath ($($signature.Status))")
+                } elseif ($signature.SignerCertificate.Subject -ne $expectedSubject) {
+                  $failures.Add(
+                    "$($package.Name): $relativePath (unexpected signer '$($signature.SignerCertificate.Subject)'; " +
+                    "expected '$expectedSubject')")
```
bernardnormier:
```suggestion
```

.github/workflows/build-packages.yml
```diff
@@ -95,6 +95,74 @@ jobs:
...
+              if (Test-Path -LiteralPath $extractPath) {
+                Remove-Item -LiteralPath $extractPath -Recurse -Force
+              }
+            }
+          }
+          if ($failures.Count -gt 0) {
+            foreach ($failure in $failures) {
+              Write-Host "::error::$failure"
+            }
+            throw "Found $($failures.Count) packaged binaries without valid, timestamped Authenticode signatures from the expected publisher."
```
bernardnormier:
```suggestion
            throw "Found $($failures.Count) packaged binaries without valid, timestamped Authenticode signatures."
```

pepone:
I don't see what is the point of checking DLLs we don't build are signed, if we don't check they are signed by the expected author.

bernardnormier:
The check isn't about authenticity, it's about whether the package works. In an App Control or AppLocker environment
an unsigned DLL doesn't load, whoever authored it, so an unsigned third-party binary in our package breaks our own
DLLs and tools there just as much as an unsigned IceRpc.dll would. That's a gap we'd rather learn about at release
time than from a blocked build. Who signed a binary is a supply-chain question, and pinned versions plus NuGet package
signatures already cover that.

Ideally the exception list goes to zero: slicec will be signed, Dunet is going away, and Google signs protoc in
3.35.x and 4.0, so 3.36 looks like a publishing slip on their side.

pepone:
Fair enough, the App Control/AppLocker angle is the convincing part. I dropped the publisher check in 1c8574b86; the step now only requires that every packaged DLL/EXE outside the exception list has a valid, timestamped Authenticode signature.

bernardnormier (approved):
Looks good.

.github/workflows/build-packages.yml
```diff
@@ -23,7 +23,7 @@ on:
         required: true
         default: "nightly"
       authenticode_sign:
-        description: "Sign the assemblies with authenticode"
+        description: "Sign the assemblies with Authenticode (upstream only)"
```
bernardnormier: "Upstream" reads as the dependencies we bundle, which is the opposite of what this means.

```suggestion
        description: "Sign the assemblies with Authenticode (ignored on forks)"
```

.github/workflows/build-packages.yml
```diff
@@ -95,6 +95,59 @@ jobs:
         working-directory: src/IceRpc.Templates
         run: dotnet pack --configuration Release --output ../../
 
+      - name: Verify packaged Authenticode signatures
+        if: github.repository == 'icerpc/icerpc-csharp' && inputs.authenticode_sign
+        shell: pwsh
+        run: |
+          $exceptions = @(
+            # These upstream binaries are not Authenticode-signed.
```
bernardnormier, line 103:
```suggestion
            # These third-party binaries are not Authenticode-signed.
```

PR #4971, opened 2026-09-17
Move dismissed audit patterns out of AGENTS.md

pepone:
`AGENTS.md` is loaded into the context of every agent task, and the "Dismissed audit patterns" section made up about 130 of its 185 lines. That content only matters when auditing, reviewing, or touching one of the affected areas, and its size diluted the short conventions that apply to every task.

This PR:

- Moves the 15 dismissed audit patterns to a new `DESIGN-DECISIONS.md`, keeping the full rationale and issue references. The entries are reworded as design decisions ("The host is assumed to be little-endian") rather than audit instructions ("Don't file findings about…"), so they apply equally to an agent writing code, and are grouped by area (trust boundaries, code generators, protocols and encodings, runtime conventions, triage policy).
- Replaces the section in `AGENTS.md` with a short pointer that says when to read `DESIGN-DECISIONS.md`, plus one-line summaries of the five decisions most likely to matter during ordinary coding tasks.
- Documents the branch conventions: trunk-based development, release-series branches, and source/binary compatibility on release branches (replacing the "warnings worth flagging" bullet).
- Adds a rule against referencing this repository's issues or PRs in source code and code comments, while allowing a workaround for a bug in another project to link that project's issue, and scopes the existing `#1234`-over-URL rule to PR descriptions, commit messages, and Markdown docs.
- Adds a "Product documentation" section: when preparing a PR, consider whether the change needs new or updated documentation, which calls for a follow-up issue in `icerpc/icerpc-docs`.

Anything outside the repository that refers to the "Dismissed audit patterns" section of `AGENTS.md` should now point to `DESIGN-DECISIONS.md`.

## What's Changed entry

None — guidance for AI coding agents only; no change to the product.


bernardnormier (approved):

AGENTS.md
```diff
@@ -50,135 +57,29 @@ These conventions apply to all AI coding assistants (Copilot, Claude Code, etc.)
...
   with `gh pr edit`.
 
-## Dismissed audit patterns
+## Product documentation
 
-This section captures the reasoning behind `ai-audit` findings that have been closed as "not planned". Before opening a
-new audit finding (or reporting an existing concern), check whether it falls under one of these patterns — the
-rationale has already been considered and rejected. Cite the canonical issue when relevant.
+The IceRPC documentation published at [docs.icerpc.dev](https://docs.icerpc.dev) lives in a separate repository,
+`icerpc/icerpc-docs`. Only the API reference is generated from this repository (from the XML doc comments).
```
InsertCreativityHere: Minor, but in the 2nd sentence "this repository" would mean the repository we just talked about, now the repository this file lives in. Which isn't what we want.

Either swap the order of the two sentences (so "this" means icerpc-csharp), or add an explicit "(`icerpc-icerpc-docs)" after "this repository" in the 2nd sentence.

AGENTS.md
```diff
@@ -50,135 +57,29 @@ These conventions apply to all AI coding assistants (Copilot, Claude Code, etc.)
...
 
-For dismissals newer than this section, also consult:
-`gh issue list --repo icerpc/icerpc-csharp --label ai-audit --state closed --search "reason:not-planned"`.
+- When preparing a PR, consider how the change affects the product documentation: a new feature that needs to be
+  documented, or a change to the behavior, API, or configuration of a feature the documentation already describes.
+  Internal changes (refactoring, tests, build, CI) usually need nothing.
+- A change that affects the documentation requires a follow-up in `icerpc/icerpc-docs`: create the companion PR, or
+  file an issue there describing the pages or topics to add or update. Link the companion PR or issue in the PR
+  description, and link back to this PR from it.
```
InsertCreativityHere: I'm not sure what the right plan is here, but I've told my agents they don't have free reign to open PRs to repositories, to avoid it opening bad PRs, or pushing code before I can review it.

I'm worried that if they read this, it will override my preferences and lead to possibly unnecessary or bad PRs being opened without my control.

AGENTS.md
```diff
@@ -15,8 +15,10 @@ These conventions apply to all AI coding assistants (Copilot, Claude Code, etc.)
 
 ## Comments and documentation
 
-- Prefer `#1234` over full GitHub URLs when referencing issues or PRs in this repo
-  (`icerpc/icerpc-csharp`). Use full URLs only when linking to other repositories.
+- In PR descriptions, commit messages, and Markdown docs, prefer `#1234` over full GitHub URLs when referencing issues
+  or PRs in this repo (`icerpc/icerpc-csharp`). Use full URLs only when linking to other repositories.
+- Don't reference GitHub issues or PRs in source code or code comments (no `#1234`, no URLs). A comment must stand
+  on its own; the issue belongs in the commit message and PR description.
```
InsertCreativityHere: Maybe we've evolved past this, but should we add an exception for comments that are linked to issues? Like:
```
// We need this fix until msquic/msquic/issues/1234 is released,
// after which this should be deleted.
```
for a not-well-formatted example.

InsertCreativityHere (approved):
Looks good to me!

AGENTS.md
pepone: Fair enough, maybe filing the issues is enough here, and who ever picks the issue in docs would create the PR.

externl (approved):

AGENTS.md
pepone: 🤖 **Claude (Opus 5):** Swapped the two sentences so "this repository" comes first and means icerpc-csharp. Done in 327f2963a.

AGENTS.md
pepone: 🤖 **Claude (Opus 5):** Dropped the companion PR. The bullet now says the change needs a follow-up issue in icerpc/icerpc-docs, without telling agents to open or file anything, so each developer's own rules still decide who files it. Done in 327f2963a.

AGENTS.md
pepone: 🤖 **Claude (Opus 5):** Agreed, and the code already does this (dotnet/runtime#82704, microsoft/msquic#3880, …). The rule now covers only this repository's issues and PRs, and allows linking another project's issue from a workaround for its bug. Done in 327f2963a.

PR #4977, opened 2026-09-18
Identify the peer's control stream by its ID during connect

### LINKED ISSUE #4975: icerpc connect fails when the peer's first request stream is accepted before its control stream
`IceRpcProtocolConnection.PerformConnectAsync` takes the first stream accepted from the peer as the peer's control stream ([IceRpcProtocolConnection.cs:171](https://github.com/icerpc/icerpc-csharp/blob/4d1b80632/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L171)). QUIC orders data within a stream, not across streams, so when the peer's first request stream is surfaced first, `ReceiveControlFrameHeaderAsync` reads the request header's leading byte as a control frame type and the server's connect fails:

```
Server connection failed to connect from 'icerpc://[::]?transport=quic' to '13.105.117.110:35642'
IceRpc.IceRpcException: The connection establishment was aborted by an icerpc protocol error.
 ---> System.IO.InvalidDataException: Received frame type 21 but expected Settings.
   at IceRpc.Internal.IceRpcProtocolConnection.ReceiveControlFrameHeaderAsync(IceRpcControlFrameType expectedFrameType, CancellationToken cancellationToken)
   at IceRpc.Internal.IceRpcProtocolConnection.<>c__DisplayClass36_0.<<ConnectAsync>g__PerformConnectAsync|0>d.MoveNext()
```

The exception is thrown on the server, 90 ms after `AcceptConnectionAsync` returned the connection (`0x15` is the leading byte of the request header's size varint). The server then aborts the QUIC connection, and the client's `ConnectAsync` fails with `The connection was aborted by the peer.` The exposure is one-directional: a server opens only its control stream during connect, so a client cannot hit it.

Captured on both ends:

- Client: a GitHub-hosted macOS 26.6.2 runner (behind a NAT pool, `13.105.117.110`/`.111`), .NET 10.0.401, libmsquic 2.6.0, the `IceRpc.Protobuf.BuildTelemetry` generator making a two-way call over QUIC.
- Server: build-telemetry.icerpc.dev, IceRpc 0.6.0.
- The 38-byte packet carrying the client's Settings frame, sent right after its Finished, never reached the server. The retransmitted Settings left in the same 268-byte packet as the first request — a healthy sibling flow's equivalent packet is 262 bytes — and the request stream was surfaced first.
- Frequency: about once in 10,000 QUIC connections from GitHub-hosted runners. The 2026-06-12 CI failure on macOS reported the same client-side error.

**Packet sequence** (runner clock; the server's is about 150 ms ahead)

Client side, failing flow (`192.168.64.14:52092` → `54.235.154.193:4062`, seen by the server as `13.105.117.110:35642`):

```
15:07:00.623  C>S 1220            Initial
15:07:00.774  S>C 1220, 1220, 376 server Initial + Handshake
15:07:00.775  C>S 1220            ACK, padded            — absent from the server capture
15:07:00.795  C>S 99              Finished               — reached the server 175 ms late
15:07:00.795  C>S 1252            PMTU probe
15:07:00.803  C>S 38              Settings, control stream — absent from the server capture
15:07:01.054  S>C 1220, 1252, 29  server Settings
15:07:01.073  C>S 268             request + retransmitted Settings
15:07:01.143  S>C 32, 24          close
```

Server side, same flow:

```
15:07:00.851  C>S 1220            Initial
15:07:00.852  S>C 1220, 1220, 376
15:07:01.198  C>S 99, 1252
15:07:01.199  S>C 1220, 1252, 29  accepted; server Settings
15:07:01.288  C>S 268
15:07:01.288  S>C 32, 24          connect failed; connection aborted
```

The healthy sibling flow (`62291`, started 55 ms later) sent the same 38-byte packet after its Finished and later a 262-byte request packet; its Settings arrived first and the connect succeeded. The 6-byte difference is one STREAM frame for stream 2 carrying the 2-byte Settings frame. Payloads are encrypted, so the frame order inside the 268-byte packet is established by what the server read: `0x15`, a request header's size varint, as the first byte of the stream it accepted.

Fix: accept streams until the peer's unidirectional control stream arrives, and hand any bidirectional stream accepted before it to the request accept loop.

bernardnormier:
Fixes #4975.

During connection establishment, each side sends its Settings frame on its control stream and then waits for the peer's Settings frame. It never waits for the peer to have received its own Settings frame. So once a client reads the server's Settings frame, its connect completes and it can send a request while its own Settings frame is still in flight. QUIC orders bytes only within a stream, so the server can accept the request stream before the control stream. `PerformConnectAsync` took the first accepted stream as the peer's control stream and read the request header as a control frame, which failed the server's connect with:

```
IceRpc.IceRpcException: The connection establishment was aborted by an icerpc protocol error.
 ---> System.IO.InvalidDataException: Received frame type 21 but expected Settings.
```

The server then aborted the QUIC connection and the client's connect failed with "The connection was aborted by the peer". This was seen about once per 10,000 QUIC connections from GitHub-hosted runners.

With this PR, connect identifies the peer's control stream by its ID rather than by accept order. Each side opens its control stream before any other stream, so it's the peer's first unidirectional stream: ID 2 when the peer is the client, ID 3 when it's the server. Connect keeps accepting streams until that stream arrives and queues the other accepted streams for the accept requests loop, which dispatches them before accepting more streams. The same code path serves client and server connections.

A malicious or buggy peer that opens streams other than its control stream can't make the queue grow forever. Each queued stream stays open and counts against the transport's stream limits. Once the peer exhausts them, connect blocks in the accept call until the connect timeout fires and the connection is disposed, like with a peer that completes the transport handshake and never opens a stream. A peer that exceeds the limits is aborted by the transport: Slic fails the connection with a protocol error, and msquic closes a QUIC connection with a stream limit error. There's no earlier abort point: over QUIC, a bidirectional request stream or a one-way request stream can legitimately arrive before the control stream.

The new test emulates the reordering with a multiplexed connection decorator that hands the peer's control stream to the protocol connection after the next accepted stream. Both test cases fail with the error above without the fix.

icerpc/icerpc-docs#562 states this rule in the [connection establishment](https://docs.icerpc.dev/icerpc/icerpc-protocol/connection-establishment) page of the protocol documentation: the control stream is the first stream each side opens, and streams accepted before the peer's control stream are dispatched once the connection is established.

## What's Changed entry

Area: Core

- Fixed an icerpc connection establishment failure ("Received frame type N but expected Settings") that occurred over QUIC when the peer's first request stream was received before its control stream.


src/IceRpc/Internal/IceRpcProtocolConnection.cs
```diff
@@ -167,9 +168,25 @@ await SendControlFrameAsync(
                     throw;
                 }
 
-                // Wait for the remote control stream to be accepted and read the protocol Settings frame
-                _remoteControlStream = await _transportConnection.AcceptStreamAsync(
-                    connectCts.Token).ConfigureAwait(false);
+                // The peer's control stream is its first unidirectional stream: ID 2 when the peer is the client, ID 3
+                // when it's the server. With QUIC, a request stream sent by the peer can reach us before its control
+                // stream.
+                ulong remoteControlStreamId = IsServer ? 2ul : 3ul;
```
externl: This comment is hard to read: it gives the ID by the peer's role, while the code switches on ours. Something like:

```suggestion
                // The peer's control stream is the first unidirectional stream it opens, so its ID is 2 when we're the
                // server and 3 when we're the client. Over QUIC, the peer's request streams can be accepted before it.
                ulong remoteControlStreamId = IsServer ? 2ul : 3ul;
```

externl (approved):
Looks good.

- Do the streams left in `streamsAcceptedBeforeRemoteControlStream` when connect fails (e.g., a connect timeout) or when the accept loop exits early need to be completed?

src/IceRpc/Internal/IceRpcProtocolConnection.cs
bernardnormier: Done, with a small rewording.

bernardnormier:
Updated answer: the transports don't require it today. Slic closes all its streams when the connection is closed or disposed, and disposing a QUIC connection aborts its streams. But this contract isn't documented, and QUIC disposes the underlying `QuicStream` only once both pipes are completed, so a stream that is never completed keeps its native handle until finalization. We decided that the caller must complete every stream it creates or accepts, including after the connection is closed or disposed. #4979 tracks documenting this contract and completing the leftover streams in the protocol connection, including the ones this PR queues.

src/IceRpc/Internal/IceRpcProtocolConnection.cs
```diff
@@ -167,9 +168,25 @@ await SendControlFrameAsync(
                     throw;
                 }
 
-                // Wait for the remote control stream to be accepted and read the protocol Settings frame
-                _remoteControlStream = await _transportConnection.AcceptStreamAsync(
-                    connectCts.Token).ConfigureAwait(false);
+                // The peer's control stream is the first unidirectional stream it opens, so its ID is 2 when we're the
+                // server and 3 when we're the client. Over QUIC, we can accept a peer's request stream before the
+                // control stream.
```
pepone, line 176: What about:

```
// The peer's first unidirectional stream is its control stream. Its ID is 2 when we're the server
// and 3 when we're the client. A peer request stream may arrive before the control stream.
```

src/IceRpc/Internal/IceRpcProtocolConnection.cs
```diff
@@ -167,9 +168,25 @@ await SendControlFrameAsync(
...
+                        await _transportConnection.AcceptStreamAsync(connectCts.Token).ConfigureAwait(false);
+
+                    if (stream.Id == remoteControlStreamId)
+                    {
+                        _remoteControlStream = stream;
+                        break;
+                    }
+
+                    // The queue can't grow beyond the transport's stream limits: a queued stream remains open and
+                    // counts against these limits until it's dispatched.
```
pepone: is "dispatched" the right term here? Isn't until the stream is closed more correct?

pepone:
🤖 **Codex (GPT-6):**

- Please bound the pre-control-stream queue independently of the transport's active-stream limits. At [IceRpcProtocolConnection.cs:186–188](https://github.com/icerpc/icerpc-csharp/blob/fa3f0bb9c8a53f38aa019e383a1d1b96b43f8026/src/IceRpc/Internal/IceRpcProtocolConnection.cs#L186), the comment assumes queued streams keep their transport slots. With Slic, a peer can repeatedly open and abort bidirectional streams while withholding its control stream: receiving both closure frames releases each transport slot, but this queue retains the stream. I reproduced 1,000 accepted streams with `MaxBidirectionalStreams = 1` while connect remained pending. The connect timeout limits the duration of this accumulation, but the claimed stream-count bound does not hold. This is separate from the cleanup tracked in #4979.
- Please add this comment immediately before `.AddSlicTransport()` in the new test to explain why its decorator can safely assume the first accepted stream is the control stream:

  ```csharp
  // Use Slic so the first accepted stream is reliably the control stream. The decorator below moves it
  // behind a request stream to simulate QUIC's cross-stream reordering.
  .AddSlicTransport()
  ```

bernardnormier:
Right, with Slic the peer's closure frames alone release the slot, so the transport limits didn't bound the queue. Connect now accepts at most `MaxIceRpcBidirectionalStreams + MaxIceRpcUnidirectionalStreams` streams before the peer's control stream and fails with `LimitExceeded` beyond that; a well-behaved peer can't exceed it without aborting invocations during the handshake. On the test comment: the decorator's own comment already explains why the first accepted stream is the control stream over Slic, so I'd rather not add a second one at the call site.

src/IceRpc/Internal/IceRpcProtocolConnection.cs
bernardnormier: The comment is gone: the queue is now bounded explicitly rather than by the transport limits.

src/IceRpc/Internal/IceRpcProtocolConnection.cs
bernardnormier: I prefer the current wording.

pepone (approved):

externl (approved):

